PR #918, opened 2013-08-31
Redmine#3279: Evil container iteration acceptance test

tzz:
see https://cfengine.com/dev/issues/3279

This acceptance test shows some issues with container iteration:

```
2013-08-31T11:05:24-0400   notice: R: Found data 0 => 0
...
2013-08-31T11:05:24-0400   notice: R: FAIL 0: Expected '
...
```


kacfengine:
@tzz: Is this still relevant? And if so, which bug does it expose? The one mentioned further up is just a generic catchall for container support.

tzz:
not needed anymore

PR #934, opened 2013-09-05
Acceptance test for passing a container to the template_data attribute

tzz:
@sigurdteigen the notation is `data` as discussed (also see https://cfengine.com/dev/issues/3334 for symbolic references in general).


kacfengine:
This will not get merged as it is. If it depends on a feature/bug fix, it's better to open a Redmine ticket and discuss it there.

tzz:
Addressed in https://github.com/cfengine/core/pull/2156

PR #1169, opened 2013-12-05
Select fix

jimis:
Switched all select() code with poll().

TODO create libcompat poll() wrapper for old OS.

### DIFF
--- a/cf-serverd/cf-serverd-functions.c
+++ b/cf-serverd/cf-serverd-functions.c
@@ -43,6 +43,7 @@
 #include <time_classes.h>
 #include <connection_info.h>
 #include <file_lib.h>
+#include <poll.h>
 
 static const size_t QUEUESIZE = 50;
 int NO_FORK = false;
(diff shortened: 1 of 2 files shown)


jimis:
Must include in 3.7, will try to update.

ediosyncratic:
On Fri, Mar 20, 2015 at 3:00 PM, Dimitrios Apostolou <
notifications@github.com> wrote:

> Must include in 3.7, will try to update.
>

Let me know when it's time to review the updated fix ...

    Eddy.

kacfengine:
Closing for now. We should do this at some point, since ``poll()`` is both safer and faster, but we will need more code here to cater to the handicapped platforms.

worr:
Hey, before I go ahead and submit a new PR for this, could we import libressl's `poll(2)` wrapper? This would solve this problem, without having to reinvent the wheel. It's public domain, so licensing shouldn't be a problem.

https://github.com/libressl-portable/portable/blob/a7f031ba55ac4a69263000357eb7f6d7fb88101a/apps/openssl/compat/poll_win.c

jimis:
@worr nice find! Yes feel free to import it.

PR #1223, opened 2013-12-20
cf-key: New option `-T`/`--trust-server-key`.

riccardomurri:
Also rename option `-t` to `--trust-client-key`.

Client and server use different key verification methods,
so we need to use two different options.

I believe this fixes Redmine issue #3074, but more testing is actually necessary.


jimis:
Will review for 3.7 inclusion.

jimis:
Closing this, since I'm bringing it up to date in a new pull request.

@riccardomurri I'm retaining most of your commits, but for compatibility I'll keep the old `--trust-key` option, which will do both, always: write into `ppkeys` and update lastseen db.

Let me know if you disagree.

riccardomurri:
@jimis It's been quite a while since I looked at this code, but my 2nd comment from Jan 3, 2014 states that you need different roles for adding to the lastseen database -- that was the reason for having two separate `--trust-*-key`s options.  How do you decide whether you are trusting a client or server key with only one option?

jimis:
@ricardomurri: AFAICT to trust the server as cf-agent you need an entry in the lastseen database (only for the `classic` protocol), but to trust the client as cf-serverd, lastseen db can be empty. So I'll just add the lastseen db entry in both cases, if the user provided the proper syntax, i.e. `[[USER@]IPADDRESS:]FILENAME`. This should work for both server and client.

Of course I intend to test, and I'll link the new pull request from here so that you can review!

PR #1224, opened 2013-12-20
cf-key: New option `-T`/`--trust-server-key`.

riccardomurri:
Also rename option `-t` to `--trust-client-key`.

Client and server use different key verification methods,
so we need to use two different options.

Fixes Redmine issue #3074

This is a backport of pull request #1223 to the `3.5.x` branch.


jimis:
This pull request refers to 3.5.x, so closing. Pull request #1223 is the one for master.

PR #1272, opened 2014-01-11
Add acceptance test for cf-upgrade Note: this is just a mock to test cf_upgrade advertized functionality. It does _not_ actually test a package upgrade, that is dependant on the install and backup scripts.

nickanderson:


kacfengine:
+1 for effort, but we won't put any more work into cf-upgrade. All platforms support doing the same operations using package scripts, which is both safer and simpler.

PR #1412, opened 2014-02-12
Log messages improvements

lpefferkorn:
I'm trying to improve log messages on the following points:
  * Messages added where they were missing in case of error
  * Log level: adjust some level (ERR, DEBUG, ...) to be consistent with the other ones
  * Context: sometimes messages are not easy to understand without any context, for example ```Error parsing entity in 'cf_ace'``` Not easy to figure out this is from ACL stuff, so I prefix the message with ACL:
  * Consistent reporting of underlying functions errors (fopen, acl_set_perm, ...)
  * Cosmetic changes to be more consistent across messages from a same feature

I will issue several commits, one per domain (acl, xml, db, and so on), with more details in the commit message.

Please do not merge at the moment, many commits are on the way :)

### DIFF
--- a/cf-monitord/mon_cpu.c
+++ b/cf-monitord/mon_cpu.c
@@ -50,7 +50,7 @@ void MonCPUGatherData(double *cf_this)
 
     if ((fp = fopen("/proc/stat", "r")) == NULL)
     {
-        Log(LOG_LEVEL_VERBOSE, "Didn't find proc data");
+        Log(LOG_LEVEL_ERR, "Could not open /proc/stat while gathering CPU data (fopen: %s)", GetErrorStr());
         return;
     }
 
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -775,7 +775,7 @@ static void Get3Environment(EvalContext *ctx)
 
     if (stat(env, &statbuf) == -1)
     {
-        Log(LOG_LEVEL_VERBOSE, "Unable to detect environment from cf-monitord");
+        Log(LOG_LEVEL_ERR, "Unable to detect environment from cf-monitord");
         return;
     }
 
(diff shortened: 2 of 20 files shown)


jimis:
There are a bunch of changes here, some easy to merge, some not. @lpefferkorn can you update the pull request since it's not mergeable? Even better, can you submit the commits as separate pull requests? That will really help to quickly merge some of the stuff.

I'm a bit sceptical about the log level changes, from VERBOSE to ERR. The agent must log ERR only when a promise can't even be repaired. If it gets repaired, then it should be INFO. If it causes cf-agent to abort it should be CRIT. In other cases it should be VERBOSE at most, that the user's are not flooded with emails from 1000 hosts for trivial things.

I haven't looked carefully into your changes, I'm just mentioning our policy in case it's not clear.

libenv/sysinfo.c
```diff
@@ -775,7 +775,7 @@ static void Get3Environment(EvalContext *ctx)
 
     if (stat(env, &statbuf) == -1)
     {
-        Log(LOG_LEVEL_VERBOSE, "Unable to detect environment from cf-monitord");
+        Log(LOG_LEVEL_ERR, "Unable to detect environment from cf-monitord");
```
kacfengine: It is legal to simply not run cf-monitord. I think this should not be an error.

cf-monitord/mon_cpu.c
```diff
@@ -50,7 +50,7 @@ void MonCPUGatherData(double *cf_this)
 
     if ((fp = fopen("/proc/stat", "r")) == NULL)
     {
-        Log(LOG_LEVEL_VERBOSE, "Didn't find proc data");
+        Log(LOG_LEVEL_ERR, "Could not open /proc/stat while gathering CPU data (fopen: %s)", GetErrorStr());
```
kacfengine: I believe this code is shared with systems that don't have ``/proc/stat`` so it should not be an error not to be able to open it.

kacfengine:
Lots of good stuff here.

kacfengine:
Superseded by https://github.com/cfengine/core/pull/2151.

PR #1413, opened 2014-02-12
Add speical variable 'user' which can be used to access information about the current user during execution

skreuzer:
This will currently introduce 5 new special variables that contain some basic information about the user who is currently running cf-agent. This can be useful so you have have folks run cf-agent by hand to configure their environment

```
bundle agent user
{

    reports:

        freebsd::

            "$(const.dollar)(user.name) = $(user.name)";
            "$(const.dollar)(user.uid) = $(user.uid)";
            "$(const.dollar)(user.gid) = $(user.gid)";
            "$(const.dollar)(user.homedir) = $(user.homedir)";
            "$(const.dollar)(user.shell) = $(user.shell)";
}
```

```
$ cf-agent -K -f ./test.cf -b user
R: $(user.name) = skreuzer
R: $(user.uid) = 1002
R: $(user.gid) = 1002
R: $(user.homedir) = /home/skreuzer
R: $(user.shell) = /usr/local/bin/zsh
```


jimis:
Reopening this pull request to raise discussion, after discussing on IRC with @skreuzer. He is willing to bring this up to date if we consider this a good addition, which I think it is.

@kacf @nickanderson what do you think?

tzz:
I still think it would be much more useful, IMO, to provide a function to get this info for the current or any other user in a data container.

jimis:
> I still think it would be much more useful, IMO, to provide a function to get this info for the current or any other user in a data container.

That would be a different feature, providing info for any user. However having something as basic as current user info in a "sys" variable sounds essential to me, as it is basic POSIX environment.

tzz:
jimis: the amount of work is the same, but providing a single variable with that information for just the current user is really not very useful. The UID and GID are already available. Are the shell, home directory, and GECOS fields of the **current** user useful? They'll be either root's information (which is well known already) or a specific user's information. What's the use case that would utilize that information?

In the original comment @skreuzer said

> This can be useful so you have have folks run cf-agent by hand to configure their environment

Maybe he should provide more detail, I just don't see how that's useful.

In contrast, a user+group info function would be immediately useful to complement `users` promises.

kacf:
I agree with Ted that this information should be obtained via a function. It just doesn't seem very useful most of the time. A very large percentage will always be running as root.

jimis:
guys, knowing the UID is basic environment detection. The rest, I can argue. But the UID/GID is **wrong** to take it from a function, it should be detected before evaluation to ensure it hasn't changed.

tzz:
@jimis yes, since 3.6 we have `this.promiser_uid` and `this.promiser_gid` for this already in the core.

jimis:
@tzz, alright I didn't know. BTW the description and implementation are a bit confusing. It seems it's something attached to the `EvalContext` which would mean it's the **promiser uid**, as the name implies, but I'm not sure it's correctly re-computed for every new bundle (it's filled in durine `EvalContextNew()`).

`sys.uid` would be the starting agent UID, most likely zero, detected during the initial phase of environment detection. Would be useful in a future cf-agent able to drop/elevate privileges.

Anyway I'm OK closing this again, even though I think it would be a nice feature that makes sense.

skreuzer:
> Maybe he should provide more detail, I just don't see how that's useful.

Lets say you have hundreds of thousands of machines and for the most part you never log into them at all or you continually spin virtual machines up and down. Having these variables would mean that it would be possible for someone to be able to run cfengine with their own account and you can write promises that will copy zshrc into my home directory if $user.shell has zsh in it and do something else if the user wants to use bash.

While you already have the this.promiser_uid, it is only providing a small part of what getpwuid can provide and since it is posix you aren't going to have to jump through a lot of hoops to keep this up to date and working on unix and unix like platforms.

My other argument is from a documentation standpoint. It seems clearer if this type of information was provided under a "user" variable instead of "this" because someone who is new to cfengine is more likely to say I need to get the uid ID of the person currently running cfegnine, I am pretty sure the last place they are going to look is under the "this" and will mostly likely to gather this by shelling out to userland tools.

Also, considering someone else took the time to comment on this pull request that "they can think of some good use cases" makes me think that if you actually do provide this feature in cfengine other people make use of it. I have no idea why such a trivial change is being met with such opposition but if you are worried that if going forward this is going to be tons of maintenance that only benefits my narrow use case perhaps we should move this discussion to cfengine-help to see if other users would find this feature useful.

tzz:
@skreuzer it's not that it's not useful to have this information in a variable, it's just that I think the function will be a better starting point. So I'm not opposed to the feature, only to the limited utility it has vs. the general function. Let me explain what I'm proposing and I think you'll see how it actually implements your proposal and more.

Let's start with the assumption that if a core function is implemented, it will be easy to call it on startup (in the C code or in policy) to fill  in a `sys` variable as you were proposing. So implementing a function doesn't sacrifice the variable.

With that assumption, I think it's better to put the information under `sys.init_user` in a data container instead of a new `user` namespace because new namespaces require much more documentation and care. It's not a trivial change in any case. This variable and the C code have to be maintained and supported so it's better to limit the spread of top-level namespaces.

My proposal from IRC: a function `getuserinfo()` which returns a data container parallel to the `users` promises (mostly the same keys). It takes an optional UID, defaulting to the current UID. The code is already in the core for the `users` promises so it's not new code, just exposing it in a function. That's another advantage over the brand new code you proposed.

```json
{
 username: "jimis",
 description: "Dimitrios",
 home_dir: "/home/jimis",
 group_primary: "jimis",
 uid: 10000,
 groups_secondary: [ "one", "two" ],
 gid_primary: 100,
 gids_secondary: [ 1, 2 ],
 shell: "/bin/bash",
 password: { format: "hash", data: "328902390238" },
 status: "locked"
}
```

So just imagine all that (filled in the C code or in policy on startup) in the data container `sys.init_user` you can use like so: `$(sys.init_user[uid])`. I think that's very close to your proposal, except for variable name and field names.

I would provide something similar in `getgroupinfo()` for groups, but we don't have `groups` promises so I don't know the fields that will be useful.

(Also note I don't work for CFEngine at this point, so I am only trying to help you get this PR accepted and to minimize the amount of new code and difficulty for everyone involved. I'm happy to write code and docs to implement my proposal above. But I don't decide what goes in.)

kacf:
I like Ted's proposal.

jimis:
Closing again as there is no consensus. @skreuzer it's good that you didn't start to work on it at least.

Generally if any of you feel hesitant to start working on something, feel free to bring this up on the "dev-cfengine" mailing list before writing code, or in a relevant pull request, like this one (but ping always the right persons).

PR #1425, opened 2014-02-15
--with-statedir compile time option

cdituri:
Mostly interesting to package mantainers but there is some cleanup here as well.

This is the continuation of #1400 and it's ready for review.


kacfengine:
Finally resurrected this PR. The new one is at https://github.com/cfengine/core/pull/2154.

PR #1529, opened 2014-03-06
Add acceptance test for namespaced classes

nickanderson:
Ref:https://cfengine.com/dev/issues/4050


kacfengine:
Sigurd is right, the test case is invalid.

PR #1536, opened 2014-03-08
example for ensuring line presence

estenberg:
Referenced as a policy example in our documentation.


kacfengine:
Fixed indentation and merged. Cannot be autotested due to the use of ``stdlib.cf``.

PR #2008, opened 2014-09-26
Fixes #6604: handle SIGHUP for cf-serverd

ncharles:
See https://dev.cfengine.com/issues/6604


kacfengine:
Merged with many additional fixes in https://github.com/cfengine/core/commit/d0547c25247a7a9fec07585e2e2f93a27b98b8cb.

PR #2037, opened 2014-11-21
Redmine#6484 fix storing promise handles during pre-evaluation step.

pasinskim:
While doing 'Preliminary variable/class-context convergence' all the
promise handles were stored inside context. While evaluating the promise
all the handles were already stored so that every dependency was already
resolved.
Right now while doing pre-evaluation step handles are not stored.


kacfengine:
@pasinskim: ping. I'm ok with merging this. I just want to know if you have an opinion on the test coverage I mentioned in the previous comment.

pasinskim:
trigger build

pasinskim:
@kacfengine @jimis now this should be properly fixed. There is not much more we can do without rewriting evaluator but all cases except depending_on classes or variables should be covered.

Please review!

pasinskim:
@kacfengine test case added!

PR #2050, opened 2014-12-03
Fixes line truncation in insert_lines statements at 4k

worr:
This includes a few fixes for insert_lines promises, where the lines get truncated at 4k. I've also included a basic acceptance test.

This change also includes a fix for SplitString, as it's used in one of the insertion functions. There's more to come for this file, but since it was brought up in help-cfengine, I think it's worth submitting as-is right now for review and possible inclusion.


worr:
Any issues here?

libpromises/matching.c
```diff
@@ -308,6 +308,27 @@ void EscapeSpecialChars(const char *str, char *strEsc, int strEscSz, char *noEsc
...
 }
 
+size_t EscapeRegexCharsLen(const char *str)
+{
+    size_t ret = 0;
+    for (const char *sp = str; *sp != '\0'; sp++)
+    {
+        switch (*sp)
+        {
+            case ' ':
```
kacfengine: Shouldn't this be ``'.'``, like in the function below?

Both functions are actually quite bad at escaping regex patterns, there are lots of other special characters too: ``[, ], ?, (, )``, not to mention ``\`` itself. But at least this way they are in sync.

libpromises/item_lib.c
```diff
@@ -515,37 +515,26 @@ Item *SplitString(const char *string, char sep)
...
-    while (*sp != '\0')
+    while ((sp = strchr(before, sep)) != NULL)
     {
-        if (*sp != sep)
-        {
-            before[i] = *sp;
-            i++;
-        }
-        else if (sp > string && sp[-1] == '\\')
+        if (sp[-1] != '\\')
```
kacfengine: There is still this issue.

worr:
Updated with fixes

cf-agent/files_editline.c
```diff
@@ -1387,21 +1395,28 @@ static int SanityCheckDeletions(Attributes a, const Promise *pp)
 
 /***************************************************************************/
 
+/* XXX */
```
jimis: XXX?

jimis:
> Updated with fixes

@worr there is still the out of bounds issue that me and kristian pointed out, in item_lib.c.

worr:
Sorry, I don't know how I missed that comment. I've updated the PR with the new changes.

jimis:
@worr since each commit is independant, could you break them into separate pull requests, at least the ones that are easily separated? This will hopefully speed-up review and merging! Because at first two commits were reviewed, then a couple more were added and some were changed, and we've lost track what is blocked by what!

If so then please reference the new pull requests with comments in this one. Thanks!

worr:
Sorry, it's been a busy month at work. Finally got back to the comments today.

kacfengine:
trigger build

kacfengine:
Looks good to me.

kacfengine:
Thanks William!

PR #2055, opened 2014-12-04
Add: Check that we don't segfault with incomplete package_methods

nickanderson:
This test only covers redhat and only the package_add command. Probably more
permutations of this test would be valuable.


kacfengine:
Close due to inactivity. Please reopen if needed.

PR #2064, opened 2014-12-09
Fix stale processes lingering after test and reenable it.

kacfengine:
It still doesn't work under fakeroot, but the test is skipped if you
use it.


kacfengine:
Superseded by #2081.

PR #2069, opened 2014-12-14
Do not fallback on unsafe seed data for the PRNG

worr:
Seeding OpenSSL's PRNG with the output of rand() isn't safe. It's better to
abort early than to continue with unsafe entropy.

Also write out a new randseed file on a deinitialization, and free our error
strings.

### DIFF
--- a/libpromises/crypto.c
+++ b/libpromises/crypto.c
@@ -87,10 +87,25 @@ void CryptoDeInitialize()
 {
     if (crypto_initialized)
     {
+        char randfile[CF_BUFSIZE];
+        snprintf(randfile, CF_BUFSIZE, "%s%crandseed",
+                 CFWORKDIR, FILE_SEPARATOR);
+
+        /* Only write out a seed if the file doesn't exist
+         * and we have enough entropy to do so. If RAND_write_File
+         * returns a bad value, delete the poor seed.
+         */
+        if (access(randfile, R_OK) && errno == ENOENT && RAND_write_file(randfile) != 1024)
+        {
+            Log(LOG_LEVEL_WARNING,
+                "Could not write randomness to '%s'", randfile);
+            unlink(randfile); /* do not reuse entropy */
+        }
+
+        chmod(randfile, 0600);
         EVP_cleanup();
         CleanupOpenSSLThreadLocks();
-        // TODO: Is there an ERR_unload_crypto_strings() ?
-        // TODO: Are there OpenSSL_clear_all_{digests,algorithms} ?
+        ERR_free_strings();
         crypto_initialized = false;
     }
 }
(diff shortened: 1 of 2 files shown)


libpromises/crypto.c
```diff
@@ -87,10 +87,25 @@ void CryptoDeInitialize()
...
+         * returns a bad value, delete the poor seed.
+         */
+        if (access(randfile, R_OK) && errno == ENOENT && RAND_write_file(randfile) != 1024)
+        {
+            Log(LOG_LEVEL_WARNING,
+                "Could not write randomness to '%s'", randfile);
+            unlink(randfile); /* do not reuse entropy */
+        }
+
+        chmod(randfile, 0600);
```
jimis: Good catch this `chmod()`. This brought to my attention that the existing `chmod(644)` in KeepKeyPromises is bogus. Also I'd expect that OpenSSL is doing the right thing with permissions, if so, do you want to remove both chmod()s?

libpromises/crypto.c
```diff
@@ -112,51 +127,39 @@ static void RandomSeed(void)
 
     /* 2. Seed the strong OpenSSL PRNG. */
 
-    /* randseed file is written by cf-key. */
+    /* randseed file is written on deinitialization of crypto system */
     char randfile[CF_BUFSIZE];
     snprintf(randfile, CF_BUFSIZE, "%s%crandseed",
              CFWORKDIR, FILE_SEPARATOR);
     Log(LOG_LEVEL_VERBOSE, "Looking for a source of entropy in '%s'",
         randfile);
```
jimis: This whole block (especially the log message) should be moved further down.

libpromises/crypto.c
```diff
@@ -112,51 +127,39 @@ static void RandomSeed(void)
...
-            "A workaround is to copy 1KB of random bytes to '%s'",
-            randfile);
-
-        unsigned char rand_buf[128];
-        for (size_t i = 0; i < sizeof(rand_buf); i++)
+        if (RAND_load_file(randfile, -1) != 1024)
         {
-            rand_buf[i] = rand() % 256;
+            Log(LOG_LEVEL_CRIT,
+                "Could not read randomness from '%s'", randfile);
```
jimis: Please print the workaround like before, if we abort we should at least give a clue on how to fix it.

jimis:
Easy to merge fixes are checking `RAND_write_file()` return value, the cleanups in DeInitialize, doing RAND_load_file() only if self-seeding has failed etc.

But writing the entropy file on every program exit, and aborting the agent in case of low entropy are controversial.

Can you put them in separate commits?

libpromises/crypto.c
worr: I don't think printing the workaround is a good idea. I don't think the users should be trusted to come up with seed-grade entropy.

worr:
I'll put them in separate commits and update the PR. The current patch does not write the entropy file on every program exit, it just writes it if it doesn't exist. The entropy file is deleted after use.

worr:
Updated

kacfengine:
@jimis: Are you happy with this?

jimis:
@kacfengine It's good, so lets put it in master. It has the potential of breaking some obscure embedded platform with limited entropy, so not for 3.6.x for now.

PR #2074, opened 2014-12-24
bundlestate: new function

tzz:
This is not finished because I'm unable to test on Mac OS X (error: 

```
duplicate symbol _DEFAULT_CONTAINER_CAPACITY in:
    ../libutils/.libs/libutils.a(json.o)
    ../libutils/.libs/libutils.a(json-yaml.o)
ld: 1 duplicate symbol for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
)

Thus the acceptance test is not finished and the code may be slightly broken, but I hope the general idea is acceptable: this function allows all the variables of a bundle to be returned in a single data container, formatted correctly (especially nested data structures).

### DIFF
--- a/cf-agent/verify_files.c
+++ b/cf-agent/verify_files.c
@@ -595,7 +595,7 @@ static PromiseResult RenderTemplateMustache(EvalContext *ctx, const Promise *pp,
     JsonElement *default_template_data = NULL;
     if (!a.template_data)
     {
-        a.template_data = default_template_data = DefaultTemplateData(ctx);
+        a.template_data = default_template_data = DefaultTemplateData(ctx, NULL);
     }
 
     Buffer *output_buffer = BufferNew();
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -2666,15 +2666,21 @@ static FnCallResult FnCallMergeData(EvalContext *ctx, ARG_UNUSED const Policy *p
     assert(false);
 }
 
-JsonElement *DefaultTemplateData(const EvalContext *ctx)
+JsonElement *DefaultTemplateData(const EvalContext *ctx, const char *wantbundle)
 {
     JsonElement *hash = JsonObjectCreate(30);
-    JsonElement *classes = JsonObjectCreate(50);
-    JsonElement *bundles = JsonObjectCreate(50);
-    JsonObjectAppendObject(hash, "classes", classes);
-    JsonObjectAppendObject(hash, "vars", bundles);
+    JsonElement *classes = NULL;
+    JsonElement *bundles = NULL;
 
+    bool want_all_bundles = (NULL == wantbundle);
+
+    if (want_all_bundles) // no specific bundle
     {
+        classes = JsonObjectCreate(50);
+        bundles = JsonObjectCreate(50);
+        JsonObjectAppendObject(hash, "classes", classes);
+        JsonObjectAppendObject(hash, "vars", bundles);
+
         ClassTableIterator *it = EvalContextClassTableIteratorNewGlobal(ctx, NULL, true, true);
         Class *cls;
         while ((cls = ClassTableIteratorNext(it)))
@@ -2684,11 +2690,8 @@ JsonElement *DefaultTemplateData(const EvalContext *ctx)
             free(key);
         }
         ClassTableIteratorDestroy(it);
-    }
 
-    {
-        ClassTableIterator *it = EvalContextClassTableIteratorNewLocal(ctx);
-        Class *cls;
+        it = EvalContextClassTableIteratorNewLocal(ctx);
         while ((cls = ClassTableIteratorNext(it)))
         {
             char *key = ClassRefToString(cls->ns, cls->name);
@@ -2705,17 +2708,30 @@ JsonElement *DefaultTemplateData(const EvalContext *ctx)
         {
             // TODO: need to get a CallRef, this is bad
             char *scope_key = ClassRefToString(var->ref->ns, var->ref->scope);
-            JsonElement *scope_obj = JsonObjectGetAsObject(bundles, scope_key);
-            if (!scope_obj)
+
+            JsonElement *scope_obj = NULL;
+            if (want_all_bundles)
             {
-                scope_obj = JsonObjectCreate(50);
-                JsonObjectAppendObject(bundles, scope_key, scope_obj);
+                scope_obj = JsonObjectGetAsObject(bundles, scope_key);
+                if (!scope_obj)
+                {
+                    scope_obj = JsonObjectCreate(50);
+                    JsonObjectAppendObject(bundles, scope_key, scope_obj);
+                }
             }
+            else if (0 == strcmp(scope_key, wantbundle))
+            {
+                scope_obj = hash;
+            }
+
             free(scope_key);
 
-            char *lval_key = VarRefToString(var->ref, false);
-            JsonObjectAppendElement(scope_obj, lval_key, RvalToJson(var->rval));
-            free(lval_key);
+            if (NULL != scope_obj)
+            {
+                char *lval_key = VarRefToString(var->ref, false);
+                JsonObjectAppendElement(scope_obj, lval_key, RvalToJson(var->rval));
+                free(lval_key);
+            }
         }
         VariableTableIteratorDestroy(it);
     }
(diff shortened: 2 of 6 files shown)


tzz:
Although the compilation works, I get more errors and most of the acceptance tests are failing on Mac OS X.  Please test the updated PR, which should work.

Standalone test policy:

```
bundle agent main
{
  methods:
      "" usebundle => holder;
      "" usebundle => print;
}

bundle agent holder
{
  vars:
      "s" string => "Hello!";
      "d" data => parsejson('[4,5,6]');
      "list" slist => { "element1", "element2" };
      "h" data => parsejson('{ "a": "x", "b": "y"}');
}

bundle agent print
{
  vars:
      "holder" data => bundlestate("holder");
      "out" string => format("%S", holder);
  reports:
      "holder state = $(out)";
}
```

Output:

```
R: holder state = {"list":["element1","element2"],"h":{"a":"x","b":"y"},"s":"Hello!","d":[4,5,6]}
```

tzz:
I verified the acceptance tests pass on Linux and rebased against the latest master.

tzz:
Remaining questions:

Should `bundlestate` fail if it doesn't find the bundle?

Should `bundlestate` require a namespace (currently it will match the
bundle name in any namespace)?  Should it take a regex or a fixed string?

estenberg:
Is there a Redmine for this? I know there was a long discussion on the mailing list, but not everyone can follow that.

The value of this function is unclear to me. At least what's in the acceptance test can be (more) easily achieved by using classic CFEngine references like $(holder.s), $(holder.d) and so on.

tzz:
@estenberg there is no Redmine issue, unless you make it part of the one for bundle return data.  You and @kacfengine should decide how to track it, he can summarize it better :)

The use case is "I want to return data from a bundle in a nice data container."  This has the following advantages:

* all the variable data (strings, slists, etc.) go inside the data container, in a single JSON data holder
* you don't need to know the bundle name after you capture this data container, so you can pass it around or merge it with other data.  It's **true** return data, not just a variable lookup.
* the bundle return data is captured in the data container, so you don't have to worry about changes to the underlying variables
* there is no new syntax and just a function call, yet we have a way to return any amount of data from a bundle
* it's **very** useful for debugging

Feel free to use the above in Redmine :)

tzz:
To be consistent with the naming conventions we've discussed, maybe the function should be called `data_bundlestate` or something else starting with `data_`?

libpromises/evalfunction.c
```diff
@@ -2666,14 +2666,23 @@ static FnCallResult FnCallMergeData(EvalContext *ctx, ARG_UNUSED const Policy *p
...
+
+    if (want_all_bundles) // no specific bundle
+    {
+        classes = JsonObjectCreate(50);
+        bundles = JsonObjectCreate(50);
+        JsonObjectAppendObject(hash, "classes", classes);
+        JsonObjectAppendObject(hash, "vars", bundles);
+    }
+
+    if (want_all_bundles) // no specific bundle
```
kacfengine: All same condition. The three ``if (want_all_bundles)`` can be consolidated into one (there is one more below as well).

libpromises/evalfunction.c
```diff
@@ -2733,10 +2756,29 @@ static FnCallResult FnCallDatastate(EvalContext *ctx,
...
+                                      ARG_UNUSED const FnCall *fp,
+                                      ARG_UNUSED const Rlist *args)
+{
+    JsonElement *state = DefaultTemplateData(ctx, RlistScalarValue(args));
+
+    if (NULL == state ||
+        JsonGetElementType(state) != JSON_ELEMENT_TYPE_CONTAINER ||
+        JsonLength(state) < 1)
+    {
+        return FnFailure();
```
kacfengine: Very corner case, I know, but perhaps it should distinguish between a failure to find the bundle and a bundle that's empty. With this code you would get an unexpanded data container if you tried to use it after returning data from an empty bundle.

That is, unless there is some bundle member that is always returned.

tests/acceptance/01_vars/02_functions/bundlestate.cf
```diff
@@ -0,0 +1,65 @@
...
+["element1","element2"]
+{"a":"x","b":"y"}
+{"external_s":"External Hello!"}
+$(test.holder_s)
+{"external_s":"External Hello!"}
+$(test.holder2_defined)
+$(test.holder3_defined)',
+                                           $(this.promise_filename),
+                                           "no");
+}
```
kacfengine: Heh, I found this indentation confusing. But it *is* correct, and I can't think of a better to indent it. :-)

tests/acceptance/01_vars/02_functions/bundlestate.cf
```diff
@@ -0,0 +1,65 @@
...
+",
+                                           '"Hello!"
+[4,5,6]
+["element1","element2"]
+{"a":"x","b":"y"}
+{"external_s":"External Hello!"}
+$(test.holder_s)
+{"external_s":"External Hello!"}
+$(test.holder2_defined)
+$(test.holder3_defined)',
```
kacfengine: Nitpick: For these two you swapped the expected/actual values, the expected values are in the first block. For all the rest they are in the second block.

kacfengine:
Overall very nice code. One thing I'd like to see is a test checking that only common bundles are accepted, like we talked about on the mailing list. This would of course need some additional code changes as well.

libpromises/evalfunction.c
kacfengine: Oh, and since ``NULL == state`` is not the only condition, you may need to destroy ``state`` before failing.

tzz:
In the mailing list we discussed that `runbundle` (when implemented) should accept only common bundles.  But `bundlestate` can get the data state of any bundle, so I think it's OK as it is.

tests/acceptance/01_vars/02_functions/bundlestate.cf
tzz: Yeah, sorry.  It sucks either way.

tests/acceptance/01_vars/02_functions/bundlestate.cf
tzz: Oh, yeah.  Fixed.

libpromises/evalfunction.c
tzz: OK; consolidated.

libpromises/evalfunction.c
tzz: I *think* calling `bundlestate` on an empty bundle should be an error, because it **could** be too early to get all the variables from that bundle (it hasn't converged yet).  I'm not sure though.

I added `JsonDestroy(state)` as you suggested, thanks!

kacfengine:
> In the mailing list we discussed that runbundle (when implemented) should accept only common bundles. But bundlestate can get the data state of any bundle, so I think it's OK as it is.

Right, good point. No need to restrict the bundlestate function.

libpromises/evalfunction.c
kacfengine: Ah yes, in case convergence will make it succeed later... I'm not entirely sure either, to be honest, but I assume your proposed behavior is the safest in that case.

cf-agent/verify_files.c
```diff
@@ -595,7 +595,7 @@ static PromiseResult RenderTemplateMustache(EvalContext *ctx, const Promise *pp,
     JsonElement *default_template_data = NULL;
     if (!a.template_data)
     {
-        a.template_data = default_template_data = DefaultTemplateData(ctx);
+      a.template_data = default_template_data = DefaultTemplateData(ctx, NULL);
```
kacfengine: Oh right, I forgot this one, bogus indentation. I can fix it when merging though. This is more note-to-self.

kacfengine:
trigger build

kacfengine:
Assuming the build goes through, this is pretty much ready now, so only documentation remains.

tzz:
documentation in https://github.com/cfengine/documentation/pull/983

tzz:
I also added an example here, which is included by the documentation page.  I hope it's helpful.

tzz:
...and I rebased+squashed, and fixed that indentation in `verify_files.c`.  HTH.

kacfengine:
Thanks!

I got an unrelated Jenkins hiccup. Will run again.

kacfengine:
trigger build

kacfengine:
Thanks a bunch, Ted!

PR #2079, opened 2015-01-05
Fix parser for lists larger than 4994 items

jimis:
Redmine#6672


jimis:
@kacfengine feel free to review if you feel comfortable with yacc, else I'll just go ahead and merge. The important commit is only a28b28e.

kacfengine:
Nice!

And +1 for using single characters.

kacfengine:
trigger build

jimis:
Please don't merge yet, I'll amend a commit with acceptance test.

kacfengine:
Ok, the build queue is quite long right now, so I cancelled the build so you can run it with the acceptance test instead.

jimis:
Done, dear jenkins please trigger build.

kacfengine:
trigger build

kacfengine:
You need a plain string. I tried that before.

kacfengine:
PS! You might want to make the list considerably longer actually. I'm assuming that 4994 items is tied specifically to Linux with a certain stack setting in the kernel. It could easily be different.

jimis:
> You need a plain string. I tried that before.

Strange, I've seen that Jenkins source code uses ```String.contains()``` (or something like that), at least the way I configured it two months ago (previously it didn't work anyway).

jimis:
> PS! You might want to make the list considerably longer actually. I'm assuming that 4994 items is tied specifically to Linux with a certain stack setting in the kernel. It could easily be different.

Indeed, but I wouldn't know what size to choose. This (stack size) is a tweakable knob in Bison and /probably/ tweakable via `ulimit()` as well, but I'm not sure about other exotic yacc implementations. And the file size and slowness (parsing is way too slow!) made me go with this size.

kacfengine:
Ok, sounds wise.

kacfengine:
Jenkins looks good. Merge?

jimis:
yes please :-)

PR #2080, opened 2015-01-05
Child should exit if parent is dead.

jimis:
To avoid stale processes when test fails, relevant ticket is Redmine#6907.

### DIFF
--- a/tests/load/lastseen_threaded_load.c
+++ b/tests/load/lastseen_threaded_load.c
@@ -18,6 +18,7 @@ unsigned int ROUND_DURATION = 10;          /* how long to run each loop */
  * from child threads and read from the main one. It's only a test, so
  * @ediosyncratic please spare me. */
 time_t START_TIME;
+pid_t PPID;
 char CFWORKDIR[CF_BUFSIZE];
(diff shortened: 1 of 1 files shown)


kacfengine:
Unfortunately the child appears to hang inside the database code, so it never returns to that loop. This is still a good fix though, so feel free to merge, but we can't remove the escape hatch in the script that runs this test.

jimis:
OK merging. Next thing I can think of is having a periodic SIGALRM checking if parent is dead.

BTW what do you mean by "escape hatch in the script"?

kacfengine:
> BTW what do you mean by "escape hatch in the script"?

Oh right, I haven't merged that yet. Here: https://github.com/kacfengine/core/commit/7ef4017a17949eab943ae0a5bcc80107198e0efc

jimis:
> Oh right, I haven't merged that yet. Here: kacfengine@7ef4017

Interesting, I think it's good enough for a test. Maybe some other time I can try different solutions on the HP-UX system, like alarm() or setsid() on the parent process.

kacfengine:
Cool!

PR #2081, opened 2015-01-06
Make backgrounding in tests easier by introducing no_fds command.

kacfengine:
The problem is that we don't know which file descriptors have been
opened behind our back, and most will be inherited by the new process.
The original process or its parent may then hang because some file
descriptor is still open, even though we didn't intend that. The
FD_CLOEXEC mechanism is not very helpful for descriptors we don't own,
so use the closefrom call to close all the descriptors we don't need
in the new process instead.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -766,6 +766,13 @@ AC_C_BIGENDIAN
 AC_CHECK_HEADERS([endian.h])
 AC_CHECK_DECLS(le32toh, [], [], [[#include <endian.h>]])
 
+AC_CHECK_DECLS(closefrom, [], [], [[#include <unistd.h>
+                                    #include <stdlib.h>]])
+AC_REPLACE_FUNCS(closefrom)
+
+AC_CHECK_HEADERS([sys/pstat.h])
+AC_CHECK_FUNCS(pstat_getfile2)
+
 CF3_PATH_ROOT_PROG([CHPASSWD], [chpasswd], [], [/sbin:/usr/sbin:/bin:/usr/bin:$PATH])
 AS_IF([test "x$CHPASSWD" != "x"],
       [AC_DEFINE(HAVE_CHPASSWD, 1, [Define if chpasswd tool is present])]
--- a/libcompat/closefrom.c
+++ b/libcompat/closefrom.c
@@ -1,0 +1,120 @@
+/*
+   Copyright (C) CFEngine AS
+
+   This file is part of CFEngine 3 - written and maintained by CFEngine AS.
+
+   This program is free software; you can redistribute it and/or modify it
+   under the terms of the GNU General Public License as published by the
+   Free Software Foundation; version 3.
+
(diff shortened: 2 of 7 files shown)


kacfengine:
I still don't know for sure if this will help, but it's worth a shot.

kacfengine:
trigger build

configure.ac
```diff
@@ -766,6 +766,12 @@ AC_C_BIGENDIAN
 AC_CHECK_HEADERS([endian.h])
 AC_CHECK_DECLS(le32toh, [], [], [[#include <endian.h>]])
 
+AC_CHECK_DECLS(closefrom, [], [], [[#include <unistd.h>]])
+AC_REPLACE_FUNCS(closefrom)
```
jimis: I have no manpage for `closefrom()`, is it BSD only?

jimis:
+1 for your persistence in fixing the if_elapsed test :-)

BTW any idea why CLOEXEC is not helping, i.e. which file descriptors we do not own? Another issue is that some platforms don't implement CLOEXEC flag at all, maybe this is what you're seeing?

configure.ac
kacfengine: BSD and Solaris. AIX equivalent is the ``F_CLOSEM`` call. Linux has no equivalent.

kacfengine:
> +1 for your persistence in fixing the if_elapsed test :-)

It's not just for the if_elapsed test. My idea is that this can be used in order to get parallel cf-serverd tests, by allowing the server to be backgrounded. Granted, this might have worked anyway if the server daemonizes in one test and is used in another, but with ``no_fds`` we'll be completely sure. See this recent commit https://github.com/cfengine/core/commit/bf8670cf85b757db29cdd84c3e86c04cc9d62e2e (particularly the README part) for some more info on the plan.

> BTW any idea why CLOEXEC is not helping, i.e. which file descriptors we do not own? Another issue is that some platforms don't implement CLOEXEC flag at all, maybe this is what you're seeing?

I don't know. I did see two unknown file descriptors open in the sub process of cf-agent, but I have not been able to determine where they come from. At least in my Windows debugging of the problem, it seemed fairly nontrivial to prevent file descriptors from being inherited in general. ``CLOEXEC`` is hardly used in our code at all.

kacfengine:
Dammit, I still saw some stale processes on AIX. It's possible they were old though. I will gather some more data before merging this.

Plus Windows chokes on this code.

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
Finally!

PR #2083, opened 2015-01-09
Fix some compiler warnings (gcc -Wextra)

stweil:


kacfengine:
Less noise, I love it!

PR #2085, opened 2015-01-09
Fix some errors (all reported by cppcheck)

stweil:
This is a first bunch of fixes for some real errors which were reported by the static code analyser cppcheck.

### DIFF
--- a/cf-agent/verify_networks.c
+++ b/cf-agent/verify_networks.c
@@ -415,7 +415,7 @@ static int GetArpInfo(ARPState **list, const Promise *pp)
     size_t line_size = CF_BUFSIZE;
     char *line = xmalloc(line_size);
     char comm[CF_BUFSIZE];
-    char ip[CF_MAX_IP_LEN], mac[CF_MAX_IP_LEN], device[CF_SMALLBUF], state[CF_SMALLBUF];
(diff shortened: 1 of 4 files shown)


cf-agent/verify_networks.c
```diff
@@ -415,7 +415,7 @@ static int GetArpInfo(ARPState **list, const Promise *pp)
     size_t line_size = CF_BUFSIZE;
     char *line = xmalloc(line_size);
     char comm[CF_BUFSIZE];
-    char ip[CF_MAX_IP_LEN], mac[CF_MAX_IP_LEN], device[CF_SMALLBUF], state[CF_SMALLBUF];
+    char ip[CF_MAX_IP_LEN + 1], mac[CF_MAX_IP_LEN], device[CF_SMALLBUF], state[CF_SMALLBUF];
```
kacfengine: Why not ``mac`` as well?

stweil: mac is filled with a maximum of 31 characters plus terminating \0, so only up to 32 bytes from the 64 byte buffer are used: the declaration allocates more memory than needed (that does not harm much).

kacfengine:
trigger build

kacfengine:
Thanks, nice fixes!

kacfengine:

PR #2086, opened 2015-01-12
LogRaw cleanups.

jimis:


jimis:
@kacfengine I define here CF_MAX_BUFSIZE which ideally should be the maximum size of arrays on the stack (in order to control unexpected stack overflows using VLAs).

I'm thinking that this, together with some defines from `libpromises/cf3.defs.h` should go in a `libutlis/cf3_defs.h`.

jimis:
trigger build

libutils/logging.c
```diff
@@ -277,25 +277,33 @@ void VLog(LogLevel level, const char *fmt, va_list ap)
...
+ * @param #buflen must be no more than CF_MAX_BUFSIZE
  */
 void LogRaw(LogLevel level, const char *prefix, const void *buf, size_t buflen)
 {
-    /* Translate non printable characters to printable ones. */
-    const unsigned char *src = (const unsigned char *) buf;
+    const unsigned char *src = buf;
     unsigned char dst[buflen+1];
-    size_t i;
+    assert(sizeof(dst) <= CF_MAX_BUFSIZE);
```
kacfengine: ``LogRaw`` is anything that comes from ``Log``, isn't it? Since this is potentially dependent on user data (if you have some long JSON string or something), I think we shouldn't assert on overflowing the length.

I do agree that allocating on the stack with no check at all is bad though. Maybe truncate message instead, like ``What does the fox sa...``.

kacfengine: Hmm, I see now that I got it the wrong way around, ``LogRaw`` calls ``Log``. My comment may still be relevant though. Where is ``LogRaw`` used?

jimis: I totally agree with truncating extra long messages, but `LogRaw()` is not
called from `Log()`, but it calls `Log()` itself. For now it's used to
print network buffers, which are guaranteed to be less than CF_BUFSIZE.

libutils/logging.c
```diff
@@ -277,25 +277,33 @@ void VLog(LogLevel level, const char *fmt, va_list ap)
...
 
-    for (i = 0; i < buflen; i++)
+    LoggingContext *lctx = GetCurrentThreadContext();
+    if (level <= lctx->report_level)
```
kacfengine: There appears to be two log levels, ``log_level`` and ``report_level``. Are you sure it's correct of ``LogRaw`` to look only at ``report_level``?

jimis: Hmmm maybe I should check both `log_level` and `report_level`.

jimis:
Rebased and force-pushed, checking both `report_level` and `log_level`.

jimis:
This should be cherry-picked to 3.6.x.

https://dev.cfengine.com/issues/6918

libutils/logging.c
kacfengine: Ok!

kacfengine:
trigger build

PR #2088, opened 2015-01-13
Create global set of keys for bodies

worr:
The new global set of keys currently has only one entry (meta),
but can be extended into multiple entries later.

### DIFF
--- a/libpromises/mod_access.c
+++ b/libpromises/mod_access.c
@@ -56,6 +56,7 @@ static bool AccessParseTreeCheck(const Promise *pp, Seq *errors);
 
 static const ConstraintSyntax report_data_select_constraints[] =
 {
+    CONSTRAINT_SYNTAX_GLOBAL,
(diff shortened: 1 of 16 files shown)


jimis:
So this adds "meta" attribute to all bodies?

basvandervlies:
the discussion is here: https://groups.google.com/forum/#!msg/dev-cfengine/xXU8GqPKt0U/fM1S5Yg5zrwJ

kacfengine:
So I suppose this only adds the syntax, and doesn't make any use of the tags?

I'd still like to know what people want to achieve with this. The only reason I've heard so far is a "deprecated" attribute, for which I think this is a little overkill. Is there anything else you have in mind?

kacfengine:
trigger build

kacfengine:
Given the fact that we have used ``meta`` to deprecate things in masterfiles already, and the fact that we're merely extending a concept into bodies which already exists for bundles, I'm gonna merge this.

kacfengine:
Thanks, @worr!

PR #2089, opened 2015-01-14
Enable support for Busybox tools.

kacfengine:
When running on Linux, CFEngine core assumes that the GNU utilities are
used.  Howver on embedded devices we commonly (always?) use busybox.
Thus 'out-of-the-box' certain functions do not work.  For example, it's
not possible to bootstrap a device because busybox /bin/ps accepts
different arguments.  The patch requires __BUSYBOX__ to be defined in
the build environment for busybox targets - hopefully there is a better
way where it can be detected automatically at runtime.


kacfengine:
trigger build

libpromises/systype.c
```diff
@@ -95,6 +97,7 @@ const char *const VPSOPTS[] =
     [PLATFORM_CONTEXT_AIX] =  "-N -eo user,pid,ppid,pgid,pcpu,pmem,vsz,ni,stat,st=STIME,time,args",       /* aix */
     /* Note: keep in sync with GetProcessOptions()'s hack for Linux 2.4 */
     [PLATFORM_CONTEXT_LINUX] = "-eo user,pid,ppid,pgid,pcpu,pmem,vsz,ni,rss:9,nlwp,stime,etime,time,args",/* linux */
+    [PLATFORM_CONTEXT_BUSYBOX] = "",                  /* linux / busybox */
```
jimis: If I recall, it just needs the "w" argument to avoid truncation. I'll login to our router to find out. :-p

jimis:
+1

kacfengine:
trigger build

libpromises/systype.c
jimis: Indeed, "w" seems to be the only acceptable parameter, at least for recent versions of busybox. I'll directly commit that change, OK @kacfengine?

kacfengine: Ok with me!

PR #2090, opened 2015-01-14
Redmine#6920: fix with acceptance test

tzz:
Attempt to fix https://dev.cfengine.com/issues/6920


libutils/sequence.c
```diff
@@ -350,3 +350,45 @@ Seq *SeqGetRange(const Seq *seq, size_t start, size_t end)
...
+                prev = cur + 1;
+            }
+
+            cur++;
+        }
+
+        if (cur > prev)
+        {
+            SeqAppend(seq, xstrndup(prev, cur - prev));
+        }
```
kacfengine: If I understand correctly, if the final character in the string is a delimiter, then a final empty element will not be added, since both ``cur`` and ``prev`` will be at the final ``\0``. I would expect such an element to be added in that case, but I haven't checked if it's appropriate for ``data_readstringarrayidx``.

kacfengine:
Looks nice though.

libutils/sequence.c
tzz: Empty strings are generally not useful in CFEngine, so I would leave the current behavior.

kacfengine: I see now that I misunderstood, I thought the function was used to delimit substrings on one line, but it's only used to delimit lines. Then I agree.

kacfengine:
trigger build

kacfengine:
Thanks!

PR #2091, opened 2015-01-17
Redmine #6582: Fix deletion of dangling symlinks

worr:
`DepthSearch` isn't entered when there's a dangling symlink. This is usually a good thing, however if a user wants to delete a dangling symlink with CFEngine, this poses a problem.

Also includes acceptance test.

### DIFF
--- a/cf-agent/verify_files.c
+++ b/cf-agent/verify_files.c
@@ -264,6 +264,7 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
     struct stat osb, oslb, dsb;
     CfLock thislock;
     int exists;
+    bool link = false;
 
     Attributes attr = GetFilesAttributes(ctx, pp);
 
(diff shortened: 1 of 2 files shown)


worr:
Redmine: https://dev.cfengine.com/issues/6582

nickanderson:
Nice test

nickanderson:
Might want to make sure that "$(G.testdir)/sourcefile" doesn't exist in init.

tdlee:
Thanks for looking at this.  And thanks also for ensuring there is an acceptance test!

worr:
@nickanderson fixed acceptance test per your recommendation

kacfengine:
> Might want to make sure that "$(G.testdir)/sourcefile" doesn't exist in init.

Actually this isn't necessary when running under the test framework, since it always starts with a blank testdir. It doesn't harm though.

kacfengine:
> It doesn't harm though.

Or actually in this case it did. It was missing the body definition. I fixed that and one other issue and resubmitted the PR here: #2108

PR #2095, opened 2015-01-20
foo!bar should not be interpretted as a valid class statement

worr:
Adds stricter token parsing. Only certain symbols should legally follow
valid tokens. ParseToken now checks the following character to ensure
that we aren't writing invalid class statements.

Before, `foo!bar` or `default:!foo` would be interpreted as valid class statements. It's unlikely that a user would encounter the former, but we had a case at LinkedIn where we encountered the latter. The old behavior was to just resolve that class as false *all* the time. Now, it triggers a parse error.

@nickanderson: this was the bug that I was talking to you about at the bar

### DIFF
--- a/libpromises/string_expressions.c
+++ b/libpromises/string_expressions.c
@@ -169,7 +169,18 @@ static StringParseResult ParseToken(const char *expr, int start, int end)
         endlit++;
     }
 
-    if (endlit > start)
+    if (endlit > start &&
+        (expr[endlit] == ':' ||
+        expr[endlit] == '.' ||
+        expr[endlit] == '|' ||
+        expr[endlit] == ')' ||
+        expr[endlit] == '(' ||
+        expr[endlit] == '$' ||
+        expr[endlit] == '@' ||
+        expr[endlit] == '{' ||
+        expr[endlit] == '}' ||
+        expr[endlit] == '&' ||
+        expr[endlit] == '\0'))
     {
         StringExpression *ret = xcalloc(1, sizeof(StringExpression));
 
--- a/tests/acceptance/00_basics/04_bundles/bad-namespace-placement.x.cf
+++ b/tests/acceptance/00_basics/04_bundles/bad-namespace-placement.x.cf
@@ -1,0 +1,27 @@
+body common control
+{
+      inputs => { "../../default.cf.sub", "namespaced.cf.subx" };
+      bundlesequence => { test,
+                          default("$(this.promise_filename)") };
+      version => "1.0";
+}
+
+bundle agent init
+{
+}
+
+bundle agent test
+{
+  classes:
+    default:!any::
+      "ok" not => "any";
+}
+
+bundle agent check
+{
+  reports:
+    ok::
+      "$(this.promise_filename) Pass";
+    !ok::
+      "$(this.promise_filename) FAIL";
+}


worr:
All of the unit and acceptance tests pass on my machine

basvandervlies:
@worr I think ```default:!foo::``` is a valid class statement and should behave exact the same as ```!foo::```. It only prefixed by a namespace.  This should give the same result:
```
    reports:
        default:!foo::
            "namespace prefixed not foo";
        !foo::
            "not foo";
```

worr:
This is not how this is interpreted by CFEngine today, and would be a different bug.

Today, this works by doing:

```
    reports:
        !default:foo::
            "namespace prefixed not foo";
        !foo::
            "not foo";
```

Your example, on the other hand, is always false.

basvandervlies:
```foo!bar``` must be detected and is not a a valid class statement. That is because the class operators are also included in lexer to detect a a class statement.

**Comment by worr** [comment]
Why shouldn't that be a syntax error? It was a huge pain point at LinkedIn, since we discovered that we had tons of dead code that we had assumed was live. Making this a syntax error would have tripped our pre-commit hook, blocking any such change in the first place.

Since it's never a valid class statement, I don't see the value in not erroring. Please correct me if I'm missing something big.

**Comment by basvandervlies** [comment]
@worr you are right. I got the wrong impression of the namespace ;-(.  Then its the same problem as with ```foo|bar```. class operators are included to detect a class statement.

**Comment by worr** [comment]
This change still permits the use of `|`, `&` and `.` in the middle of a class statement so that you can have complex class expressions. Furthermore, it still permits variable expansions. All unit and acceptance tests pass on my machine with this change, and the entire LinkedIn codebase also works correctly. It just catches cases where `!` is embedded in the middle of a class statement, which is invalid.

**Comment by basvandervlies** [comment]
@worr I agree that it should be a syntax error. I was just explaining why. We could consider that class names can not have the class operators chars. Then we could write a statement for a valid class expression as class_name maybe followed by class_operators and so on. This is another solution that detect the error and is fine by me.

**Comment by worr** [comment]
Ah, ok. Doing this in the lexer makes a lot of sense. This seemed like the least intrusive place to make this change (I'm also not familiar with yacc, so that influenced my decision as well).

I can take a stab at implementing this in the lexer, if that's the preferred way to make such a change. The problem is that this will increase the complexity quite a bit, since like you said, class statements will now be composed of class names and class operators.

**Comment by basvandervlies** [comment]
I had to ping @jimis ;-).  From Fosdem at Ghent

**Comment by jimis** [comment]
Well the lexer seems to catch one whole class expression as a single token, so changing it there requires refactoring a big part of the lexer and the parser, I wouldn't recommend it. For reference:
```
class          [.|&!()a-zA-Z0-9_\200-\377:]+::
```

jimis:
The parser of class expressions is not beatiful at all, kudos to @worr for diving in there. Instead of leveraging yacc and lex tokenisation and recursive parser capabilities, it has it's own recursive parser implemented in `logic_expressions.c`.

However it looks like this is not the right place for the fix. I'm trying to figure out what's the proper place, if it proves too complex I'll merge the pull request.

In the meantime, @worr could you write a couple of acceptance tests that include the invalid class expressions? FYI the filenames that end in `.x.cf` are PASSing only if cf-agent fails, which should be good enough to check invalid syntax.

worr:
Acceptance test added

kacfengine:
I'm not an expert in this area, but I agree that it doesn't seem like the right place. It's too far down in the parsing, at the token level, it should be at the level that determines the next operator, perhaps at ``ParseExpression`` or somewhere around there, where the search for the next operator and token begins.

Otherwise if we add operators later, we will have to modify the token parser, which doesn't seem right at all.

basvandervlies:
@kacfengine @jimis I have to look into it, but maybe there is an solution in the lexer/parser area ;-)

jimis:
@basvandervlies be my guest, accepting "foo!bar" class is certainly a bug so lets figure out the best way to fix it!

basvandervlies:
The implementation by @worr is the quick fix.  Maybe we/I can make it "easier" to detect.  I know something about parsers ;-)

kacfengine:
Closing for now, the fix is not in the right place, and may complicate the parser at a later stage. The whole thing is quite convoluted, I have problems figuring it all out myself just by reading it, but certainly the correct place for the fix is where the operators are parsed. Two classes should not be allowed next to each other without a valid operator in between.

PR #2096, opened 2015-01-21
On a badly configured host, hostname might return "(none)", and ...

jimis:
...thos parenthesis are erring the shell, so they need to be quoted.


jimis:
This fixes ill-configured buildslaves failing spuriously.

jimis:
@tzz that's what I was asking you on IRC, thanks for helping! :-)

tzz:
cool!

jimis:
Great, merging myself because of urgency!

PR #2098, opened 2015-01-23
Fix off-by-one read.

jimis:


jimis:
@kacfengine that's the only issue address sanitizer spotted.

kacfengine:

jimis:
Thanks!

PR #2100, opened 2015-01-27
Solaris build fixes

kacfengine:
Various fixes for Solaris. This was used to build the 3.6.4 package so the commit IDs should not be altered. If fixes are needed they should be put on top.


kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
The last commit should be cherry-picked on top of #2101 before merging that.

PR #2103, opened 2015-01-31
Fix improper use of #!/bin/bash

rootwyrm:
This fixes unit test failures on *BSD environments where bash is at `/usr/local/bin/bash` or similar. (Also certain AIX and HP-UX configurations.)

### DIFF
--- a/tests/unit/dynamic_dependency_test.sh
+++ b/tests/unit/dynamic_dependency_test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Tests that the symbols in our static libraries do not occur twice in the
 # output binaries. Most platforms don't warn about this, but it has potential


tests/unit/dynamic_dependency_test.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
kacfengine: Even better, I think this can simply be switched to ``/bin/sh``, I don't see any bash specific syntax in there.

kacfengine: In fact, this is already fixed on master, just a missing cherry-pick, which I fixed here: https://github.com/cfengine/core/commit/0000c548d3c70204c68d897fa99f6468272396d0

Thanks for bringing it up though!

PR #2105, opened 2015-02-10
Remove classes, variables, bundle lmdb databases to reduce I/O and disk space usage at the clients.

maciejmrowiec:


kacfengine:
Will merge 3.6.x tomorrow after master is green.

PR #2109, opened 2015-02-16
Edit xml hang redmine6992

jimis:


jimis:
@kacfengine @pasinskim please review commit by commit.

I'm merging now though (and /possibly/ cherry-picking to 3.6.x) because we need the special package for our CI platform.

kacfengine:
"Another bug caught while reading this code." neither tells me what was fixed, nor why the fix works, so I can't really review it. Would be good with a better description. Rest looks ok.

jimis:
Its a logic error, based on the function's description
, sorry can't describe it any better. I guess I need to work a bit on
expressing myself!

PR #2110, opened 2015-02-17
Implement assertion when mixing cursors with other transactions.

kacfengine:
This has been a problem in the past that often goes unnoticed and can
lead to database corruption or hangs.


kacfengine:
For @jimis.

@basvandervlies: The result of my late night hacking after arguing with Jimis at CfgMgmtCamp! :-)

basvandervlies:
@kacfengine belgium beer did its work ;-D!

jimis:
+1, merging, that doesn't mean you won though... ;-)

kacfengine:
The battle goes on!

PR #2111, opened 2015-02-19
Fix evaluation of promises in server bundles.

pasinskim:
Normal ordering so that variables, classes, roles and user promises will
be evaluated is kept.
All promises evaluated in the same loop so that we are not pushing and
popping stack frames and destroying local promises between subsequent
passes.

### DIFF
--- a/cf-serverd/server_transform.c
+++ b/cf-serverd/server_transform.c
@@ -44,6 +44,7 @@
 #include <cf-serverd-enterprise-stubs.h>
 #include <syslog_client.h>
 #include <verify_classes.h>
+#include <verify_vars.h>
 #include <generic_agent.h> /* HashControls */
 #include <file_lib.h>      /* IsDirReal */
 #include <matching.h>      /* IsRegex */
@@ -53,7 +54,6 @@
 #include "strlist.h"
 
 
-static void KeepContextBundles(EvalContext *ctx, const Policy *policy);
 static PromiseResult KeepServerPromise(EvalContext *ctx, const Promise *pp, void *param);
 static void InstallServerAuthPath(const char *path, Auth **list, Auth **listtail);
 static void KeepServerRolePromise(EvalContext *ctx, const Promise *pp);
@@ -267,7 +267,6 @@ void KeepPromises(EvalContext *ctx, const Policy *policy, GenericAgentConfig *co
         exit(255);
     }
 
-    KeepContextBundles(ctx, policy);
     KeepControlPromises(ctx, policy, config);
     KeepPromiseBundles(ctx, policy);
 }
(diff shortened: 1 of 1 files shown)


cf-serverd/server_transform.c
```diff
@@ -543,10 +542,37 @@ static void KeepControlPromises(EvalContext *ctx, const Policy *policy, GenericA
...
+    NULL
+};
+
+/* Check if promise is NOT belonging to default server types 
+ * (see SERVER_TYPESEQUENCE)*/
+static bool IsPromiseTypeNotInServerTypeSequence(const char *promise_type)
+{
+    for (int type = 0; SERVER_TYPESEQUENCE[type] != NULL; type++)
+    {
+        if (strncmp(promise_type, SERVER_TYPESEQUENCE[type],
```
jimis: Why `strncmp()` instead of `strcmp()`?

pasinskim: I think no specific reason. Can change to strcmp as both strings are null terminated for sure.

jimis: On Thu, Feb 19, 2015 at 4:24 PM, pasinskim <notifications@github.com> wrote:

> In cf-serverd/server_transform.c
> <https://github.com/cfengine/core/pull/2111#discussion_r24995120>:
>
> > +static const char *const SERVER_TYPESEQUENCE[] =
> > +{
> > +    "vars",
> > +    "classes",
> > +    "roles",
> > +    "access",
> > +    NULL
> > +};
> > +
> > +/* Check if promise is NOT belonging to default server types
> > + * (see SERVER_TYPESEQUENCE)*/
> > +static bool IsPromiseTypeNotInServerTypeSequence(const char *promise_type)
> > +{
> > +    for (int type = 0; SERVER_TYPESEQUENCE[type] != NULL; type++)
> > +    {
> > +        if (strncmp(promise_type, SERVER_TYPESEQUENCE[type],
>
> I think no specific reason. Can change to strcmp as both strings are null
> terminated for sure.
>
Yes, because `strncmp()` has a different purpose: it compares to true if
one string is **prefix** of another. E.g.
```strncmp("123", "12345", 3);```

kacfengine:
trigger build

kacfengine:
Is there a Redmine ticket associated with this?

kacfengine:
Passed the build, so this can be merged (as of 2502a37), but if there's a Redmine ticket I'd like to attach it (for the changelog later). Is there?

pasinskim:
There is one describing part of the issue: https://dev.cfengine.com/issues/6947

kacfengine:
Thanks, merged manually!

PR #2114, opened 2015-02-20
Add: Test no errors when selecting by symlink

nickanderson:
When selecting files that are symlinks, plain files should not generate 
errors.

Ref: https://dev.cfengine.com/issues/6996


kacfengine:
trigger build

PR #2116, opened 2015-02-21
SplitString now dynamically allocates a buffer rather than relying on a ...

worr:
...4k buffer on the stack.

From #2050

### DIFF
--- a/libpromises/item_lib.c
+++ b/libpromises/item_lib.c
@@ -515,37 +515,34 @@ Item *SplitString(const char *string, char sep)
     into a linked list of separate items, */
 {
     Item *liststart = NULL;
-    const char *sp = string;
-    char before[CF_BUFSIZE];
-    int i = 0;
+    char *before = xstrdup(string);
+    char *sp = before;      /* selection pointer */
+    char *chunk = before;   /* next chunk to insert into list */
+    char *start = before;   /* start of string to free */
 
-    while (*sp != '\0')
+    while ((sp = strchr(before, sep)) != NULL)
     {
-        if (*sp != sep)
-        {
-            before[i] = *sp;
-            i++;
-        }
-        else if (sp > string && sp[-1] == '\\')
+        if (sp != before && sp[-1] == '\\')
         {
             /* Escaped use of list separator; over-write the backslash
-             * we copied last time round the loop (and don't increment
-             * i, so next time round we'll continue in the right
-             * place). */
-            before[i - 1] = sep;
+             * we copied last time round the loop
+             */
+            memmove(sp - 1, sp, strlen(sp) + 1);
+            before = sp;
+            continue;
         }
-        else
+        else if (sp != before && sp[-1] != '\\')
         {
-            before[i] = '\0';
-            PrependItem(&liststart, before, NULL);
-            i = 0;
+            *sp = '\0';
+            PrependItem(&liststart, chunk, NULL);
         }
 
-        sp++;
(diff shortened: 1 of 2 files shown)


kacfengine:
It looks like before, if you had ``string = "abc\:def"`` and ``sep = ':'``, then the resulting element would be ``abc:def``, since the backslash is skipped. But with the new code, the element will be ``abc\:def``. You should probably make sure to remove the backslash.

jimis:
I believe that if for the string `"abc\:def"` the new code will insert only `"def"` which is certainly a bug.

pasinskim:
This function seems to be deprecated since 3.6. Don't know if it's worth to spend time changing it...

worr:
@pasinskim: This function is used internally in a few different places, so it's worthwhile to update it imo.

@jimis, @kacfengine: It looks like neither assertion is correct, and it doesn't even escape the separator. I've written an acceptance test for this case, and will submit that along with my fixed code.

jimis:
@worr great, acceptance or unit test is the proper way to prove your point! Just append a commit with it and I'll merge right away, there is no other reason to hold this pull request. :-)

worr:
Welp, after taking some time to realize that `splitstring` does not actually use `SplitString`, I've fixed my bad code, and added a unit test.

tests/unit/item_test.c
```diff
@@ -110,6 +110,33 @@ static void test_list_compare(void)
...
+
+    DeleteItemList(expected);
+    DeleteItemList(actual);
+
+    actual = SplitString("foo\\:bar", ':');
+    AppendItem(&expected, "foo:bar", NULL);
+
+    result = ListsCompare(actual, expected);
+    puts(actual->name);
+    printf("%p\n", actual->next);
```
jimis: I guess the puts() and printf() calls are forgotten pieces of debugging?

jimis: The unit test fails for me!

worr:
Sorry, forgot to push the updated unit tests. These pass (for me, at least).

libpromises/item_lib.c
```diff
@@ -515,37 +515,34 @@ Item *SplitString(const char *string, char sep)
...
         }
 
-        sp++;
+        before = sp + 1;
+        chunk = sp + 1;
     }
 
-    before[i] = '\0';
-    PrependItem(&liststart, before, "");
+    PrependItem(&liststart, chunk, "");
```
jimis: @worr @kacfengine Any idea why this last `PrependItem()` call has the last argument non-NULL?

jimis:
So I was trying to merge and added a couple more testcases, since you've already started unit testing this function. But it was failing on some corner cases, so I dove into and fixed it. And then I wrote some test cases for escaping the escape character and went fully berserk and reimplemented the whole function, in order to support that. 

Have a look [at my branch](https://github.com/jimis/core/tree/worr-4k_splitstring) and let me know what you think, the progress is visible in the commit logs.

jimis:
Alright, I'm merging and on top of this I'll apply my changes. @worr thanks, especially for the unit tests that initiated looking into all corner cases!

jimis:
For reference, [here](https://github.com/cfengine/core/pull/2140) are my changes that apply on top of yours.

worr:
Yeah, the new function looks way better than what I came up with. I'm glad that it's now possible to escape the backslash when a literal `\sep` is intended.

PR #2118, opened 2015-02-23
Remove limit in SeqSort and SeqSoftSort for sorting only up to 1000 item...

maciejmrowiec:
... sequence.


maciejmrowiec:
required by https://github.com/cfengine/enterprise/pull/249

pasinskim:
Looks OK even though I am not convinced that using recursive version of quick sort is best option. But this is unrelated to the change.

kacfengine:
Cherry-picked to 3.6.x.

PR #2119, opened 2015-02-23
Systemd hardclass redmine6995

jimis:


jimis:
Tested with custom, almost empty policy, and "systemd" hard class is correctly set. 

@kacfengine this could be useful for your policy rewriting.

kacfengine:
Nice! Will definitely come in handy! Should we deprecate the one in masterfiles? I suppose old clients still need it...

kacfengine:
trigger build

kacfengine:
Cherry-picked to 3.6.x.

PR #2121, opened 2015-02-24
Pre eval 3 passes

pasinskim:
Add 3 passes in pre-evaluation for common bundles. This should help with resolving dependencies between classes and variables.


pasinskim:
Before merging this to not break enterprise builds this is needed: https://github.com/cfengine/enterprise/pull/250

pasinskim:
@jimis the change for pre-evaluation here.

jimis:
You mentioned about tests that are breaking with this change. Are they the ones you remove in this pull request?

pasinskim:
@jimis there are some in enterprise: cfengine/enterprise#250

kacfengine:
This has been shown to cause problems in real installations, and some Enterprise tests are failing as well. We might revisit this later, but close for now.

PR #2122, opened 2015-02-24
Process kill redmine6946

jimis:


jimis:
@kacfengine see the commit message. In other words I'm not sure *why* the change was needed, but it seems OK to me to ignore `pr_why` since we know process is stopped.

kacfengine:
Nice! ``pr_why`` seems ok to remove. I'll give it a go in the pipeline since the process_test was changed.

kacfengine:
trigger build

PR #2126, opened 2015-03-02
Fix systemd detection if launch path has two "systemd" substrings.

kacfengine:
For example if launched from /usr/lib/systemd/systemd.


jimis:
Confusing but I guess it should do the job, thanks for fixing!

PR #2128, opened 2015-03-03
More debug messages to detect possible blocking in cf-serverd main loop.

jimis:


kacfengine:
What about that bug that you and Eddy discovered in the agent a while ago, where the descriptor would overflow ``FD_SETSIZE``? Perhaps you should put in a check for that as well. It could happen on a heavily loaded server, couldn't it?

jimis:
IIRC here we select() only on one file descriptor, the one for the listening socket. So no risk of overflow.

kacfengine:
But the file descriptor itself may still overflow ``FD_SETSIZE`` if there are generally many open descriptors in the server, which I assume there will be under load.

kacfengine:

kacfengine:
For clarity: ``FD_SETSIZE`` is a limit on how high the file descriptor number can be, not on how many file descriptors you can have in the set.

PR #2129, opened 2015-03-06
Fix intermittent test failure in filesystem tests on HPUX.

kacfengine:
Caused by overflow if there is more than 2GB available. off_t is
verified to be only 32 bits on HPUX. There is way more than 2GB
available on our test server, so most likely this passed earlier only
because it wrapped around multiple times and got back to the positive
side.


kacfengine:
trigger build

PR #2130, opened 2015-03-10
Redmine#7031: string_mustache() function

tzz:
See https://dev.cfengine.com/issues/7031

This is a very simple function that is useful when you don't want to create an external `.mustache` file.

Test policy:

```
bundle agent main
{
  vars:
      "state" data => datastate();
      "out" string => string_mustache(state, "sys.os = {{vars.sys.os}}");
      "out2" string => string_mustache(state, "sys.os = {{vars.sys.os");
      "out3" string => string_mustache(nosuchstate, "sys.os = {{vars.sys.os}}");

  reports:
      "mustache out: $(out)";
      "mustache out: $(out2)"; # no expansion
      "mustache out: $(out3)"; # no expansion
}
```

Output:

```console
% cf-agent -KI -f test_string_mustache.cf
R: mustache out: sys.os = linux
R: mustache out: $(out2)
R: mustache out: $(out3)
warning: Broken Mustache template, could not find end delimiter after reading start delimiter at 'sys.os = {{vars.sys.os'...
```

### DIFF
--- a/tests/acceptance/01_vars/02_functions/string_mustache.cf
+++ b/tests/acceptance/01_vars/02_functions/string_mustache.cf
@@ -1,0 +1,48 @@
+body common control
+{
+      inputs => { "../../default.cf.sub" };
+      bundlesequence => { default("$(this.promise_filename)") };
+      version => "1.0";
+}
+
+bundle agent init
+{
+  vars:
+      "desert" string => "Sahara";
+}
+
+bundle agent test
+{
+  vars:
+      "state1" data => datastate();
+      "out11" string => string_mustache("desert = {{vars.init.desert}}", state1);
+      "out12" string => string_mustache("desert = {{vars.init.desert}}");
+
+      "state2" data => parsejson('{"x": 1, "y": [{"item": 2}, {"item": 3}, {"item": 4}]}');
+      "out21" string => string_mustache("desert = {{vars.init.desert}}", state2);
+      "out22" string => string_mustache("x = {{x}}", state2);
+      "out23" string => string_mustache("{{#y}}{{item}} {{/y}}", state2);
+}
+
+bundle agent check
+{
+  vars:
+      "actual" string => "
+out11 = $(test.out11)
+out12 = $(test.out12)
+out21 = $(test.out21)
+out22 = $(test.out22)
+out23 = $(test.out23)
+";
+      "expected" string => "
+out11 = desert = Sahara
+out12 = desert = Sahara
+out21 = desert = 
+out22 = x = 1
+out23 = 2 3 4 
+";
+
+  methods:
+      "" usebundle => dcs_check_strcmp($(actual), $(expected), $(this.promise_filename), "no");
+
+}
--- a/examples/string_mustache.cf
+++ b/examples/string_mustache.cf
@@ -1,0 +1,70 @@
+#  Copyright (C) Cfengine AS
+
+#  This file is part of Cfengine 3 - written and maintained by Cfengine AS.
+
+#  This program is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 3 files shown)


tzz:
If this is acceptable, I will write tests and docs for it.

nickanderson:
This is useful. Of course still bound to the 4k limit, but for many small templates this is nice.

tzz:
ping @kacfengine @jimis @estenberg ?

estenberg:
What does this add on top of parsejson()?

Also, can it be used to solve https://dev.cfengine.com/issues/7031?
Seems like this is just a part of it?

tzz:
@estenberg this is a way to apply Mustache templates in a string without having to make an intermediate `whatever.mustache` file for the template or another file for the output.  It's extremely useful for quick templating from policy.

It would solve https://dev.cfengine.com/issues/7031 but subject to the 4K limit; the user would just call `file_make(target_file, expanded_string)`.  It's not what the user proposed, but IMO the `string_mustache` approach is much more generally useful than just passing the template inline in that one use case.

estenberg:
Is file_make() a typo? I can't find this function. :-/
Functions that change the system are breaking current design..

tzz:
`file_make` is a bundle, sorry I forgot to explain! See https://docs.cfengine.com/latest/reference-standard-library-files.html#file_make

tzz:
So the actual usage may be something like

```
methods:
  "" usebundle => file_make("/etc/myfile", string_mustache(...parameters...));
```

kacfengine:
I think it looks quite useful, and I think it complements ``parsejson`` rather than trying to replace it.

However, I propose swapping the order of the arguments, and making the ``state`` argument optional. In many cases you will just want to use the ``datastate`` without any fuzz.

tzz:
OK, I pushed an untested change to make the second argument optional and if it's missing, get it from `datastate()`.

If that's acceptable, let me know and I'll finish testing, then add tests and docs.

tzz:
I added a test and fixed a minor problem.  This is ready for review.

tzz:
ping?  The merge conflicts are minor, do you need them resolved first?

kacfengine:
Yes, please! :-)

It makes the test run easier. Looks good!

tzz:
OK; rebased.

kacfengine:
trigger build

tests/acceptance/01_vars/02_functions/string_mustache.cf
```diff
@@ -0,0 +1,46 @@
...
+      "actual" string => "
+out11 = $(test.out11)
+out12 = $(test.out12)
+out21 = $(test.out21)
+out22 = $(test.out22)
+out23 = $(test.out23)
+";
+      "expected" string => "
+out11 = sys.os = linux
+out12 = sys.os = linux
```
kacfengine: Heh, I guess it should have been no surprise this failed on non-Linux platforms. You can just replace "linux" with "$(sys.os)" though, we're testing the mustache function here, not the sys variable.

tzz:
OK, acceptance test changed to use a local variable instead of `sys.os`.

kacfengine:
Thanks!

kacfengine:
trigger build

kacfengine:
Alright, all lights are green, can be merged as soon as documentation is ready!

tzz:
Example added (also included by docs)

tzz:
documentation in https://github.com/cfengine/documentation/pull/1018

kacfengine:
Thanks!

kacfengine:
trigger build

tzz:
**celebration**

PR #2131, opened 2015-03-10
Ignore HAVE_STRUCT_SOCKADDR_SA_LEN on NetBSD.

bahamat:
This has been used on NetBSD for quite some time.
See https://github.com/jsonn/pkgsrc/blob/trunk/sysutils/cfengine3/patches/patch-src_unix.c

Merging this will reduce the number of patches that need to be maintained for NetBSD in pkgsrc.

### DIFF
--- a/libenv/unix_iface.c
+++ b/libenv/unix_iface.c
@@ -54,7 +54,7 @@
 
 #ifndef __MINGW32__
 
-# ifdef HAVE_STRUCT_SOCKADDR_SA_LEN
+# if defined(HAVE_STRUCT_SOCKADDR_SA_LEN) && !defined(__NetBSD__)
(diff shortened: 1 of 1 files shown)


bahamat:
See also, an early bug report: http://gnats.netbsd.org/40092

> cfengine has trouble finding interfaces due to the way it computes the size of ifreq entries from SIOCGIFCONF.  It bases its method for the computation on "HAVE_STRUCT_SOCKADDR_SA_LEN" and "ifdef _SIZEOF_ADDR_IFREQ".

kacfengine:
I don't have the means to test on NetBSD so I'll trust you! ;-)

kacfengine:

PR #2132, opened 2015-03-11
Remove dependency that is not needed for community and enterprise.

pasinskim:
IMPORTANT: this will break rhel7 hub package so needs to be fixed
when hub package support will be added.


jimis:
This is good for 3.6 but as you mentioned in master it will probably break the hub package. Pinging @kacfengine.

PR #2138, opened 2015-03-16
define extra class for server_started when using systemd, as systemd tak...

estenberg:
...es care of this for us. Redmine #7043.


kacfengine:
+1, this is even better than what I had in mind. I have cherry-picked this to my RHEL7 branch and will merge it along with other fixes after testing it.

PR #2140, opened 2015-03-18
SplitString() now works for all corner cases.

jimis:


kacfengine:
+1 for nice tests!

kacfengine:
trigger build

libpromises/item_lib.c
```diff
@@ -510,40 +510,76 @@ void InsertAfter(Item **filestart, Item *ptr, const char *string)
...
             continue;
         }
-        else if (sp != before && sp[-1] != '\\')
+        else                   /* separator was found, and it's not escaped */
         {
-            *sp = '\0';
-            PrependItem(&liststart, chunk, NULL);
+            assert(string[z] == sep);
+            assert((z == 0) ||
+                   (z > 0 && string[z-1] != '\\'));
```
ediosyncratic: Bad assertion.
`SplitString("item\\\\:whatever", ':');` has a \ at the end of its first item but does indeed split on the : that follows the even number of backslashes.  Your assertion is going to complain about the last character before the ':' being a backslash.

libpromises/item_lib.c
```diff
@@ -510,40 +510,76 @@ void InsertAfter(Item **filestart, Item *ptr, const char *string)
     ip->classes = NULL;
 }
 
+/**
+ *  Splits a string containing a separator like ':' into a linked list of
+ *  separate items,
+ *
+ *  @NOTE backslashes can be used to escape either the separator, or the
+ *        backslash itself, e.g. "\:" or "\\" (ofcourse proper C strings need
+ *        double backslash).
```
ediosyncratic: Note that backslash cannot be used as separator.
This is unfortunate when you consider splitting a DOS-path to get a list of components.

tests/unit/item_test.c
```diff
@@ -113,41 +113,245 @@ static void test_list_compare(void)
...
+    DeleteItemList(expected); expected = NULL;
+    test_progress();
+
+    actual = SplitString("\\", ':');
+    AppendItem(&expected, "\\", NULL);
+    assert_true(ListsCompare(actual, expected));
+    DeleteItemList(actual);   actual   = NULL;
+    DeleteItemList(expected); expected = NULL;
+    test_progress();
+
```
ediosyncratic: Please include a string whose last character (after escaping) before a separator is a backslash.
Of course, that means the separator is preceded by an even number of backslashes; so the C source contains a multiple of four backslashes, followed by the separator.

jimis: @ediosyncratic I already have a test for `"blah\\\\:"`.

libpromises/item_lib.c
jimis: There is a similar test that doesn't complain (`"\\\\:blah"`). FWIW I've just run the exact string you suggest and it doesn't complain either.

tests/unit/item_test.c
ediosyncratic: Good :-)

libpromises/item_lib.c
ediosyncratic: I'm intrigued - my analysis of the code (so long ago I forget details) seemed to indicate it would complain.  The proof of the pudding is in the eating, though; so if it works where I thought it wouldn't, all's well ;-)

PR #2143, opened 2015-03-20
Use F4 as RSA exponent - it's common practice.

jimis:
Redmine 7053. Thanks to Tomáš Chvátal for submitting the patch.


kacfengine:
+1!

kacfengine:
trigger build

ncharles:
won't this break compatibility between new server and older agent ?

kacfengine:
The build is green, and it looks good from my side.

I have no opinion on @ncharles's question, I'll let @jimis address that.

jimis:
@ncharles no it shouldn't break anything, the only line that changes is in cf-key's key generation routine, the rest are tests. All functionality should remain unchanged.

I'm merging this in master, it's harmless so it might go in 3.6.x later on, please let me know if anyone finds performance issues - I believe this is the only possible regression.

PR #2144, opened 2015-03-22
Define a hard class with just the OS major version on FreeBSD

skreuzer:
On FreeBSD, VSYSNAME.release will look some like ’11.0-CURRENT’ 

When defining the $(sys.flavour) you read up to the ‘-‘ character so it will become freebsd-11.0
and that will also gets converted into the hard class ‘freebsd_11_0’

It would be helpful to also define a hard class that is just the major version such as freebsd_10 or
freebsd_11 so you wouldn’t have to do any additional work in an agent by calling regcmp() or
the like

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -1151,6 +1151,26 @@ static void OSClasses(EvalContext *ctx)
     snprintf(context, CF_BUFSIZE, "%s_%s", VSYSNAME.sysname, vbuff);
     SetFlavour(ctx, context);
 
+#ifdef __FreeBSD__
+    /*
+     * Define a hard class with just the version major number on FreeBSD
+     *
+     * For example, when being run on either FreeBSD 10.0 or 10.1 a class
+     * called freebsd_10 will be defined
+     */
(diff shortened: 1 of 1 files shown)


jimis:
All these #ifdefs are making this file unmaintainable, can you add a small comment describing what the class name is?

Otherwise I have no problem to merge, maybe it can be applied to other BSDs as well...

skreuzer:
> All these #ifdefs are making this file unmaintainable, can you add a small comment describing what the class name is?

I updated the pull request with a small comment

> Otherwise I have no problem to merge, maybe it can be applied to other BSDs as well...

Chances are this is correct but at this time I don't have access to any other BSD platform. Can you commit this as is right now and at some point I can spin up some VMs and see if this works on other BSDs?

jimis:
Thanks, merged! @kacfengine what do you think about cherry-picking to 3.6.x?

kacfengine:
I suppose this is a class that should have been there to begin with, and the fix is quite isolated, so I'm ok with cherry-picking to 3.6.

PR #2146, opened 2015-03-23
cfengine.el: Fix compilation error.

riccardomurri:
Fixes error "Eager macro-expansion failure: (void-variable quote)"
and a similar one happening while byte-compiling the file
on Emacs 24.3.1 (but possibly other versions too).

### DIFF
--- a/contrib/cfengine.el
+++ b/contrib/cfengine.el
@@ -1299,7 +1299,7 @@ see.  Use it by executing `turn-on-eldoc-mode'."
          ;; In the main syntax-table, \ is marked as a punctuation, because
          ;; of its use in DOS-style directory separators.  Here we try to
          ;; recognize the cases where \ is used as an escape inside strings.
-         (syntax-propertize-rules '("\\(\\(?:\\\\\\)+\\)\"" (1 "\\"))))
+         (syntax-propertize-rules ("\\(\\(?:\\\\\\)+\\)\"" (1 "\\"))))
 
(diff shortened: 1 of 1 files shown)


jimis:
Thanks, I'll merge once the maintainer of the file gives the green light. Pinging @tzz! :-)

tzz:
The first change is fine.

The second one looks OK but I honestly don't know why it used to work and now it doesn't.  Do you?  Maybe @leoliu knows?

Finally, please target master.  It will get backported into 3.6.x.  I'll also merge it into the Emacs master and 24.x branches.

leoliu:
@tzz,

My guess is when macros were changed to do eager expansion. So the code was incorrect but the compiler didn't detect the issue then.

`(fboundp #'XXXX)` is incorrect. `fboundp` takes symbols. So `(fboundp 'XXXX)` should be preferred.

Cheers,
Leo

tzz:
OK; @riccardomurri can you please remove the `fboundp` change and the other one is cool as long as you target master?  Thanks!

jimis:
Thanks @tzz! @riccardomurri yes please target master and we'll cherry-pick it to 3.6.x..

riccardomurri:
The bug has already been fixed in master ...

riccardomurri:
I've updated the PR and removed the `fboundp` change.

kacfengine:
Looks like this one is good to go!

kacfengine:
Thanks!

PR #2147, opened 2015-03-23
Do not use insecure MD5 but rather go with SHA256

scarabeusiv:
Simple swap MD5 with SHA256.

There is still problem with password protected PEM files. But this should be okay to merge.


jimis:
Wow, doesn't this break a bunch of things? Like the key names, or the copy_from digest verification?

kacfengine:
@scarabeusiv: I'm afraid it's not quite that simple. This would break compatibility between 3.6 and 3.7 which is unacceptable. This would require modifying the protocol to use MD5 as a fallback mechanism if the SHA256 key doesn't match, which is considerably more work.

PR #2149, opened 2015-03-25
Add new eval() option to return a class name

tzz:
This is a trivial change to `eval()` to add a second argument of `class` which:

* returns `any` if the result is not an error and not 0
* returns `!any` if the result is 0 or an error

I need it very often.

If it's accepted, I'll write docs and examples.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -5205,16 +5205,21 @@ static FnCallResult FnCallEval(EvalContext *ctx, ARG_UNUSED const Policy *policy
     char *input = RlistScalarValue(finalargs);
     char *type = RlistScalarValue(finalargs->next);
     char *options = RlistScalarValue(finalargs->next->next);
-    if (0 != strcmp(type, "math") || 0 != strcmp(options, "infix"))
-    {
-        Log(LOG_LEVEL_ERR, "Unknown %s evaluation type %s or options %s", fp->name, type, options);
-        return FnFailure();
-    }
+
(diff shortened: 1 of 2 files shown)


tzz:
Here is a use case:

```
  vars:
   "length" int => length(mylist);
   "length2" int => length(mylist2);

  classes:
    "diff_over5" expression => context_eval("abs($(length) - ($(length2)) > 5");
```

Using the built-in `isgreaterthan()` and `islessthan()` function, you'd still have to use `eval()` to get the absolute value of the difference and then find if it's greater than 5.

tzz:
ping?

kacfengine:
Looks useful.

Did you look at the possibility of overloading ``eval``? I haven't checked how difficult it is myself. Otherwise I think this is quite nice, but the name is not intuitive. Maybe ``boolean_eval`` or ``if_eval``?

tzz:
It turned out to be easiest to make the second argument a menu of `math` and `class`.  In `class` mode we just return `any` or `!any`.  Much nicer and cleaner!

tzz:
I've updated the example and PR intro accordingly!

kacfengine:
Aha, that's even nicer! This looks good to me. Will you adjust the documentation?

kacfengine:
trigger build

kacfengine:
Build is green.

tzz:
Documentation is in https://github.com/cfengine/documentation/pull/1010 thanks!

kacfengine:
Great, thanks!

PR #2150, opened 2015-03-25
Redmine#2504: allow string contexts

tzz:
This PR addresses https://dev.cfengine.com/issues/2504 to allow the syntax sugar of specifying a string as a context.  The underlying implementation simply creates an `ifvarclass` promise attribute with that string, and if such an attribute is also specified directly, the two will be appended.

Because we relay the logic to `ifvarclass`, this should have **no effect** on the CFEngine evaluation behavior.  But that needs to be thoroughly tested, so community help is warmly welcome.

Sample usage:

```
bundle agent main
{
  reports:
      "hello, this is OS $(sys.os)"
      ifvarclass => "$(sys.os)";
    any::
      "any hello";
    "any"::
      "string hello";
    "$(sys.os)"::
      "variable context hello!!!";
    "not$(sys.os)"::
      "BAD variable context hello!!!";
    "$(sys.os)"::
      "augmented variable context hello!!!"
      ifvarclass => "any";

    "$(sys.os)"::
      "BAD augmented variable context hello!!!"
      ifvarclass => "nonesuch";
}
```

Output:

```console
% cf-agent -KI -f ./test_variable_contexts.cf
R: hello, this is OS darwin
R: any hello
R: string hello
R: variable context hello!!!
R: augmented variable context hello!!!
```

Issues and concerns:

* in `policy.c` I may have a leak because of the rval item manipulation (marked with a TODO).  Needs review.
* in `promises.c` I shut up the "apparent bundle" warning if the constraint was `ifvarclass` or `if`.  That seems flimsy.
* touching these internals is always tricky and I don't know if I covered all the cases


tzz:
Acceptance test added, please review.

tzz:
I added a test from @markburgess for iterated promises.

tzz:
I added a commit to forbid variable string contexts inside bodies.  Before they were allowed but had no effect, so the body attributes ended up under `any`.  That's because body attributes can't iterate or have `ifvarclass` attributes, of course--they are evaluated in a special way.

tzz:
ping

libpromises/cf3lex.l
```diff
@@ -116,6 +116,7 @@ arrow      ->
 qstring        \"((\\(.|\n))|[^"\\])*\"|\'((\\(.|\n))|[^'\\])*\'|`[^`]*`
 
 class          [.|&!()a-zA-Z0-9_\200-\377:]+::
+varclass       (\"[^"]*\"|\'[^"]*\')::
```
ediosyncratic: The second half of this regex appears to be broken:surely it should be `\'[^']\'` excluding single quote, instead of double, between the single quote delimiters.

libpromises/policy.c
```diff
@@ -1425,6 +1430,16 @@ Constraint *PromiseAppendConstraint(Promise *pp, const char *lval, Rval rval, bo
         Constraint *old_cp = SeqAt(pp->conlist, i);
         if (strcmp(old_cp->lval, lval) == 0)
         {
+            if (strcmp(old_cp->lval, "ifvarclass") == 0 ||
+                strcmp(old_cp->lval, "if") == 0)
+            {
+                Buffer *grow = BufferNew();
+                BufferAppendF(grow, "(%s).", old_cp->rval);
+                BufferAppend(grow, RvalScalarValue(rval), strlen(RvalScalarValue(rval)));
```
ediosyncratic: Why doesn't rval also need parentheses round it, the same as `old_cp->rval` ?

libpromises/promises.c
```diff
@@ -170,7 +170,7 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
                         "Apparent body '%s' was undeclared or could "
                         "have incorrect args, but used in a promise near "
                         "line %zu of %s (possible unquoted literal value)",
-                        RvalScalarValue(cp->rval), pp->offset.line,
+                        RvalScalarValue(cp->rval),pp->offset.line,
```
ediosyncratic: Why have you removed a space here ?
It's required by out coding standards ...

libpromises/policy.c
tzz: Because there's no ambiguity: `(x).y` is the same as `(x).(y)` for any `y`.  But I changed it anyway to be clear :)

libpromises/cf3lex.l
tzz: Right, fixed.

libpromises/promises.c
tzz: Accident, sorry... fixed

tzz:
ping?

libpromises/policy.c
```diff
@@ -1425,6 +1430,17 @@ Constraint *PromiseAppendConstraint(Promise *pp, const char *lval, Rval rval, bo
...
         if (strcmp(old_cp->lval, lval) == 0)
         {
+            if (strcmp(old_cp->lval, "ifvarclass") == 0 ||
+                strcmp(old_cp->lval, "if") == 0)
+            {
+                Buffer *grow = BufferNew();
+                BufferAppendF(grow, "(%s).(", old_cp->rval);
+                BufferAppend(grow, RvalScalarValue(rval), strlen(RvalScalarValue(rval)));
+                BufferAppendChar(grow, ')');
+                rval.item = xstrdup(BufferData(grow)); // TODO: will this leak?
```
kacfengine: Yes, I think so. And looking at how ``Rval`` works I wouldn't even expect this to work correctly. ``Rval`` is passed by value, so if you change the ``rval`` variable here, the ``cp`` that you created further up should still have the old ``Rval``. On the other hand I might be confused, does this work for you?

As for the memory leak, what you can do is the following:

<pre>
...
BufferAppendChar(grow, ')');
RvalDestroy(rval);
rval.item = BufferClose(grow); // Avoids xstrdup.
</pre>

tzz: Right, OK.  Besides destroying `cp->rval` we also need to make a new rval, see my new commit which I believe does the right thing.

kacfengine:
trigger build

kacfengine:
Some incompatibilities with Enterprise. Fixed and rerunning tests on a private branch.

kacfengine:
Merged manually.

I messed up the merge commit, so it's going to say "canonical-json", but it is really string contexts. Too many PRs in the air... :-P

tzz:
Thank you!  Do you want a documentation patch?

kacfengine:
Yes, please!

tzz:
OK, see https://github.com/cfengine/documentation/pull/1020

PR #2152, opened 2015-03-27
Redmine#1815: libcurl integration, curl() function

tzz:
See https://dev.cfengine.com/issues/6480

This implements the function `url_get(URL, OPTIONS)`.

Usage:
```
bundle agent main
{
  vars:
      "options_str" string => '
{
  "url.max_content": 512,
  "url.verbose": 0,
  "url.headers": [ "Foo: bar" ]
}';
      "options" data => parsejson($(options_str));
      "url" string => "http://cfengine.com";
      "res" data => url_get($(url), options);
      "out" string => format("%S", res);

      "url2" string => "http://nosuchcfenginehost.com";
      "res2" data => url_get($(url2), options);
      "out2" string => format("%S", res2);

  reports:
      "$(this.bundle): from $(url) with options $(options_str) we got $(out)";
      "$(this.bundle): from $(url2) with options $(options_str) we got $(out2)";
}
```

output:

```console
R: main: from http://cfengine.com with options 
{
  "cfengine.max_content": 512,
  "curl.verbose": 0,
  "curl.headers": [ "Foo: bar" ]
} we got {"returncode":200,"rc":0,"success":true,"content":"\n<!DOCTYPE html>\n<!--[if lt IE 7]>\n<html class=\"no-js lt-ie9 lt-ie8 lt-ie7\" lang=\"en-US\" prefix=\"og: http://ogp.me/ns#\"> <![endif]-->\n<!--[if IE 7]>\n<html class=\"no-js lt-ie9 lt-ie8\" lang=\"en-US\" prefix=\"og: http://ogp.me/ns#\"> <![endif]-->\n<!--[if IE 8]>\n<html class=\"no-js lt-ie9\" lang=\"en-US\" prefix=\"og: http://ogp.me/ns#\"> <![endif]-->\n<!--[if gt IE 8]><!-->\n<html class=\"no-js\" lang=\"en-US\" prefix=\"og: http://ogp.me/ns#\"> <!--<![endif]-->\n<head>\n\n    \n    <meta charset=\"utf-8\">\n\n    <title>\n        CFEng","headers":"HTTP/1.1 200 OK\r\nDate: Fri, 27 Mar 2015 18:13:01 GMT\r\nServer: Apache\r\nX-Powered-By: PHP/5.3.3\r\nX-Pingback: http://cfengine.com/xmlrpc.php\r\nConnection: close\r\nTransfer-Encoding: chunked\r\nContent-Type: text/html; charset=UTF-8\r\n\r\n"}

R: main: from http://nosuchcfenginehost.com with options 
{
  "cfengine.max_content": 512,
  "curl.verbose": 0,
  "curl.headers": [ "Foo: bar" ]
} we got {"returncode":0,"rc":6,"success":false,"content":"","headers":""}
```

There are several things to note:

* the options are mostly simple string keys, except for the headers.  When they begin with `curl` they are passed straight on.  When they begin with `cfengine` they control CFEngine's behavior.
* caching is implemented with a simple global cache. The normal caching won't work for this function.  The cache uses the URL plus the JSON dump of the options as a key.
* we never clean up the global resources on exit.  That should happen on agent exit, I think.  But then the initialization should happen on startup too?
* right now it uses signals in `SIGALRM` but we can turn that off
* default timeout is 3 seconds which is pretty short
* the headers are returned as a big string but could be an array... I can't think of a use for that right now.
* no `PUT` and `POST`, no SSL right now
* it could support writing to a file but currently doesn't, and wouldn't be a probe if it did


tzz:
I can write docs and attempt tests for this.  The tests may be tricky...

ncharles:
Awesome, thank you Ted !

tzz:
ping

libpromises/evalfunction.c
```diff
@@ -1668,6 +1673,193 @@ static FnCallResult FnCallSplayClass(EvalContext *ctx,
...
+
+static FnCallResult FnCallCurl(ARG_UNUSED EvalContext *ctx,
+                               ARG_UNUSED const Policy *policy,
+                               ARG_UNUSED const FnCall *fp,
+                               const Rlist *finalargs)
+{
+#ifdef HAVE_LIBCURL
+    if (NULL == CURL_CACHE)
+    {
+        CURL_CACHE = JsonObjectCreate(10);
```
ediosyncratic: Is there a nice way to tell valgrind to ignore the leak-on-exit implied by this ?

libpromises/evalfunction.c
```diff
@@ -1668,6 +1673,193 @@ static FnCallResult FnCallSplayClass(EvalContext *ctx,
...
+    if (NULL != old_result)
+    {
+        Log(LOG_LEVEL_VERBOSE, "%s: found cached request for %s", fp->name, url);
+        WriterClose(cache_w);
+        return (FnCallResult) { FNCALL_SUCCESS, (Rval) { JsonCopy(old_result), RVAL_TYPE_CONTAINER } };
+    }
+
+    if (!CURL_INITIALIZED && curl_global_init(CURL_GLOBAL_DEFAULT) != 0)
+    {
+        Log(LOG_LEVEL_ERR, "%s: libcurl initialization failed, sorry", fp->name);
```
ediosyncratic: Is there no value to reporting what curl_global_init() returned ?

tzz: Nah, it's extremely unlikely to happen, like `malloc` fails and such.

libpromises/evalfunction.c
tzz: Sorry, I don't know :)

libpromises/evalfunction.c
```diff
@@ -1668,6 +1673,193 @@ static FnCallResult FnCallSplayClass(EvalContext *ctx,
...
+
+        WriterClose(cache_w);
+        return FnFailure();
+    }
+
+    //TODO: register curl_global_cleanup() to run on agent exit
```
ediosyncratic: Ideally, you'd wrap curl_global_cleanup() in a local function that also frees the cache; use that as tidy-up.
However ...

Your curl activity is currently:
* initialise cache if NULL
* (check argument sanity)
* check for cached answer, use if present
* initialise curl if not already initialised
* set initialised flag [and register tidy-up]

This means that, if curl_global_init fails, you'll still create the cache, that you'll never use [and you'll leak it, even if your tidy-up function would have cleared it away].
That could be avoided by changing the order to:
* (check argument sanity)
* if curl not initialised { initialise it; on success set CURL_INITIALIZED, register tidy-up }
* if no cach, initialise it
* check for cached value, use it.

I don't see any loss of efficiency here, and it'd ensure the tidy-up can also release the cache.
Not urgent, though, as you don't yet have the tidy-up in place ;-)

libpromises/evalfunction.c
```diff
@@ -1668,6 +1673,193 @@ static FnCallResult FnCallSplayClass(EvalContext *ctx,
...
+        Log(LOG_LEVEL_ERR, "%s: libcurl easy_init failed, sorry", fp->name);
+
+        WriterClose(cache_w);
+        return FnFailure();
+    }
+
+    Buffer *content = BufferNew();
+    Buffer *headers = BufferNew();
+    curl_easy_setopt(curl, CURLOPT_URL, url);
+    // TODO: this uses SIGALRM, is that OK?
```
ediosyncratic: The default action for SIGALRM is termination.
We don't generally catch SIGALRM; in one place we do transiently (and set to SIG_DFL after; this should probably restore whatever signal said was there before, instead); in another place we do catch it but the handler calls alarm(0) as its first action (which may well be non-portable, as some OSen shall block the resulting recursive signal until the handler returns, others shall terminate before running the handler's body).

So: is it OK for whatever's processing this function call to terminate on time-out ?

libpromises/evalfunction.c
```diff
@@ -1668,6 +1673,193 @@ static FnCallResult FnCallSplayClass(EvalContext *ctx,
...
+            max_headers = IntFromString(value);
+        }
+    }
+
+    JsonElement *options_headers = JsonObjectGetAsArray(options, "curl.headers");
+    iter = JsonIteratorInit(options_headers);
+    struct curl_slist *header_list = NULL;
+    while ((e = JsonIteratorNextValueByType(&iter, JSON_ELEMENT_TYPE_PRIMITIVE, true)))
+    {
+        header_list = curl_slist_append(header_list, JsonPrimitiveGetAsString(e));
```
ediosyncratic: Is curl_slist_append() an O(1) operation, or an O(length(header_list)) operation ?
If the latter (making this loop O(length**2) overall), consider a prepend instead (assuming that's O(1)); if order matters a lot, hopefully a curl_slist_reverse() is available; this should be O(length), so calling it once at the end leaves the whole operation O(length) instead of O(length**2).

libpromises/evalfunction.c
ediosyncratic: Less important if you can arrange (see below) for tidy-up to make this not a leak.

libpromises/evalfunction.c

libpromises/evalfunction.c
tzz: I've turned off the use of signals with a new commit.

libpromises/evalfunction.c
tzz: I wouldn't worry about it.  This is a user-provided list of headers, not megabytes of data.

tzz:
Rebased and pushed again.

kacfengine:
I'm not sure I see the benefit of having this in core, since this is so easy to achieve using the curl tool itself. Are there any problematic cases where a simple ``curl $(options) -O $(output) $(url)`` wouldn't do the trick? It could even be wrapped in a bundle.

tzz:
Oh, definitely, there are many advantages to the function:

* timeouts
* proper caching
* limiting the length
* controlling the HTTP and HTTP/S protocol in detail
* passing complex data without going through strings or files
* proving the request success, the actual HTTP code, and the returned content and headers in a data container.  This is much harder with `curl` from the command line.
* no need to call an external binary, better security and reliability
* much easier to use and automate for a beginner, for all the above reasons plus having to learn all the command-line `curl` options

tzz:
Rebased so it can merge cleanly.

estenberg:
Thanks for this!
We should name the function after what it does, not the library it uses, though.

So if you could rename it from curl() to httpget(), we should be able to merge this for 3.7.

kacfengine:
Alright, you've got some good points. But incorporating this into the package as a dependency is a bit of work and we've already got a tight schedule, so this will only be considered as a "community can compile it themselves" kind of thing for 3.7. Then we can create a ticket to get the dependency in for 3.8.

I agree with @estenberg about the renaming, though I would use an underscore, ``http_get``.

libpromises/evalfunction.c
```diff
@@ -1674,6 +1680,211 @@ static FnCallResult FnCallSplayClass(EvalContext *ctx,
...
+        return FnFailure();
+    }
+
+    JsonElement *options = VarRefValueToJson(ctx, fp, ref, NULL, 0);
+    VarRefDestroy(ref);
+
+    if (JsonGetElementType(options) != JSON_ELEMENT_TYPE_CONTAINER ||
+        JsonGetContainerType(options) != JSON_CONTAINER_TYPE_OBJECT)
+    {
+        return FnFailure();
```
kacfengine: Will ``options`` leak here?

kacfengine: On all returns, in fact.

libpromises/evalfunction.c
```diff
@@ -1674,6 +1680,211 @@ static FnCallResult FnCallSplayClass(EvalContext *ctx,
...
+    size_t max_content = 4096;
+    size_t max_headers = 4096;
+    JsonIterator iter = JsonIteratorInit(options);
+    const JsonElement *e;
+    while ((e = JsonIteratorNextValueByType(&iter, JSON_ELEMENT_TYPE_PRIMITIVE, true)))
+    {
+        const char *key = JsonIteratorCurrentKey(&iter);
+        const char *value = JsonPrimitiveGetAsString(e);
+
+        if (0 == strcmp(key, "curl.timeout"))
```
kacfengine: Like the function name, I don't see anything curl specific about these options, so a name like ``http.timeout`` might be better. OTOH, I see that other protocols are allowed too, so maybe something like ``web_get`` and ``web.timeout``? I'm open to suggestions. :-)

libpromises/evalfunction.c
```diff
@@ -1674,6 +1680,211 @@ static FnCallResult FnCallSplayClass(EvalContext *ctx,
...
+    curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &returncode);
+    JsonObjectAppendInteger(result, "returncode",  returncode);
+
+    curl_easy_cleanup(curl);
+
+    JsonObjectAppendInteger(result, "rc",  0+res);
+
+    JsonObjectAppendBool(result, "success",  res == 0);
+
+    BufferTrimToMaxLength(content, max_content);
```
kacfengine: One problem I see here is that the download is not bounded while downloading, but rather trimmed afterwards, which means you can easily download a 10GB file straight into memory, and only truncating that to 4KB afterwards. This would first of all be incredibly inefficient, but most importantly dangerous for memory consumption.

ediosyncratic: HTTP, at least, supports GETting ranges within a resource, so there's some hope that libcurl might support saying you want the first 4kB of the resource, rather than all of it.  See:
http://tools.ietf.org/html/rfc2616#section-14.35
with supporting details in other sections of the RFC.

tzz:
I will rename it to `url_get` because it will handle many URL types like `https` and `ftp`.

libpromises/evalfunction.c
tzz: Fixed all the leaks.

libpromises/evalfunction.c
tzz: I used `url.timeout` and all the parameters have `url` now (no more `cfengine.xyz`).

libpromises/evalfunction.c
tzz: We can't control the chunking that libcurl does, but we can refuse to accept its data.  If the write function returns 0, as it will when we're over the limit, libcurl will take the hint.

Unfortunately if we limit the size of the headers, libcurl assumes it can't write the content **at all**.  I think that's not fixable on our side.

tzz:
* rebased and squashed in a single commit
* options are destroyed before every exit
* enforce max content and header earlier
* options renamed for clarity
* function renamed to `url_get`

Ready for another review, then I can add acceptance test if we're in agreement on naming and options.

Also I'd like to stretch for HTTP/S requests for 3.7, if there's time.  Let me know if you think so, or we can leave it for a later patch 3.7.x release.

tzz:
ping?

estenberg:
Looks good to me.

kacfengine:
We should get going on this one again, there's still plenty of time before 3.8, but better to be early. It needs rebasing, but I haven't checked what the conflicts are. Anything else you'd like to do here before I start reviewing?

Agree with ``url_get`` name btw.

kacfengine:
Btw, can you change the Redmine number in the commit messages to #6480? The original number is a private ticket number.

tzz:
OK, rebased and compiles cleanly.  Testing libcurl is tricky, like testing `readtcp`.  Any suggestions?  I can test `file:///` resources but that's not going to exercise the network layer or talking to a real web server.

kacfengine:
Thanks!

Indeed, testing libcurl is tricky. A ``file:///`` test would still be useful, at least we test that we are using the curl API correctly.

Otherwise, I'm inclined to leave network tests out for this one, given that they're quite hard to implement, and that this isn't the core of the product (if a bad bug turns up after all, you can always recover as long as the policy can be updated, IOW you won't break your installation). I welcome input from @estenberg and @jimis though. If you have good ideas for testing this, I'm all ears.

tzz:
@kacfengine I added a test using some fairly static websites.

tzz:
I added tests and updated comments.  Ready for review!

libpromises/evalfunction.c
```diff
@@ -1695,6 +1701,265 @@ static FnCallResult FnCallSplayClass(EvalContext *ctx,
...
+
+    Writer *cache_w = StringWriter();
+    WriterWriteF(cache_w, "url = %s; options = ", url);
+    JsonWriteCompact(cache_w, options);
+
+    if (NULL == CURL_CACHE)
+    {
+        CURL_CACHE = JsonObjectCreate(10);
+    }
+    atexit(&CurlCleanup);
```
kacfengine: This should be called inside the if to avoid registering it every time someone calls ``url_get``, or?

libpromises/evalfunction.c
```diff
@@ -1695,6 +1701,265 @@ static FnCallResult FnCallSplayClass(EvalContext *ctx,
...
+
+    JsonObjectAppendObject(CURL_CACHE, StringWriterData(cache_w), JsonCopy(result));
+    WriterClose(cache_w);
+
+    JsonDestroy(options);
+    return (FnCallResult) { FNCALL_SUCCESS, (Rval) { result, RVAL_TYPE_CONTAINER } };
+#else
+    Log(LOG_LEVEL_INFORM,
+        "%s: libcurl integration is not compiled into CFEngine, sorry", fp->name);
+    return FnFailure();
```
kacfengine: FYI, the correct log level is ``LOG_LEVEL_INFO``, but in any case, I think ``LOG_LEVEL_ERR`` is more appropriate here, because the user tried to do something that can't be done.

kacfengine:
Thanks! Nice work on the test.

A couple of comments:

* Do you know what the policy of the ``ajax.googleapis.com`` page is in terms of automated querying? I'm a bit worried about being blacklisted because once our tests start running they are called *a lot*, and this has in fact happened before. Using ``cfengine.com`` is smart though, then we have only ourselves to blame. :-)
* How do you feel about moving the ``max_content`` element out of the JSON and into its own function argument? It seems it would be extremely common to use, and it would be inconvenient to specify it inside JSON every time, esp. since you have to merge it with the rest of the options.

kacfengine:
FYI, I will start looking at building libcurl as part of our package. This is a bit of work, so this PR will be my main focus right now.

libpromises/evalfunction.c
tzz: Yeah, it's fine either way.  Adjusted.

libpromises/evalfunction.c
tzz: Agreed on both items, fixed, thanks.

tzz:
I think it's OK to query that URL a lot, it's a CDN built for open use.  But you can ask Google if you're concerned.  Worst case, our tests start failing.

@estenberg recommended for `readdata()` to eliminate the size from the parameters and I agreed then.  I think we should be consistent about it and keep it out from the parameters here too.  We could add a global setting for it I guess?

I don't know if `max_content` will be frequently needed.  When would you want to use it?  For testing it's convenient but generally I think you want the whole document.

kacfengine:
> I don't know if ``max_content`` will be frequently needed. When would you want to use it? For testing it's convenient but generally I think you want the whole document.

Well, this is all downloaded into memory, so limiting it seems important to prevent a rogue web server from exhausting the memory on all your hosts with a large file. But maybe you're right that it's better to specify one sensible limit in the JSON and then use that everywhere. I'm fine with that as well.

basvandervlies:
@kacfengine if a limit is specified it must be overridable via een config setting.  No hard coded boundaries.

kacfengine:
Yes, I didn't mean there will be one global limit. I just mean that with the current spec it is not possible to specify the limit without merging it together with the rest of the options in the JSON container, which makes it awkward to specify if you need to change it often. But most likely you don't need to, so it's ok to specify the JSON once, and then you use it everywhere.

Just to be clear it's still *possible* to specify the limit, you just have to construct a container containing it first.

basvandervlies:
@kacfengine thanks for the clarification

tzz:
So we're going with the current scheme?  OK.

The main problem is really that CFEngine functions lack named parameters with defaults, so you always end up adding optional or required positional parameters.  But that's a common problem in many languages... there's no easy fix.

kacfengine:
I'm currently in the process of building libcurl so that it's included in the package. Some platforms are (as usual) being difficult though, so might take some more time.

In the meantime, I was thinking we should probably add a ``feature_url_get`` or something, like we did for YAML, so that it can be selected during execution? If the exotic platforms turn out to be too difficult, I may end up disabling curl there. What do you think?

tzz:
OK; I added the `feature_url_get` and added a new `library=xyz` informational tag to the feature hard classes.

kacfengine:
Thanks!

kacfengine:
Do you have docs for this, btw? :-)

tzz:
I'll write docs last as usual, when you're ready to merge.

kacfengine:
Ok! :-)

kacfengine:
Alright, *finally* I was able to integrate libcurl, it was surprisingly tricky, but at least I solved some other problems along the way. I made some additional modifications to this PR, so it is superseded by https://github.com/cfengine/core/pull/2352. In addition the library integration here: https://github.com/cfengine/buildscripts/pull/89.

All lights are green, so if you can provide the documentation, we can merge this ASAP!

tzz:
docs in https://github.com/cfengine/documentation/pull/1191

you can close this, I think

kacfengine:
Thanks for a job well done, Ted! Merged in the other PR.

PR #2153, opened 2015-03-28
Redmine#4309: implement common body attribute inherit_from for body inheritance

tzz:
After #2088 was merged thanks to @worr I realized it could be used to implement a universal `inherit_from` slist body attribute.

Example:

```
bundle agent main
{
  reports:
      "def" classes => classes_generic_2("def");

    def_ok::
      "def was OK!";
}

body classes classes_generic(x)
# @brief Define `x` prefixed/suffixed with promise outcome
# @param x The unique part of the classes to be defined
{
      promise_repaired => { "promise_repaired_$(x)", "$(x)_repaired", "$(x)_ok", "$(x)_reached" };
      repair_failed => { "repair_failed_$(x)", "$(x)_failed", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      repair_denied => { "repair_denied_$(x)", "$(x)_denied", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      repair_timeout => { "repair_timeout_$(x)", "$(x)_timeout", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      promise_kept => { "promise_kept_$(x)", "$(x)_kept", "$(x)_ok", "$(x)_not_repaired", "$(x)_reached" };
}

body classes classes_generic_2(x)
{
      inherit_from => { classes_generic };
}
```

Output:

```console
R: def
R: def was OK!
```

The key to making this happen was that body attributes are actually appended to the promise, so CFEngine doesn't keep track of bodies per se.  So to implement inheritance, we just have to walk through a list of parent bodies and insert their attributes verbatim.  Thus this is a simple splice, **without** evaluation implications.  The only requirement is that derived bodies **must** use the same parameter names as their parents, if the parameter names are in the attributes.  In other words, if the parent is `classes_generic` as in the above example, then the child has to have the parameter `x` or the parent's attributes won't work.

Duplicate attributes are overwritten by the child.

Acceptance test is included.  If this is accepted, I'll add docs as well.


worr:
This is awesome to see. :+1:

tzz:
ping

tzz:
ping?

kacfengine:
I like that this is a quite clean design. However, this opens up several corner cases, and takes the language in a slightly new direction, so we need to make sure we have thought of everything here.

1. I'm not sure about using a list of bodies to inherit from. This is mostly a gut feeling from programming languages; multiple inheritance tends to lead to problems and are really hard to follow when reading. For example, what to do if an attribute is defined in two bodies, both inherited from?
2. How many levels of inheritance should we support? I haven't looked closely enough at the code to determine how many you have provided, but regardless I'm tempted to say that we should limit the inheritance to one level. Readability suffers for every level and I have a hard time seeing use for more than one.
3. I'd like example use cases of the body inheritance, to verify that this is actually worth it. Package bodies are not good examples, since we are replacing those anyway.
4. I don't think that the mandatory and magically passed parameter to the inherited body is a good solution, this is an implementation detail which should not leak into the language. The expectation from users would certainly be that parameters are local to the particular body, and any parameters passed to the base body would have to be passed explicitly in the ``inherit_from`` attribute. So something like this:
```
body classes classes_generic(x)
# @brief Define `x` prefixed/suffixed with promise outcome
# @param x The unique part of the classes to be defined
{
      promise_repaired => { "promise_repaired_$(x)", "$(x)_repaired", "$(x)_ok", "$(x)_reached" };
      repair_failed => { "repair_failed_$(x)", "$(x)_failed", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      repair_denied => { "repair_denied_$(x)", "$(x)_denied", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      repair_timeout => { "repair_timeout_$(x)", "$(x)_timeout", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      promise_kept => { "promise_kept_$(x)", "$(x)_kept", "$(x)_ok", "$(x)_not_repaired", "$(x)_reached" };
}

body classes classes_generic_2(x)
{
      inherit_from => classes_generic($(x));
}
```

@estenberg: Your comments welcome as well!

tzz:
* I think it's OK to inherit from multiples since the inheritance here is just a copy, not a true inheritance (we don't resolve the parents at runtime!).  But if you really dislike it, I'll adjust.
* similarly to the above, any level of inheritance is just copying, so I think it's OK as is.  It's like calling a bundle from another bundle--readability suffers but it's really convenient.
* I'd like @nickanderson to help with the demo of the package bodies.  Although there are **many** use cases for this, I'd like to point out that `classes_generic` and `scoped_classes_generic`, both incredibly useful, have almost the same contents but for one attribute:

```
body classes scoped_classes_generic(scope, x)
# @brief Define `x` prefixed/suffixed with promise outcome
# **See also:** `scope`
#
# @param scope The scope in which the class should be defined
# @param x The unique part of the classes to be defined
{
      scope => "$(scope)";
      promise_repaired => { "promise_repaired_$(x)", "$(x)_repaired", "$(x)_ok", "$(x)_reached" };
      repair_failed => { "repair_failed_$(x)", "$(x)_failed", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      repair_denied => { "repair_denied_$(x)", "$(x)_denied", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      repair_timeout => { "repair_timeout_$(x)", "$(x)_timeout", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      promise_kept => { "promise_kept_$(x)", "$(x)_kept", "$(x)_ok", "$(x)_not_repaired", "$(x)_reached" };
}

body classes classes_generic(x)
# Define x prefixed/suffixed with promise outcome
{
      promise_repaired => { "promise_repaired_$(x)", "$(x)_repaired", "$(x)_ok", "$(x)_reached" };
      repair_failed => { "repair_failed_$(x)", "$(x)_failed", "$(x)_not_ok", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      repair_denied => { "repair_denied_$(x)", "$(x)_denied", "$(x)_not_ok", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      repair_timeout => { "repair_timeout_$(x)", "$(x)_timeout", "$(x)_not_ok", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
      promise_kept => { "promise_kept_$(x)", "$(x)_kept", "$(x)_ok", "$(x)_not_repaired", "$(x)_reached" };
}
```

* rewriting the parameters will IMO be much harder than asking users to match them (in effect forcing them to use an API).  But if you feel strongly about this, I can try to implement it.

kacfengine:
> I'd like @nickanderson to help with the demo of the package bodies. Although there are many use cases for this, I'd like to point out that classes_generic and scoped_classes_generic, both incredibly useful, have almost the same contents but for one attribute:

I'd still like more concrete use cases for this. As I indicated, package bodies are not a good example because that body type is going away entirely. And what you have displayed is just two bodies, both part of our own standard library, so it's rarely seen by an SA, just used.

My question boils down to: How would an SA use this? When would he benefit from this?

tzz:
Well, the primary use of this is in libraries to avoid duplication.  It's essential in order to provide flexible bodies without duplicating attributes.  So the end users benefit less than library authors.

For an end user, bodies are rarely written, but here's an easy example.  The following two bodies in the stdlib differ by one attribute!

```
body edit_field col(split,col,newval,method)
# @brief Edit tabluar data with comma-separated sub-values
# @param split The separator that defines columns
# @param col The (1-based) index of the value to change
# @param newval The new value
# @param method The method by which to edit the field
{
      field_separator    => "$(split)";
      select_field       => "$(col)";
      value_separator    => ",";
      field_value        => "$(newval)";
      field_operation    => "$(method)";
      extend_fields      => "true";
      allow_blank_fields => "true";
}

##

body edit_field line(split,col,newval,method)
# @brief Edit tabular data with space-separated sub-values
# @param split The separator that defines columns
# @param col The (1-based) index of the value to change
# @param newval The new value
# @param method The method by which to edit the field
{
      field_separator    => "$(split)";
      select_field       => "$(col)";
      value_separator    => " ";
      field_value        => "$(newval)";
      field_operation    => "$(method)";
      extend_fields      => "true";
      allow_blank_fields => "true";
}
```

OK, so in this case we could make it a body parameter.  Let's see... how about `copy_from` bodies?

```
body copy_from secure_cp(from,server)
# @brief Download a file from a remote server over an encrypted channel
#
# Only copy the file if it is different from the local copy, and verify
# that the copy is correct.
# 
# @param from The location of the file on the remote server
# @param server The hostname or IP of the server from which to download
{
      source      => "$(from)";
      servers     => { "$(server)" };
      compare     => "digest";
      encrypt     => "true";
      verify      => "true";
}

body copy_from remote_dcp(from,server)
# @brief Download a file from a remote server if it is different from the local copy.
#
# @param from The location of the file on the remote server
# @param server The hostname or IP of the server from which to download
{
      servers     => { "$(server)" };
      source      => "$(from)";
      compare     => "digest";
}
```

You can see clearly here than `secure_cp` is simply a copy of `remote_dcp` but with encryption and verification.  You can find similar examples in `perms` and `changes` bodies among others.

kacfengine:
As tempting as it is to encourage features that benefit us, I'm reluctant to do so if it doesn't also benefit the user. And in this case it exposes a language feature in an already difficult language that actually makes the policy less readable for the user. I expect that even though the user may not change our internal masterfiles very often, he may read them from time to time, and this will not help.

So you know where I stand, but then again, I'm not the primary one maintaining masterfiles here. I'd really like input from @nickanderson and @estenberg on this. Do you think this is worth the extra language complexity?

tzz:
Would it help if I cited the dozens of requests for body inheritance from the mailing list?  I've been convinced we need this for years...

As for less readable, I disagree.  It's like saying that class inheritance makes C++ code less readable.  It hides information, yes, but you (the reader) don't want to see **all** the information, only the delta from a base class that you know.  Otherwise you (the reader) have to visually and mentally check all the duplicate information, which is exhausting and very error-prone.

worr:
tbh, this would have benefited us quite a bit at LinkedIn quite a bit. A feature like this would really have eliminated a lot of duplication in our own stdlib. That's why I was so excited to see this change proposed.

kacfengine:
> Would it help if I cited the dozens of requests for body inheritance from the mailing list? I've been convinced we need this for years...

@tzz: It might. It would be good to hear why people are requesting this, especially if it's outside of masterfiles maintenance.

> tbh, this would have benefited us quite a bit at LinkedIn quite a bit. A feature like this would really have eliminated a lot of duplication in our own stdlib.

@worr: That's interesting. Why do you need to maintain your own stdlib?

tzz:
@kacfengine I found a few threads with support at https://groups.google.com/forum/#!searchin/help-cfengine/body$20inheritance but didn't count all the supporters.  I am pretty sure Normation is strongly in favor, for instance.

In addition I recall 5-6 times body inheritance was suggested by trainees during my training sessions (together with defining bodies inline, which I'll avoid discussing here :)

worr:
@kacfengine: We have some 1k lines of internal bodies that we use that manage LinkedIn-specific behavior. A lot of it are different `package_method` bodies we've implemented that we used to work around bugs in masterfiles (I think I submitted a fix for one of these upstream some time ago).

Some of it is, admittedly, due to the fact that we haven't tracked masterfiles closely at all.

Another common use are `contains` bodies that contain information about linkedin-specific site information. We also have some `edit_line` bundles that modify LinkedIn-specific files in certain ways.

Most of this stuff doesn't make a lot of sense to submit upstream. Some of it could be reimplemented with body parameters, rather than hard-coding information. That said, body inheritance would reduce the cases of repeating ourselves.

kacfengine:
> Another common use are ``contains`` bodies that contain information about linkedin-specific site information.

This is the most relevant use case I've heard so far, but I have a hard time believing this can be many lines?

It's pretty clear from the given answers that body inheritance would primarily benefit "most upstream" maintainers of a policy set, IOW those that maintain some kind of standard library for others to use, whether it is just internally or through the official masterfiles. In that sense my position hasn't changed much; I still believe that readability for "the downstream" policy maintainers is the most important.

But you do have some valid points, and I don't want to be the only CFEngine voice in this decision, so what I'm going to do is to take this up with the whole group in the planning meeting we have tomorrow, and get a final decision on whether we want this feature.

basvandervlies:
@kacfengine it is like the include all the bundle files option, that is now applied and make the life easier for some people.  I feel this option has the same potential. You do not have to use it for your masterfiles its an option for people that write their own libs, the "power users". we also have our own SURFsara library and the feature will really be used in this lib.

kacfengine:
We talked this over in our planning meeting, and although the table was a little divided there as well, we eventually agreed that we'll accept the inheritance feature. However, we want to keep it as simple and intuitive as possible, which means in particular:

* Body parameters must be independent (each body has to pass parameters explicitly to the next body if it needs to, just like methods and bundles would do). See also point 4 in my first comment to this pull request.
* No multiple inheritance (each body can only inherit from one other body).
* Only one level of inheritance (the body inherited from cannot itself inherit from somewhere else).

Our thinking with the last point is that this restriction can easily be relaxed later if we see that actual usage demands it, but it's harder to go the other way.

tzz:
I will try to implement rewriting the body parameters on inheritance.  I can also restrict to single inheritance, that's fine.

The last item (only one inheritance level) is IMHO an artificial limitations that does not reflect real-world usage.  In particular, it will make it much harder for third parties to extend the standard library's bodies because the standard library will need to extend its own bodies at least once.  As a former standard library maintainer, I assure you even internally to the standard library this will be a crippling limitation.  I hope you will reconsider.

kacfengine:
Please understand that this is about keeping the language simple. One level will already give maintainers a lot to work with, but it does reduce readability, and each level will add less value, but the same amount of reduced readability. Sorry, but the decision was quite clear yesterday, we will only accept one-level inheritance.

tzz:
I looked around but can't find any other languages that enforce one-level inheritance.  It's just not done.  Does that not indicate that the limitation is unwarranted?

Each level of inheritance provides another layer of abstraction, not value per se.  Abstraction in our case can be quantified as the number of of things you have to keep in your head to achieve some behavior.  I disagree that more than one layer of abstraction will confuse the reader.  By the same logic, Java classes are horribly designed because they go 5-6 levels deep and have much more hidden behavior.

To your second point about readability: I think that is an arbitrary measure.  Regular expressions are heavily used in CFEngine and are completely unreadable to casual visitors (e.g. https://docs.cfengine.com/docs/master/reference-standard-library-files.html#set_line_based).

Sorry to keep insisting on this after your final statement, but I do feel the one-level limitation is not going to serve a practical purpose.

neilhwatson:
This feature will help cut down on the sprawl of stdlib and our own custom libraries, but only if the level limit isn't arbitrarily short. One level only is silly. Yes, more levels can lead to complexity but we should steer users away from that buy guidance not by setting a hard limit.

It's wrong to set such limits and think that we know best. Remove the limits and let users innovate.

basvandervlies:
I agree with @tzz and @neilhwatson. Do not limit it and let the users decide.

tzz:
Rebased and converted to single inheritance.  Parameter rewriting still TODO.

kacfengine:
Ok, if you are not content with the current proposal then I suggest moving this either to Redmine or to the mailing list for further discussion. But note that we don't really have time to discuss this matter internally another time for 3.7. This feature is likely to be the last to go into 3.7 from the community, but that's assuming the currently proposed spec is implemented. If not it's gonna take more time and it will be too late for 3.7.

worr:
I'm also with @basvandervlies, @neilhwatson and @tzz.

Single-level inheritance is meaningless to users that consume other's stdlibs. If they ever need a custom body, and try and use inheritance to make it easier to define, they will almost certainly hit the one-level inheritance wall.

Since no other commonly used language works like this, this rule is another special exception that people have to keep in mind when working with CFEngine. tbh, I think we should avoid those unless they really do serve the user.

tzz:
I implemented rewritten parameters as requested but it's tremendously complicated compared to my original proposal because at the body definition stage, variable expansion is not available.  So I had to implement a special Rval rewrite mechanism just for this feature!  The simpler code without parameters was a much, much smaller patch.

It's one level of inheritance only for now, but I left it so it's pretty easy to extend to more levels if we agree on that.

It currently works for static strings like this:

```
body classes scoped_classes_generic_none_inherit
{
      inherit_from => scoped_classes_generic("namespace", "test_inherit_none");
}
```

but not for parameterized attributes like this:

```
body classes scoped_classes_generic_some_inherit(myx)
{
      inherit_from => scoped_classes_generic("namespace", $(myx));
}
```

You can see the failing test in `00_basics/03_bodies/inherit_from_parameters.cf`.  I left ERR-level log statements that show the rval value is replaced correctly, e.g.

```
   error: args case: rewritten rval from  {'promise_repaired_$(x)','$(x)_repaired','$(x)_ok','$(x)_reached'} to  {'promise_repaired_$(myx)','$(myx)_repaired','$(myx)_ok','$(myx)_reached'}
```

I am positive the new Rval is correctly written, but the variable inside (like `$(myx)` above) doesn't seem to be expanded correctly later.  This seems to be an evaluation quirk.  So I can work on it some more (and maybe get help from the developers), or we can say it only works with static strings for now.

tzz:
@kacfengine I'd rather get this into 3.7 with one level of inheritance than not get it in at all, since those are the only choices.  We can discuss multiple levels later as you said.

Also generic `BufferReplaceStr()` and `BufferReplaceRegex()` functions would be really sweet.  You can see what I did in `RvalNewRewriter()` and it's not pretty.

tzz:
Fixes pushed, all tests pass.  The expansion rewrite **only** triggers with `inherit_from` (when `args_rewrite` is not NULL) so I think it's a safe addition in general.

The scope and expansion internals are pretty complex.  Working on them definitely builds one's character.

kacfengine:
It may be inappropriate of me to try to defend single level inheritance, since I am still against this feature in general, my vote goes to zero level inheritance.. :-P

The only thing I can say is that we can't compare inheritance of general purpose language classes to body inheritance. Bodies are completely rigid and set (unlike bundles). We know exactly what body types we have, how many attributes they have, the user cannot add even a single attribute, and even between versions they almost never change. And most bodies only have a few of attributes.

I have to say, that the fact that so many people think that this feature will be useless in its current form has me quite worried. If single level inheritance does not solve the problem for a body with a few attributes, how will multiple levels?

I'm tempted to say we should simply make a best practice guide for how to reuse data in bodies by referring to variables in global bundles. It's both more transparent and explicit, only uses existing features, and AFAICS accomplishes the same thing. I invite you to come up with an example where this would not work, but multiple levels of inheritance would.

kacfengine:
trigger build

kacfengine:
Sending it through a test build. I haven't looked at the code yet.

tzz:
@kacfengine the current `packages` bodies use common bundles for shared knowledge exactly as you describe.  It creates a situation where the common bundle becomes a frozen API that all implementers of `packages` bodies have to use.  And the biggest problem, the redundancy and unreadable code caused by having to explicitly state **all** the desired body attributes in every body, still remains.

Multi-level inheritance solves the problem of: standard library creates a base body `B` with attributes `a1` and `a2`.  Standard library extends `B` to `B1` with attribute `a3`.  Then standard library **or a third party** wants to extend `B1` but override `a2`.

With single inheritance, you'd have to extend `B`, override `a2`, and restate `a3`.  OK, so you create `B2` with that behavior, but **crucially** you were really interested in `B1`.  Now the standard library says "wait, `B1` has a bug, it also needs attribute `a4` and `a3` has to change."

Suddenly both the standard library and every third party has to be aware of the bug in `B1` and to modify `a3` and add `a4` to every place that was chasing `B1` behavior and just wanted to override it.  In other words, the complexity goes up significantly because you can't rely on inheritance to flow from the top.  This scenario is even worse with global "common knowledge" bundles like what I described for `packages`.

To answer your point about body rigidity, the body **elements** are set in stone, but the body **definitions** that use those elements can change easily.  Every new body attribute grows new possibilities for the body type combinatorially (and that factorial growth makes chasing the desired body behavior and knowing the potential bugs a tough job).  Inheritance reduces that growth by creating base behaviors that can be extended, instead of forcing the user to specify all the attributes he needs every time.

Anyhow.  I've prepared for multiple inheritance by using a `Seq` of bodies and args, which currently will only have 4 elements (the base body, then a NULL, then the parent, then its args).  When and if we support multiple inheritance, we'll just pass a longer `Seq`, so a single place in the code has to be modified.

tzz:
I added a staging test for multiple inheritance that we can use at some point.  But commit 9a65e0cc2d8e6b496e192731a737a9852581d7ec can be dropped, it's not needed for this.

kacfengine:
I got a lot of failures in Jenkins on this one:

<pre>
./01_vars/02_functions/nth_datacontainer.cf.nth_datacontainer.cf
./10_files/copy_local_file.cf.copy_local_file.cf
./10_files/this_promiser.cf.this_promiser.cf
./16_cf-serverd/serial/001.cf.001.cf
./16_cf-serverd/serial/004.cf.004.cf
./16_cf-serverd/serial/007.cf.007.cf
./16_cf-serverd/serial/008.cf.008.cf
./16_cf-serverd/serial/010.cf.010.cf
./16_cf-serverd/serial/011.cf.011.cf
./16_cf-serverd/serial/allow_path1_then_deny_path2_a.cf.allow_path1_then_deny_path2_a.cf
./16_cf-serverd/serial/copy_from_digest_different.cf.copy_from_digest_different.cf
./16_cf-serverd/serial/copy_from_digest_different_expand_ip_and_shortcut.cf.copy_from_digest_different_expand_ip_and_shortcut.cf
./16_cf-serverd/serial/copy_from_digest_different_expand_ip.cf.copy_from_digest_different_expand_ip.cf
./16_cf-serverd/serial/copy_from_digest_different_expand_shortcut.cf.copy_from_digest_different_expand_shortcut.cf
./16_cf-serverd/serial/copy_from_encrypted_md5_zero_length_file.cf.copy_from_encrypted_md5_zero_length_file.cf
./16_cf-serverd/serial/copy_from_expand_ip_directory.cf.copy_from_expand_ip_directory.cf
./16_cf-serverd/serial/copy_from_md5_zero_length_file.cf.copy_from_md5_zero_length_file.cf
./16_cf-serverd/serial/simple_copy_from_admit_localhost.cf.simple_copy_from_admit_localhost.cf
</pre>

At this point I'm sorry to say that this feature won't make it for 3.7. We're preparing to branch very soon and there are still code issues here. In addition I don't think it's wise to rush in a feature when there is still disagreement over how it should work. We will of course continue the discussion of going into 3.8.

> Multi-level inheritance solves the problem of: standard library creates a base body B with attributes a1 and a2. Standard library extends B to B1 with attribute a3. Then standard library or a third party wants to extend B1 but override a2.
> 
> With single inheritance, you'd have to extend B, override a2, and restate a3. OK, so you create B2 with that behavior, but crucially you were really interested in B1. Now the standard library says "wait, B1 has a bug, it also needs attribute a4 and a3 has to change."
> 
> Suddenly both the standard library and every third party has to be aware of the bug in B1 and to modify a3 and add a4 to every place that was chasing B1 behavior and just wanted to override it. In other words, the complexity goes up significantly because you can't rely on inheritance to flow from the top.

This is a good point, I can see how this would increase complexity. The crucial, missing link here for me is: Which body needs this? What you're saying sounds very logical and well argued, but that's seen in the light of data types that you can extend freely. Can you give me an actual body from "real life", inherited in the described way, where B2 inherits from B1 and B1 fixes a bug in B? What would be an example of such a case in CFEngine?

tzz:
The test errors are from places where we say:

`copy_from => local_cp($(this.promise_filename))`

which breaks with inheritance because body attributes seem to be evaluated sooner now.  I think that's fixable but agree, this feature shouldn't go into 3.7 right now.

> What you're saying sounds very logical and well argued, but that's seen in the light of data types that you can extend freely. Can you give me an actual body from "real life", inherited in the described way, where B2 inherits from B1 and B1 fixes a bug in B? What would be an example of such a case in CFEngine?

I can't give real-world examples of body inheritance because it hasn't ever been available!  I think `packages` and `commands` bodies are good potential uses of the feature.  `classes` are another.  Users copy them again and again (3-10 attributes) to make a change in 1 or 2 attributes.  As soon as the standard library uses inheritance just once (and it should, to keep the code clean), users are locked out of inheriting standard library bodies.

kacfengine:
> I can't give real-world examples of body inheritance because it hasn't ever been available! I think packages and commands bodies are good potential uses of the feature. classes are another. Users copy them again and again (3-10 attributes) to make a change in 1 or 2 attributes.

Ok, let me rephrase the question: If you had body inheritance today, where would be the first place you would use it?

> As soon as the standard library uses inheritance just once (and it should, to keep the code clean), users are locked out of inheriting standard library bodies.

This is a very good point, since it could in fact break users' policies, if the user started using inheritance before we did in masterfiles. I think you have me convinced now that single level inheritance probably isn't a good idea.

tzz:
> Ok, let me rephrase the question: If you had body inheritance today, where would be the first place you would use it?

Probably the first standard library targets would be `copy_from`, `packages` (or whatever replaces it), `contains`, `classes`, and `edit_field`.

> I think you have me convinced now that single level inheritance probably isn't a good idea.

All right!  There's plenty of time to convince the other developers too :)

libpromises/promises.c
```diff
@@ -170,15 +172,69 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
...
+                            continue;
+                        }
+
+                        if (IsDefinedClass(ctx, scp->classes))
+                        {
+                            Rval returnval = ExpandPrivateRval(ctx, NULL, "body", scp->rval.item, scp->rval.type);
+
+                            Rval rewrite = RvalCopyRewriter(returnval, arg_rewrite);
+                            RvalDestroy(returnval);
+                            returnval = ExpandPrivateRval(ctx, NULL, "body", rewrite.item, rewrite.type);
```
tzz: The bug is here: we rewrite the body parameters twice to get inherited parameters expanded.  Working on it.

tzz:
I found the inheritance bug: I was expanding the Rval contents twice and that was causing early expansion of `$(this.promise_filename)`.

This is ready for review.  I'd like to go back to the multiple inheritance model, but otherwise it's working and the tests show it.

kacfengine:
> I found the inheritance bug: I was expanding the Rval contents twice and that was causing early expansion of $(this.promise_filename).

Great!

> This is ready for review. I'd like to go back to the multiple inheritance model, but otherwise it's working and the tests show it.

I think given the latest discussion that we can proceed with *multi-level* inheritance (not to be confused with multiple inheritance though). Is that included in the latest changes?

tzz:
OK, I didn't get a final yes on multi-level inheritance so I was holding off.  I'll see if I can implement it.

tzz:
I've implemented multi-level inheritance:

* fairly simple, and works by recursively finding each body's parent and arguments
* depth check: if we go over 30, abort.  This also prevents circular dependencies cheaply (unlike a Set of body names).  I think this will be a rare occurrence.
* self-inheritance check with acceptance test: if a body tries to inherit from itself, abort with a helpful message.  I think this will happen occasionally so the message tries to be helpful.

tzz:
ping?

estenberg:
In the current state it looks good.

kacfengine:
Alright, 3.7 is in a good place, I believe we can start looking at this again.

tests/acceptance/00_basics/03_bodies/inherit_from.cf
```diff
@@ -0,0 +1,54 @@
...
+{
+  reports:
+      "test_inherit" classes => classes_generic_inherit("test_inherit");
+      "test_nonesuch" classes => classes_generic_nonesuch("test_nonesuch");
+      "test_overwrite" classes => classes_generic_overwrite("test_overwrite");
+}
+
+body classes classes_generic_inherit(x)
+{
+      inherit_from => classes_generic;
```
kacfengine: This is wrong behavior, IMHO. Here I would expect an error message "Body takes 1 argument but 0 given" or something of the sort. Letting named arguments "bleed through" to the next body without being explicitly specified breaks locality of variables and is unexpected when compared to other languages.

libpromises/rlist.c
```diff
@@ -325,18 +325,50 @@ Rlist *RlistAppendRval(Rlist **start, Rval rval)
     return rp;
 }
 
-Rval RvalNew(const void *item, RvalType type)
+Rval RvalNewRewriter(const void *item, RvalType type, JsonElement *map)
 {
+    int max_size = 10*CF_BUFSIZE+1;
+    char buffer_from[max_size];
+    char buffer_to[max_size];
```
kacfengine: This is almost 100KB of stack space. Considering the possibility of CFEngine being used on embedded platforms, I would make this a heap allocation instead, since stack space can be considerably more limited on some platforms.

libpromises/rlist.c
```diff
@@ -325,18 +325,50 @@ Rlist *RlistAppendRval(Rlist **start, Rval rval)
...
+            const char *key;
+
+            while ((key = JsonIteratorNextKey(&iter)))
+            {
+                BufferPrintf(format, "$(%s)", key);
+                ReplaceStr(buffer_from, buffer_to, max_size, BufferData(format), JsonObjectGetAsString(map, key));
+                strncpy(buffer_from, buffer_to, max_size);
+                BufferPrintf(format, "${%s}", key);
+                ReplaceStr(buffer_from, buffer_to, max_size, BufferData(format), JsonObjectGetAsString(map, key));
+            }
```
kacfengine: This might give unexpected results. Since you are only iterating over each key once, you might not catch recursively defined variables, and whether or not you do will depend on the order of the map. For example:

```
map = { a=b, b=c }

Iterations:
    key="a":
        "$($(a))" -> "$(b)"
    key="b":
        "$(b)" -> "c"
```

However, if we flip the order:

```
map = { b=c, a=b }

Iterations:
    key="b":
        "$($(a))" -> "$($(a))"
    key="a":
        "$($(a))" -> "$(b)"
```

which will not resolve into "c".

Such cases are surely handled elsewhere in the code. Being no evaluator expert, I don't know exactly where, but the function ``ExpandAndMapIteratorsFromScalar`` may be a place to start looking.

If it proves too difficult to reuse existing code, then this could be implemented differently: Instead of iterating over the keys, find all occurrences of ``$(...)`` in the string, and replace those after looking up the key in the map. When you are finished, repeat the whole process, since the replacements may have revealed new variables that need resolving. Only stop once no more replacements can be made, either because there are no ``$(...)`` references left, or because the remaining ones can't be found in the map.

kacfengine: Perhaps with an upper limit on number of iterations as well, since someone could be evil and put ``map = { a="$(b)", b="$(a)" }``.

libpromises/rlist.c
tzz: OK, will do.

libpromises/rlist.c
tzz: I'd prefer to avoid evaluation here.  If you feel otherwise, tell me why.  Ideally show a test case because I don't see a common need for this kind of replacement.  I added a 3-time iteration to the rewriting loop but hope I can avoid the rest.

tests/acceptance/00_basics/03_bodies/inherit_from.cf
tzz: I think it's a nice syntactic shortcut.  Are you sure you want it changed?

libpromises/rlist.c
```diff
@@ -325,18 +325,56 @@ Rlist *RlistAppendRval(Rlist **start, Rval rval)
...
-Rval RvalNew(const void *item, RvalType type)
+Rval RvalNewRewriter(const void *item, RvalType type, JsonElement *map)
 {
     switch (type)
     {
     case RVAL_TYPE_SCALAR:
-        return (Rval) { xstrdup(item), RVAL_TYPE_SCALAR };
+        if (NULL != map && JsonLength(map) > 0)
+        {
+            // TODO: replace with BufferSearchAndReplace when the
```
tzz: The `string_replace` PR from https://dev.cfengine.com/issues/7346 has a `BufferSearchAndReplace` function I will use here, once that PR is merged.  So please consider these workarounds temporary.

tests/acceptance/00_basics/03_bodies/inherit_from.cf
kacfengine: Yes, it's deceptive, because it makes it look like ``classes_generic`` takes no arguments. Better to be explicit about such things.

libpromises/rlist.c
```diff
@@ -325,18 +325,56 @@ Rlist *RlistAppendRval(Rlist **start, Rval rval)
...
+            // string_replace code is merged.
+            // Sorry about the CF_BUFSIZE ugliness.
+            int max_size = 10*CF_BUFSIZE+1;
+            const char* buffer_from = malloc(max_size);
+            const char* buffer_to = malloc(max_size);
```
kacfengine: Should use ``xmalloc``.

libpromises/rlist.c
```diff
@@ -325,18 +325,56 @@ Rlist *RlistAppendRval(Rlist **start, Rval rval)
...
+            {
+                while ((key = JsonIteratorNextKey(&iter)))
+                {
+                    BufferPrintf(format, "$(%s)", key);
+                    ReplaceStr(buffer_from, buffer_to, max_size, BufferData(format), JsonObjectGetAsString(map, key));
+                    strncpy(buffer_from, buffer_to, max_size);
+                    BufferPrintf(format, "${%s}", key);
+                    ReplaceStr(buffer_from, buffer_to, max_size, BufferData(format), JsonObjectGetAsString(map, key));
+                }
+            }
```
kacfengine: Sorry, but I don't think this is a good approach, because it is incredibly inefficient, and this is going to get evaluated a lot. In a body with say, 4 arguments, you will be scanning the string 3 * 4 * 2 = 24 times, even when there is not a single replacement. What about something like this (untested):

```
for (int iteration = 0; iteration < 10; iteration++)
{
    bool replacement_made = false;
    int var_start = -1;
    for (int c = 0; c < buffer_from[c]; c++)
    {
       if (buffer_from[c] == '$')
       {
          c++;
          if (buffer_from[c] == '(' || buffer_from[c] == '{')
          {
             var_start = c-1;
          }
       }
       else if (var_start >= 0 && buffer_from[c] == ')' || buffer_from[c] == '}')
       {
          // Before the replacement.
          memcpy(buffer_to, buffer_from, var_start);

          // The actual replacement.
          char saved = buffer_from[c];
          buffer_from[c] = '\0';
          const char *repl = JsonObjectGetAsString(map, buffer_from[var_start+2]);
          int repl_len = strlen(repl);
          buffer_from[c] = saved;
          memcpy(buffer_to + var_start, repl, repl_len);

          // The text after.
          strlcpy(buffer_to + var_start + repl_len, buffer_from + c + 1, max_size - var_start - repl_len);

          // Reset location to immediately after the replacement.
          c = var_start + repl_len - 1;
          var_start = -1;
          strcpy(buffer_from, buffer_to);
          replacement_made = true;
       }
    }

    if (!replacement_made)
    {
        break;
    }
}
```

I'm not taking into account that the string can look like ``$(x}`` or ``${x)``, but this is just a suggestion.

kacfengine: Updated example.

tzz: I've modified the code to break early and to do 10 iterations, but would prefer if you make the optimizations you propose after it's merged.

tests/acceptance/00_basics/03_bodies/inherit_from.cf
tzz: OK, it's not done, but I pushed a commit with tests and 2 TODO items in `promises.c` that collects the given argument count and when it's done, will check it against the body prototype.  Check that the tests match what you expect, please.

kacfengine: Yes, that's exactly right! :-)

tzz:
@kacfengine OK, this is updated as follows:

* when at the last position of the `bodies_and_args` list (which is the current body) we don't check argument counts.  Those are checked elsewhere.
* otherwise, we compare the `Body.args Rlist` length with the `Rval Fncall` given args.
* bareword references like `inherit_from => mybody` are also checked
* we fail with a `FatalError` because this is a run-time check, like vararg functions.  Example: `error: Fatal CFEngine error: Argument count mismatch for body reference 'classes_generic_broken_inherit' (gave 0 arguments) vs. inherited body 'default:classes_generic' (requires 1 arguments) in promise at line 17 of file './00_basics/03_bodies/inherit_from_noargs.x.cf'`
* the acceptance tests pass.

kacfengine:
Ok, sounds good! I'll take a detailed look tomorrow.

kacfengine:
Great work, looks good now! I'll try my performance patch on top.

kacfengine:
I have opened https://github.com/cfengine/core/pull/2320 with my extra fixes on top, if you can do a final sanity check?

tzz:
Yup, feel free to close this one.

kacfengine:
Thanks.

kacfengine:
The test build failed in package promises, I suspect this line: https://github.com/tzz/core/blob/feature/redmine4309-inherits_from/libpromises/attributes.c#L1051

I'll leave it to you to fix it, you know the sequence structure best. Just push it to this pull request, and I can port it over to mine.

tzz:
Which tests failed?

kacfengine:
The unsafe ones in masterfiles.

* ./17_packages/10_new/unsafe/the_great_package_test.cf.the_great_package_test.cf
* ./17_packages/11_old/unsafe/named_multi_pkg_absent.cf.named_multi_pkg_absent.cf
* ./17_packages/11_old/unsafe/named_multi_pkg_present.cf.named_multi_pkg_present.cf
* ./17_packages/11_old/unsafe/named_pkg_absent.cf.named_pkg_absent.cf
* ./17_packages/11_old/unsafe/named_pkg_present.cf.named_pkg_present.cf
* ./17_packages/12_new/timed/installed-list-cached.cf.installed-list-cached.cf
* ./17_packages/12_new/timed/local-updates-list-cached.cf.local-updates-list-cached.cf
* ./17_packages/12_new/timed/updates-list-cached.cf.updates-list-cached.cf

Most of them segfaulted, but there are also a few "safe" ones that failed. They didn't segfault, but failed for other reasons, but they may still be related.

kacfengine:
I don't think it's necessary to analyze those in detail though. The line I mentioned is clearly wrong, since the signature of the function has changed in your implementation. I just want you to change it, since you know best how to handle the new ``Seq`` that it now returns.

tzz:
Oh, I see.  I missed that in the original change set.  OK, pushed the fix with a hopefully helpful comment.

kacfengine:
Great, thanks!

PR #2154, opened 2015-03-31
Statedir compile time option

kacfengine:


kacfengine:
Merge together with https://github.com/cfengine/enterprise/pull/260 and https://github.com/cfengine/nova/pull/908.

basvandervlies:
Nice one. Finally ;-)

cdituri:
Resurrected - thanks!
Also good call with 9107879; implementing a fallback like that hadn't occurred to me at the time :-)

kacfengine:
@cdituri: Better late than never! Thanks for waiting this long.

PR #2155, opened 2015-03-31
Add: Test dry-run mode does not assert it changes perms

nickanderson:
When run in dry run mode, the agent should not say that it changes any 
permissions, because in fact it does not.

Ref: https://dev.cfengine.com/issues/7082

### DIFF
--- a/tests/acceptance/00_basics/02_switches/dry_run_perms_doesnt_lie.cf
+++ b/tests/acceptance/00_basics/02_switches/dry_run_perms_doesnt_lie.cf
@@ -1,0 +1,38 @@
+body common control
+{
+      inputs => { "../../default.cf.sub" };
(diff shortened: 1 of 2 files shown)


cdituri:
@nickanderson -
dry-run flag sets DONTDO global [here](https://github.com/cfengine/core/blob/a17a584f4e74f40aee5c878c0af71250b34c29b9/cf-agent/cf-agent.c#L861).
The log statements in question are found [here](https://github.com/cfengine/core/blob/dabfb4610f30342407dc27fb5b9738dbff2fb5f7/cf-agent/verify_files_utils.c#L2107)

IINM, the cfPS() call just needs to move inside the `if (!DONTDO) { ... }` conditional block.

kacfengine:
Thanks @cdituri!

Attempting to fix the problem as well: https://github.com/cfengine/core/pull/2184.

PR #2156, opened 2015-04-01
Redmine#3660: inlined JSON

tzz:
This PR implements inlined JSON/YAML as a literal string that gets either directly parsed or synthetically replaced with the equivalent `parsejson()` or `parseyaml()` call.  The trigger for this behavior is that:

* a `SCALAR` is passed to a `data` attribute
* the string **begins** with either `---` for YAML or else it's JSON
* if the string begins with the YAML marker and contains no unexpanded variables (found by just looking for `$(` or `${`) we try to parse the YAML directly
    * if the YAML parse fails or unexpanded variables were found, we insert a synthetic fncall to `parseyaml()`
* else, we try JSON: if the string contains no unexpanded variables (found by just looking for `$(` or `${`) we try to parse the JSON directly
    * if the JSON parse fails or unexpanded variables were found, we insert a synthetic fncall to `parsejson()`

The example below will demonstrate:

```
bundle agent main
{
  vars:
      "options1" data => '
{
  "cfengine.max_content": 512,
  "curl.verbose": 0,
  "curl.headers": [ "Foo: bar" ]
}';

      "options2" data => '---
a: b
c: d
e:
  - 1
  - 2
';

      "options3" data => '
{
  "cfengine.max_content": 512,
  "curl.verbose": "$(sys.os)",
  "curl.headers": [ "Foo: bar" ]
}';

      "options4" data => '---
a: $(sys.os)
c: d
e:
  - 1
  - 2
';

      "out1" string => format("%S", options1);
      "out2" string => format("%S", options2);
      "out3" string => format("%S", options3);
      "out4" string => format("%S", options4);

  reports:
      "$(this.bundle): we got $(out1)";
      "$(this.bundle): we got $(out2)";
      "$(this.bundle): we got $(out3)";
      "$(this.bundle): we got $(out4)";
}
```

Output:
```console
R: main: we got {"cfengine.max_content":512,"curl.verbose":0,"curl.headers":["Foo: bar"]}
R: main: we got {"a":"b","c":"d","e":[1,2]}
R: main: we got {"cfengine.max_content":512,"curl.verbose":"linux","curl.headers":["Foo: bar"]}
R: main: we got {"a":"linux","c":"d","e":[1,2]}
```

There are some minor problems with this approach:

* if the JSON or YAML are malformed, the error points to a nonexistent position on the trigger line
* inserting synthetic function calls is kind of weird
* the `cf-promises` JSON output needs to be adapted to deal with containers in the policy itself, as shown here with our earlier example.  That should be an easy fix.

```console
% cf-promises -p json .//test_inline_json.cf
policy.c:1579: Programming Error: Attempted to export attribute of type: c
```

### DIFF
--- a/libpromises/cf3parse.y
+++ b/libpromises/cf3parse.y
@@ -600,11 +600,9 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
                                                Buffer *copy = BufferNewFrom(P.rval.item, strlen(P.rval.item));
 
                                                const char* fname = NULL;
-                                               int offset = 0;
                                                if (strlen(P.rval.item) > 3 && strncmp("---", P.rval.item, 3) == 0)
                                                {
                                                    fname = "parseyaml";
-                                                   offset = 0;
 
                                                    // look for unexpanded variables
                                                    if (NULL == strstr(P.rval.item, "$(") && NULL == strstr(P.rval.item, "${"))
@@ -617,7 +615,6 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
                                                else
                                                {
                                                    fname = "parsejson";
-                                                   offset = 0;
                                                    // look for unexpanded variables
                                                    if (NULL == strstr(P.rval.item, "$(") && NULL == strstr(P.rval.item, "${"))
                                                    {
@@ -631,11 +628,11 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
 
                                                if (json_parse_attempted && res != JSON_PARSE_OK)
                                                {
-                                                   // no error, we'll try with the fname fncall
+                                                   // Parsing failed, insert fncall so it can be retried during evaluation
                                                }
                                                else if (NULL != json && JsonGetElementType(json) == JSON_ELEMENT_TYPE_PRIMITIVE)
                                                {
-                                                   // no error, we'll try with the fname fncall
+                                                   // Parsing failed, insert fncall so it can be retried during evaluation
                                                    JsonDestroy(json);
                                                    json = NULL;
                                                }
@@ -645,7 +642,7 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
                                                    if (NULL == json)
                                                    {
                                                        Rlist *synthetic_args = NULL;
-                                                       RlistAppendScalar(&synthetic_args, xstrdup(P.rval.item+offset));
+                                                       RlistAppendScalar(&synthetic_args, xstrdup(P.rval.item));
                                                        RvalDestroy(P.rval);
 
                                                        P.rval = (Rval) { FnCallNew(xstrdup(fname), synthetic_args), RVAL_TYPE_FNCALL };
--- a/libpromises/cf3parse.y
+++ b/libpromises/cf3parse.y
@@ -576,7 +576,23 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
                                        // Intentional fall
                                    case SYNTAX_STATUS_NORMAL:
                                        {
-                                           if (P.rval.type == RVAL_TYPE_SCALAR && strcmp(P.lval, "data") == 0)
+                                           const char *item = P.rval.item;
+                                           // convert @(x) to mergedata(x)
+                                           if (P.rval.type == RVAL_TYPE_SCALAR &&
+                                               (strcmp(P.lval, "data") == 0 || strcmp(P.lval, "template_data") == 0) &&
+                                               strlen(item) > 3 &&
+                                               item[0] == '@' &&
+                                               (item[1] == '(' || item[1] == '{'))
+                                           {
+                                               Rlist *synthetic_args = NULL;
+                                               RlistAppendScalar(&synthetic_args, xstrndup(P.rval.item+2, strlen(P.rval.item)-3 ));
+                                               RvalDestroy(P.rval);
+
+                                               P.rval = (Rval) { FnCallNew(xstrdup("mergedata"), synthetic_args), RVAL_TYPE_FNCALL };
+                                           }
+                                           // convert 'json or yaml' to direct container or parsejson(x) or parseyaml(x)
+                                           else if (P.rval.type == RVAL_TYPE_SCALAR &&
+                                                    (strcmp(P.lval, "data") == 0 || strcmp(P.lval, "template_data") == 0))
                                            {
                                                JsonElement *json = NULL;
                                                JsonParseError res;
--- a/libpromises/cf3parse.y
+++ b/libpromises/cf3parse.y
@@ -36,6 +36,7 @@
 #include "mod_files.h"
 #include "string_lib.h"
 #include "logic_expressions.h"
+#include <json-yaml.h>
 
 // FIX: remove
 #include "syntax.h"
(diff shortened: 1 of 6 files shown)


tzz:
If the syntax and behavior are acceptable, I will fix the `cf-promises` issue and add acceptance tests and docs.

tzz:
I added a followup commit that:

* enables inlined JSON and YAML for the `template_data` attribute (although I'd like to find out if the type of the `lval` is container, not hard-code the constraint name from the lval)
* enables using `@(anything)` and `@{anything}` for `data` and `template_data` promise attributes, in which case a synthetic call to `mergedata(anything)` is done.  Note that that means you can use classic CFE arrays and slists and even wrap containers in `data` and `template_data` attributes, as long as you surround them with `@()` or `@{}`.  That remedies https://github.com/cfengine/core/pull/934

Example of both of the above (note that `x` is copied into `y` and then `y` is used):

```
bundle agent main
{
  vars:
      "x" data => parsejson('{ "os": "$(sys.os)" }');
      "y" data => @(x);

  files:
      "/tmp/myfile.txt"
      create => "true",
      edit_template => "$(this.promise_filename).mustache",
      template_data => @(y);
}
```

tzz:
ping

tzz:
ping?

libpromises/cf3parse.y
```diff
@@ -573,6 +574,88 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
                                        // Intentional fall
                                    case SYNTAX_STATUS_NORMAL:
                                        {
+                                           const char *item = P.rval.item;
+                                           // convert @(x) to mergedata(x)
+                                           if (P.rval.type == RVAL_TYPE_SCALAR &&
+                                               (strcmp(P.lval, "data") == 0 || strcmp(P.lval, "template_data") == 0) &&
+                                               strlen(item) > 3 &&
+                                               item[0] == '@' &&
+                                               (item[1] == '(' || item[1] == '{'))
```
kacfengine: This doesn't check that the final character is ')', so this would be legal, and a reference to ``myvar``:

```
data => @(myvarX;
```

kacfengine: Or '}' for that matter.

libpromises/cf3parse.y
```diff
@@ -573,6 +574,88 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
...
+                                                    (strcmp(P.lval, "data") == 0 || strcmp(P.lval, "template_data") == 0))
+                                           {
+                                               JsonElement *json = NULL;
+                                               JsonParseError res;
+                                               bool json_parse_attempted = false;
+                                               Buffer *copy = BufferNewFrom(P.rval.item, strlen(P.rval.item));
+
+                                               const char* fname = NULL;
+                                               int offset = 0;
+                                               if (strlen(P.rval.item) > 3 && strncmp("---", P.rval.item, 3) == 0)
```
kacfengine: What do we expect from the initial '---' string here? Maybe we should enforce a newline after it, just so that there's no ambiguity?

libpromises/cf3parse.y
```diff
@@ -573,6 +574,88 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
...
+                                                       res = JsonParse(&copy_data, &json);
+                                                       json_parse_attempted = true;
+                                                   }
+                                               }
+
+                                               BufferDestroy(copy);
+
+                                               if (json_parse_attempted && res != JSON_PARSE_OK)
+                                               {
+                                                   // no error, we'll try with the fname fncall
```
kacfengine: Slightly confusing comment, maybe instead: ``Parsing failed, insert fncall so it can be retried during evaluation``.

libpromises/cf3parse.y
```diff
@@ -573,6 +574,88 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
...
+                                               }
+                                               else if (NULL != json && JsonGetElementType(json) == JSON_ELEMENT_TYPE_PRIMITIVE)
```
kacfengine: Why is ``JSON_ELEMENT_TYPE_PRIMITIVE`` a failure?

libpromises/cf3parse.y
```diff
@@ -573,6 +574,88 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
...
+                                               {
+                                                   // no error, we'll try with the fname fncall
```
kacfengine: Again a confusing comment.

libpromises/cf3parse.y
```diff
@@ -573,6 +574,88 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
...
+                                                   JsonDestroy(json);
+                                                   json = NULL;
+                                               }
+
+                                               if (NULL != fname)
+                                               {
+                                                   if (NULL == json)
+                                                   {
+                                                       Rlist *synthetic_args = NULL;
+                                                       RlistAppendScalar(&synthetic_args, xstrdup(P.rval.item+offset));
```
kacfengine: ``offset`` appears to always be zero. It looks like it can be removed.

kacfengine:
I think this is quite nice. But it needs acceptance tests.

libpromises/cf3parse.y
tzz: I think that's OK.  In that case we simply try to parse with a function instead of directly, resulting in the same effect if those were not variable references.  I really don't want to write a unexpanded var ref detector--if you can suggest one, I'll use it...

libpromises/cf3parse.y
tzz: We know `---` is not valid JSON, so it must be YAML.

libpromises/cf3parse.y
tzz: OK, fixed

libpromises/cf3parse.y
tzz: It's actually not allowed by the JSON standard.  But I can allow it if you want...

tzz:
Acceptance test added.

libpromises/cf3parse.y
kacfengine: Right, I just tried it, and it seems the parser will not allow you to get into that situation anyway.

tests/acceptance/01_vars/04_containers/inline_json.cf
```diff
@@ -0,0 +1,70 @@
...
+c: d
+e:
+  - 1
+  - 2
+';
+
+      "options3" data => '
+{
+  "max_content": 512,
+  "verbose": "$(sys.os)",
```
kacfengine: This will fail on all non-Linux platforms. In general it's not a good idea to use ``$(sys.os)`` in acceptance tests, better to use some local variable you have defined.

kacfengine: See 01ca00c1363d7dc94fef59120050116ef144e3f5 for example.

kacfengine:
I'm missing a test case for ``"var" data => @(othervar);``, but otherwise nice test.

tzz:
Sorry, I don't know what `"var" data => @(othervar)` should do.  Do you mean `$(othervar)`?

tests/acceptance/01_vars/04_containers/inline_json.cf
tzz: OK; fixed

tzz:
Acceptance test updated as noted; sorry for that.  I'll wait for the other test you mentioned, but this is still not ready: `cf-promises` will segfault when dumping the policy in JSON format, if it has data containers directly inside.  Also column offsets are wrong in reports.  Should I proceed with that work, assuming the syntax is all right?

kacfengine:
> Sorry, I don't know what ``"var" data => @(othervar)`` should do. Do you mean ``$(othervar)``?

I mean just using it in the test case so that we know it doesn't cause a parsing error and it does the right thing. Something ala the example you posted earlier:

<pre>
bundle agent main
{
  vars:
      "x" data => parsejson('{ "os": "$(sys.os)" }');
      "y" data => @(x);

  files:
      "/tmp/myfile.txt"
      create => "true",
      edit_template => "$(this.promise_filename).mustache",
      template_data => @(y);
}
</pre>

Syntax seems good to me, so go ahead!

tzz:
Ha ha, @sigurdteigen had a bug open for passing containers to promises with `@(myvar)` but never got to it.  I wish it worked too.  But it doesn't; currently you have to call a **function** for a container promise attribute.  If you want to pass a container without changes, you'd say

```
"y" data => mergedata(x); # makes a copy
```

Currently `@(container)` is only usable when passing containers as a bundle or body parameter, which is really passing by reference.

kacfengine:
Hehe, that's indeed quite awkward, but fair enough, we can fix that later, it's no worse than it already was.

kacfengine:
trigger build

tzz:
Rebased against master to use state-based acceptance test.

tzz:
`cf-promises -p json` on policy with inline JSON works fine now, with acceptance test.  Also moved memory allocations closer to the usage site in `policy.c:AttributeValueToJson()`.

tzz:
I don't think I can fix the column offsets with inline JSON when a FnCall is inserted synthetically, because that would require some deep changes to the error messages.  If you think that's a requirement, let me know.  Otherwise this is ready for final review and works well for me.

kacfengine:
trigger build

kacfengine:
> I don't think I can fix the column offsets with inline JSON when a FnCall is inserted synthetically, because that would require some deep changes to the error messages.

I think it's ok, it won't break anything existing, right? Only if you start using inline JSON with variables?

We can handle it as a bug later if it turns out that people need it.

kacfengine:
@tzz: The slightly tweaked #2203 is green. Care to add some docs?

tzz:
documentation in https://github.com/cfengine/documentation/pull/1027

kacfengine:
Thanks, merged in #2203.

PR #2159, opened 2015-04-01
Add: Test that isvariable works with array keys that have spaces

nickanderson:
Arrays and data having spaces in a key is valid, so isvariable should be able
to test for them like any other variable.

Ref: https://dev.cfengine.com/issues/7088


kacfengine:
trigger build

kacfengine:
Damn, the soft failure mechanism doesn't work for this one, because the parser barfs before even getting to the evaluation stage. So this is not easily testable until the bug is fixed. I'll close this one for now, the ticket still has a reference to this test and we can pick it up later if we fix the bug.

PR #2160, opened 2015-04-02
libutils/logging.c: fix non-machine color output with --color=auto or always

tzz:
This was broken when the logging was changed.  It's a trivial fix to make the agent work like it used to, plus factoring out the ANSI color reset sequence since it's used twice.

### DIFF
--- a/libutils/logging.c
+++ b/libutils/logging.c
@@ -138,6 +138,11 @@ static const char *LogLevelToColor(LogLevel level)
     }
 }
 
+static const char *LogLevelColorReset()
+{
+    return "\x1b[0m";
+}
+
 bool LoggingFormatTimestamp(char dest[64], size_t n, struct tm *timestamp)
 {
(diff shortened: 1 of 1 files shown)


tzz:
ping

libutils/logging.c
```diff
@@ -174,6 +179,11 @@ static void LogToConsole(const char *msg, LogLevel level, bool color)
     }
     else
     {
+        if (color)
+        {
+            fprintf(stdout, "%s", LogLevelToColor(level));
```
ediosyncratic: General pattern: any use of `fprintf(tgt, "%s", blah)` is more naturally written `fputs(blah, tgt)` which skips parsing of the format string and uses faster simpler code paths.  (A good compiler might even make this transformation for you, but it's better to write it the efficient way to begin with.)

tzz:
OK; `fputs` used.  Thanks.

tzz:
ping?

tzz:
I believe this has been remedied in master, so closing

PR #1485, opened 2014-02-26
Redmine#4703: new mapjson() function and support data containers in maparray()

tzz:
see https://cfengine.com/dev/issues/4703

* `maparray()` can now process slists, classic arrays, and data containers
* new function `mapjson()` exactly like `maparray()` but the output is parsed as JSON into a data container
* new `this.k2` variable, for when we iterate through a second-level slist or container

Documentation and examples will follow upon merge.

The fixes for https://cfengine.com/dev/issues/4704 (marked as TODO in the acceptance test) will follow upon merge.

PR #2162, opened 2015-04-03
Add readdata() function for generic data input

tzz:
This PR has three components:

* unify CSV, JSON, and YAML parsing in one `FnCallReadData` function
* fix a bug with CSV parsing, where a null result from the parse could result in a segfault.  Also the byte limit is raised to 50 MB.
* introduce the function `readdata()` which takes two parameters: the filename and the mode (`auto` or `CSV` or `YAML` or `JSON`).  In `auto` mode the file extension is checked.  Note that `readdata()` has no file size parameter.

`readdata()` is handy for cases where the policy author wants to handle multiple types of files without special cases, like in https://github.com/cfengine/masterfiles/pull/397 as @nickanderson will confirm.

If this function is acceptable, I will write docs and acceptance tests.


nickanderson:
This is neat, it can make def.json support yaml and I guess csv without any code differences.

So maxbytes for JSON and YAML is inf, but CSV is 50MB? Is there a good reason for the difference? If there is no technical reason to limit CSV then I would suggest it operates the same as JSON and YAML, now we are just missing XML ;-0

nickanderson:
Also, does this mean we should deprecate readyaml, readjson, readcsv ?

tzz:
The current `readyaml`, `readcsv`, and `readjson` functions have two reasons to stay: first, explicit size limits (except for `readcsv`).  Second, using them gives you static validation that you'll read the right format instead of deciding it dynamically.

I can make CSV unlimited too, I just thought 50MB was quite large.

I did think about XML!  XML is much harder to read into a data container.  The mapping is not one-to-one and CFEngine doesn't yet have the tools to extract both content and attributes if I mix them in the data (the `mapjson()` proposal would help).  Also I don't know of anyone that needs it.  But we do have libxml linkage... so maybe a future feature :)

I also thought about naming the function `data_read`.  Not sure, what do you think?

nickanderson:
For naming I prefer readdata or read_data. @estenberg opinion on CSV data limit?

50MB csv is big and annoying, but so is knowing that CSV is a special case. So unless there is a technical reason to limit it, I would vote for consistency (yes, there is a hobgoblin in there).

ediosyncratic:
It is only petty consistency that's a hobgoblin.  Where consistency is actually useful, as here, there are no hobgoblins.

tzz:
ping?

tzz:
ping? This is a pretty safe isolated change IMO.

kacfengine:
I agree that CSV should not be size limited, so that we are consistent with the others. Other than that this is great, no concerns from me!

What's missing is an acceptance test and some docs.

tzz:
For the acceptance test, I'll wait for https://github.com/cfengine/core/pull/1485 to get merged because it will make the acceptance test much easier.

tzz:
Added examples.  Docs in https://github.com/cfengine/documentation/pull/1026

tzz:
The auto-guessing feature didn't feel right so I've removed it.  Now we simply try reading as JSON if the mode is `auto` and the extension is unknown.

tzz:
For JSON and YAML I had to set a max read size, and chose 50MB just like CSV.  I think that's reasonable but we can raise it.

tzz:
Acceptance test added.  A small fix for `bundlestate()` in `DefaultTemplateData()` was needed to remove mangled var refs from the bundle state.

tzz:
The state-based acceptance test functionality is really nice here--almost entirely data-driven.

tzz:
Docs updated as well.

libpromises/evalfunction.c
```diff
@@ -2931,7 +2931,10 @@ JsonElement *DefaultTemplateData(const EvalContext *ctx, const char *wantbundle)
             if (NULL != scope_obj)
             {
                 char *lval_key = VarRefToString(var->ref, false);
-                JsonObjectAppendElement(scope_obj, lval_key, RvalToJson(var->rval));
+                if (NULL == strchr(lval_key, '#')) // don't collect mangled refs
+                {
+                    JsonObjectAppendElement(scope_obj, lval_key, RvalToJson(var->rval));
+                }
```
kacfengine: I didn't quite understand this. Can you explain why this happens?

kacfengine:
I tried adding some test cases for failed parsing, like here: https://gist.github.com/kacfengine/0e0abe107a8b63061394, but this caused an infinite loop. Would be nice if you included it in your test.

kacfengine:
And btw you're right, the state based testing is really nice!

libpromises/evalfunction.c
tzz: `this#promise_filename` shows up in the bundle's variables (claims its scope is the bundle).  I don't know why it happens since I didn't write the variable-gathering code, but since we can't have variables with `#` in the name, the fix is good regardless of the underlying issue.

kacfengine: I vaguely recall that there are some exceptions to this rule, and '#' is used to mark those exceptions, but I don't remember any details (or even if I remember correctly..). I suppose we're ok if the tests pass, and this is anyway limited to the template data.

tzz:
OK; acceptance test expanded as you suggested.  That exposed a pre-existing bug in the YAML parser: it wouldn't break on `YAML_NO_EVENT` like it's supposed to (on malformed JSON).  Fixed; thanks for the suggestion.

tzz:
I think it makes sense to have a failed CSV parse return the empty array `[]` but for failed JSON and YAML parses, to fail the variable promise.

The reason is that CSV has no structure: anything can be CSV.  Whereas JSON and YAML have to obey strict rules to be parseable.

If you disagree, I can change that behavior.

tzz:
@nickanderson the data limit is 50MB for all formats now.

libpromises/evalfunction.c
tzz: `#` is used to replace `.` when variables are "mangled."  Do `git grep -i mangle` in core.git and prepare yourself :)

kacfengine:
> I think it makes sense to have a failed CSV parse return the empty array [] but for failed JSON and YAML parses, to fail the variable promise.
> 
> The reason is that CSV has no structure: anything can be CSV. Whereas JSON and YAML have to obey strict rules to be parseable.
> 
> If you disagree, I can change that behavior.

Nope, I agree as well. Empty CSV is still valid CSV, so the only possible error is being unable to open the file.

libpromises/evalfunction.c
kacfengine: Haha, you're right, I wish I hadn't. :-)
I suppose this is appropriate then, since the data for the bundle is not supposed to include any variables from outside the bundle. The ``this#promise_filename`` is an exception merely because it is in fact local to that bundle.

kacfengine:
trigger build

tzz:
Thank you!

PR #2163, opened 2015-04-07
Remove: Inventory tag on ipv4 addresses

nickanderson:
This has moved to policy so that undesired addresses like the local loopback
can be easily filtered out using a regular expression that is user accessible.

Ref: https://dev.cfengine.com/issues/7097

This should be merged with a change to masterfiles: https://github.com/cfengine/masterfiles/pull/444


nickanderson:
@estenberg for your review

estenberg:
Looks good.

PR #2164, opened 2015-04-07
Redmine#7098: let filestat() support xattr

tzz:
https://dev.cfengine.com/issues/7098

Adds a `xattr` parameter to `filestat()` which also will return SELinux attributes in addition to extended attributes.

```
bundle agent main
{
  vars:
      "files" slist => { "/etc/passwd", "/etc/shadow", "/tmp/t" };
      "info[$(files)]" string => filestat($(files), "xattr");
  reports:
      "$(this.bundle): $(files): xattr=$(info[$(files)])";
}
```

Note that `/etc/shadow` doesn't exist.

```console
% mkdir /tmp/t
% xattr -w foo bar /tmp/t
R: main: /etc/passwd: xattr=
R: main: /etc/shadow: xattr=$(info[/etc/shadow])
R: main: /tmp/t: xattr=foo=bar
```

I think it's fairly minimal.  I'd really like feedback, especially whether this is acceptable.

If accepted, it would also require modifying the `fileinfo` bundle in the masterfiles.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -29,6 +29,7 @@
 #include <dir.h>
 #include <dbm_api.h>
 #include <lastseen.h>
+#include <files_copy.h>
 #include <files_names.h>
 #include <files_interfaces.h>
 #include <files_hashes.h>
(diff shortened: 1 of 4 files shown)


tzz:
This has been tested to return SELinux contexts and also works on Mac OS X.

tzz:
ping?

tzz:
ping?  This is IMO a safe change, no change in existing functionality or touching evaluation etc.

libpromises/evalfunction.c
```diff
@@ -3152,6 +3153,69 @@ static FnCallResult FnCallFileStatDetails(ARG_UNUSED EvalContext *ctx,
...
+                int datasize = getxattr(path, current, data, sizeof(data));
+                if (datasize < 0)
+                {
+                    if (errno == ENOTSUP)
+                    {
+                        continue;
+                    }
+                    else
+                    {
+                        Log(LOG_LEVEL_INFO, "Can't read extended attribute '%s' of '%s'. (getxattr: %s)",
```
kacfengine: Should be ``LOG_LEVEL_ERR``.

libpromises/evalfunction.c
```diff
@@ -3152,6 +3153,69 @@ static FnCallResult FnCallFileStatDetails(ARG_UNUSED EvalContext *ctx,
...
+#if defined(WITH_XATTR)
+        // Extended attributes include both POSIX ACLs and SELinux contexts.
+        char attr_raw_names[CF_BUFSIZE];
+        ssize_t attr_raw_names_size = listxattr(path, attr_raw_names, sizeof(attr_raw_names));
+
+        if (attr_raw_names_size < 0)
+        {
+            if (errno != ENOTSUP && errno != ENODATA)
+            {
+                Log(LOG_LEVEL_INFO, "Can't read extended attributes of '%s'. (listxattr: %s)",
```
kacfengine: Should be ``LOG_LEVEL_ERR``.

libpromises/evalfunction.c
```diff
@@ -3152,6 +3153,69 @@ static FnCallResult FnCallFileStatDetails(ARG_UNUSED EvalContext *ctx,
...
+                else
+                {
+                  if (!StringIsPrintable(data))
+                  {
+                      Log(LOG_LEVEL_INFO, "Skipping extended attribute of '%s', it has non-printable data: '%s=%s'",
+                          path, current, data);
+                      continue;
+                  }
+
+                  BufferPrintf(printattr, "%s=%s", current, data);
```
kacfengine: If there is more than one entry, perhaps there should be newlines between them?

libpromises/evalfunction.c
```diff
@@ -3152,6 +3153,69 @@ static FnCallResult FnCallFileStatDetails(ARG_UNUSED EvalContext *ctx,
...
+                }
+            }
+
+            snprintf(buffer, CF_MAXVARSIZE, "%s", BufferData(printattr));
```
kacfengine: ``printattr`` leaks here.

kacfengine:
Nice! Do you think it's hard to make an acceptance test for this?

libpromises/evalfunction.c
tzz: OK.  I don't think of a system probe error as an actual error, that's why I had `INFO` before.

libpromises/evalfunction.c
tzz: OK

libpromises/evalfunction.c
tzz: Oh, right.  I had that originally and forgot to add it back after a minor refactoring!

libpromises/evalfunction.c

tzz:
I'll see about the acceptance test but the bugs are fixed.

tzz:
I added a simple acceptance test that works on Mac OS X (which has the xattr utility).  I couldn't figure it out for Linux but let me know if you do...

tzz:
@neilhwatson checked on this too.  The fundamental problem is that to test SELinux, you must run it.  So CFE has to run dedicated VMs with SELinux to test this properly.  I don't think I can write a useful acceptance test for this PR otherwise.

kacfengine:
Right. Don't worry about the acceptance test. I have modified it to report the proper result, and we will run it opportunistically! This is quite exotic platform support anyway.

Modified version: #2193.

tzz:
OK, moving to #2193

PR #2165, opened 2015-04-08
list_ifelse: new function

tzz:
This is a new function that builds slists similarly to `ifelse()` but accepting multiple true conditions and splitting each value on `\n`.  It's very convenient for cases where you want to say "build a list with elements x and y on Linux and element z everywhere".

Empty strings are ignored.

```
bundle agent main
{
  vars:
      "build" slist => list_ifelse("linux",
                                  "abc
def",

                                  "aix",
                                  "ignore",

                                  "darwin",
                                  "more
mac os x
arguments",
                                  "any",
                                  "any_argument_1",

                                  "any",
                                  "any_argument_2",

                                  "last
arguments");

      "build_str" string => format("%S", build);

  reports:
      "$(this.bundle): built list $(build_str)";
}
```

Output on Mac OS X:

```console
R: main: built list { "more", "mac os x", "arguments", "any_argument_1", "any_argument_2", "last", "arguments" }
```

If this is acceptable, I will make tests and docs.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -773,10 +773,11 @@ static FnCallResult FnCallIfElse(EvalContext *ctx,
                                  ARG_UNUSED const FnCall *fp,
                                  const Rlist *finalargs)
 {
+    bool list_mode = (0 == strcmp(fp->name, "list_ifelse"));
     int argcount = 0;
     char id[CF_BUFSIZE];
 
-    snprintf(id, CF_BUFSIZE, "built-in FnCall ifelse-arg");
+    snprintf(id, CF_BUFSIZE, "built-in FnCall %s-arg", fp->name);
 
     /* We need to check all the arguments, ArgTemplate does not check varadic functions */
     for (const Rlist *arg = finalargs; arg; arg = arg->next)
@@ -792,8 +793,10 @@ static FnCallResult FnCallIfElse(EvalContext *ctx,
     /* Require an odd number of arguments. We will always return something. */
     if (argcount%2 != 1)
     {
-        FatalError(ctx, "in built-in FnCall ifelse: even number of arguments");
+      FatalError(ctx, "in built-in FnCall %s: even number of arguments", fp->name);
     }
+
+    Rlist *ret = NULL;
 
     const Rlist *arg;
     for (arg = finalargs;        /* Start with arg set to finalargs. */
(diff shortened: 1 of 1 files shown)


nickanderson:
I can see how this would be useful in some cases. I assume it also works if I used $(const.n) instead of an actual newline:

```
"build" slist => list_ifelse("linux",
                                  "abc$(const.n)def",
```

tzz:
ping?

tzz:
I still think this would be useful but maybe there's no time for it for the next release...

kacfengine:
I'm working my way through the list, but yeah at this point I can't guarantee that everything you have submitted will make it for 3.7. If you think some PRs are more important, feel free to direct me to those first.

kacfengine:
I find the general list_ifelse concept useful, but I think this is not the right design. Nothing else in CFEngine depends on newlines in given parameters, and I think it breaks a bit with expectations. Another possibility would be to somehow solve this using regular '::' classes syntax in the middle of a list to split it, but this may be language wise a bit awkward too.

The ideal case would be to do it like this:

```
"my" slist => list_ifelse("myclass", { "item1", "item2" }, "myclass2", @(list));
```

But I suspect the list-as-argument syntax would require some substantial changes.

tzz:
I could make it an entirely new syntax.  The fundamental issue is that the current way to set slists incrementally is not convergent.  So we could try something else:

* a new `slist_append` vars promise attribute
* (my preferred way) a new inline syntax like

```
"mylist" slist => { linux:: "item1", "item2", !linux:: "item3" };
```
* something else?

We can discuss further here or in Redmine, let me know which you prefer.

kacfengine:
> The fundamental issue is that the current way to set slists incrementally is not convergent.

I agree, it goes against "the convergent way", and "the declarative way" as well.

> ``"mylist" slist => { linux:: "item1", "item2", !linux:: "item3" };``

Ah cool, that's even better than my suggestion! The syntax may be slightly ambiguous if we allow string contexts in there, since then it will look like a string, but I think as long as the '::' is outside we should be able to make the parser work it out.

Do you know if there already is a Redmine ticket for this? It's probably a good idea to move the discussion there. I should also mention that the cutoff for 3.7 features is likely to be somewhere towards the end of next week, which is even better reason to keep this in a ticket if it doesn't make it.

kacfengine:
Hmm, one technical problem I foresee with that proposal is that classes may change between passes, but CFEngine will not normally reevaluate a promise that has already been evaluated successfully. I'm not sure how to solve that off the top of my head...

tzz:
OK, clearly we should move this to Redmine, can you please make a ticket?  When you've created the ticket we can close this PR.  I don't think there's an existing one, though I recall some e-mail discussions.

To answer your question about changing classes: the `ifelse()` function has the same problem but is still useful.  We just tell users not to use classes that may change during bundle evaluation and it works out all right.

tzz:
https://dev.cfengine.com/issues/4335 is relevant but closed.

kacfengine:
Yes, relevant, but not specifically about ``list_ifelse``. I have created https://dev.cfengine.com/issues/7155 to discuss further.

PR #2166, opened 2015-04-08
Add: Test that copying files locally works

nickanderson:
Ref: https://dev.cfengine.com/issues/5576

### DIFF
--- a/tests/acceptance/10_files/copy_local_file.cf
+++ b/tests/acceptance/10_files/copy_local_file.cf
@@ -1,0 +1,48 @@
+bundle common test_meta
+{
+  vars:
+      "description" string => "Test that copying a file locally works";
+      "story_id" string => "5576";
+      "covers" string => "operational_repaired";
+}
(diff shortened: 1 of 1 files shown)


tests/acceptance/10_files/copy_local_file.cf
```diff
@@ -0,0 +1,48 @@
+bundle common test_meta
+{
+  vars:
+      "description" string => "Test that copying a file locally works";
+      "story_id" string => "5576";
+      "covers" string => "copy_local_file";
```
estenberg: covers should be "operational_repaired" here (but this string is still unused).

operational = not dry-run
repaired = you're testing that something not in compliance became in compliance

nickanderson: ok fixed.

kacfengine:
I feel confident enough that local file copy will work that I'm merging this straight away! :-)

PR #1241, opened 2014-01-06
libcfnet: bandwidth limit on the network

xrg:
This time, bwlimit is implemented with integer arithmetics only. It is written as a separate function, applied both to classic SendStream and TLS methods.

PR #2167, opened 2015-04-09
libcfnet: bandwidth limit on the network

kacfengine:
Original pull request #1241.


kacfengine:
@xrg: This looks good, and assuming it passes the test build, I will merge this. Do you mind writing some documentation for it here? https://github.com/cfengine/documentation/blob/master/reference/components.markdown

kacfengine:
trigger build

kacfengine:
Documentation in https://github.com/cfengine/documentation/pull/1013.

libcfnet/net.c
```diff
@@ -225,6 +225,90 @@ int ReceiveTransaction(const ConnectionInfo *conn_info, char *buffer, int *more)
...
+
+    if (!bwlimit_kbytes)
+    {
+        /* early return, before any expensive syscalls */
+        return;
+    }
+
+    if (pthread_mutex_lock(&bwlimit_lock) == 0)
+    {
+        clock_gettime(CLOCK_MONOTONIC, &clock_now);
```
kacfengine: It appears that ``CLOCK_MONOTONIC`` is Linux-specific. I propose that you use ``CLOCK_REALTIME`` instead and store the old value, so you can detect backwards jumps and reset the next time. If the clock is adjusted there may be small breaks or bursts, but this effect will be temporary.

kacfengine: @xrg: ^

xrg: > clock_gettime(CLOCK_MONOTONIC, &clock_now);

According to IEEE standards, CLOCK_MONOTONIC is not Linux specific, but 
"optional" for platforms [1] . There was no "Linux specific" marking in Linux's 
programming manual (man 2 clock_gettime)

Can you check non-Linux platforms (out of those you support) for this feature?

Still, I would prefer that we use this optimal solution for platforms 
supporting it, and #ifdef a workaround only for the handicapped ones that 
don't.


[1] 
http://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_getres.html

kacfengine: > Can you check non-Linux platforms (out of those you support) for this feature?

Linux: Supported
AIX: Supported
Solaris: Not supported
HP-UX: Not supported
Windows: Not supported

> Still, I would prefer that we use this optimal solution for platforms supporting it, and #ifdef a workaround only for the handicapped ones that don't.

Agreed.

Windows doesn't support ``clock_gettime`` at all, it seems. I'm ok with simply ifdeffing it out on that platform and give a warning that it is not supported. If you do the ``CLOCK_REALTIME`` bit for the other platforms, I can do the Windows part. Just push it to your original branch name, then let me know here and I can pick up from there.

xrg:
I have updated the "feature-bwlimit" branch to tip f5700ee09e69 , where clock jumping backward is addressed. Can you check that, please?

kacfengine:
Thanks!

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
Thanks, and sorry for the long delay in handling this!

PR #2168, opened 2015-04-09
[3.6.x] server_common: remove double stat() of file, on non-link paths

xrg:
In an audit to way server "stat" method is implemented, discovered that `sendbuffer` would be uninitialized in the /extremely rare/ case that 2nd `stat()` call would fail for the same regular file.

In fact, the 2nd call to `stat()` is redundant, at least on POSIX systems (according to documentation of `lstat()` ) . We can ommit the buggy lines altogether.

### DIFF
--- a/cf-serverd/server_common.c
+++ b/cf-serverd/server_common.c
@@ -1003,20 +1003,10 @@ int StatFile(ServerConnectionState *conn, char *sendbuffer, char *ofilename)
 
         cfst.cf_readlink = linkbuf;
     }
-#endif /* !__MINGW32__ */
-
-    if ((!islink) && (stat(filename, &statbuf) == -1))
-    {
-        Log(LOG_LEVEL_VERBOSE, "BAD: unable to stat file '%s'. (stat: %s)",
-            filename, GetErrorStr());
-        SendTransaction(conn->conn_info, sendbuffer, 0, CF_DONE);
-        return -1;
-    }
-
-    Log(LOG_LEVEL_DEBUG, "Getting size of link deref '%s'", linkbuf);
 
     if (islink && (stat(filename, &statlinkbuf) != -1))       /* linktype=copy used by agent */
     {
+        Log(LOG_LEVEL_DEBUG, "Getting size of link deref '%s'", linkbuf);
         statbuf.st_size = statlinkbuf.st_size;
         statbuf.st_mode = statlinkbuf.st_mode;
         statbuf.st_uid = statlinkbuf.st_uid;
@@ -1024,6 +1014,8 @@ int StatFile(ServerConnectionState *conn, char *sendbuffer, char *ofilename)
         statbuf.st_mtime = statlinkbuf.st_mtime;
         statbuf.st_ctime = statlinkbuf.st_ctime;
     }
+
+#endif /* !__MINGW32__ */
 
     if (S_ISDIR(statbuf.st_mode))
     {


cf-serverd/server_common.c
```diff
@@ -1003,20 +1003,10 @@ int StatFile(ServerConnectionState *conn, char *sendbuffer, char *ofilename)
 
         cfst.cf_readlink = linkbuf;
     }
-#endif /* !__MINGW32__ */
```
ediosyncratic: Moving this means skipping the following code on MINGW, which we didn't previously.
How is that right ?

xrg: islink = true;
can only be set within the `#ifndef __MINGW32__ ` block,, makes sense only on POSIX systems.
So, why have code checking that outside the block?

cf-serverd/server_common.c
```diff
@@ -1003,20 +1003,10 @@ int StatFile(ServerConnectionState *conn, char *sendbuffer, char *ofilename)
...
-
-    if ((!islink) && (stat(filename, &statbuf) == -1))
-    {
-        Log(LOG_LEVEL_VERBOSE, "BAD: unable to stat file '%s'. (stat: %s)",
-            filename, GetErrorStr());
-        SendTransaction(conn->conn_info, sendbuffer, 0, CF_DONE);
```
xrg: This is where `sendbuffer` would be left with content from previous response, btw.

cf-serverd/server_common.c
ediosyncratic: Ah !  Yes, now I understand :-)
Thanks for the explanation.

jimis:
trigger build

kacfengine:
Note that this is for 3.6.x, so "trigger build" probably won't work (the other repo branches will be out of sync).

@xrg: You should always target the master branch first. Fixes will be cherry-picked from there to 3.6.x afterwards.

xrg:
On Wednesday 22 of April 2015, Kristian Amlie wrote:
> @xrg: You should always target the master branch first. Fixes will be
> cherry-picked from there to 3.6.x afterwards.

IMHO the other way is a bit safer: patches against stable series are more 
likely to apply to feature-enhanced branches, than the other way around.

I would /advise/ to adopt a workflow, where bugs are fixed on the earliest 
series they affect. Then, a "git merge" of the stable -> master should always 
ensure that these bugs are eliminated on later branches.
The other way will mean cherry-picking each commit, could potentially miss 
some.

kacfengine:
Those are the two disciplines! :-)

In CFEngine we have adopted the "master first, stable series after" workflow for several reasons. The primary reason is ownership. The stable branch, 3.6.x, is owned by the person(s) doing the next release of core, so they decide what goes into the branch, and developers are not supposed to submit to it without getting agreement from them (currently me and @jimis, with @estenberg as tie breaker).

Another reason is practical, our test pipeline is currently optimized for developing code on top of master, and changing this is not trivial, so it's not something we will do unless we change our policy altogether.

That being said, I have worked with "the other discipline" successfully as well, and both ways have merits. The "stables series first" approach has the advantage that you can merge the stable branch into master instead of cherry-picking from it, which generally makes conflict resolution easier. The downside is that the branch must be owned by everyone.

jimis:
@kacfengine: good thing you noticed! @xrg since we're already strictly following the policy that @kacfengine mentions, it is hard to change now. Can you submit this change against master branch?

Closing.

jimis:
This was small and conflict-free, so I cherry-picked and re-submitted at: https://github.com/cfengine/core/pull/2186

PR #2169, opened 2015-04-09
Fix Redmine #7094 and add acceptance test

worr:
This takes the basic test that @kacfengine added in Redmine #7094 and expands it into a proper acceptance test. On master, this seems to pass.


worr:
I've updated this a bit to better reflect the discussion in the ticket.

worr:
I've added a fix to this problem as well, now that I'm back from vacation

kacfengine:
Thanks for locating the problem. I have added one small optimization in https://github.com/cfengine/core/pull/2183, I'll merge that one instead if it passes.

kacfengine:
Worked nicely!

PR #2171, opened 2015-04-14
Fix array-size inconsistency in mon_network.c

ediosyncratic:
The ECGSOCKS[] array size was given by a macro that had to be updated
to stay in sync with the actual contents of the array.  Growing the
array without updating the macro would have produced a compile-time
error.  Just for once, we recently shrank it; this lead to the array
containing NULL-filled entries.  Since code accessing it used the
size, rather than a check for NULL-filled entries, cf-monitord crashed
on a strcmp(, NULL).

Change the array to unspecified size, so the compiler gets it right
reliably; and redefine the size to be derived from the actual array.
This'll avoid needing to change the size define every time we add an
entry to the array.

### DIFF
--- a/cf-monitord/mon_network.c
+++ b/cf-monitord/mon_network.c
@@ -35,8 +35,6 @@
 
 /* Globals */
 
-#define ATTR        22
-
 Item *ALL_INCOMING = NULL;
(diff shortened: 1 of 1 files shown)


kacfengine:
trigger build

kacfengine:
I got some spurious errors, but I don't think it's related to this fix.

PR #2173, opened 2015-04-15
YAML: Make sure that key is reset when adding non-string elements.

kacfengine:
Otherwise CFEngine will assert when you try to mix key/values and
arrays inside another key/value and array.

Originally discovered and fixed approx. five min before presenting it at CfgMgmtCamp. :-)


kacfengine:
@tzz: For your consideration.

kacfengine:
trigger build

tzz:
Nice!

PR #2174, opened 2015-04-15
Use LMDB NOSYNC mode during test runs to improve performance.

kacfengine:


kacfengine:
Don't bother reviewing this just yet, I'd like to see if it improves running time first, if not, then a review is wasted.

kacfengine:
trigger build

kacfengine:
Nah, no difference, must be CPU or network bound.

kacfengine:
trigger build

kacfengine:
This still made no difference.

jimis:
reopening to send some tests run a bit later

jimis:
Let's see if now it makes any difference!

jimis:
trigger build

jimis:
HP-UX: 70 min
AIX:       67 min
Solaris10: 49 min

kacfengine:
Master last finished in 58 min on HP-UX, IOW this has no effect.

jimis:
HP-UX: 73 min
AIX: 62 min
Solaris10: 44 min

You're right, I guess the effect is minor.

jimis:
I don't really believe that SYNC is as fast as NOSYNC, it should be much slower. 

I'd just explain that as our most frequent offender (cf_lock) is already NOSYNC, and the rest of the databases are seldom accessed.

PR #2176, opened 2015-04-16
Redmine#7111: @if macro

tzz:
See https://dev.cfengine.com/issues/7111

* `@if after_version(a.b)` is hard-coded to exactly that format
* you can give multiple `@if` statements and they "stack up"
* `@endif` ends **any** `@if` state, and it's safe to give multiples

The acceptance tests should make it clear.  I'm not sure if the stacking behavior is best but I'd be OK with it as a user, since typically you'd only give one `@if` at a time.

### DIFF
--- a/libpromises/cf3lex.l
+++ b/libpromises/cf3lex.l
@@ -69,6 +69,8 @@ P.offsets.current += yyleng;
 
 %}
 
+%x if_ignore_state
+
 space      [ \t]+
 
 newline    ([\n]|[\xd][\xa])
@@ -76,6 +78,9 @@ newline    ([\n]|[\xd][\xa])
 line       ^.*$
 
 comment    #[^\n]*
+
+macro_if    ^@if\ version_after\([0-9]{1,10}\.[0-9]{1,10}\)
+macro_endif ^@endif
 
 promises   bundle
 
@@ -132,6 +137,80 @@ promise_type   [a-zA-Z_]+:
                           P.line_no++;
                           P.line_pos = 0;
                       }
+
+{macro_if}            {
+                        const char* version_text = yytext+18;
+                        ParserDebug("\tL:macro @if %d:version=%s\n", P.line_pos, version_text);
+                        {
+                          int request_major = 0;
+                          int request_minor = 0;
+                          int major = 0;
+                          int minor = 0;
+                          int patch = 0;
+
+                          if (P.if_depth > 0)
+                          {
+                            yyerror("fatal: nested @if macros are not allowed");
+                            return 0;
+                          }
+
+                          P.if_depth++;
+
+                          if (3 == sscanf(Version(), "%d.%d.%d", &major, &minor, &patch))
+                          {
+                            if (2 == sscanf(version_text, "%d.%d", &request_major, &request_minor))
+                            {
+                            if ((request_major > major) ||
+                                (request_major == major && request_minor > minor))
+                              {
+                                /* ignore to the next @endif */
+                                ParserDebug("\tL:macro @if %d:ignoring to next @endif or EOF\n", P.line_pos);
+                                BEGIN(if_ignore_state);
+                              }
+                              else /* we're OK with the version */
+                              {
+                                ParserDebug("\tL:macro @if %d:accepted to next @endif\n", P.line_pos);
+                              }
+                            }
+                            else
+                            {
+                              yyerror("fatal: macro @if requested an unparseable version");
+                              return 0;
+                            }
+                          }
+                          else
+                          {
+                            yyerror("fatal: Version() was unparseable");
+                            return 0;
+                          }
+                        }
+                      }
+
+{macro_endif}         {
+                        ParserDebug("\tL:macro @endif %d\n", P.line_pos);
+                        BEGIN(INITIAL);
+                        if (P.if_depth <= 0)
+                        {
+                          yyerror("fatal: @endif macros without a matching @if are not allowed");
+                          return 0;
+                        }
+                        P.if_depth--;
+                      }
+
+<if_ignore_state>{macro_endif} {
+                                 ParserDebug("\tL:macro @endif %d\n", P.line_pos);
+                                 BEGIN(INITIAL);
+                                 if (P.if_depth <= 0)
+                                 {
+                                   yyerror("fatal: @endif macros without a matching @if are not allowed");
+                                   return 0;
+                                 }
+                                 P.if_depth--;
+                               }
+
+<if_ignore_state>.*          {
+                               ParserDebug("\tL:inside macro @if, ignoring text:%s\n", yytext);
+                             }
 
 {promises}            {
                           /* Note this has to come before "id" since it is a subset of id */
@@ -298,6 +377,21 @@ promise_type   [a-zA-Z_]+:
                           return yytext[0];
                       }
 
+<if_ignore_state><<EOF>>     {
+                               if (P.if_depth > 0)
+                               {
+                                 yyerror("EOF seen while @if was waiting for @endif");
+                                 return 0;
+                               }
+                             }
+
+<<EOF>>     {
+              if (P.if_depth > 0)
+              {
+                yyerror("EOF seen while @if was waiting for @endif");
+              }
+              return 0; // loops forever without this
+            }
 
 %%
 
--- a/tests/acceptance/00_basics/macros/if.cf
+++ b/tests/acceptance/00_basics/macros/if.cf
@@ -1,0 +1,48 @@
+######################################################
+#
+#  Test that @if works
+#
+#####################################################
+
+body common control
+{
+      inputs => { "../../default.cf.sub" };
+      bundlesequence  => { default("$(this.promise_filename)") };
+      version => "1.0";
+}
+
+bundle common test
+{
+@if version_after(3.7)
+  classes:
+      "expected" expression => "any";
+@endif
+
+@if version_after(3.6)
+  classes:
+      "expected2" expression => "any";
+@endif
+
+@if version_after(2.100)
+  classes:
+      "expected_2_100" expression => "any";
+@endif
+
+@if version_after(300.700)
+  classes:
+      "not_expected" expression => "any";
+@endif
+}
+
+bundle agent check
+{
+  methods:
+      "" usebundle => dcs_passif_expected("expected,expected2,expected_2_100",
+                                         "not_expected",
+                                         $(this.promise_filename));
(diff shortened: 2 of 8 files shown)


libpromises/cf3lex.l
```diff
@@ -133,6 +138,57 @@ promise_type   [a-zA-Z_]+:
...
+                          int request_major = 0;
+                          int request_minor = 0;
+                          int major = 0;
+                          int minor = 0;
+                          int patch = 0;
+                          if (3 == sscanf(Version(), "%d.%d.%d", &major, &minor, &patch))
+                          {
+                            if (2 == sscanf(version_text, "%d.%d", &request_major, &request_minor))
+                            {
+                              if (request_major > major || request_minor > minor)
```
kacfengine: This expression will make ``@if after_version(2.8)`` ignore the block on 3.7. You'll need:

<pre>
if (request_major > major || (request_major == major && request_minor > minor))
</pre>

libpromises/cf3lex.l
```diff
@@ -133,6 +138,57 @@ promise_type   [a-zA-Z_]+:
...
+                              }
+                              else /* we're OK with the version */
+                              {
+                                ParserDebug("\tL:macro @if %d:accepted to next @endif or EOF\n", P.line_pos);
+                              }
+                            }
+                            else
+                            {
+                              yyerror("fatal: macro @if requested an unparseable version");
+                              exit(1);
```
kacfengine: Why exit on one and not the other? We should probably let it finish by returning error in the main parsing loop, or?

libpromises/cf3lex.l
```diff
@@ -133,6 +138,57 @@ promise_type   [a-zA-Z_]+:
...
+                          {
+                            yyerror("fatal: Version() was unparseable\n");
+                          }
+                        }
+                      }
+
+{macro_endif}         {
+                        ParserDebug("\tL:macro @endif %d\n", P.line_pos);
+                        BEGIN(INITIAL);
+                      }
```
kacfengine: Should this be an error if encountered on its own?

It's possible that I'm misunderstanding how the parsing works, but the next section (``<if_state>{macro_endif}``) is the path that's taken if we are both in the if_state and the endif is found, correct? So if a lone endif is found we should produce an error.

libpromises/cf3lex.l
```diff
@@ -298,7 +354,10 @@ promise_type   [a-zA-Z_]+:
                           return yytext[0];
                       }
 
-
+<if_state><<EOF>>     {
+                         ParserDebug("\tL:EOF, ending any macros:%d\n", P.line_pos);
+                         BEGIN(INITIAL);
+                      }
```
kacfengine: I think this should also be an error. We don't want unbalanced ifs.

tests/acceptance/00_basics/macros/if.cf
```diff
@@ -0,0 +1,42 @@
...
+body common control
+{
+      inputs => { "../../default.cf.sub" };
+      bundlesequence  => { default("$(this.promise_filename)") };
+      version => "1.0";
+}
+
+bundle common test
+{
+@if version_after(3.7)
```
kacfengine: Should include a 2.100 test case as well, following my earlier comment.

libpromises/cf3lex.l
kacfengine: Right, now I see that this is the stacking behavior you were talking about. I think it's best to avoid that. If you really want stackability they should at least balance with the endifs, but I'm ok with having no stackability at all (we should barf on it in that case).

kacfengine:
Some small issues, but overall nice work, Ted!

libpromises/cf3lex.l
tzz: OK, fixed.

libpromises/cf3lex.l
tzz: OK, fixed.

libpromises/cf3lex.l
tzz: All of the above fixed and tests added to ensure the stacking behavior is forbidden.

The EOF behavior when not in `if_ignored_state` was the only problem: I had to add `return 0` to avoid an endless loop.  I think I'm doing the right thing, just need confirmation.

libpromises/cf3lex.l
tzz: OK, fixed.

tests/acceptance/00_basics/macros/if.cf
tzz: Added, thanks.

tzz:
I think all the issues have been resolved, take a look at the tests.

libpromises/cf3lex.l
kacfengine: Sorry, I wasn't being very clear there, what I meant was that we shouldn't be exiting at all, but return an error to the calling function. There are no other code paths in ``cf3lex.l`` that call exit, and the best thing is to let the caller handle it.

For example, if you put invalid policy in ``/var/cfengine/inputs``, cf-serverd might try to load it, fail, and keep running with the old policy. But with this it would exit instead.

kacfengine: Goes for the other exits as well.

libpromises/cf3lex.l
```diff
@@ -298,6 +377,22 @@ promise_type   [a-zA-Z_]+:
...
+                             }
+
+<<EOF>>     {
+              if (P.if_depth > 0)
+              {
+                yyerror("EOF seen while @if was waiting for @endif");
+                exit(1);
+              }
+              return 0; // loops forever without this
+            }
```
kacfengine: When you say "loops forever without this", do you mean that the section ``<<EOF>>`` loops forever without ``return 0``, or the parser loops forever without an ``<<EOF>>`` section? If it's the former, then you can remove the whole section, since ``P.if_depth > 0`` can't be true while ``<if_ignore_state>`` is false, so all you're doing is returning 0.

kacfengine:
Nice tests!

kacfengine:
Giving it an early spin.

kacfengine:
trigger build

libpromises/cf3lex.l
tzz: Without the explicit `<<EOF>>` section, the lexer accepts cases like `/00_basics/macros/if_eof_without_endif2.x.cf`.

With the explicit `<<EOF>>` section but without an explicit `return 0`, the lexer never finishes parsing any of the tests.

I don't want to assume the `P.if_depth` logic is perfect, so I'd rather leave the check in.

tzz:
Amended and pushed with just the change from `exit 1` to `return 0` as requested.

libpromises/cf3lex.l
kacfengine: Ok!

kacfengine:
trigger build

kacfengine:
Build is green! I wonder where this is best documented... perhaps very top level: https://docs.cfengine.com/docs/master/reference-components.html ? What do you think?

tzz:
It will need docs in many places because it affects the stdlib, general usage, etc.

Are you ready for docs, then? :)

kacfengine:
Yep!

basvandervlies:
+1 for me and consider this pull request also for the 3.6.X branch

tzz:
OK, docs in https://github.com/cfengine/documentation/pull/1021

tzz:
@basvandervlies assuming we can backport this to the next 3.6 release, the docs in the PR say it was introduced in 3.6.6

kacfengine:
Sorry, no features in a patch release. So it will be kind of useless for 3.7, but I'm glad we got it in, this will be great to have for later versions. Thanks for the work, Ted.

I adjusted the documentation to say 3.7 and not 3.6.

basvandervlies:
@kacfengine. I not totally agree with this rejection for 3.6. This "feature" would  really be nice for 3.6 and has minimal impact. It will only skip the offending lines. It solve all the splitting into  several config files and duplicate code + the configuration code is easier to read.

kacfengine:
We want to release feature versions more often, not squeeze features into patch releases. The 3.6 series has been a stable series exactly because we reject changes like these. Let's not throw that away now. The 3.6 series is already quite far along in its patch number, and only important bug fixes will go in there now.

Note that the twice-a-year feature releases that I presented at the Configuration Management Camp is something we still intend to do, so six months should be the maximum time to wait for a new feature. I expect we will give a more formal update on release schedule soon.

basvandervlies:
@kacfengine We are on the same line with the 3.6. series. I only disagree on this patch/feature, because it would also make the upgrade from 3.6 clients to the 3.7 easy without the hassle of splitting config files because we want to want to use a new cfengine feature or wait till all clients have upgraded to 3.7. This patch will solve this kind of sysadmin problems and we can mix 3.6/3.7/...  environments. This is my last comment ;-)

libpromises/cf3lex.l
```diff
@@ -77,6 +79,9 @@ line       ^.*$
 
 comment    #[^\n]*
 
+macro_if    ^@if\ version_after\([0-9]{1,10}\.[0-9]{1,10}\)
```
naksu: Does the user get a clear or confusing error if there is, e.g., two spaces after if. The rest of the language is very lenient about whitespace; perhaps this should also be. For example, @if\s+version_after\s*\(\s*[0-9]{1,10}\.[0-9]{1,10}\s*\). Forcing @if to start at the beginning of the line is too strict, in my opinion.

libpromises/cf3lex.l
```diff
@@ -133,6 +138,80 @@ promise_type   [a-zA-Z_]+:
...
+                          {
+                            yyerror("fatal: Version() was unparseable");
+                            return 0;
+                          }
+                        }
+                      }
+
+{macro_endif}         {
+                        ParserDebug("\tL:macro @endif %d\n", P.line_pos);
+                        BEGIN(INITIAL);
```
naksu: Is the lexer already in INITIAL state at this point?

libpromises/cf3lex.l
tzz: I'd be OK with that change but it will be slightly harder to detect the boundaries mechanically.

libpromises/cf3lex.l
tzz: Yes.  From the flex docs:

> `BEGIN(0)` returns to the original state where only the rules with no start conditions are active. This state can also be referred to as the start-condition "INITIAL", so `BEGIN(INITIAL)` is equivalent to `BEGIN(0)`

PR #2178, opened 2015-04-20
Redmine #6446: Remove "method_" prefix from methods promises.

kacfengine:
It's more intuitive that the name matches the bundle exactly.


kacfengine:
trigger build

PR #2179, opened 2015-04-20
Bootstrap trust

jimis:


kacfengine:
Looks like both of us fixed many of the same problems. See in particular https://github.com/cfengine/core/commit/8ceaa0029ca4a62b6ef90abd2f082f0037ff9578 and https://github.com/cfengine/core/commit/14c2303e2698d434bf9eca53638ebdc3eccc0228 (in cf-key.c).

libpromises/Makefile.am
```diff
@@ -216,5 +216,5 @@ CLEANFILES = *.gcno *.gcda $(BUILT_SOURCES)
 #
 MOSTLYCLEANFILES = *~ *.orig *.rej
 
-bootstrap.inc:
+bootstrap.inc: failsafe.cf
```
kacfengine: Heh, a lot simpler than my solution: https://github.com/cfengine/core/commit/e51703ae4d9bc1971c5316aa2abe81ed09b22e82. I think this should be ok, as long as tar.gz keeps timestamps intact, which I believe it does.

cf-agent/cf-agent.c
```diff
@@ -301,9 +304,11 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
      */
     bool cfruncommand = false;
 
-    while ((c = getopt_long(argc_new, argv_new, "tdvnKIf:D:N:VxMB:b:hlC::E", OPTIONS, NULL)) != EOF)
+    while ((c = getopt_long(argc_new, argv_new, "tdvnKIf:D:N:VxMB:b:hlC::ET::",
```
kacfengine: Double colon means optional argument, whereas in the long option struct you have put ``required_argument``.

libpromises/failsafe.cf
```diff
@@ -252,6 +252,7 @@ body copy_from u_scp(from)
 {
       source      => "$(from)";
       compare     => "digest";
+    trust_server::
```
kacfengine: +1 for the trust_server class! Nice way of doing it. I would put a comment here though, explaining how it is set.

(Recommenting in PR instead of in commit)

cf-agent/cf-agent.c
jimis: You're right, I wasn't sure which one I wanted. I ended up with optional argument, so I'll change the struct, thanks!

jimis:
Updated with separate commits. I'll rebase though, since it's not mergeable!

jimis:
So many conflicts, indeed we were fixing the exact same stuff!

jimis:
Rebased, removed commit fixing already fixed stuff, squashed-in the review corrections.

kacfengine:
Nice!

kacfengine:
trigger build

kacfengine:
Dammit, master appears broken for other reasons, something slipped through the cracks. But all the other core tests passed, so I'll merge this.

PR #2180, opened 2015-04-20
Add: Acceptance test for getvalues

nickanderson:
Getvalues should return similar results when used on a datacontainer and 
similar classic array.

Ref: https://dev.cfengine.com/issues/7116


kacfengine:
trigger build

kacfengine:
trigger build

PR #2185, opened 2015-04-23
Redmine #6994: Fix HPUX LMDB corruption due to non-unified file cache.

kacfengine:
On HP-UX, a unified file cache was not introduced until version 11.31.
This means that on 11.23 there are separate file caches for mmap()'ed
files and open()'ed files. When these two are mixed, changes made using
one mode won't be immediately seen by the other mode, which is an
assumption LMDB is relying on. The MDB_WRITEMAP flag causes LMDB to use
mmap() only, so that we stay within one file cache.


kacfengine:
trigger build

PR #2186, opened 2015-04-23
server_common: remove double stat() of file, on non-link paths

jimis:
The algorithm first calls `lstat()`, which should bring all required
fields in the plain-file case.
Only on posix systems, and when `lstat()` indicates we have a symlink,
we need to call `stat()` again to read destination fields.

Note that the removed lines contain a bug: in the extremely rare case
that a regular file vanishes between first and second `stat()` calls,
we would be sending an uninitialized sendbuffer!


jimis:
trigger build

jimis:
trigger build

jimis:
All green, merging. Thanks @xrg for the fix!

PR #2187, opened 2015-04-23
Add: Acceptance test for parsejson when there are undefined vars

nickanderson:
Ref: https://dev.cfengine.com/issues/7130


kacfengine:
The original bug is rejected. Closing this.

PR #2188, opened 2015-04-23
4k/match policy

worr:
This finishes off the feature of having insert_lines not care about the 4k limit

### DIFF
--- a/cf-agent/files_editline.c
+++ b/cf-agent/files_editline.c
@@ -1401,16 +1401,22 @@ static int MatchPolicy(EvalContext *ctx, const char *camel, const char *haystack
     Rlist *rp;
     char *sp, *spto, *firstchar, *lastchar;
     InsertMatchType opt;
-    char work[CF_BUFSIZE], final[CF_BUFSIZE];
     Item *list = SplitString(camel, '\n'), *ip;
     int direct_cmp = false, ok = false, escaped = false;
+    char *final = NULL;
+    size_t final_size = 0;
+    char *work = NULL;
+    size_t work_size = 0;
 
-//Split into separate lines first
-
+    //Split into separate lines first
     for (ip = list; ip != NULL; ip = ip->next)
     {
         ok = false;
         direct_cmp = (strcmp(camel, haystack) == 0);
+        final = xstrdup(ip->name);
+        final_size = strlen(final) + 1;
+        work = NULL;
+        work_size = final_size;
 
         if (insert_match == NULL)
         {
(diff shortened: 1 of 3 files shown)


cf-agent/files_editline.c
```diff
@@ -1441,27 +1444,33 @@ static int MatchPolicy(EvalContext *ctx, const char *camel, const char *haystack
...
-                // Strip initial and final first
-
-                for (firstchar = final; isspace((int)*firstchar); firstchar++)
+                // Need to escape the original string once here in case it contains regex chars when non-exact match
+                // Check size of escaped string, and realloc if necessary
+                size_t escape_regex_len = EscapeRegexCharsLen(ip->name);
+                if (escape_regex_len + 1 > final_size)
                 {
+                    final = xrealloc(final, escape_regex_len + 1);
+                    final_size = escape_regex_len;
```
kacfengine: You're not being consistent here. Earlier you did:

<pre>
final = xstrdup(ip->name);
final_size = strlen(final) + 1;
</pre>

..which would make ``final_size`` exactly the size of the buffer, including the '\0', but here you assign it the length of the string. Should probably be changed to ``final_size = escape_regex_len + 1;``.

cf-agent/files_editline.c
```diff
@@ -1470,34 +1479,76 @@ static int MatchPolicy(EvalContext *ctx, const char *camel, const char *haystack
...
                     {
-                        *spto++ = *sp;
+                        required_size++;
+                        toadd[0] = *sp;
+                        toadd[1] = '\0';
+                    }
+
+                    if (required_size > work_size)
+                    {
+                        work_size = required_size + 12;
```
kacfengine: I take it that this ``+ 12``, and the ``+ 6`` further up is simply heuristics, expecting some characters to replaced, but not too many, and then expanding if necessary?

cf-agent/files_editline.c
```diff
@@ -1470,34 +1479,76 @@ static int MatchPolicy(EvalContext *ctx, const char *camel, const char *haystack
...
+                        work = xrealloc(work, work_size);
+                    }
+
+                    if (strlcat(spto, toadd, work_size) >= work_size - 1)
```
kacfengine: It's a bit hard to follow the logic, but I think there may be a problem here: After this, the for loop ends and you increment ``spto`` by one. This will work if the character wasn't a space. However, if it was, and you added ``\s+`` here, ``spto`` will only have moved forward by one character, and the next write will overwrite the ``s``. Surely not the intention!

kacfengine:
Nice!

cf-agent/files_editline.c
worr: Yes, though that should be documented. I'll add a comment explaining that.

kacfengine:
@worr: FYI, github doesn't inform me when you update your branch so make sure you post a comment as well to get my attention! I just happened to notice.

kacfengine:
trigger build

kacfengine:
Thanks! Another 4K bug squashed!

PR #2190, opened 2015-04-24
A key must now be loaded at the beginning of the agent run.

jimis:
For example, if we try to initialise a TLS connection but a key is not
loaded, we will fail. Previously, it was possible to create a key
*while* the system was converging, so the first iteration it would fail,
the second iteration it would succeed.

The reason for changing this is that proper TLS initialisation involves
much more than just reading the keys (for example we need to set ciphers
and TLS version to be used), thus it can only happen during program
initialisation where that information (usually configured in body common
control) is easily accessible.

Most obvious side-effect in practice: "cf-agent --bootstrap".

If you have no key generated while bootstrapping, the bootstrap
procedure can't generate one for you, but it should inform you to
generate on by running cf-key, and then retry to bootstrap.

### DIFF
--- a/cf-runagent/cf-runagent.c
+++ b/cf-runagent/cf-runagent.c
@@ -68,7 +68,8 @@ static void ThisAgentInit(void);
 static GenericAgentConfig *CheckOpts(int argc, char **argv);
 
 static void KeepControlPromises(EvalContext *ctx, const Policy *policy);
-static int HailServer(EvalContext *ctx, char *host);
+static int HailServer(const EvalContext *ctx, const GenericAgentConfig *config,
+                      char *host);
 static void SendClassData(AgentConnection *conn);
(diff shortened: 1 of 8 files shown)


jimis:
This important change is needed for the features described at [redmine 3690](https://dev.cfengine.com/issues/3690) and [redmine 6883](https://dev.cfengine.com/issues/6883).

@nickanderson  what would you say about changing our failsafe.cf to detect if we have keys installed, and just abort with a message to run "cf-key" if there are no keys?

kacfengine:
Also note that packages' postinstall scripts are supposed to create a key for you, so this should not affect automated installs/bootstraps.

jimis:
trigger build

jimis:
I believe the feedback about this change was mostly positive. However a lot more changes are needed after all to fix the failing enterprise tests. 

I'll resubmit this together with a bunch of other changes once it's done, closing.

PR #2191, opened 2015-04-24
Add: Acceptance test for abortbundleclasses

nickanderson:
Test that the agent aborts the current bundle when an exact abortclass is
matched and that the agent continues on to the next bundle.

Test that the agent aborts the current bundle when a regular expression is
matched and that the agent continues on to the next bundle.

### DIFF
--- a/cf-serverd/server_common.c
+++ b/cf-serverd/server_common.c
@@ -1010,20 +1010,10 @@ int StatFile(ServerConnectionState *conn, char *sendbuffer, char *ofilename)
 
         cfst.cf_readlink = linkbuf;
     }
-#endif /* !__MINGW32__ */
-
-    if ((!islink) && (stat(filename, &statbuf) == -1))
-    {
-        Log(LOG_LEVEL_VERBOSE, "BAD: unable to stat file '%s'. (stat: %s)",
-            filename, GetErrorStr());
-        SendTransaction(conn->conn_info, sendbuffer, 0, CF_DONE);
-        return -1;
-    }
-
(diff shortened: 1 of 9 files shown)


kacfengine:
trigger build

kacfengine:
I'm a bit confused, all tests have ``"test_soft_fail" string => "!any"``, which means that they all need to pass (note "!any"). Does that mean that the bug is fixed? In that case the ticket should be closed, and ``test_soft_fail`` should be removed entirely.

The build was green, btw.

kacfengine:
Ping.

nickanderson:
Sorry didnt see the ping.

This was just a test to prove that abortbundleclasses work (it does, but there was no test).

I can remove the test_soft_fail. I just set it to !any, because it should fail softly nowhere, basically a no-op.

kacfengine:
I guess it's ok to leave the soft_fail in, it's actually a nice way to document which ticket is associated with the test case. But the ticket is still open, shouldn't it be closed if the test passes?

kacfengine:
@pasinskim said he found the "!any" soft failures confusing, so maybe they should be removed instead. It's easy to miss that it says "!any".

nickanderson:
yeah I kind of liked the documentation aspect of the !any, but I will remove it since you ask.

It could be !cfengine instead of !any but i don't suppose that would be very different.

nickanderson:
UGH! will re-open

nickanderson:
Replaced by https://github.com/cfengine/core/pull/2359

PR #2192, opened 2015-04-24
Redmine #7030: Don't hang when operating on fifos

worr:
The bug is caused by basic fifo semantics. CFEngine uses `openat(2)` to get an fd to `fchown(2)`. However, `open(2)`ing a fifo `RDONLY` blocks until at least one other process has opened it `WRONLY`.

The fix will involve using `chown(2)` rather than `fchown(2)`, since we cannot act on an fd reliably in this case. Fix attached.

### DIFF
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -453,6 +453,13 @@ int safe_open(const char *pathname, int flags, ...)
                 return -1;
             }
 
+            // We can't safely open(2) a FIFO without blocking infinitely
+            if (S_ISFIFO(stat_before.st_mode))
+            {
+                close(currentfd);
+                return -1;
+            }
+
             TEST_SYMLINK_SWITCH_POINT
 
             if (!next_component)
@@ -665,6 +672,13 @@ int safe_chown(const char *path, uid_t owner, gid_t group)
     int fd = safe_open(path, 0);
     if (fd < 0)
     {
+        // Do additional checking for FIFOs
+        struct stat st;
+        if (stat(path, &st) == 0 && S_ISFIFO(st.st_mode))
+        {
+            return chown(path, owner, group);
+        }
+
         return -1;
     }
 
@@ -739,6 +753,13 @@ int safe_chmod(const char *path, mode_t mode)
     int fd = safe_open(path, 0);
     if (fd < 0)
     {
+        // Do additional checking for FIFOs
+        struct stat st;
+        if (stat(path, &st) == 0 && S_ISFIFO(st.st_mode))
+        {
+            return chmod(path, mode);
+        }
+
         return -1;
     }
 
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -26,6 +26,7 @@
 #include <dir.h>
 
 #include <alloc.h>
+#include <libgen.h>
 #include <logging.h>
 
 bool FileCanOpen(const char *path, const char *modes)
@@ -657,6 +658,32 @@ int safe_chdir(const char *path)
 #endif // !__MINGW32__
 }
 
+int safe_chown_impl(const char *path, uid_t owner, gid_t group, int flags)
+{
+    int dirfd = -1;
+    int ret = -1;
+
+    char *parent_dir_alloc = xstrdup(path);
+    char *leaf_alloc = xstrdup(path);
+    char *parent_dir = dirname(parent_dir_alloc);
+    char *leaf = basename(leaf_alloc);
+
+    if ((dirfd = safe_open(parent_dir, O_RDONLY)) == -1)
+        goto bad;
+
+    if ((ret = fchownat(dirfd, leaf, owner, group, flags)))
+        goto bad;
+
+bad:
+    free(parent_dir_alloc);
+    free(leaf_alloc);
+
+    if (dirfd != -1)
+        close(dirfd);
+
+    return ret;
+}
+
 /**
  * Use this instead of chown(). It changes file owner safely, using safe_open().
  * @param path Path to operate on.
@@ -669,22 +696,7 @@ int safe_chown(const char *path, uid_t owner, gid_t group)
 #ifdef __MINGW32__
     return chown(path, owner, group);
 #else // !__MINGW32__
-    int fd = safe_open(path, 0);
-    if (fd < 0)
-    {
-        // Do additional checking for FIFOs
-        struct stat st;
-        if (stat(path, &st) == 0 && S_ISFIFO(st.st_mode))
-        {
-            return chown(path, owner, group);
-        }
-
-        return -1;
-    }
-
-    int ret = fchown(fd, owner, group);
-    close(fd);
-    return ret;
+    return safe_chown_impl(path, owner, group, 0);
 #endif // !__MINGW32__
 }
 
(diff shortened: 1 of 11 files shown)


worr:
I've attached a fix for this problem

libutils/file_lib.c
```diff
@@ -665,6 +672,13 @@ int safe_chown(const char *path, uid_t owner, gid_t group)
     int fd = safe_open(path, 0);
     if (fd < 0)
     {
+        // Do additional checking for FIFOs
+        struct stat st;
+        if (stat(path, &st) == 0 && S_ISFIFO(st.st_mode))
+        {
+            return chown(path, owner, group);
+        }
```
kacfengine: The problem with this is that it introduces a security hole, or actually two security holes:

1. Both ``stat``, ``chmod`` and ``chown`` will gladly follow any symlink that the user has inserted in the middle of the path, so if you're operating on a user directory, he can set the owner/permissions of any fifo on the system.
2. There is a race condition between the ``stat`` and the ``chown``, where the user can switch the file after it has been stat'ed with a different symlink, and using this he can change the owner/permissions of any file on the system, not just fifos.

The first vulnerability is easy to exploit, the second is harder, but still possible.

The correct way to solve this problem for chown is to use ``safe_open`` to open the directory that contains the final path component, and then use ``fstatat(safe_opened_fd, path_basename, &stat, AT_SYMLINK_NOFOLLOW)`` as well as ``fchownat(safe_opened_fd, path_basename, owner, group, AT_SYMLINK_NOFOLLOW)``, to prevent dangerous symlinks from being followed along the way.

Unfortunately, the problem *can't* be fixed for chmod without introducing a security hole, since the ``fchmodat`` call doesn't honor ``AT_SYMLINK_NOFOLLOW``. Since this is used extensively in CFEngine, in all operations that involve file manipulations in untrusted directories, we have to stay with the secure option.

We can avoid the hanging though, using the topmost patch, but we can't use that to change the fifo later.

I believe this there is an earlier ticket where this was dicussed as well, but I can't find it now.

kacfengine: > Unfortunately, the problem can't be fixed for chmod without introducing a security hole, since the ``fchmodat`` call doesn't honor ``AT_SYMLINK_NOFOLLOW``.

Actually, it can be fixed if we resort to dropping effective privileges temporarily to the owner of the fifo, and group "nobody". A user can still switch the file with a symlink, but it will only succeed if he points it at one of his own files.

But this is more work. I'm perfectly ok with keeping only the part that fixes the hanging.

kacfengine: So conceptually something like (not correct syntax/API, but you get the idea):

<pre>
fd = safe_open(dirname_path, O_RDONLY);
fstatat(fd, basename_path, &stat, AT_SYMLINK_NOFOLLOW);
gid = getgrnam("nogroup");
if (!gid) {
    gid = getgrnam("nobody"); // One of those should exist on all systems.
}
if (!gid) {
    return -1;
}
seteuid(stat.st_uid);
setegid(gid);
fchmodat(fd, basename_path, 0755)
seteuid(0); // Actually should save the old value.
setegid(0);
</pre>

worr:
Non-Linux OSes handle `AT_SYMLINK_NOFOLLOW` just fine. Maybe implement a basic wrapper around `fchmodat(2)` and only link it in if we're running on a broken OS? We already have to do this for OSes that don't have `fchownat(2)`.

In the case of Linux, we can instead implement this by doing a `openat(..., O_RDWR, AT_NO_FOLLOW_SYMLINKS)` followed by an `fchmod(2)`. This will work on fifos, since Linux will not block when doing an `open(2)` with `O_RDWR` on a fifo. That behavior is linux-specific, so in the case of a portable `fchmodat(2)`, we'd have to implement slightly different behavior. Since we don't already have a portable `fchmodat(2)`, I doubt that it's a problem that we'd have a linux-specific impl.

worr:
Ahh, I see we don't use `fchmodat(2)` and that AIX 6.*, OS X and possibly Solaris 10 don't support it. In that case, we could just abort on fifos in our implementation? I don't like the failure mode of sometimes following symlinks vs. failing on fifos.

kacfengine:
Being able to do ``openat()`` followed by ``fchmod()`` on Linux is a good find. I believe we should still do that, since it improves behavior, even if not on all platforms.

What do you think about the ``seteuid()`` approach though?

worr:
The `seteuid(2)` approach is certainly interesting and solves the problem, but I'm worried about the non-obvious semantics. Whereas most of the time we're very explicit about whether or not we follow and operate on symlinks, this particular implementation *sometimes* follows symlinks (iff they're owned by the same user as the link itself).

It probably works well enough for AIX ≤ 6 and Solaris ≤ 10 though.

kacfengine:
> Whereas most of the time we're very explicit about whether or not we follow and operate on symlinks, this particular implementation sometimes follows symlinks (iff they're owned by the same user as the link itself).

Yes, but this is already the case for most of CFEngine's file operations, and certainly those for file promises. Due to using ``safe_open``, CFEngine will never follow a link whose target has a different owner, so this will just extend on that.

worr:
Ah, I guess I was conflating `safe_open(..., O_NOFOLLOW)` with the default case of `safe_open()`. In that case, that renders my problem irrelevant.

There are the problems of errant files being owned by `nobody` and having write permissions, which happens, even though it's bad practice. I do think that that's good for the OSes that don't support `fchmodat(2)` at all though.

kacfengine:
Indeed, but I just realized that group permission makes no difference for chmod calls, since you still have to be the actual owner of the file, so group is irrelevant, we can just use ``seteuid()`` and that'll be enough.

worr:
I've updated this with an implementation of `safe_chown` and `safe_chmod` that use `fchownat(2)` and `fchmodat(2)`. I've also added libcompat functions for `fchmodat(2)` for OSes that don't have it, as well as a workaround for Linux's broken implementation.

kacfengine:
I realize that we are dealing with a particularly poor aspect of POSIX here. There are a couple of problems with the current approach:

* ``lchmod`` only exists for BSD derivatives, it seems. Platforms like HP-UX, which don't have ``fchmodat``, also don't have ``lchmod``.
* The approach of opening files with ``O_RDONLY`` breaks chmod operations if cf-agent is run as a non-root user. He may not have the rights to open a file read-only, but may still be able to change its permissions. I'm tempted to say this is less important though.
* The ``safe_open`` -> ``fchmodat`` sequence in ``safe_chmod`` does not protect against symlink attacks, because you are not using ``AT_SYMLINK_NOFOLLOW``. Of course normal ``chmod`` is supposed to follow links, but in CFEngine's case we need to put extra restrictions that the symlink and its target are owned by the same UIDs/GIDs. An alternative approach here would be to call ``readlink`` first, to see if it is a link, then follow it, and retry the whole operation on the new path it produces, making sure the owner is the same.

However, particularly in light of the last point, I really believe that the ``seteuid()`` approach will be much simpler here. Open the parent dir in ``safe_chmod``, check whether the leaf file is a FIFO (without following links), and if it is, switch EUID to that user and proceed normally. If it is not a FIFO, use the old codepath. The same can be done for ``chown``.

The fact that POSIX lets root blindly follow symlinks pointing to different users is quite terrible, if you ask me.

worr:
In light of these, I'll implement the seteuid approach within `fchmodat(2)`. `fchmodat(2)` and `fchownat(2)` serve us very well if here iff they're implemented. It seems much saner to implement compat functions if they aren't available, or if they weren't implemented correctly.

kacfengine:
Yep, I agree.

PR #2194, opened 2015-04-29
Add support for adding/removing fifos in CFEngine

worr:
It would be nice if CFEngine could create and manage FIFOs. This functionality could be extended to include doors as well on Solaris, as well as other special files.

### DIFF
--- a/cf-agent/verify_files_utils.c
+++ b/cf-agent/verify_files_utils.c
@@ -3779,6 +3779,52 @@ bool CfCreateFile(EvalContext *ctx, char *file, const Promise *pp, Attributes at
             return false;
         }
     }
+    else if (attr.file_type && !strncmp(attr.file_type, "fifo", 5))
+    {
+#ifndef _WIN32
+        if (!DONTDO)
+        {
+            mode_t saveumask = umask(0);
+            mode_t filemode = 0600;
+
+            if (PromiseGetConstraintAsRval(pp, "mode", RVAL_TYPE_SCALAR) == NULL)
+            {
+                /* Relying on umask is risky */
+                filemode = 0600;
+                Log(LOG_LEVEL_VERBOSE, "No mode was set, choose plain file default %04jo", (uintmax_t)filemode);
+            }
+            else
+            {
+                filemode = attr.perms.plus & ~(attr.perms.minus);
+            }
+
+            MakeParentDirectory(file, attr.move_obstructions);
+
+            char errormsg[CF_BUFSIZE];
+            if (!mkfifo(file, filemode))
+            {
+                snprintf(errormsg, sizeof(errormsg), "(mkfifo: %s)", GetErrorStr());
+                cfPS(ctx, LOG_LEVEL_ERR, PROMISE_RESULT_FAIL, pp, attr, "Error creating file '%s', mode '%04jo'. %s",
+                     file, (uintmax_t)filemode, errormsg);
+                umask(saveumask);
+                return false;
+            }
+
+            umask(saveumask);
+            return true;
+        }
+        else
+        {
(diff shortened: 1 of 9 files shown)


kacfengine:
I think before think can be truly useful, https://github.com/cfengine/core/pull/2192 will have to be addressed.

kacfengine:
I think now this can be considered again.

cf-agent/verify_files_utils.c
```diff
@@ -3694,6 +3694,44 @@ bool CfCreateFile(EvalContext *ctx, char *file, const Promise *pp, Attributes at
...
+            }
+
+            MakeParentDirectory(file, attr.move_obstructions);
+
+            char errormsg[CF_BUFSIZE];
+            if (!mkfifo(file, filemode))
+            {
+                snprintf(errormsg, sizeof(errormsg), "(mkfifo: %s)", GetErrorStr());
+                umask(saveumask);
+                return false;
```
kacfengine: ``errormsg`` is not used anywhere. Remove?

cf-agent/verify_files_utils.c
```diff
@@ -3694,6 +3694,44 @@ bool CfCreateFile(EvalContext *ctx, char *file, const Promise *pp, Attributes at
...
+
+            umask(saveumask);
+            return true;
+        }
+        else
+        {
+            cfPS(ctx, LOG_LEVEL_ERR, PROMISE_RESULT_WARN, pp, attr, "Warning promised, need to create fifo '%s'", file);
+            *result = PromiseResultUpdate(*result, PROMISE_RESULT_WARN);
+            return false;
+        }
```
kacfengine: Can you put a ``#ifndef _WIN32`` around the whole block (starting from the ``DONTDO`` statement), and print a warning that the operation is unsupported on Windows?

tests/acceptance/10_files/07_delete_lines/700.cf
```diff
@@ -0,0 +1,38 @@
...
+{
+  classes:
+    "exists" expression => fileexists("$(G.testfile)");
+    "ok" not => "exists";
+
+  reports:
+    ok::
+      "$(this.promise_filename) Pass";
+    !ok::
+      "$(this.promise_filename) Fail";
```
kacfengine: Should be ``FAIL``.

tests/acceptance/10_files/09_insert_lines/400.cf
```diff
@@ -0,0 +1,43 @@
...
+{
+  classes:
+    "exists" expression => fileexists("$(G.testfile)");
+    "ok" not => "exists";
+
+  reports:
+    ok::
+      "$(this.promise_filename) Pass";
+    !ok::
+      "$(this.promise_filename) Fail";
```
kacfengine: Should be ``FAIL``.

kacfengine:
I'm not quite clear on what will happen in the final test. If you try to edit a fifo, then I presume CFEngine will read whatever's on the pipe, however it will create a new, regular file for editing, and will then replace the fifo with the edited file. Editing a fifo seems a bit wrong to me, perhaps we should disallow it instead? What do you think?

worr:
Addressed the inline comments.

re: erroring, we do error if we try to edit a fifo. Both tests try and create a file before they edit it, and since we error because we're trying to edit a fifo, the file does not get tested.

kacfengine:
Ah ok, now I understand.

Looks good, I will proceed with testing then. Do you mind putting in a word about this here? https://github.com/cfengine/documentation/blob/master/reference/promise-types/files.markdown

kacfengine:
trigger build

kacfengine:
All green, good stuff! Thanks!

PR #2195, opened 2015-04-29
4k/execresult

worr:
This is a bit of a hack to remove the 4k limit from `execresult`. Unfortunately, this does change the prototype of `GetExecOutput`. It now requires that the caller dynamically allocate `buffer`. No other functions rely on it, so it doesn't seem to be that big of a problem.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -1600,9 +1600,10 @@ static FnCallResult FnCallExecResult(ARG_UNUSED EvalContext *ctx, ARG_UNUSED con
         return FnFailure();
     }
 
-    char buffer[CF_EXPANDSIZE];
+    size_t buffer_size = CF_EXPANDSIZE;
+    char *buffer = xcalloc(1, buffer_size);
 
-    if (GetExecOutput(RlistScalarValue(finalargs), buffer, shelltype))
+    if (GetExecOutput(RlistScalarValue(finalargs), &buffer, &buffer_size, shelltype))
     {
         Log(LOG_LEVEL_VERBOSE, "%s ran '%s' successfully", fp->name, RlistScalarValue(finalargs));
         return FnReturn(buffer);
--- a/tests/acceptance/01_vars/02_functions/execresult_4k.cf
+++ b/tests/acceptance/01_vars/02_functions/execresult_4k.cf
@@ -1,0 +1,38 @@
+#######################################################
+#
+# Test returnszero()
+#
+#######################################################
+
+body common control
+{
+      inputs => { "../../default.cf.sub" };
+      bundlesequence  => { default("$(this.promise_filename)") };
+      version => "1.0";
+}
+
+bundle agent init
+{
+}
+
+bundle agent test
+{
+  vars:
+    any::
+      "cwd" string => dirname("$(this.promise_filename)");
+      "cat_4k" string => "$(G.cat) $(cwd)$(const.dirsep)text_xform.cf.txt";
+      "result" string => execresult("$(cat_4k)", "useshell");
+
+  classes:
+    any::
+      "ok" expression => regcmp(".*3456789abcdefghij$", "$(result)");
+
+  reports:
+    DEBUG::
+      "$(result)";
+    ok::
+      "$(this.promise_filename) Pass";
+    !ok::
+      "$(this.promise_filename) FAIL";
+
+}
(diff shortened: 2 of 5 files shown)


worr:
There's a bug here, where lines greater than 16k will silently truncate. Need to fix before merge.

worr:
Updated to remove previously mentioned bug.

kacfengine:
This is great! However, I'm thinking that we should probably have *some* kind of upper limit, to prevent runaway processes from draining the memory completely, even if that limit is quite high. Maybe something like 50MB or 100MB?

worr:
I brought this up in some other bug, iirc. Yes, I totally agree. I think we should probably define a hard upper limit globally, and use that in subsequent patches where memory could runaway. Something suitably large like 100MB works for me, but I'd like to hear other opinions.

kacfengine:
I think 100MB is a good limit, it's high, but not so high that it's likely to disturb other services on the system. And centralizing the limit also sounds good, that will make it easier to change the limit if we need a different limit on an embedded system for example.

There are probably other candidates for using this global limit as well.

kacfengine:
Btw, the recently merged ``readdata`` function can probably make use of the global limit: https://github.com/cfengine/core/blob/master/libpromises/evalfunction.c#L5714

basvandervlies:
@worr and @kacfengine. Is this a hard limit or can we configure this with an option in body agent control? I am not a big fan of hard coded limits.

worr:
While I hate knobs, I definitely see the value in this being a knob. Maybe a 100MB default with an option to increase it?

kacfengine:
I agree that this should probably be configurable, but I think we shouldn't introduce such an option until we know that it's supported relatively well throughout CFEngine. It's kinda silly to have an option which works in one place, but not another.

@worr, you asked earlier on the mailing list what 4K issues remain, and I didn't see too many concrete answers. Do you think this limit has been lifted in most areas of CFEngine now?

worr:
It seems like that's the case. I took @tzz's advice and looked through the code for more instances of `CF_BUFSIZE`, but saw very little left that wasn't network buffers, paths or lock names. I'll keep looking for more.

worr:
I've updated the patch to make use of a large maximum size. Currently, it's a macro, but we can revisit that later.

kacfengine:
trigger build

kacfengine:
> Currently, it's a macro, but we can revisit that later.

Indeed, it's still better than it was before!

worr:
Is there anything else that needs doing before this can get merged?

kacfengine:
I'm getting quite a lot of test failures for this branch. Some examples:

* ./00_basics/01_compiler/100.cf
* ./00_basics/02_switches/dry_run_perms_doesnt_lie.cf
* ./00_basics/04_bundles/call_namespaced_bundle_from_namespace_without_specifying_namespace.cf
* ./00_basics/04_bundles/dynamic_bundlesequence/app_based_inputs.cf

tests/acceptance/01_vars/02_functions/execresult_4k.cf
```diff
@@ -0,0 +1,43 @@
...
+{
+}
+
+bundle agent test
+{
+  vars:
+    any::
+      "cwd" string => dirname("$(this.promise_filename)");
+    !windows::
+      "cat_4k" string => "cat $(cwd)$(const.dirsep)text_xform.cf.txt";
```
kacfengine: Should probably use ``$(G.cat)`` instead of ``cat``. That also saves you from the Windows specific section below, since it will locate the Windows MSYS tools correctly.

worr:
My bad, I've fixed the problem and pushed a new version that passes all tests.

kacfengine:
Thanks!

kacfengine:
trigger build

kacfengine:
Some build errors in Enterprise because ``GetExecOutput`` changed signature. I will fix it shortly, the rest of the build is green.

libpromises/evalfunction.c
```diff
@@ -1600,9 +1600,10 @@ static FnCallResult FnCallExecResult(ARG_UNUSED EvalContext *ctx, ARG_UNUSED con
         return FnFailure();
     }
 
-    char buffer[CF_EXPANDSIZE];
+    size_t buffer_size = CF_EXPANDSIZE;
+    char *buffer = xcalloc(1, buffer_size);
```
kacfengine: There was a missing ``free()`` as well. Added in https://github.com/cfengine/core/commit/cccbed6d8fcb8b9522d8da3eba7502109220df9a.

PR #2197, opened 2015-04-30
Ciphers redmine3690

jimis:


jimis:
Lots of small details here, plus changes needed for fixing the enterprise compliance tests.

jimis:
@pasinskim can you review commit 557c3fd ?

jimis:
Hmmm commit 49a3032 broke the check-outputs test. I'll be fixing that perl script...

jimis:
Fixed `remake_outputs.pl` and squashed into the relevant commit.

jimis:
trigger build

jimis:
The enterprise tests are failing, but the should be fixed as soon as the relevant pull request in Enterprise is merged: https://github.com/cfengine/enterprise/pull/263

libpromises/mod_common.c
```diff
@@ -262,6 +262,7 @@ const ConstraintSyntax CFG_CONTROLBODY[COMMON_CONTROL_MAX + 1] =
     ConstraintSyntaxNewReal("bwlimit", CF_VALRANGE, "Limit outgoing protocol bandwidth in Bytes per second", SYNTAX_STATUS_NORMAL),
     ConstraintSyntaxNewBool("cache_system_functions", "Cache the result of system functions. Default value: true", SYNTAX_STATUS_NORMAL),
     ConstraintSyntaxNewOption("protocol_version", "0,undefined,1,classic,2,latest", "CFEngine protocol version to use when connecting to the server. Default: classic", SYNTAX_STATUS_NORMAL),
+    ConstraintSyntaxNewOption("tls_ciphers", ".*", "List of acceptable ciphers in TLS connections that are initiated. For syntax help see man page for \"openssl ciphers\"", SYNTAX_STATUS_NORMAL),
```
kacfengine: Can we set this to a predefined list to improve syntax checking, or does it need to be freeform in case OpenSSL is upgraded separately?

libpromises/eval_context.c
```diff
@@ -1865,10 +1871,7 @@ VariableTableIterator *EvalContextVariableTableFromRefIteratorNew(const EvalCont
 
 const void *EvalContextVariableControlCommonGet(const EvalContext *ctx, CommonControl lval)
 {
-    if (lval >= COMMON_CONTROL_MAX)
-    {
-        return NULL;
-    }
+    assert(lval >= 0 && lval < COMMON_CONTROL_MAX);
```
kacfengine: ``assert`` is debug-only. Maybe keep the ``return NULL`` for release builds, or else make it ``ProgrammingError`` instead?

libpromises/crypto.c
```diff
@@ -538,20 +541,6 @@ char *PrivateKeyFile(const char *workdir)
...
-        NULL == getenv("CFENGINE_TEST_OVERRIDE_WORKDIR"))
-    {
-        return LOG_LEVEL_ERR;
-    }
-    else
-    {
-        return LOG_LEVEL_VERBOSE;
-    }
-}
-
```
kacfengine: +1! Good to remove this hack!

libutils/logging.c
```diff
@@ -173,7 +173,7 @@ static void LogToConsole(const char *msg, LogLevel level, bool color)
         fprintf(stdout, "%s ", formatted_timestamp);
     }
 
-    fprintf(stdout, "%s: %s\n", LogLevelToString(level), msg);
+    fprintf(stdout, "%8s: %s\n", LogLevelToString(level), msg);
```
kacfengine: I had this when I was working on the logging branch actually, but I thought it looked funny since there is nearly always at least one space (when do you ever hit critical?), and removed it. But if you want it too, let's do it! :-)

libpromises/eval_context.c
jimis: If it triggers we'll know this from the debug builds. I didn't find a reason for this check to be in production, the debug build should easily catch a programming mistake (famous last words)...

libpromises/mod_common.c
jimis: Indeed, the string is passed for parsing to OpenSSL, so it's better to be free-form. If OpenSSL complains then an error we output an error anyway.

libutils/logging.c
jimis: It could be `%7s`, if saving one character is important, I have no problem with that. :-)

libpromises/mod_common.c
jimis: Hmmm maybe `.*` should be just changed to the empty string, I can see that being used elsewhere.

PR #2198, opened 2015-04-30
Add: Acceptance test for setting variables remotely

nickanderson:
You should not be allowed to set variables in another bundle.

Ref: https://dev.cfengine.com/issues/7145


kacfengine:
trigger build

kacfengine:

PR #2201, opened 2015-05-05
Redmine#7119: clarify remote copy_from failure.

jimis:


kacfengine:
+1 to replace the message, it's a weird expression. But I do think it should include the word "no server *responded*" since with your version it sounds like we didn't even try to connect.

kacfengine:
Or maybe "Not able to connect to any suitable server"?

jimis:
We tried to connect, servers /possibly/ responded, but no *suitable* server was among them.

jimis:
I'm merging it as is, since it's a clear improvement. Feel free to suggest further improvements in the ticket.

PR #2206, opened 2015-05-07
We now require tls_protocol to successfully bootstrap.

jimis:


jimis:
@estenberg, it would be nice if this huge message was just a link to some "connection error" FAQ.

jimis:
It still remains to be investigated whether the specific version mismatch failure can be specifically detected, so that a nicer warning can be printed. But for this sprint this message is what we planned for.

estenberg:
Noted and backlogged.

We should still keep *common* problems here, though. And at the time a 3.7 client bootstrapping to a 3.5 server is a somewhat common problem. However when we're at 3.10 this message can definitely be removed. :)

kacfengine:
This won't have any effect unless you also change ``bootstrap.inc`` (and there is a test which checks that they are equal).

jimis:
On Fri, May 8, 2015 at 9:45 AM, Kristian Amlie <notifications@github.com>
wrote:

> This won't have any effect unless you also change bootstrap.inc (and
> there is a test which checks that they are equal).
>
Running "make" automatically generates the `.inc` file, so the change will
have effect. I'll also commit the generated file for completeness, but why
are we doing that?

kacfengine:
> I'll also commit the generated file for completeness, but why are we doing that?

We have never tested if the required Perl module is available everywhere, and we avoided it to make our build system not depend on Perl. I think it's probably available in most setups so feel free to change that and remove bootstrap.inc altogether, but run it through Jenkins first.

jimis:
> I think it's probably available in most setups so feel free to change that and remove bootstrap.inc altogether

Hmm it seems that currently the file is generated during compilation. I would like that to be generated during "make dist" phase, so that it's pre-generated in the source tarball (just like our `cf3lex.c` and `cf3parse.c` files). I'll probably have to add the file to `DIST_COMMON` for that.

jimis:
`DIST_COMMON` is not to be touched after all, and `EXTRA_DIST` had become a mess of manual includes because of badly written dependencies in the makefile. I believe my last commit should fix all the issues. I'm starting to surprise myself with my automake-foo...

jimis:
trigger  build

jimis:
trigger build

jimis:
build green, @kacfengine want to review and merge?

kacfengine:
Wow! That's serious automake stuff!

Looks good to me!

PR #2207, opened 2015-05-07
Redmine#7166: data_expand() function

tzz:
see https://dev.cfengine.com/issues/7166

Implements `data_expand()` as discussed, with example.  If this is acceptable I'll write docs and tests.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -5934,6 +5934,88 @@ static FnCallResult DataRead(EvalContext *ctx, const FnCall *fp, const Rlist *fi
 
 /*********************************************************************/
 
+JsonElement* DataExpandElement(EvalContext *ctx, const JsonElement *source)
+{
+    if (JsonGetElementType(source) == JSON_ELEMENT_TYPE_PRIMITIVE)
+    {
+        Buffer *expbuf;
+        JsonElement *expanded_json;
+
+        switch (JsonGetPrimitiveType(source))
+        {
+        case JSON_PRIMITIVE_TYPE_STRING:
+            expbuf = BufferNew();
+            ExpandScalar(ctx, NULL, "this", JsonPrimitiveGetAsString(source), expbuf);
+            expanded_json = JsonStringCreate(BufferData(expbuf));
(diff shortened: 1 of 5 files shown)


nickanderson:
@kacfengine @estenberg this would be *very* useful to have in 3.7. Right now we have to either reconstruct data containers into classic arrays. Or we have to use parsejson and we are limited to 4K. And looping over "large" arrays has shown dramatic performance implications.

tzz:
The function is very contained and does not touch any of the rest of the core.  I think it's fairly safe even at this late stage.

neilhwatson:
This will be very useful for EFL. I hope it can make it to 3.7.

libpromises/evalfunction.c
```diff
@@ -5934,6 +5934,87 @@ static FnCallResult DataRead(EvalContext *ctx, const FnCall *fp, const Rlist *fi
...
+    return NULL;
+}
+
+static FnCallResult FnCallDataExpand(EvalContext *ctx,
+                                     ARG_UNUSED const Policy *policy,
+                                     ARG_UNUSED const FnCall *fp,
+                                     const Rlist *args)
+{
+    const char *varname = RlistScalarValue(args);
+    VarRef *ref = ResolveAndQualifyVarName(fp, varname);
```
kacfengine: I believe you will need to ``VarRefDestroy`` this.

kacfengine:
Indeed, this looks quite useful and self contained. If we get docs and acceptance test quickly and the testing goes through, then we can include this, but note that Tuesday next week is a hard deadline for this. No contributions will go in after that.

tzz:
State-based acceptance test and `VarRefDestroy()` call added.  If tests pass I'll add docs, they'll be trivial.

kacfengine:
trigger build

kacfengine:
Passed! Only docs remain.

tzz:
Docs in https://github.com/cfengine/documentation/pull/1038

nickanderson:
yay! @kacfengine can this be merged now before the freeze?

kacfengine:
I has been merged already! :-)

PR #2210, opened 2015-05-12
CFSTARTTIME should only be modified when starting the process or re-reading policies.

jimis:


jimis:
Maybe related to a bug in windows that @nickanderson is seeing.

jimis:
trigger build

jimis:
Green light from jenkins and it's simple enough, merging!

PR #2212, opened 2015-05-12
Redmine#6180: add feature_yaml and feature_xml and adjust tests

tzz:
This fixes the test falures noted by @jimis in https://dev.cfengine.com/issues/6180#change-38438 as follows:

* add new classes `feature_yaml` and `feature_xml` to show if CFEngine was compiled with support for those data formats
* adjust the YAML tests to pass with or without YAML support

I don't see why `examples/variablesmatching.cf` fails without YAML support?

Please let me know if any other tests need adjustment.  With the minor C change, this is pretty trivial.


tzz:
I saw we use this in the acceptance tests:

```
bundle common yaml_check
{
  classes:
      "no_yaml_support"
        not => regline(" *# *define *HAVE_LIBYAML.*", "$(this.promise_dirname)/../../../../libutils/config.h");
}
```

but I think having `feature_yaml` as a hard class is more accurante and useful, especially for users.

kacfengine:
Thanks for fixing this.

kacfengine:
trigger build

jimis:
I like the feature classes idea! However I'm still getting failure in the examples test, see ticket for details.

tzz:
`examples/readdata.cf` adjusted.

tzz:
@kacfengine @jimis it may be helpful to allow me to trigger builds, if that's possible.  Then you don't have to do it every time I push changes.

jimis:
For this pull request I should trigger a build on my own workstation, that lacks libyaml. :-p

jimis:
Build green! :-)

kacfengine:
Build is green. I will merge shortly, just waiting for some Jenkins maintenance.

kacfengine:
@tzz: I appreciate the incentive, but we are quite resource constrained on the build side. Jenkins is often 100% busy during a workday, so it's better if we trigger builds when we have reviewed and have capacity to handle the results.

PR #2213, opened 2015-05-13
Fix context update when iterating on individual package actions

amousset:
Ticket: https://dev.cfengine.com/issues/7181

`EvalContextStackPushPromiseFrame` is always called with the first promise, whereas it should be called with the current promise.


kacfengine:
trigger build

kacfengine:
Thanks!

PR #2214, opened 2015-05-18
Add: Test for bundlesmatching tags in params 3-n

nickanderson:
Ref: https://dev.cfengine.com/issues/7192

### DIFF
--- a/tests/acceptance/01_vars/02_functions/bundlesmatching_multiple_tags.cf
+++ b/tests/acceptance/01_vars/02_functions/bundlesmatching_multiple_tags.cf
@@ -1,0 +1,47 @@
+# Test that bundlesmatching works correctly for parameters 3-n
+
+# Ref: https://dev.cfengine.com/issues/7192
+body common control
+{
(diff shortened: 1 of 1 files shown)


kacfengine:
The docs are not 100% clear on this, but my take on it is that the tag search is ANDed, hence you should only find one bundle, the one marked with both tags. Your test case implies that the results are ORed together, IOW find bundles that have *any* of the tags.

Perhaps @tzz has an opinion on this.

kacfengine:
Relevant section in the docs: "When any tags are given, only the bundles with those tags are returned."

tzz:
It's AND, because otherwise there would be no way to specify AND (except by calling `intersection`).

To get OR, call `bundlesmatching` again and merge the results.

tzz:
The docs should probably clarify this...

kacfengine:
Sounds like the test case is wrong then. Does the bug really exist in that case? Closing for now, discussion can continue in the ticket.

PR #2215, opened 2015-05-20
New packages promise

pasinskim:
Complete implementation of new package promise feature.

### DIFF
--- a/cf-agent/package_module.c
+++ b/cf-agent/package_module.c
@@ -95,7 +95,7 @@ void UpdatePackagesCache(EvalContext *ctx, bool force_update)
 {
     Log(LOG_LEVEL_DEBUG, "Updating package cache.");
     
-    PackagePromoseGlobalLock package_lock = 
+    PackagePromiseGlobalLock package_lock = 
             AcquireGlobalPackagePromiseLock(ctx);
                         
     if (package_lock.g_lock.lock == NULL)
@@ -146,7 +146,7 @@ void UpdatePackagesCache(EvalContext *ctx, bool force_update)
     YieldGlobalPackagePromiseLock(package_lock);
 }
 
-PackagePromoseGlobalLock AcquireGlobalPackagePromiseLock(EvalContext *ctx)
+PackagePromiseGlobalLock AcquireGlobalPackagePromiseLock(EvalContext *ctx)
 {
     Bundle bundle = {.name = "package_global"};
     PromiseType promise_type = {.name = "package_global",
@@ -162,11 +162,11 @@ PackagePromoseGlobalLock AcquireGlobalPackagePromiseLock(EvalContext *ctx)
                                               .expireafter = VEXPIREAFTER},
                         &pp, false);
                         
-    return (PackagePromoseGlobalLock) {.g_lock = package_promise_global_lock,
+    return (PackagePromiseGlobalLock) {.g_lock = package_promise_global_lock,
                                        .lock_ctx = ctx};
 }
 
-void YieldGlobalPackagePromiseLock(PackagePromoseGlobalLock lock)
+void YieldGlobalPackagePromiseLock(PackagePromiseGlobalLock lock)
 {
     Bundle bundle = {.name = "package_global"};
     PromiseType promise_type = {.name = "package_global",
--- a/cf-agent/package_module.c
+++ b/cf-agent/package_module.c
@@ -57,10 +57,6 @@ PackageModuleWrapper *NewPackageModuleWrapper(PackageModuleBody *package_module)
     assert(package_module && package_module->name);
 
     PackageModuleWrapper *wrapper = xmalloc(sizeof(PackageModuleWrapper));
-    if (!wrapper)
-    {
-        return NULL;
-    }
     
     wrapper->path = GetPackageModuleRealPath(package_module->name);
     wrapper->name = SafeStringDuplicate(package_module->name);
@@ -100,7 +96,7 @@ void UpdatePackagesCache(EvalContext *ctx, bool force_update)
                         
     if (package_lock.g_lock.lock == NULL)
     {
-        Log(LOG_LEVEL_INFO, "Can not aquire global lock for package promise. "
+        Log(LOG_LEVEL_INFO, "Can not acquire global lock for package promise. "
             "Skipping updating cache.");
         return;
     }
@@ -120,7 +116,7 @@ void UpdatePackagesCache(EvalContext *ctx, bool force_update)
         PackageModuleBody *module = GetPackageModuleFromContext(ctx, pm_name);
         if (!module)
         {
-            Log(LOG_LEVEL_VERBOSE,
+            Log(LOG_LEVEL_ERR,
                 "Can not find body for package module: %s", pm_name);
             continue;
         }
@@ -129,7 +125,7 @@ void UpdatePackagesCache(EvalContext *ctx, bool force_update)
 
         if (!module_wrapper)
         {
-            Log(LOG_LEVEL_VERBOSE,
+            Log(LOG_LEVEL_ERR,
                 "Can not set up wrapper for module: %s", pm_name);
             continue;
         }
@@ -137,8 +133,8 @@ void UpdatePackagesCache(EvalContext *ctx, bool force_update)
         UpdateSinglePackageModuleCache(ctx, module_wrapper,
                                        UPDATE_TYPE_INSTALLED, force_update);
         UpdateSinglePackageModuleCache(ctx, module_wrapper,
-                                       force_update ? UPDATE_TYPE_LOCAL_UPDATES : 
-                                           UPDATE_TYPE_UPDATES,
+                                       force_update ? UPDATE_TYPE_UPDATES : 
+                                           UPDATE_TYPE_LOCAL_UPDATES,
                                        force_update);
 
         DeletePackageModuleWrapper(module_wrapper);
@@ -211,6 +207,8 @@ static int IsReadWriteReady(const IOData *io, int timeout_sec)
         .tv_usec = 0,
     };
 
+    //TODO: For Windows we will need different method and select might not 
+    //      work with file descriptors.
     int ret = select(io->read_fd + 1, &rset, NULL, NULL, &tv);
 
     if(ret < 0)
@@ -242,7 +240,6 @@ static int IsReadWriteReady(const IOData *io, int timeout_sec)
 static Rlist *ReadDataFromPackageScript(const IOData *io)
 {
     char buff[CF_BUFSIZE] = {0};
-    int red_data_size = 0;
     
     Buffer *data = BufferNew();
     if (!data)
@@ -276,7 +273,7 @@ static Rlist *ReadDataFromPackageScript(const IOData *io)
                 }
                 else
                 {
-                    Log(LOG_LEVEL_VERBOSE,
+                    Log(LOG_LEVEL_ERR,
                         "Unable to read output from package module: %s",
                         GetErrorStr());
                     BufferDestroy(data);
@@ -287,9 +284,8 @@ static Rlist *ReadDataFromPackageScript(const IOData *io)
             {
                 break;
             }
-            Log(LOG_LEVEL_DEBUG, "Data red from package module: %zu [%s]",
+            Log(LOG_LEVEL_DEBUG, "Data read from package module: %zu [%s]",
                 res, buff);
-            red_data_size += res;
 
             BufferAppendString(data, buff);
             memset(buff, 0, sizeof(buff));
@@ -301,9 +297,9 @@ static Rlist *ReadDataFromPackageScript(const IOData *io)
         }
     }
     
-    char *red_string = BufferClose(data);
-    Rlist *response_lines = RlistFromSplitString(red_string, '\n');
-    free(red_string);
+    char *read_string = BufferClose(data);
+    Rlist *response_lines = RlistFromSplitString(read_string, '\n');
+    free(read_string);
     
     return response_lines;
 }
@@ -467,7 +463,7 @@ static PackageInfo *ParseAndCheckPackageDataReply(const Rlist *data)
             {
                 /* Some error occurred as we already have name for 
                  * given package. */
-                Log(LOG_LEVEL_VERBOSE, "duplicated package name "
+                Log(LOG_LEVEL_ERR, "Duplicated package name "
                     "received: [%s] %s", line, package_data->name);
                 free(package_data);
                 return NULL;
@@ -481,7 +477,7 @@ static PackageInfo *ParseAndCheckPackageDataReply(const Rlist *data)
             {
                 /* Some error occurred as we already have version for 
                  * given package. */
-                Log(LOG_LEVEL_VERBOSE, "duplicated package version "
+                Log(LOG_LEVEL_ERR, "Duplicated package version "
                     "received: [%s] %s", line, package_data->version);
                 free(package_data);
                 return NULL;
@@ -495,7 +491,7 @@ static PackageInfo *ParseAndCheckPackageDataReply(const Rlist *data)
             {
                 /* Some error occurred as we already have arch for 
                  * given package. */
-               Log(LOG_LEVEL_VERBOSE, "duplicated package architecture "
+               Log(LOG_LEVEL_ERR, "Duplicated package architecture "
                     "received: [%s] %s", line, package_data->arch);
                 free(package_data);
                 return NULL;
@@ -619,7 +615,7 @@ static int IsPackageInCache(EvalContext *ctx,
         if (!UpdateSinglePackageModuleCache(ctx, module_wrapper,
                                             UPDATE_TYPE_INSTALLED, false))
         {
-            Log(LOG_LEVEL_INFO, "Can not update cache.");
+            Log(LOG_LEVEL_ERR, "Can not update cache.");
         }
     }
     
@@ -863,17 +859,18 @@ bool UpdateCache(Rlist* options, const PackageModuleWrapper *wrapper,
             "Received empty packages list after requesting: %s", req_type);
     }
     
+    bool ret = true;
+    
     /* We still need to update DB with empty data. */
     if (UpdatePackagesDB(response, wrapper->name, type) != 0)
     {
         Log(LOG_LEVEL_INFO, "Error parsing package cache.");
-        free(options_str);
-        return false;
+        ret = false;
     }
     
     RlistDestroy(response);
     free(options_str);
-    return true;
+    return ret;
 }
 
 
@@ -982,7 +979,7 @@ static PromiseResult InstallPackageGeneric(Rlist *options,
     else
     {
         /* If we end up here something bad has happened. */
-        assert(0 && "unsupported package type");
+        ProgrammingError("Unsupported package type");
     }
     
     Rlist *error_message = NULL;
@@ -1088,13 +1085,10 @@ Seq *GetVersionsFromUpdates(EvalContext *ctx, const PackageInfo *info,
     Seq *updates_list = NULL;
     
     /* Make sure cache is updated. */
-    if (ctx)
+    if (!UpdateSinglePackageModuleCache(ctx, module_wrapper,
+                                        UPDATE_TYPE_UPDATES, false))
     {
-        if (!UpdateSinglePackageModuleCache(ctx, module_wrapper,
-                                            UPDATE_TYPE_UPDATES, false))
-        {
-            Log(LOG_LEVEL_INFO, "Can not update packages cache.");
-        }
+        Log(LOG_LEVEL_INFO, "Can not update packages cache.");
     }
     
     if (OpenSubDB(&db_updates, db_id, module_wrapper->package_module->name))
@@ -1328,7 +1322,7 @@ static bool CheckPolicyAndPackageInfoMatch(const NewPackages *packages_policy,
             !StringSafeEqual(info->arch, packages_policy->package_architecture))
     {
         Log(LOG_LEVEL_VERBOSE, 
-            "package arch and one specified in policy doesn't match: %s -> %s",
+            "Package arch and one specified in policy doesn't match: %s -> %s",
             info->arch, packages_policy->package_architecture);
         return false;
     }
@@ -1336,15 +1330,15 @@ static bool CheckPolicyAndPackageInfoMatch(const NewPackages *packages_policy,
     {
         if (StringSafeEqual(packages_policy->package_version, "latest"))
         {
-            Log(LOG_LEVEL_VERBOSE, "unsupported 'latest' version for package "
+            Log(LOG_LEVEL_WARNING, "Unsupported 'latest' version for package "
                     "promise of type file.");
             return false;
         }
         if (packages_policy->package_version && 
             !StringSafeEqual(info->version, packages_policy->package_version))
         {
-            Log(LOG_LEVEL_VERBOSE,
-                "package version and one specified in policy doesn't "
+            Log(LOG_LEVEL_WARNING,
+                "Package version and one specified in policy doesn't "
                 "match: %s -> %s",
                 info->version, packages_policy->package_version);
             return false;
@@ -1374,7 +1368,7 @@ PromiseResult HandlePresentPromiseAction(EvalContext *ctx,
         {
             if (!CheckPolicyAndPackageInfoMatch(policy_data, package_info))
             {
-                Log(LOG_LEVEL_ERR, "package data and policy doesn't match");
+                Log(LOG_LEVEL_ERR, "Package data and policy doesn't match");
                 FreePackageInfo(package_info);
                 return PROMISE_RESULT_FAIL;
             }
@@ -1425,7 +1419,7 @@ PromiseResult HandlePresentPromiseAction(EvalContext *ctx,
                 /* We shouldn't end up here. If we are having unsupported 
                  package type this should be detected and handled
                  in ParseAndCheckPackageDataReply(). */
-                assert(0 && "unsupported package type");
+                ProgrammingError("Unsupported package type");
         }
         
         FreePackageInfo(package_info);
@@ -1519,7 +1513,9 @@ bool UpdateSinglePackageModuleCache(EvalContext *ctx,
         if (module_wrapper->package_module->installed_ifelapsed == CF_NOINT ||
             module_wrapper->package_module->updates_ifelapsed == CF_NOINT)
         {
-            Log(LOG_LEVEL_ERR, "Package module body constraints error: %s %d %d",
+            Log(LOG_LEVEL_ERR,
+                "Invalid or missing arguments in package_module body '%s':  "
+                "query_installed_ifelapsed = %d query_updates_ifelapsed = %d",
                 module_wrapper->package_module->name, 
                 module_wrapper->package_module->installed_ifelapsed,
                 module_wrapper->package_module->updates_ifelapsed);
--- a/cf-agent/package_module.c
+++ b/cf-agent/package_module.c
@@ -63,12 +63,12 @@ PackageModuleWrapper *NewPackageModuleWrapper(PackageModuleBody *package_module)
     wrapper->package_module = package_module;
     
     /* Check if file exists */
-    struct stat sb;
-    if (!wrapper->path || (stat(wrapper->path, &sb) != 0))
+    if (!wrapper->path || (access(wrapper->path, X_OK) != 0))
     {
         Log(LOG_LEVEL_VERBOSE,
-            "can not find package wrapper in provided location: %s",
-            wrapper->path);
+            "can not find package wrapper in provided location '%s' "
+            "or access to file is restricted: %d",
+            wrapper->path, errno);
         DeletePackageModuleWrapper(wrapper);
         return NULL;
     }
@@ -83,6 +83,11 @@ PackageModuleWrapper *NewPackageModuleWrapper(PackageModuleBody *package_module)
         DeletePackageModuleWrapper(wrapper);
         return NULL;
     }
+    
+    Log(LOG_LEVEL_DEBUG,
+        "Successfully created package module wrapper for '%s' package module.",
+        package_module->name);
+    
     return wrapper;
 }
 
@@ -186,12 +191,12 @@ static void ParseAndLogErrorMessage(const Rlist *data)
         }
         else if (StringStartsWith(line, "ErrorMessage="))
         {
-            Log(LOG_LEVEL_VERBOSE, "Caught error message: %s", 
+            Log(LOG_LEVEL_INFO, "Caught error message: %s", 
                 line + strlen("ErrorMessage="));
         }
         else
         {
-            Log(LOG_LEVEL_VERBOSE, "Caught unsupported error info: %s", line);
+            Log(LOG_LEVEL_INFO, "Caught unsupported error info: %s", line);
         }
     }
 }
@@ -211,28 +216,26 @@ static int IsReadWriteReady(const IOData *io, int timeout_sec)
     //      work with file descriptors.
     int ret = select(io->read_fd + 1, &rset, NULL, NULL, &tv);
 
-    if(ret < 0)
+    if (ret < 0)
     {
         Log(LOG_LEVEL_VERBOSE, "Failed checking for data. (select: %s)",
             GetErrorStr());
         return -1;
     }
-
     else if (FD_ISSET(io->read_fd, &rset))
     {
         return io->read_fd;
     }
 
     /* We have reached timeout */
-    if(ret == 0)
+    if (ret == 0)
     {
         Log(LOG_LEVEL_DEBUG, "Timeout reading from package module.");
         return 0;
     }
 
     Log(LOG_LEVEL_VERBOSE,
-        "Unknown outcome (ret > 0 but our only fds is not set). (select: %s)",
-        GetErrorStr());
+        "Unknown outcome (ret > 0 but our only fd is not set).");
 
     return -1;
 }
@@ -251,7 +254,7 @@ static Rlist *ReadDataFromPackageScript(const IOData *io)
     
     int timeout_seconds_left = PACKAGE_PROMISE_SCRIPT_TIMEOUT_SEC;
     
-    while(!IsPendingTermination() && timeout_seconds_left > 0)
+    while (!IsPendingTermination() && timeout_seconds_left > 0)
     {
         int fd = IsReadWriteReady(io, PACKAGE_PROMISE_TERMINATION_CHECK_SEC);
         
@@ -335,7 +338,7 @@ int WriteDataToPackageModule(const char *args, const char *data,
     IOData io = cf_popen_full_duplex(command, false);
     free(command);
     
-    if (io.write_fd == 0 || io.read_fd == 0)
+    if (io.write_fd == -1 || io.read_fd == -1)
     {
         Log(LOG_LEVEL_VERBOSE, "Error occurred while opening pipes for "
             "communication with package module.");
@@ -372,7 +375,7 @@ static int ReadWriteDataToPackageScript(const char *args, const char *request,
     IOData io = cf_popen_full_duplex(command, false);
     free(command);
     
-    if (io.write_fd == 0 || io.read_fd == 0)
+    if (io.write_fd == -1 || io.read_fd == -1)
     {
         Log(LOG_LEVEL_INFO, "Some error occurred while communicating "
             "package module.");
@@ -437,11 +440,11 @@ static PackageInfo *ParseAndCheckPackageDataReply(const Rlist *data)
     
     for (const Rlist *rp = data; rp != NULL; rp = rp->next)
     {
-        char *line = RlistScalarValue(rp);
+        const char *line = RlistScalarValue(rp);
                    
         if (StringStartsWith(line, "PackageType="))
         {
-            char *type = line + strlen("PackageType=");
+            const char *type = line + strlen("PackageType=");
             if (StringSafeEqual(type, "file"))
             {
                 package_data->type = PACKAGE_TYPE_FILE;
@@ -463,8 +466,9 @@ static PackageInfo *ParseAndCheckPackageDataReply(const Rlist *data)
             {
                 /* Some error occurred as we already have name for 
                  * given package. */
-                Log(LOG_LEVEL_ERR, "Duplicated package name "
-                    "received: [%s] %s", line, package_data->name);
+                Log(LOG_LEVEL_ERR,
+                    "Extraneous package name line received: [%s] %s",
+                    line, package_data->name);
                 free(package_data);
                 return NULL;
             }
@@ -477,8 +481,9 @@ static PackageInfo *ParseAndCheckPackageDataReply(const Rlist *data)
             {
                 /* Some error occurred as we already have version for 
                  * given package. */
-                Log(LOG_LEVEL_ERR, "Duplicated package version "
-                    "received: [%s] %s", line, package_data->version);
+                Log(LOG_LEVEL_ERR,
+                    "Extraneous package version line received: [%s] %s",
+                    line, package_data->version);
                 free(package_data);
                 return NULL;
             }
@@ -491,8 +496,9 @@ static PackageInfo *ParseAndCheckPackageDataReply(const Rlist *data)
             {
                 /* Some error occurred as we already have arch for 
                  * given package. */
-               Log(LOG_LEVEL_ERR, "Duplicated package architecture "
-                    "received: [%s] %s", line, package_data->arch);
+               Log(LOG_LEVEL_ERR,
+                   "Extraneous package architecture line received: [%s] %s",
+                   line, package_data->arch);
                 free(package_data);
                 return NULL;
             }
@@ -502,12 +508,12 @@ static PackageInfo *ParseAndCheckPackageDataReply(const Rlist *data)
         /* For handling errors */
         else if (StringStartsWith(line, "Error="))
         {
-            Log(LOG_LEVEL_VERBOSE, "have error: %s",
+            Log(LOG_LEVEL_VERBOSE, "package_module - have error: %s",
                 line + strlen("Error="));
         }
         else if (StringStartsWith(line, "ErrorMessage="))
         {
-            Log(LOG_LEVEL_VERBOSE, "have error message: %s",
+            Log(LOG_LEVEL_VERBOSE, "package_module - have error message: %s",
                 line + strlen("ErrorMessage="));
         }
         else
@@ -521,14 +527,16 @@ static PackageInfo *ParseAndCheckPackageDataReply(const Rlist *data)
 
 static int NegotiateSupportedAPIVersion(PackageModuleWrapper *wrapper)
 {
+    Log(LOG_LEVEL_DEBUG, "Getting supported API version.");
+    
     int api_version = -1;
 
     Rlist *response = NULL;
     if (ReadWriteDataToPackageScript("supports-api-version", "",
                                      &response, wrapper) != 0)
     {
-        Log(LOG_LEVEL_INFO, "Some error occurred while communicating with "
-            "wrapper while getting supported API version.");
+        Log(LOG_LEVEL_INFO,
+            "Error occurred while getting supported API version.");
         return -1;
     }
     
@@ -550,6 +558,8 @@ static
 PackageInfo *GetPackageData(const char *name, Rlist *options,
                             const PackageModuleWrapper *wrapper)
 {   
+    Log(LOG_LEVEL_DEBUG, "Getting package '%s' data.", name);
+    
     char *options_str = ParseOptions(options);
     char *request = StringFormat("%sFile=%s\n", options_str, name);
     
@@ -573,10 +583,11 @@ PackageInfo *GetPackageData(const char *name, Rlist *options,
         
         if (package_data)
         {
-            /* At this point at least package name MUST be known (if no error) */
+            /* At this point at least package name and type MUST be known
+             * (if no error) */
             if (!package_data->name || package_data->type == PACKAGE_TYPE_NONE)
             {
-                Log(LOG_LEVEL_INFO, "Unknown package name.");
+                Log(LOG_LEVEL_INFO, "Unknown package name or type.");
                 FreePackageInfo(package_data);
                 package_data = NULL;
             }
@@ -603,7 +614,7 @@ static int IsPackageInCache(EvalContext *ctx,
     const char *version = ver;
     /* Handle latest version in specific way for repo packages. 
      * Please note that for file packages 'latest' version is not supported
-     * and check against that is made elsewhere. */
+     * and check against that is made in CheckPolicyAndPackageInfoMatch(). */
     if (version && StringSafeEqual(version, "latest"))
     {
         version = NULL;
@@ -648,7 +659,7 @@ static int IsPackageInCache(EvalContext *ctx,
     int is_in_cache = 0;
     char buff[1];
     
-    Log(LOG_LEVEL_DEBUG, "looking for key in cache: %s", key);
+    Log(LOG_LEVEL_DEBUG, "Looking for key in installed packages cache: %s", key);
     
     if (ReadDB(db_cached, key, buff, 1))
     {
@@ -682,24 +693,7 @@ void WritePackageDataToDB(CF_DB *db_installed,
     
     xsnprintf(package_key, sizeof(package_key),
               "N<%s>", name);
-    if ((type == UPDATE_TYPE_UPDATES || type == UPDATE_TYPE_LOCAL_UPDATES) && 
-            HasKeyDB(db_installed, package_key, strlen(package_key) + 1))
-    {
-        size_t val_size =
-                ValueSizeDB(db_installed, package_key, strlen(package_key));
-        char buff[val_size + strlen(arch) + strlen(ver) + 8];
-        
-        ReadDB(db_installed, package_key, buff, val_size);
-        xsnprintf(buff + val_size, sizeof(package_key), "V<%s>A<%s>\n", ver, arch);
-        WriteDB(db_installed, package_key, buff, strlen(buff));
-    }
-    else if (type == UPDATE_TYPE_UPDATES || type == UPDATE_TYPE_LOCAL_UPDATES)
-    {
-        char buff[strlen(arch) + strlen(ver) + 8];
-        xsnprintf(buff, sizeof(package_key), "V<%s>A<%s>\n", ver, arch);
-        WriteDB(db_installed, package_key, buff, strlen(buff));
-    }
-    else /* UPDATE_TYPE_INSTALLED */
+    if (type == UPDATE_TYPE_INSTALLED)
     {
         WriteDB(db_installed, package_key, "1", 1);
         xsnprintf(package_key, sizeof(package_key),
@@ -711,6 +705,25 @@ void WritePackageDataToDB(CF_DB *db_installed,
         xsnprintf(package_key, sizeof(package_key),
                 "N<%s>V<%s>A<%s>", name, ver, arch);
         WriteDB(db_installed, package_key, "1", 1);
+    }
+    else if (HasKeyDB(db_installed, package_key, strlen(package_key) + 1))
+    {
+        /* type == UPDATE_TYPE_UPDATES || type == UPDATE_TYPE_LOCAL_UPDATES */
+        size_t val_size =
+                ValueSizeDB(db_installed, package_key, strlen(package_key));
+        char buff[val_size + strlen(arch) + strlen(ver) + 8];
+
+        ReadDB(db_installed, package_key, buff, val_size);
+        xsnprintf(buff + val_size, sizeof(package_key), "V<%s>A<%s>\n",
+                  ver, arch);
+        WriteDB(db_installed, package_key, buff, strlen(buff));
+    }
+    else
+    {
+        /* type == UPDATE_TYPE_UPDATES || type == UPDATE_TYPE_LOCAL_UPDATES */
+        char buff[strlen(arch) + strlen(ver) + 8];
+        xsnprintf(buff, sizeof(package_key), "V<%s>A<%s>\n", ver, arch);
+        WriteDB(db_installed, package_key, buff, strlen(buff));
     }
 }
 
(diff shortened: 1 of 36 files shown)


pasinskim:
@jimis @kacfengine you can take a look at this... it's a lot though but hopefully in reviewable state.

pasinskim:
@ediosyncratic can you take a look at this?

pasinskim:
@tzz this is the quite final version of new package promise implementation.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+    
+    CfLock package_promise_global_lock;
+    
+    package_promise_global_lock =
+            AcquireLock(ctx, GLOBAL_PACKAGE_PROMISE_LOCK_NAME, VUQNAME, CFSTARTTIME,
+                        (TransactionContext) {.ifelapsed = VIFELAPSED,
+                                              .expireafter = VEXPIREAFTER},
+                        &pp, false);
+                        
+    return (PackagePromoseGlobalLock) {.g_lock = package_promise_global_lock,
```
tzz: `Promose` typo in many places

tzz:
An example of a working minimal package manager (as an acceptance test?) would be really helpful.  It doesn't have to do anything, just make it lie :)

kacfengine:
> An example of a working minimal package manager (as an acceptance test?) would be really helpful.

@tzz: @pasinskim was too quick for me. :-) The package modules for apt and yum are here: https://github.com/cfengine/masterfiles/pull/465

cf-agent/package_module.c
pasinskim: Thanks @tzz. Fixed!

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+
+PackageModuleWrapper *NewPackageModuleWrapper(PackageModuleBody *package_module)
+{
+    assert(package_module && package_module->name);
+
+    PackageModuleWrapper *wrapper = xmalloc(sizeof(PackageModuleWrapper));
+    if (!wrapper)
+    {
+        return NULL;
+    }
```
kacfengine: ``xmalloc`` will end the program if we can't allocate memory, the ``!wrapper`` check is not necessary.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+void UpdatePackagesCache(EvalContext *ctx, bool force_update)
+{
+    Log(LOG_LEVEL_DEBUG, "Updating package cache.");
+    
+    PackagePromoseGlobalLock package_lock = 
+            AcquireGlobalPackagePromiseLock(ctx);
+                        
+    if (package_lock.g_lock.lock == NULL)
+    {
+        Log(LOG_LEVEL_INFO, "Can not aquire global lock for package promise. "
```
kacfengine: ``s/aquire/acquire/``

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+            Log(LOG_LEVEL_VERBOSE,
+                "Can not set up wrapper for module: %s", pm_name);
+            continue;
+        }
+
+        UpdateSinglePackageModuleCache(ctx, module_wrapper,
+                                       UPDATE_TYPE_INSTALLED, force_update);
+        UpdateSinglePackageModuleCache(ctx, module_wrapper,
+                                       force_update ? UPDATE_TYPE_LOCAL_UPDATES : 
+                                           UPDATE_TYPE_UPDATES,
```
kacfengine: Is this correct? My understanding is that online updates should *never* be forced, not even after installing a package. What's important is to update CFEngine cache to include the updates that apt and yum know about. Going online is of less importance, and if you install several packages after one another, is close to useless, since nothing will have changed since the previous online query.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+    fd_set  rset;
+    FD_ZERO(&rset);
+    FD_SET(io->read_fd, &rset);
+
+    struct timeval tv = {
+        .tv_sec = timeout_sec,
+        .tv_usec = 0,
+    };
+
+    int ret = select(io->read_fd + 1, &rset, NULL, NULL, &tv);
```
kacfengine: I believe ``select`` does not work on file descriptors on Windows, only on sockets. We may have to consider a Windows specific section here.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+                        GetErrorStr());
+                    BufferDestroy(data);
+                    return NULL;
+                }
+            }
+            else if (res == 0) /* reached EOF */
+            {
+                break;
+            }
+            Log(LOG_LEVEL_DEBUG, "Data red from package module: %zu [%s]",
```
kacfengine: ``s/red/read/``

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+            {
+                if (errno == EINTR)
+                {
+                    continue;
+                }
+                else
+                {
+                    Log(LOG_LEVEL_VERBOSE,
+                        "Unable to read output from package module: %s",
+                        GetErrorStr());
```
kacfengine: This should be higher log level I think, since it indicates error. I would use ``LOG_LEVEL_ERR`` here.

cf-agent/package_module.c
kacfengine: In fact "red" occurs frequently. Although it's pronounced like "red", it's spelled "read" (both present and past has the same spelling, kudos to English for being so unambiguous....)

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+                return NULL;
+            }
+        }
+        else if (StringStartsWith(line, "Name="))
+        {
+            if (package_data->name)
+            {
+                /* Some error occurred as we already have name for 
+                 * given package. */
+                Log(LOG_LEVEL_VERBOSE, "duplicated package name "
```
kacfengine: I would make all unexpected scenarios, especially with regards to the protocol communication, into ``LOG_LEVEL_ERR``. They are indicative of a bug in either the code or the package module, and should stand out.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+        version = NULL;
+    }
+    
+    /* Make sure cache is updated. */
+    if (ctx)
+    {
+        if (!UpdateSinglePackageModuleCache(ctx, module_wrapper,
+                                            UPDATE_TYPE_INSTALLED, false))
+        {
+            Log(LOG_LEVEL_INFO, "Can not update cache.");
```
kacfengine: Again, all unexpected scenarios should be ``LOG_LEVEL_ERR``, since they prevent normal operation.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+        Log(LOG_LEVEL_DEBUG,
+            "Received empty packages list after requesting: %s", req_type);
+    }
+    
+    /* We still need to update DB with empty data. */
+    if (UpdatePackagesDB(response, wrapper->name, type) != 0)
+    {
+        Log(LOG_LEVEL_INFO, "Error parsing package cache.");
+        free(options_str);
+        return false;
```
kacfengine: You don't destroy ``response`` here.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+        package_install_command = "file-install";
+    }
+    else if (type == PACKAGE_TYPE_REPO)
+    {
+        package_install_command = "repo-install";
+    }
+    else
+    {
+        /* If we end up here something bad has happened. */
+        assert(0 && "unsupported package type");
```
kacfengine: We usually use ``ProgrammingError()`` for cases like these.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+        {
+            PackageInfo *update_package = SeqAt(latest_versions, i);
+            
+            Log(LOG_LEVEL_DEBUG, "Checking for package '%s' version '%s' in "
+                "available updates", package_info->name,
+                update_package->version);
+            
+            /* Just in case some package managers will report highest possible 
+             * version in updates list instead of removing entry if package is 
+             * already in the latest version. */
```
kacfengine: Good thinking! :-)

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+            "package arch and one specified in policy doesn't match: %s -> %s",
+            info->arch, packages_policy->package_architecture);
+        return false;
+    }
+    if (info->version)
+    {
+        if (StringSafeEqual(packages_policy->package_version, "latest"))
+        {
+            Log(LOG_LEVEL_VERBOSE, "unsupported 'latest' version for package "
+                    "promise of type file.");
```
kacfengine: User error, this should be at least ``LOG_LEVEL_WARNING``.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+            return false;
+        }
+        if (packages_policy->package_version && 
+            !StringSafeEqual(info->version, packages_policy->package_version))
+        {
+            Log(LOG_LEVEL_VERBOSE,
```
kacfengine: ``LOG_LEVEL_WARNING``

cf-agent/package_module.c
```diff
@@ -0,0 +1,1604 @@
...
+    
+    if (!force_update)
+    {
+        if (module_wrapper->package_module->installed_ifelapsed == CF_NOINT ||
+            module_wrapper->package_module->updates_ifelapsed == CF_NOINT)
+        {
+            Log(LOG_LEVEL_ERR, "Package module body constraints error: %s %d %d",
+                module_wrapper->package_module->name, 
+                module_wrapper->package_module->installed_ifelapsed,
+                module_wrapper->package_module->updates_ifelapsed);
```
kacfengine: A bit cryptic. I would say ``Not all attributes specified in package body '%s': query_installed_ifelapsed = %d, query_updates_ifelapsed = %d``.

cf-agent/verify_new_packages.c
```diff
@@ -0,0 +1,160 @@
...
+    if (!a.new_packages.module_body || !a.new_packages.module_body->name)
+    {
+        Log(LOG_LEVEL_ERR, "Can not find package module body in policy.");
+        return false;
+    }
+    
+    if (a.new_packages.module_body->updates_ifelapsed == CF_NOINT ||
+        a.new_packages.module_body->installed_ifelapsed == CF_NOINT)
+    {
+        Log(LOG_LEVEL_ERR, "Package module body constraints error.");
```
kacfengine: Same issue as in https://github.com/cfengine/core/pull/2215/files#r30784865.

cf-agent/verify_new_packages.c
```diff
@@ -0,0 +1,160 @@
...
+        return false;
+    }
+    
+    if (a.new_packages.package_policy == NEW_PACKAGE_ACTION_NONE)
+    {
+        Log(LOG_LEVEL_ERR, "Not supported package policy in new package"
+            " promise.");
```
kacfengine: ``s/Not supported/Unsupported/``

``s/new package promise/package promise/`` - The fact that it's new quickly becomes obsolete.

cf-agent/verify_new_packages.c
```diff
@@ -0,0 +1,160 @@
...
+    
+    CfLock package_promise_lock;
+    char promise_lock[CF_BUFSIZE];
+    snprintf(promise_lock, CF_BUFSIZE - 1, "new-package-%s-%s",
+             pp->promiser, a.new_packages.module_body->name);
+
+    if (global_lock.g_lock.lock == NULL)
+    {
+        *promise_log_msg =
+                SafeStringDuplicate("Can not aquire global lock for package "
```
kacfengine: ``s/aquire/acquire/``

libpromises/cf3parse.y
```diff
@@ -674,14 +674,6 @@ constraint:            constraint_id                        /* BUNDLE ONLY */
...
                                            cp->offset.context = P.offsets.last_class_id;
                                            P.currentstype->offset.end = P.offsets.current;
-
-                                           // Cache whether there are subbundles for later $(this.promiser) logic
-
-                                           if (strcmp(P.lval,"usebundle") == 0 || strcmp(P.lval,"edit_line") == 0
-                                               || strcmp(P.lval,"edit_xml") == 0 || strcmp(P.lval,"home_bundle") == 0)
-                                           {
-                                               P.currentpromise->has_subbundles = true;
-                                           }
```
kacfengine: ??

This looks very unrelated. Is it supposed to be here?

libpromises/dbm_api.c
```diff
@@ -197,39 +304,19 @@ void CloseAllDBExit()
...
+    
+    DynamicDBHandles *db_dynamic_handles_list = db_dynamic_handles;
+    while (db_dynamic_handles_list)
+    {
+        DBHandle *handle = db_dynamic_handles_list->handle;
+        CloseDBInstance(handle);
+        db_dynamic_handles_list = db_dynamic_handles_list->next;
+        free(handle);
+    }
+    db_dynamic_handles = NULL;
```
kacfengine: Is ``db_dynamic_handles`` static? IOW will entries there always remain until application exit?

libpromises/pipes_unix.c
```diff
@@ -142,6 +169,113 @@ static pid_t CreatePipeAndFork(const char *type, int *pd)
...
+{
+/* For simplifying reading and writing directions */
+#define READ  0
+#define WRITE 1
+    int child_pipe[2];  /* From child to parent */
+    int parent_pipe[2]; /* From parent to child */
+    pid_t pid;
+
+    fflush(NULL); /* Empty file buffers */
+    pid = CreatePipesAndFork("rt", child_pipe, parent_pipe);
```
kacfengine: This part is a bit confusing. ``type`` is supposed to signal what kind of pipe you want, following ``fopen()`` semantics. Here you are creating a bidirectional pipe, but mode is still ``rt``. The correct way to specify it according to ``fopen()`` would be ``r+t``. Whether we want to keep this semantic is debatable, but regardless I think we should choose just one, so that we don't have this inconsistent situation where ``rt`` (signaling unidirectional) is specified, but in fact bidirectional is created.

cf-agent/package_module.c
pasinskim: Nice catch. This was causing some of yours tests to fail. Fixed!

cf-agent/package_module.c
pasinskim: Yup. I will add a comment to modify this for Windows.

cf-agent/package_module.c
pasinskim: Looks like I can improve not only coding style with reviews :)

cf-agent/package_module.c
pasinskim: Fixed!

cf-agent/package_module.c
pasinskim: Sounds better indeed.

cf-agent/verify_new_packages.c
pasinskim: Fixed!

libpromises/cf3parse.y
pasinskim: We can move it from this PR. It is just something I've noticed and removed.

libpromises/dbm_api.c
pasinskim: Yes. Those are stored in global 'static DynamicDBHandles *db_dynamic_handles;' variable similar as previous one in 'static DBHandle db_handles[dbid_max]'. Those are persistent withing agent run.

libpromises/cf3parse.y
kacfengine: No, that's ok, I noticed now that it's a separate commit, that made things clear.

libpromises/dbm_api.c
kacfengine: Ok, thanks for clarifying.

libpromises/pipes_unix.c
pasinskim: Agree. In fact while opening bi-directional channel this is not used anywhere except the place where the check is performed if we have correct pipe type. I can leave it as "r" or "w" or can make the check to allow "r+t".

kacfengine: Maybe use ``r+t``, but make sure ``PipeTypeIsOk`` accepts it!

pasinskim: Done!

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+
+    PackageModuleWrapper *wrapper = xmalloc(sizeof(PackageModuleWrapper));
+    
+    wrapper->path = GetPackageModuleRealPath(package_module->name);
+    wrapper->name = SafeStringDuplicate(package_module->name);
+    wrapper->package_module = package_module;
+    
+    /* Check if file exists */
+    struct stat sb;
+    if (!wrapper->path || (stat(wrapper->path, &sb) != 0))
```
ediosyncratic: If you used `access(wrapper->path, F_OK)` you'd save yourself the need for the `struct stat` in which to store stat data you don't use; and replacing F_OK with suitable options can adjust the return to tell you whether you can read and write the path, not just whether it exists, depending on what you actually need to be able to do to the file.  This is way easier than working out the same answer from what the `struct stat` would have contained !

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+        DeletePackageModuleWrapper(wrapper);
+        return NULL;
+    }
+    return wrapper;
+}
+
+
+void UpdatePackagesCache(EvalContext *ctx, bool force_update)
+{
+    Log(LOG_LEVEL_DEBUG, "Updating package cache.");
```
ediosyncratic: cf-agent/cf-agent.c:255 describes this as a package **promise** cache - which is it ?

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+        UpdateSinglePackageModuleCache(ctx, module_wrapper,
+                                       UPDATE_TYPE_INSTALLED, force_update);
+        UpdateSinglePackageModuleCache(ctx, module_wrapper,
+                                       force_update ? UPDATE_TYPE_UPDATES : 
+                                           UPDATE_TYPE_LOCAL_UPDATES,
+                                       force_update);
+
+        DeletePackageModuleWrapper(module_wrapper);
+    }
+    YieldGlobalPackagePromiseLock(package_lock);
```
ediosyncratic: Do we have a duty to `RlistDestroy(default_inventory)` now ?  If so, we're leaking it ...

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    Rlist *default_inventory = GetDefaultInventoryFromContext(ctx);
+    
+    for (const Rlist *rp = default_inventory; rp != NULL; rp = rp->next)
+    {
+        const char *pm_name =  RlistScalarValue(rp);
+        
+        /* We don't want inventory to be reported. */
+        if (StringSafeEqual(pm_name, "cf_null"))
+        {
+            break;
```
ediosyncratic: So ignore everything after this entry (break), rather than (continue) just this entry ?
Does GetDefaultInventoryFromContext put the useful information early, followed by a cf_null entry, then the boring bits of inventory ?

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+        }
+        else if (StringStartsWith(line, "ErrorMessage="))
+        {
+            Log(LOG_LEVEL_VERBOSE, "Caught error message: %s", 
+                line + strlen("ErrorMessage="));
+        }
+        else
+        {
+            Log(LOG_LEVEL_VERBOSE, "Caught unsupported error info: %s", line);
+        }
```
ediosyncratic: Should these only be VERBOSE ?  Why are none of them even INFO ?
Should the caller pass our log-level ?

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    struct timeval tv = {
+        .tv_sec = timeout_sec,
+        .tv_usec = 0,
+    };
+
+    //TODO: For Windows we will need different method and select might not 
+    //      work with file descriptors.
+    int ret = select(io->read_fd + 1, &rset, NULL, NULL, &tv);
+
+    if(ret < 0)
```
ediosyncratic: Space between `if` and open-paren.  Likewise below for `if(ret == 0)`

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    {
+        Log(LOG_LEVEL_VERBOSE, "Failed checking for data. (select: %s)",
+            GetErrorStr());
+        return -1;
+    }
+
```
ediosyncratic: Bad place for a blank line.  The `else if` should directly follow the `if`'s close-brace.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    /* We have reached timeout */
+    if(ret == 0)
+    {
+        Log(LOG_LEVEL_DEBUG, "Timeout reading from package module.");
+        return 0;
+    }
+
+    Log(LOG_LEVEL_VERBOSE,
+        "Unknown outcome (ret > 0 but our only fds is not set). (select: %s)",
```
ediosyncratic: `s/fds/fd/` It's our only fd, so not plural !

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+        GetErrorStr());
```
ediosyncratic: `GetErrorStr()` shall, at best, return something irrelevant, which you shouldn't attribute to `select`, as it succeeded so won't have set `errno`.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    if (!data)
+    {
+        Log(LOG_LEVEL_VERBOSE, 
+            "Unable to allocate buffer for handling package module responses.");
+        return NULL;
+    }
+    
+    int timeout_seconds_left = PACKAGE_PROMISE_SCRIPT_TIMEOUT_SEC;
+    
+    while(!IsPendingTermination() && timeout_seconds_left > 0)
```
ediosyncratic: space after `while`

libpromises/pipes_unix.c
```diff
@@ -142,6 +169,113 @@ static pid_t CreatePipeAndFork(const char *type, int *pd)
...
+    pd[1] = pipes[0].pipe_desc[1];
+
+    return pid;
+}
+
+/*****************************************************************************/
+
+static pid_t CreatePipesAndFork(const char *type, int *pd, int *pdb)
+{
+    IOPipe pipes[2] = {{0}};
```
ediosyncratic: 0 is a bad choice for initialiser of descriptors, as it's a valid file descriptor (albeit usually in use as stdin); -1 is a better choice for the pipe_desk[] fields' default (and as the sentinel to compare against for failure).

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+}
+
+int WriteDataToPackageModule(const char *args, const char *data,
+                             const PackageModuleWrapper *wrapper)
+{
+    char *command = StringFormat("%s %s", wrapper->path, args);
+    IOData io = cf_popen_full_duplex(command, false);
+    free(command);
+    
+    if (io.write_fd == 0 || io.read_fd == 0)
```
ediosyncratic: 0 is a bad sentinel for 'bad descriptor' as it's actually a valid descriptor; if stdin has ever been close()d, it *shall* get used.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    return BufferClose(data);
+}
+
+static PackageInfo *ParseAndCheckPackageDataReply(const Rlist *data)
+{
+    PackageInfo * package_data = xcalloc(1, sizeof(PackageInfo));
+    
+    for (const Rlist *rp = data; rp != NULL; rp = rp->next)
+    {
+        char *line = RlistScalarValue(rp);
```
ediosyncratic: There would be a certain virtue to using `const char *` here and for `type` below.
You don't intend to actually modify it; and declaring it `const` makes it immediately obvious that we don't have ownership of (hence a duty to free) it.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+                SafeStringDuplicate(line + strlen("Name="));
+        }
+        else if (StringStartsWith(line, "Version="))
+        {
+            if (package_data->version)
+            {
+                /* Some error occurred as we already have version for 
+                 * given package. */
+                Log(LOG_LEVEL_ERR, "Duplicated package version "
+                    "received: [%s] %s", line, package_data->version);
```
ediosyncratic: Splitting the formats across lines (above, below and here) is going to be painful for someone searching for the message in source: `grep -Rw 'Duplicated package version received'` shall get no hits.
Better to put the format on a line of its own, in between log level and later parameters.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+                SafeStringDuplicate(line + strlen("Version="));
+        }
+        else if (StringStartsWith(line, "Architecture="))
+        {
+            if (package_data->arch)
+            {
+                /* Some error occurred as we already have arch for 
+                 * given package. */
+               Log(LOG_LEVEL_ERR, "Duplicated package architecture "
+                    "received: [%s] %s", line, package_data->arch);
```
ediosyncratic: Given that we haven't checked the new arch is the same as the previously given one (and, if we did that check, there'd be a case for ignoring the error), the log message mis-describes the problem; the architecture isn't **Duplicated**.  A second package architecture has been received and we don't like that.  `s/Duplicated/Extraneous/g` would perhaps be more apt.  Applies to many other Log() messages in this function.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+static int NegotiateSupportedAPIVersion(PackageModuleWrapper *wrapper)
+{
+    int api_version = -1;
+
+    Rlist *response = NULL;
+    if (ReadWriteDataToPackageScript("supports-api-version", "",
+                                     &response, wrapper) != 0)
+    {
+        Log(LOG_LEVEL_INFO, "Some error occurred while communicating with "
+            "wrapper while getting supported API version.");
```
ediosyncratic: Somewhat long-winded.  `"Some error occurred while asking wrapper for supported API version."`

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    
+    if (response)
+    {   
+        package_data = ParseAndCheckPackageDataReply(response);
+        RlistDestroy(response);
+        
+        if (package_data)
+        {
+            /* At this point at least package name MUST be known (if no error) */
+            if (!package_data->name || package_data->type == PACKAGE_TYPE_NONE)
```
ediosyncratic: Difference between comment and code: the comment demands name, the code demands name **and type**.
The comment should probably `s/name/name and type/`

ediosyncratic: The Log() below should likewise `s/name/name or type/`

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+}
+
+static int IsPackageInCache(EvalContext *ctx,
+                            const PackageModuleWrapper *module_wrapper,
+                            const char *name, const char *ver, const char *arch)
+{
+    const char *version = ver;
+    /* Handle latest version in specific way for repo packages. 
+     * Please note that for file packages 'latest' version is not supported
+     * and check against that is made elsewhere. */
```
ediosyncratic: It would be a kindness to the reader to be explicit: "elsewhere" leaves the reader to search through many MB of code for where, exactly, the check is made.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+            if (IsPackageInCache(ctx, wrapper, package_info->name,
+                                 update_package->version,
+                                 update_package->arch) == 1)
+            {
+                Log(LOG_LEVEL_VERBOSE, "Package version from updates matches "
+                    "one installed. Skipping package instalation.");
+                res = PromiseResultUpdate(res, PROMISE_RESULT_NOOP);
+                continue;
+            }
+            else
```
ediosyncratic: If IsPackageInCache() returned -1, is it still appropriate to go ahead ?
I think you're assuming it was 0 for this else branch ...

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    {
+        size_t val_size =
+                ValueSizeDB(db_installed, package_key, strlen(package_key));
+        char buff[val_size + strlen(arch) + strlen(ver) + 8];
+        
+        ReadDB(db_installed, package_key, buff, val_size);
+        xsnprintf(buff + val_size, sizeof(package_key), "V<%s>A<%s>\n", ver, arch);
+        WriteDB(db_installed, package_key, buff, strlen(buff));
+    }
+    else if (type == UPDATE_TYPE_UPDATES || type == UPDATE_TYPE_LOCAL_UPDATES)
```
ediosyncratic: Repeated condition.  May be better to structure as:
```
if (type == UPDATE_TYPE_UPDATES || type == UPDATE_TYPE_LOCAL_UPDATES)
{
    if (HasKeyDB(db_installed, package_key, strlen(package_key) + 1)) {/* block above */}
    else {/* this block */}
}
else /* UPDATE_TYPE_INSTALLED */ {...}
```
or, if that last comment is to be trusted, even as
```
if (type == UPDATE_TYPE_INSTALLED) {/* the final else block below *}
else if (HasKeyDB(db_installed, package_key, strlen(package_key) + 1)) {/* the block above */}
else {/* the following block */}
```

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    {
+        CleanDB(db_cached);
+        
+        Buffer *inventory_data = BufferNewWithCapacity(INVENTORY_LIST_BUFFER_SIZE);
+
+        char *package_data[3] = {NULL, NULL, NULL};
+
+        for (const Rlist *rp = data; rp != NULL; rp = rp->next)
+        {
+            char *line = RlistScalarValue(rp);
```
ediosyncratic: suggest: `s/char/const char/`

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+                                         package_data[1], package_data[2], type);
+                    
+                    char inventory_line[strlen(package_data[0]) +
+                                        strlen(package_data[1]) +
+                                        strlen(package_data[2]) + 4];
+                    
+                    xsnprintf(inventory_line, sizeof(inventory_line),
+                              "%s,%s,%s\n", package_data[0], package_data[1],
+                              package_data[2]);
+                    BufferAppendString(inventory_data, inventory_line);
```
ediosyncratic: This block of code is duplicated after the loop.  Suggest: move it to a noddy little static function called from the two places.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+            
+            char inventory_line[strlen(package_data[0]) +
+                                strlen(package_data[1]) +
+                                strlen(package_data[2]) + 4];
+
+            xsnprintf(inventory_line, sizeof(inventory_line),
+                      "%s,%s,%s\n", package_data[0], package_data[1],
+                      package_data[2]);
+            BufferAppendString(inventory_data, inventory_line);
+        }
```
ediosyncratic: The matching block inside the loop treats partial data as an error; but you ignore that here, though it might just as well happen.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+                    package_data[2] = NULL;
+                }
+                else if (package_data[0] && (!package_data[1] || !package_data[2]))
+                {
+                    /* some error occurred */
+                    Log(LOG_LEVEL_VERBOSE, "Malformed response from package "
+                        "manager for package %s", package_data[0]);
+                    package_data[1] = NULL;
+                    package_data[2] = NULL;
+                }
```
ediosyncratic: Redundancy in testing.  This could as well (and more clearly) be:
```
if (package_data[0])
{
    if (package_data[1] && package_data[2]) { /* handle the good case */ }
   else { /* handle the bad case */ }
}
```

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+        char *package_data[3] = {NULL, NULL, NULL};
+
+        for (const Rlist *rp = data; rp != NULL; rp = rp->next)
+        {
+            char *line = RlistScalarValue(rp);
+
+            if (StringStartsWith(line, "Name="))
+            {
+                /* We have all the information we need from previous loop
+                 * iteration. */
```
ediosyncratic: This comment belongs inside the following block; and is misleading.  It sounds like the name we've just read goes with the data from previous iterations, where actually the name we've just read forces us to check for whether we have data from earlier iterations that we need to process before going further, since the new name starts a new record to which the old data doesn't apply.

cf-agent/package_module.c
ediosyncratic: Furthermore, clearing package_data[1] and [2] happens in both branches, so should go inside the `if (package_data[0])` block after the inner `else` clause.

cf-agent/package_module.c
ediosyncratic: Also: use `BufferAppendF(inventory_data, "%s,%s,%s\n", package_data[0], package_data[1], package_data[2]);` instead of doing your own array allocation and local xsnprintf().

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+                             package_data[1], package_data[2], type);
+            
+            char inventory_line[strlen(package_data[0]) +
+                                strlen(package_data[1]) +
+                                strlen(package_data[2]) + 4];
+
+            xsnprintf(inventory_line, sizeof(inventory_line),
+                      "%s,%s,%s\n", package_data[0], package_data[1],
+                      package_data[2]);
+            BufferAppendString(inventory_data, inventory_line);
```
ediosyncratic: `BufferAppendF` would simplify this the same as above.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+                "package module while updating cache.");
+        free(options_str);
+        return false;
+    }
+    
+    if (!response)
+    {
+        Log(LOG_LEVEL_DEBUG,
+            "Received empty packages list after requesting: %s", req_type);
+    }
```
ediosyncratic: Should the `UpdatePackagesDB` call be bypassed if `!response` ?  How does it cope with being passed a NULL response ?

ediosyncratic: Should ret be false if `!response` ?  I'm suspecting code like:
```
bool ret = false;
if (!response) { Log(/* as above */); }
else
{
    if (UpdatePackgesDB(...) == 0) { ret = true; }
   else { Log(/* as below */); }
   RlistDestroy(response);
}
```

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    RlistDestroy(response);
+    free(options_str);
+    return ret;
+}
+
+
+PromiseResult ValidateChangedPackage(const NewPackages *policy_data,
+                                     const PackageModuleWrapper *wrapper,
+                                     const PackageInfo *package_info,
+                                     NewPackageAction action_type)
```
ediosyncratic: Having "New" in names is generally the set-up for future generations to laugh at you.
The college where I was a student has a "New Court" in which stands a tree, famous for the fact that two philosophers stood with their backs to that tree four hundred years ago and argued about whether a tree makes a sound in the forest when it falls, if there is no-one there to hear.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    else if (is_in_cache == 0)
+    {
+        return action_type == NEW_PACKAGE_ACTION_PRESENT ? 
+            PROMISE_RESULT_FAIL : PROMISE_RESULT_CHANGE;
+    }
+    else
+    {
+        Log(LOG_LEVEL_INFO, "Some error occurred while reading installed "
+            "packages cache.");
+        return PROMISE_RESULT_FAIL;
```
ediosyncratic: All this function's Log()s could line-break after the level and rejoin the two halves of their messages.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    if (type == PACKAGE_TYPE_FILE)
+    {
+        request = StringFormat("File=%s\n%s%s",
+            package_to_install, ver ? ver : "", arch ? arch : "");
+    }
+    else if (type == PACKAGE_TYPE_REPO)
+    {
+        request = StringFormat("Name=%s\n%s%s",
+            package_to_install, ver ? ver : "", arch ? arch : "");
+    }
```
ediosyncratic: `else { ProgrammingError(...); }` ?

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+        PackageType type, const char *packages_list_formatted,
+        const PackageModuleWrapper *wrapper)
+{
+    char *options_str = ParseOptions(options);
+    char *request = StringFormat("%s%s", options_str, packages_list_formatted);
+    
+    PromiseResult res = PROMISE_RESULT_CHANGE;
+    
+    const char *package_install_command = NULL;
+    if (type == PACKAGE_TYPE_FILE)
```
ediosyncratic: Consider using `switch` any time you have chained `if-else if` testing the same variable.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+    char *arch = architecture ? 
+        StringFormat("Architecture=%s\n", architecture) : NULL;
+    char *request = NULL;
+    
+    PromiseResult res = PROMISE_RESULT_CHANGE;
+    
+    if (type == PACKAGE_TYPE_FILE)
+    {
+        request = StringFormat("File=%s\n%s%s",
+            package_to_install, ver ? ver : "", arch ? arch : "");
```
ediosyncratic: When using complicated ?: expressions for arguments, it's kinder to the reader to put each on a separate line !

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+        return PROMISE_RESULT_NOOP;
+    }
+    
+    PromiseResult res;
+     
+    if (action == cfa_warn || DONTDO)
+    {
+         Log(LOG_LEVEL_NOTICE, "Should install file type package: %s",
+             package_file_path);
+        res = PROMISE_RESULT_FAIL;
```
ediosyncratic: I take it `DONTDO` is the dry-run flag.  I would have though `PROMISE_RESULT_NOOP` would be more apt there; but follow whatever other `DONTDO` cases do !

cf-agent/package_module.c
jimis: Indeed. Generally we try to not break strings in multiple lines, in order to improve code's grepability. If they do not fit, then we break on punctuation points.

We also do not end `Log()` lines with full stop.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+                    package->arch = SafeStringDuplicate(arch);
+                    SeqAppend(updates_list, package);
+                }
+                else
+                {
+                    Log(LOG_LEVEL_INFO, "Not able to parse available updates "
+                        "line: %s", package_line);
+                    /* Some error occurred while scanning package updates. */
+                    FreePackageInfo(package);
+                }
```
ediosyncratic: I would put the `calloc()` and setting of `package->` fields into a separate `CreatePackageInfo(version, arch)` call, done inside the `sscanf`-success branch, saving the need to `FreePackageInfo()` in its `else` clause.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+                          bool *verified)
+{
+    Log(LOG_LEVEL_DEBUG, "Installing repo type package: %d", is_in_cache);
+    
+    /* Package is not present in cache. */
+    if (is_in_cache == 0)
+    {
+        const char *version = package_info->version;
+        if (package_info->version &&
+                StringSafeEqual(package_info->version, "latest"))
```
ediosyncratic: Indentation.  A lot of your continuation lines don't line up under the first thing after the `(` that delimits what's being continued.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+            version = NULL;
+        }
+        if (action == cfa_warn || DONTDO)
+        {
+            Log(LOG_LEVEL_NOTICE, "Should install repo type package: %s",
+                package_info->name);
+            return PROMISE_RESULT_FAIL;
+        }
+        
+        *verified = false; /* Verification will be done elsewhere. */
```
ediosyncratic: Saying *where* verification happens would be far better than saying it's "elsewhere" - see above.

cf-agent/package_module.c
tzz: Moreover the pattern of "check if string starts with x, then do y if it's duplicated" is replicated in too many places.  I would factor all the repetitive pieces out, it's hard to read and really easy to have bugs.

cf-agent/package_module.c
tzz: That seems like a `ERR` message?  Why just `INFO`?

tzz:
It would have been helpful to merge small changes incrementally instead of one huge PR.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+                free(package_data);
+                return NULL;
+            }
+            package_data->arch = 
+                SafeStringDuplicate(line + strlen("Architecture="));
+        }
+        /* For handling errors */
+        else if (StringStartsWith(line, "Error="))
+        {
+            Log(LOG_LEVEL_VERBOSE, "have error: %s",
```
tzz: "have error" is not very helpful if you're reading the logs.  Maybe you can indicate the component name?

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+                    ValueSizeDB(db_updates, package_key, sizeof(package_key));
+            char buff[val_size + 1];
+            buff[val_size] = '\0';
+
+            ReadDB(db_updates, package_key, buff, val_size);
+            Seq* updates = SeqStringFromString(buff, '\n');
+            
+            for (int i = 0; i < SeqLength(updates); i++)
+            {
+                PackageInfo *package = calloc(1, sizeof(PackageInfo));
```
jimis: `xcalloc()` instead of `calloc()`, since you're not checking the return value.

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+
+        free(package_info);
+    }
+}
+
+static char *ParseOptions(Rlist *options)
+{
+    if (RlistIsNullList(options))
+    {
+        return SafeStringDuplicate("");
```
jimis: Use `xstrdup()`, why mask a perfectly recognisable libc call?

cf-agent/package_module.c
```diff
@@ -0,0 +1,1611 @@
...
+            {
+                /* Some error occurred as we already have name for 
+                 * given package. */
+                Log(LOG_LEVEL_ERR, "Duplicated package name "
+                    "received: [%s] %s", line, package_data->name);
+                free(package_data);
+                return NULL;
+            }
+            package_data->name = 
+                SafeStringDuplicate(line + strlen("Name="));
```
jimis: `xstrdup()`

cf-agent/verify_new_packages.c
```diff
@@ -0,0 +1,165 @@
...
+                                          Attributes a, char **promise_log_msg,
+                                          LogLevel *log_lvl)
+{
+    Log(LOG_LEVEL_DEBUG, "New package promise handler");
+    
+    
+    if (!NewPackagePromiseSanityCheck(a))
+    {
+        *promise_log_msg =
+                SafeStringDuplicate("New package promise failed sanity check.");
```
jimis: `xstrdup()` here too.

There are a bunch of other places that we know the argument is not NULL.

cf-agent/verify_new_packages.c
```diff
@@ -0,0 +1,165 @@
...
+        return PROMISE_RESULT_FAIL;
+    }
+    
+    PromiseBanner(ctx, pp);
+    
+    PackagePromiseGlobalLock global_lock = AcquireGlobalPackagePromiseLock(ctx);
+    
+    CfLock package_promise_lock;
+    char promise_lock[CF_BUFSIZE];
+    snprintf(promise_lock, CF_BUFSIZE - 1, "new-package-%s-%s",
```
jimis: Why are you using `CF_BUFSIZE-1` as size, I can't find that you are appending a character later. So `-1` is not needed, and anyway it's better to use `sizeof(promise_lock)`.

cf-agent/verify_new_packages.c
```diff
@@ -0,0 +1,165 @@
...
+                                                a.transaction.action);
+            *log_lvl = result == PROMISE_RESULT_FAIL ?
+                                 LOG_LEVEL_ERR : LOG_LEVEL_VERBOSE;
+            *promise_log_msg = result == PROMISE_RESULT_FAIL ?
+                StringFormat("Error installing package '%s'", pp->promiser) :
+                StringFormat("Successfully installed package '%s'", pp->promiser);
+            break;
+        case NEW_PACKAGE_ACTION_NONE:
+        default:
+            assert(0); /* This MUST be handled by sanity check. */
```
jimis: `ProgrammingError("relevant message")`

cf-agent/verify_packages.c
```diff
@@ -131,8 +144,91 @@ PackageManager *INSTALLED_PACKAGE_LISTS = NULL; /* GLOBAL_X */
...
+    Attributes a = GetPackageAttributes(ctx, pp);
+    PackagePromiseType package_promise_type =
+            GetPackagePromiseVersion(&a.packages, &a.new_packages);
+    
+    switch (package_promise_type)
+    {
+        case PACKAGE_PROMISE_TYPE_NEW:
+            result = HandleNewPackagePromiseType(ctx, pp, a, &promise_log_message,
+                    &level);
+
```
jimis: I would initialise `promise_log_message = NULL` before this call, and here I would `assert(promise_log_message != NULL)`. It was hard to follow the logic of the function call, and I bet it will be easy to regress after some time when someone forgets to set the log message.

cf-agent/verify_new_packages.c
```diff
@@ -0,0 +1,165 @@
...
+    CfLock package_promise_lock;
+    char promise_lock[CF_BUFSIZE];
+    snprintf(promise_lock, CF_BUFSIZE - 1, "new-package-%s-%s",
+             pp->promiser, a.new_packages.module_body->name);
+
+    if (global_lock.g_lock.lock == NULL)
+    {
+        *promise_log_msg =
+                SafeStringDuplicate("Can not acquire global lock for package "
+                                    "promise. Skipping promise evaluation");
```
jimis: Break the two lines at the punctuation. For example:
```c
xstrdup("Could not acquire global lock for package promise,"
        " skipping promise evaluation");
```

libpromises/dbm_api.c
```diff
@@ -180,6 +253,40 @@ static DBHandle *DBHandleGet(int id)
...
+    }
+    /* Keep mutex locked. */
+
+    /* If we exited because of timeout make sure we Log() it. */
+    if (handle->refcount != 0) {
+        Log(LOG_LEVEL_ERR,
+                "Database %s refcount is still not zero (%d), forcing CloseDB()!",
+                handle->filename, handle->refcount);
+        DBPrivCloseDB(handle->priv);
+    } else /* TODO: can we clean this up unconditionally ? */ {
```
jimis: Coding style: braces on separate lines. I don't see why you changed it from where you moved the code.

libpromises/dbm_api.c
```diff
@@ -66,6 +76,7 @@ struct DBCursor_
 static pthread_mutex_t db_handles_lock = PTHREAD_ERRORCHECK_MUTEX_INITIALIZER_NP; /* GLOBAL_T */
 
 static DBHandle db_handles[dbid_max] = { { 0 } }; /* GLOBAL_X */
+static DynamicDBHandles *db_dynamic_handles;
```
jimis: Why? Why do we need a dynamic amount of databases? Haven't we had enough trouble with only a static number of those? ;-)

cf-agent/package_module.c
pasinskim: Makes sense and looks much simpler.

cf-agent/package_module.c
pasinskim: No. default_inventory is taken from ctx which is persistent for the whole agent run and is destroyed at the end of its run.

cf-agent/package_module.c
pasinskim: We can either have some "valid" entry on the list of it can be defined as empty list in the policy. If this is an empty list there is no need to continue iteration.

cf-agent/package_module.c
pasinskim: Makes sense. At the moment ignoring error messages (no error recovery) and only action performed is logging.

cf-agent/package_module.c
pasinskim: Fixed!

cf-agent/package_module.c
pasinskim: Fixed!

cf-agent/package_module.c
pasinskim: Leftover from previous implementation when I've read data both form stdout and stderr. Now stderr is handled elsewhere.
Fixed!

cf-agent/package_module.c
pasinskim: Right, fixed!

cf-agent/package_module.c
pasinskim: Fixed!

libpromises/pipes_unix.c
pasinskim: Indeed. In fact I don't have to initialize this at all. If some descriptor won't be valid after this function exits return value will indicate error.

cf-agent/package_module.c
pasinskim: Fixed!

cf-agent/package_module.c
pasinskim: Agree, fixed!

cf-agent/package_module.c
pasinskim: Yes, I was trying to achieve this in all the places.

cf-agent/package_module.c
pasinskim: Duplicated is indicating that we have received second line with the same parameter description. We are expecting ONLY one so "Duplicated name description received" means that we already had one line describing name of the package. So it is not indicating that name is exactly the same but line containing package name description is seen 2nd time.

pasinskim: Makes sense @ediosyncratic. @tzz in fact it is only in 3 places in this particular function.

cf-agent/package_module.c
pasinskim: Fixed! General error is printed with this log message 'Log(LOG_LEVEL_ERR,
+ "Can not set up wrapper for module: %s", pm_name);'. This is just more detailed and can be seen once we are interested in more details (info level in this case).

cf-agent/package_module.c
pasinskim: Fixed!

cf-agent/package_module.c
pasinskim: Comment updated.

cf-agent/package_module.c
pasinskim: Right. Thanks for pointing this out.

cf-agent/package_module.c
pasinskim: Like second version.

cf-agent/package_module.c
pasinskim: Fixed!

cf-agent/package_module.c
pasinskim: BufferAppendF looks much simpler and better.

cf-agent/package_module.c
pasinskim: Thanks for this one. Fixed!

cf-agent/package_module.c
pasinskim: Changed!

cf-agent/package_module.c
pasinskim: Empty data is also valid data in this case. ```UpdatePackagesDB``` also calls ```CleanDB``` which we still need to do if we are having no data.

cf-agent/package_module.c
pasinskim: Fully agree. Will try to fix all leftovers.

cf-agent/package_module.c
pasinskim: Better safe than sorry. In fact this is done in other places.

cf-agent/package_module.c
pasinskim: Always for more than 3 checks. I suppose in simpler cases (or even when more checks are done) compiled code will be the same. For readability I think 3 cases is quite OK to use ```if-else if```.

cf-agent/package_module.c
pasinskim: Agree. Changed!

cf-agent/package_module.c
pasinskim: It's standard pattern to return ```PROMISE_RESULT_FAIL``` in such a cases.

cf-agent/package_module.c
pasinskim: In fact I can move ```calloc()``` inside ```if``` clause and remove ```FreePackageInfo()``` from ```else```.

cf-agent/package_module.c
pasinskim: Fixed!

cf-agent/package_module.c
pasinskim: Changed.

cf-agent/package_module.c
pasinskim: Makes a lot of sense. Fixed!

cf-agent/package_module.c
pasinskim: Fixed!

cf-agent/verify_new_packages.c
pasinskim: Probably copy-pasted line of code from here: ```snprintf(lockname, CF_BUFSIZE - 1, "package-%s-%s", pp->promiser, a.packages.package_list_command);``` (verify_packages.c).

cf-agent/verify_new_packages.c
pasinskim: Fixed!

cf-agent/verify_packages.c
pasinskim: Makes sense.

libpromises/dbm_api.c
pasinskim: Fixed! My editor has option to beautify the code after copy-paste which is not perfect all the time...

libpromises/dbm_api.c
pasinskim: This is needed as we need one DB per package module. We can not know how many of those will be used in policy. In fact dynamic is static from the durability perspective. It is dynamic from the number of instances perspective.

tzz:
When will this code switch to `modules/packages`?

tzz:
The error when the module is missing is not helpful.  Could the code be a bit more verbose and tell the user where it looked?  I'd make it at least `INFO` and maybe even `ERR` level.

kacfengine:
I have it in my test branch, but this is @pasinskim's PR, and he is out of the office right now. Do you need it, or were you just reviewing?

kacfengine:
> The error when the module is missing is not helpful. Could the code be a bit more verbose and tell the user where it looked? I'd make it at least INFO and maybe even ERR level.

Fully agree, in fact I ran into this right now. :-) I will fix it.

tzz:
@kacfengine OK, as long as you plan to  switch soon.  I've been testing in my own environment with the old location, it's not a big deal.

kacfengine:
Closing in favor of #2224.

PR #2216, opened 2015-05-20
Deprecate cf-runagent -o.

jimis:


jimis:
This is the first part of the solution, the second is the server-side EXEC sanitisation.

@kacfengine please review, so that this simple commit is ready to be merged in case I don't complete the server-side part before the Beta.

jimis:
trigger build

kacfengine:
Looks good and green!

PR #2217, opened 2015-05-22
Redmine#6978, Redmine#5715: Extra checks for EXEC command in cf-serverd

jimis:


jimis:
@kacfengine @ediosyncratic please review, per commit and ignoring whitespace!

jimis:
trigger build

cf-serverd/server_common.c
```diff
@@ -271,231 +252,170 @@ int MatchClasses(EvalContext *ctx, ServerConnectionState *conn)
...
+    int ret = snprintf(cmdbuff, CF_BUFSIZE, "%s -I -D",
+                       CFRUNCOMMAND);
+    /* TODO what if cfruncommand is not cf-agent?
+       These arguments don't apply then. */
 
-    if ((strlen(word) == 2) && (strncmp(pos, word, 2) == 0))
+    if (ret <= 0 || ret >= CF_BUFSIZE)
     {
-        return true;
+        /* impossible? */
```
kacfengine: No, I don't think so. If ``CFRUNCOMMAND -I -D`` exceeds ``CF_BUFSIZE`` (which is possible, though very unlikely), ``snprintf`` will return a value higher than it.

``ret <= 0`` should be impossible unless the formatting string is wrong.

cf-serverd/server_common.c
```diff
@@ -271,231 +252,170 @@ int MatchClasses(EvalContext *ctx, ServerConnectionState *conn)
...
+    /* Pointer to append list of classes. */
+    char  *authorized_classes     = &cmdbuff[ret];
+    size_t authorized_classes_len = 0;
 
-    if (strncmp(pos, word, len) != 0)
-    {
-        return false;
-    }
+    StringAppendDelimited(authorized_classes, &authorized_classes_len,
+                          CF_BUFSIZE, "cfruncommand", ',');
```
kacfengine: ``CF_BUFSIZE`` happens to be half of ``CF_EXPANDSIZE``, so your content will always fit. However, perhaps it is better to be explicit here, and declare ``cmdbuff[CF_BUFSIZE*2]``? That would also protect you in case either constant is changed in the future and the relationship is not half/double anymore.

kacfengine:
Looks good otherwise! (I haven't looked at Jenkins)

cf-serverd/server_common.c
jimis: RIght, I'll change this, though it still seems cryptic to me. :-(

This method with `CF_EXPANDSIZE` was being used before, but I'm wondering if I should just go with dynamic allocation and forget about the whole fit/does not fit problem.

Hmmm of course in that case I would have to check and report `malloc()` failures, so I think I'll keep it like that.

jimis:
BTW jenkins is green. I'll amend a couple of easy fixes (based on the review) and do some testing on a custom deployment, so don't merge yet.

jimis:
Minor changes done, now on to the deployment phase...

jimis:
Fixed review comments and quirks revealed by testing, and rebased.

jimis:
trigger build

jimis:
Build green!

PR #2219, opened 2015-05-22
Fix segfault while executing cf-agent -x.

maciejmrowiec:
Add check for EVP_DigestInit status,


jimis:
Thanks, I don't get why it triggered, but the patch certainly improves what was there!

PR #2222, opened 2015-05-26
redmine7219: clarify error message and condition on caring.

ediosyncratic:
No column has overflowed anything.  We *are* on the point of
overflowing the table, if there's more columns to come - except that
we side-step and don't make that mistake.  In any case, filling the
last column of the table isn't an error; the only problem would come
if there were another column to come, for which we lack space in the
table.  So stop to see if there actually *is* another column to come,
before Log()ging any error.  Also, testing > limit - 1 is just an ugly
way to test >= limit !


nickanderson:
This should be backported to 3.6.x when it goes in.

kacfengine:
trigger build

kacfengine:

PR #2223, opened 2015-05-28
redmine7138: pass open file descriptors portably

ediosyncratic:
<s>This is a preview, not yet a real pull request; I still need to fix some problems with the test.
I *shall* push -f a replacement branch before I'm ready to suggest merging this.
In particular, the Makefile.am changes need to move to the end, once the test is passing (and I've verified it fails before implementing the feature), so that git bisect never hits a point in the middle where tests fail.</s>
Tested on Jenkins and in our scale-test environment.  We need to test on MS-Win (but at least it compiles there, now).

### DIFF
--- a/libutils/passopenfile.c
+++ b/libutils/passopenfile.c
@@ -1,0 +1,34 @@
+/*
+   Copyright (C) CFEngine AS
+
+   This file is part of CFEngine 3 - written and maintained by CFEngine AS.
+
+   This program is free software; you can redistribute it and/or modify it
+   under the terms of the GNU General Public License as published by the
+   Free Software Foundation; version 3.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+#include <passopenfile.h> /* Starts with <platform.h> */
+
+bool PassOpenFile_Put(int uds, int descriptor, const char *text)
+{
+    return false;
+}
+
+int PassOpenFile_Get(int uds, char **text)
+{
+    return -1;
+}
--- a/libutils/passopenfile.c
+++ b/libutils/passopenfile.c
@@ -338,6 +338,9 @@ int PassOpenFile_Get(int uds, char **text)
 #error "No support for connection sharing on this platform :-("
 #endif
 
+#ifndef MSG_WAITALL /* Linux >= 2.2 */
+#define MSG_WAITALL 0
+#endif
 static const char NULL_MSG[] = "\0NULL";
 
 bool PassOpenFile_Put(int uds, int descriptor, const char *text)
@@ -451,7 +454,7 @@ int PassOpenFile_Get(int uds, char **text)
 #endif
 
     /* Receive message: */
-    if (recvmsg(uds, &message, 0) < 0)
+    if (recvmsg(uds, &message, MSG_WAITALL) < 0)
     {
         Log(LOG_LEVEL_ERR,
             "Can't receive descriptor (recvmsg: %s)",
--- a/libutils/passopenfile.c
+++ b/libutils/passopenfile.c
@@ -23,12 +23,499 @@
 */
 #include <passopenfile.h> /* Starts with <platform.h> */
 
+#include <logging.h>
+#include <alloc.h>
+
+/* Local tuning parameter: should anything else know about it ? */
+#define MAX_MESSAGE_SIZE  1024 /* strlen()+1 limit on text transmitted */
+
+#ifdef __MINGW32__
+#include <printsize.h>
+static const char PID_FMT[] = "PID: %lu\n";
+#define PID_MSG_SIZE (PRINTSIZE(unsigned long) + 8)
+static const char ACK_MSG[] = "ACK\n";
+typedef uint32_t MSG_LEN_T;
+
+/* TODO: replace explicit waiting with use of blocking UDS. */
+/* Package a call to select(), to avoid duplicating boilerplate code. */
+static bool wait_for(const int uds, bool write, bool *ready)
+{
+    struct timeval tv;
+    fd_set fds;
+
+    FD_ZERO(&fds);
+    FD_SET(uds, &fds);
+    tv.tv_sec = 1; /* Wait for up to a second */
+    tv.tv_usec = 0;
+
+    int ret;
+    if (write)
+    {
+        ret = select(uds + 1, NULL, &fds, NULL, &tv);
+    }
+    else
+    {
+        ret = select(uds + 1, &fds, NULL, NULL, &tv);
+    }
+
+    if (ret < 0)
+    {
+        return false;
+    }
+    *ready = FD_ISSET(uds, &fds);
+    return true;
+}
+
 bool PassOpenFile_Put(int uds, int descriptor, const char *text)
 {
+    /* From the WSADuplicateSocket() doc [0]: "A process can call
+     * closesocket on a duplicated socket and the descriptor will
+     * become deallocated. The underlying socket, however, will remain
+     * open until closesocket is called by the last remaining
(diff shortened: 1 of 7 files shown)


ediosyncratic:
@jimis please take a look.

jimis:
trigger build

jimis:
+1 for first writing an extensive test suite, and then doing the actual implementation. I should take example!

jimis:
No comments really, the implementation looks great, covering every possible aspect!

There are mainly three different code paths, which ones are actually tested? I'm quite sure windows is not tested, but AFAIK we're not even currently running unit tests on windows, right @kacfengine?

ediosyncratic:
At present, I've only tested `HAVE_MSGHDR_MSG_CONTROL` (the "Linux" variant) but I imagine the "BSD" variant shall be tested by a build on Solaris.  Testing for MS-Win is an open challenge, for which we'll need to contrive a build that *builds* the test program, so that we can run it on an MS-Win host.  I doubt I shall have time for this in the next three days ...

ediosyncratic:
Compiled on MinGW for the first time, got some errors.
Applied some fixup()s to the implementation and rebased.

ediosyncratic:
Updated description to reflect current state of branch: tested in Jenkins, compiles on MinGW.
The branch is now in its final order, so things aren't enabled until they work, ensuring that git bisect won't be thrown off by the transient period when the test failed.

libutils/passopenfile.c
```diff
@@ -0,0 +1,509 @@
...
+                Log(LOG_LEVEL_ERR,
+                    "Failed to read size of text accompanying descriptor");
+                size = 0;
+            }
+
+            if (text && size && !*text)
+            {
+                closesocket(descriptor);
+                return -1;
+            }
```
kacfengine: I thought you said that passing a string consisting of just ``\0`` was ok? Or did I misunderstand?

libutils/passopenfile.c
```diff
@@ -0,0 +1,509 @@
...
+{
+    struct msghdr message;
+    struct iovec vector;
+    size_t msglen = text ? strlen(text) + 1 : sizeof(NULL_MSG);
+    assert(MAX_MESSAGE_SIZE >= msglen);
+    Log(LOG_LEVEL_VERBOSE,
+        "Connected to peer, passing descriptor %d with %s %s",
+        descriptor,
+        text ? "text:" : "no",
+        text ? text : "text");
```
kacfengine: I think this should be debug log level, it will be quite verbose since this happens on every connection, right? It's also too technical to be useful for the user.

libutils/passopenfile.c
```diff
@@ -0,0 +1,509 @@
...
+#define INTERFACE_STYLE "Linux"
+#elif HAVE_MSGHDR_ACCRIGHTS
+/* This is the old (BSD) interface.  No Linux should use this, it's
+ * only known to be present in BSD-based Unixes. */
+#define INTERFACE_STYLE "BSD"
+#else
+#error "No support for connection sharing, this code should not be compiled"
+#endif
+
+#ifndef MSG_WAITALL /* Linux >= 2.2 */
```
kacfengine: I really hope nobody is still using that! :-)

kacfengine:
HP-UX failed on my private branch with:

```
passopenfile.c: In function 'PassOpenFile_Put':
passopenfile.c:376: error: 'struct msghdr' has no member named 'msg_flags'
```

libutils/passopenfile.c
ediosyncratic: The case you're thinking of is an empty string: size is 1, *text is a non-NULL pointer and **text is '\0'.  The size includes trailing '\0' so that we can distinguish empty string from NULL (size == 0).

libutils/passopenfile.c
ediosyncratic: Fair point.

libutils/passopenfile.c
ediosyncratic: Not sure what you mean.  Plenty of folk are using Linux >= 2.2 and some are not using Linux.
Did you mean you hope no-one is using Linux < 2.2 ?  Fair enough.
However, non-Linux won't have MSG_WAITALL, that I know of.

ediosyncratic:
Reviewed Log() levels.
Use memset() to clear structs to avoid the msg_flags problem.

libutils/passopenfile.c
kacfengine: Yes, I meant Linux < 2.2. :-)

libutils/passopenfile.c
kacfengine: Right, the part I missed was that it is a double pointer, I thought ``*text`` would be the ``\0``.

kacfengine:
When Solaris 9 (eventually) finished I also got this:

```
passopenfile.c: In function 'PassOpenFile_Put':
passopenfile.c:379:5: error: implicit declaration of function 'CMSG_SPACE' [-Werror=implicit-function-declaration]
passopenfile.c:388:5: error: implicit declaration of function 'CMSG_LEN' [-Werror=implicit-function-declaration]
```

Will it be solved already?

ediosyncratic:
Solaris hides its defines in sys/socket.h; duly added (in a fixup for now: @kacfengine please test and let me know if it works; I'll then rebase it into place).

kacfengine:
Testing...

kacfengine:
Solaris 9 again:

```
passopenfile.c:334:2: error: #error "No support for connection sharing, this code should not be compiled"
passopenfile.c: In function 'PassOpenFile_Get':
passopenfile.c:430:54: error: expected ')' before 'INTERFACE_STYLE'
```

ediosyncratic:
OK, this means that configure is seeing the `_XPG2_4` branch of `<sys/socket.h>` but when we compile we're not.  The `_XPG2_4` define is controlled by `<feature_test.h>` which says to see `man 5 standards` which claims that the standards I think `_XPG2_4` relates to aren't available until Solaris 10.
This may mean that *in fact* our Solaris 9 has support for the needed features, and we just have to set the right define (it'll be there among the ones configure.log reports as being in the failing program when it tests for `msg_accrights`; there are only a few hundred of them) somewhere sensible (probably to be seen by all compilation units on Solaris 9).  Otherwise, it means Solaris 9 lacks the features and configure is passing the wrong defines.  Fix that and it'll set the correct one of the `HAVE_MSGHDR_*` defines and all shall be well.

Good luck :-)

ediosyncratic:
I said:
> [...] see  `man 5 standards` which claims that the standards I think  `_XPG2_4`  relates to aren't available until Solaris 10.
IIRC, it was the Solaris 10 man page that claimed this.
The Solaris 9 `man 5 standards` claims XPG4v2 is supported since Solaris 2.6; and
> The application must define `_XOPEN_SOURCE` with a value other than 500 (preferably 1) and set `_XOPEN_SOURCE_EXTENDED=1`.
I see that our `config.log` reports the failed compilation for the msg_accrights check including: ```
#define _XOPEN_SOURCE 500
```
This setting selects the SUSv2 standard, supported since Solaris 7.
In `core/configure.ac` we set `_XOPEN_SOURCE` to 500 on solaris 2.8 and 2.9 but to 600 on 2.10 and 2.11; this define is propagated to `libutils/config.h` which is included by `libutils/platform.h`, so we should be seeing it.
All of which looks like it should lead to `HAVE_MSGHDR_MSG_CONTROL` being defined, which would give us the "modern" SCM_RIGHTS variant; indeed, `config.log` does report success for that check.  So why isn't `passopenfile.c` seeing the define set ?
Sure enough, `libutils/config.h` is defining it.
Ah - but the fixup I had before breaks this by over-riding the defines in `passopenfile.c`; it would seem we need to change the `_XOPEN_SOURCE` setting for Solaris 9, given that `.msg_control` didn't work for it.
So a different fixup is needed, probably in configure.in's settings for solaris [89].

ediosyncratic:
After some hacking on the build machine, I found a less drastic fixup that worked: supply the `CMESG_*` macros whose absence prompted the previous fixup.

ediosyncratic:
trigger build

kacfengine:
Passes on Solaris 9, but fails on HP-UX now. :-(

http://jenkins.usdc.cfengine.com/view/Pull%20requests/job/testing-enterprise-pr_core/168/label=PACKAGES_ia64_hpux_11.23/console

ediosyncratic:
trigger build

ediosyncratic:
This is a reprise of one of the problems I did a fixup for before but lost - the price of going away in the middle of the work ! - setting each field of the struct is a dumb way to zero-fill, memset() ensures we clear all fields, even the ones some exotic implementation has that we've never heard of; it also avoids HP-UX's problem here.

ediosyncratic:
Let me know when you've reviewed the fixups and I'll push a suitably rebased version that squashes them into the commit they amend.

kacfengine:
Yep, the fixup makes perfect sense!

kacfengine:
trigger build

kacfengine:
For the time being merged to master only. We can backport it later if we need it.

PR #2225, opened 2015-06-02
improve error message seen from cf-key -r

estenberg:
cf-key -r may output a message that lastseen is not coherent, but this does not mean db-level corruption.
Running cf-key -r -x skips coherence check but this function is generic so it did not seem appropriate to mention this.

https://cfengine.zendesk.com/agent/tickets/1964


kacfengine:
Looks good!

Btw, don't push feature branches to main cfengine/core, use a personal repository for that! :-)

kacfengine:
Cherry-picked to both 3.6.x and 3.7.x.

PR #2227, opened 2015-06-09
redmine#7242: Avoid writing hundreds of megabytes by preserving sparse areas in LMDB files.

jimis:


jimis:
Do not merge, it's only opened to check jenkins.

jimis:
trigger build

kacfengine:
+1!

kacfengine:
It shaves almost 25 min off the testing time on HP-UX, nice!

jimis:
trigger build

jimis:
@kacfengine the total test time was 3:14, even slower for some reason, before even reaching the acceptance tests!

kacfengine:
?

On HP-UX the test time was 67 min, better than ever before. This is the one I'm looking at: http://jenkins.usdc.cfengine.com/view/Pull%20requests/job/testing-enterprise-pr_core/149/label=PACKAGES_ia64_hpux_11.23/consoleFull

jimis:
Here are the last lines of that log for me:
```
03:14:30.538 
03:14:30.538 Deleting project workspace... 
03:14:30.735 done
03:14:30.735 
03:14:30.755 Finished: UNSTABLE
```

It's probably different from what you're seeing because I've set Jenkins to display *elapsed time*. But looking a bit better it seems that it started the run at 2:07 so total time is indeed about one hour! The initial delay is probably because of already queued jobs.

kacfengine:
Yep, so this looks pretty good then!

jimis:
@kacfengine then if you can review the code and merge at will!

jimis:
BTW I was hoping to put `memcchr()` in libcompat, since it exists on some BSDs, but would prefer to avoid spending more time.

kacfengine:
Yeah, let's merge it for now, we really need the speed optimization. Looks good to me!

kacfengine:

jimis:
YES!

jimis:
Cherry-picking to 3.6.x

PR #2229, opened 2015-06-11
Reduce usage of test_suppress_fail.

kacfengine:
It's not giving us useful test data to have the build contain
persistent failures. Better to save test_suppress_fail for the bad
cases and let the Jenkins build be green instead of yellow.

test_suppress_fail is still available if we need to suppress a really
bad bug.

### DIFF
--- a/tests/acceptance/01_vars/02_functions/sort.cf
+++ b/tests/acceptance/01_vars/02_functions/sort.cf
@@ -63,7 +63,7 @@ bundle agent init
 bundle agent test
 {
   meta:
-      "test_suppress_fail" string => "hpux|sunos_5_9|windows",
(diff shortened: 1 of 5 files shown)


kacfengine:
@pasinskim, @jimis: What do you think?

kacfengine:
trigger build

jimis:
Are we sure that the few soft failures that we currently have are unimportant? Are they to be fixed soon?

Else +1, it's been a long time since the build was green!

kacfengine:
I went through them, they're all just regular bugs, none of them major. But this was the reason I didn't go through the Windows failures, there are a lot more of those and we should make sure no serious bugs are hiding in any of them.

PR #2230, opened 2015-06-11
Redmine #7235 Fix global promise locking.

pasinskim:
While cf-agent is run by cf-execd global promise lock is perceived
due to ifelapsed set by default to 1 minute. As global promise lock
should be available immediately for every new process it should be
set to 0.


kacfengine:
+1!

PR #2231, opened 2015-06-11
Redmine #7216: Don't report repaired status unless we change the system

kacfengine:
Killing is changing the system, setting a class is not.


kacfengine:
trigger build

PR #2232, opened 2015-06-15
policy server cannot copy from masterfiles shortcut because cf-serverd is not yet up

estenberg:
Redmine #7251


kacfengine:

PR #2233, opened 2015-06-15
Fix upgrading packages with multiple architectures

pasinskim:


kacfengine:
Nice! I'll try it before merging.

kacfengine:
Merged manually with one more bug fix, and cherry-picked to 3.7.x.

PR #2236, opened 2015-06-16
Include the template files for extensions_template.[ch] in the source tarball

jimis:
To achieve this the files were renamed out of their .[ch] extension,
because make was trying to compile them after listing them as a
dependency to libpromises. By giving the extension .c.pre, it's clear
that they are not C files, but files that need pre-processing to become
valid C files.

Also removed their mention in the root Makefile, it doesn't seem to be
necessary anymore.


jimis:
This fixes the final problems with the source tarball.

@kacfengine I also removed the circular-dependency fix from the main Makefile, it doesn't seem to be needed anymore.

libpromises/Makefile.am
```diff
@@ -204,9 +205,9 @@ BUILT_SOURCES = cf3lex.c cf3parse.h cf3parse.c \
 EXTRA_SCRIPTS = text2cstring.pl enterprise_extension.sed
 EXTRA_DIST = $(EXTRA_SCRIPTS)
 
-enterprise_extension.c: extensions_template.c enterprise_extension.sed
+enterprise_extension.c: enterprise_extension.c.pre enterprise_extension.sed
```
kacfengine: It should be called ``extensions_template.c.pre``, because this is not only used for the enterprise extension, it's also used for the report collector plugin for cf-hub. We have two extensions.

jimis: You're right, will update.

jimis:
Updated, checking enterprise and nova now.

jimis:
I'm merging this together with enterprise. Once testing-master and nightlies-master are green, I'll cherry-pick.

kacfengine:
Ok, looks good to me!

PR #2238, opened 2015-06-16
contrib/cfengine.el: 3.7 indentation and JSON improvements

tzz:
@ediosyncratic @leoliu please take a look.  This will be merged into Emacs as well, later.

Test case, indented correctly:

```
bundle agent main
{
@if version_after(3.7)
  reports:
    any::
    "any"::
    '$(any)'::
    'any'::
      "hello 3.7!
  @this should stay indented a few spaces";

  vars:
      "mything" data => '{
  "a": 100,
  "b": [
    "c",
    200
  ]
}';
@endif
}
```

Changes:

* recognize macros like `@anything`, apply the correct syntax, and indent to column 0 (unless inside a string, note `@this should stay indented` stays indented)
* when macro `@anything` is not at column 0, apply the warning face
* allow class selectors to look like `"myclass.$(myvar)"::` (but they still have string appearance, that's very hard to override)
* new interactive function `cfengine3-reformat-json-string` will, when inside a string, reformat it as JSON.  This is very handy when you're editing.  I hesitated to make it more DWIM.  It's not bound to a key, should it be by default?  I'm also not sure about the way it finds the closing character; maybe it should be smarter about EOB and errors.

Let me know if you have further suggestions or ideas.

### DIFF
--- a/contrib/cfengine.el
+++ b/contrib/cfengine.el
@@ -100,7 +100,7 @@ will use a fallback syntax definition."
   :group 'cfengine
   :type '(choice file (const nil)))
 
-(defcustom cfengine-parameters-indent '(promise pname 0)
+(defcustom cfengine-parameters-indent '(promise pname 2)
   "Indentation of CFEngine3 promise parameters (hanging indent).
 
 For example, say you have this code:
@@ -121,15 +121,15 @@ You can also choose to indent the start of the word
 
 Finally, you can choose the amount of the indent.
 
-The default is to anchor at promise, indent parameter name, and offset 0:
+The default is to anchor at promise, indent parameter name, and offset 2:
 
 bundle agent rcfiles
 {
   files:
     any::
       \"/tmp/netrc\"
-      comment => \"my netrc\",
-      perms => mog(\"600\", \"tzz\", \"tzz\");
+        comment => \"my netrc\",
+        perms => mog(\"600\", \"tzz\", \"tzz\");
 }
 
 Here we anchor at beginning of line, indent arrow, and offset 10:
--- a/contrib/cfengine.el
+++ b/contrib/cfengine.el
@@ -50,6 +50,15 @@
 
 ;; (add-hook 'cfengine3-mode-hook 'eldoc-mode)
 
+;; You may also find the command `cfengine3-reformat-json-string'
+;; useful, just bind it to a key you prefer. It will take the current
+;; string and reformat it as JSON. So if you're editing JSON inside
+;; the policy, it's a quick way to make it more legible without
+;; manually reindenting it.  For instance:
+
+;; (global-set-key [(control f4)] 'cfengine3-reformat-json-string)
+
+
 ;; It's *highly* recommended that you use this mode with the latest
 ;; Emacs, or at least 24.1.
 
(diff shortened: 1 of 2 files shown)


kacfengine:
Great stuff!

> Let me know if you have further suggestions or ideas.

What about promise attributes? Currently the code is auto-indented like this:

```
{
  files:
      "/tmp/myfile"
      create => "true";
}
```

whereas it should be indented like this:

```
{
  files:
      "/tmp/myfile"
        create => "true";
}
```

which is the preferred standard, and used extensively (but unfortunately inconsistently) in masterfiles.

What's here is already great though, if the additional changes take considerable effort, we can split it and merge this immediately.

tzz:
@kacfengine the indentation **defaults** in this code can be overridden by `contrib/dir-locals.el`.  When you use `contrib/reindent.pl` on any policy, it will use that file, so take a look, it has C and CFEngine style settings.  At least when I was maintaining the masterfiles, we kept to that standard for new code.

But I've changed the indentation defaults in `cfengine.el` as well as you requested because it seems everyone prefers the indented style.  It's easy enough to change it back for anyone individually.

kacfengine:
Ah, nice, I didn't know about that! I think it's a good idea to have it as default, yes.

Shall I merge this then?

tzz:
I think it's ready as it is, but let @ediosyncratic or @leoliu take a look, they use this a lot so they may have more feedback.  In particular the way that `cfengine3-reformat-json-string` works may not be ideal.

One thing that's been bugging me is that function prototypes, when I show them with eldoc, can only show the type of the argument.  So for instance when I type `regcmp(` it will tell me that it expects two strings, but I don't know which one is the regex.  It would be nice if that was available in the syntax dump and then I could grab it.  But that's probably a more long-term task.

kacfengine:
Ok, I'll wait a bit.

tzz:
I added a commit that adds the function parameter `description` to the syntax dump in JSON format.  I think that's a really useful improvement and use it in `contrib/cfengine.el` to distinguish between parameters with the same type.  It looks like `function(string /desc1/, string /desc2/)`.

I don't think the change above will affect users in any way.  The JSON syntax dump is only used by a few tools: documentation generator and `contrib/cfengine.el` as far as I know.

leoliu:
I am not fully aware of cfengine's string syntax. for example, can ' appear inside ''? Why does \ have punctuation syntax?

Maybe the following is more robust:

```
(defun cfengine3-reformat-json-string ()
  "Reformat the current string as JSON using `json-pretty-print'."
  (interactive)
  (let ((ppss (syntax-ppss)))
    (when (nth 3 ppss)                  ;inside a string
      (save-excursion
        (json-pretty-print (nth 8 ppss)
                           (progn (goto-char (nth 8 ppss)) (forward-sexp 1)
                                  (point)))))))
```

tzz:
@leoliu thanks, that was very helpful.  I've reworked that function as you suggested (with some minor changes because `json-pretty-print` takes the region **inside** the string).

@kacfengine I think this is ready for merge, if you agree with the change to `syntax.c` to include the parameter `description` in the JSON syntax dump.  That `description` can and should eventually be used by the documentation generator as well; right now we have to provide it manually, duplicating information unnecessarily.  So this change allows that in the future...

tzz:
The JSON reformat will fail if the JSON is invalid, which in fact is an excellent way to ensure the JSON is parseable.

kacfengine:
I think the change to ``syntax.c`` is indeed justified, better to share that information. Running it through the test analyzer once, just in case there is a test somewhere that looks for exact JSON output.

kacfengine:
trigger build

kacfengine:
Thanks, cherry-picked to 3.7.x as well.

PR #2240, opened 2015-06-17
Prototypes circular dependency redmine7290

jimis:


jimis:
do not merge yet, I'm just testing it in pull request analyzer.

jimis:
trigger build

kacfengine:
+1! It's much better to avoid ``cf3.defs.h`` in libcfnet altogether if we can!

jimis:
And this is not over yet. In my private branch I tried a bunch of shuffling around in order to avoid the dependency to the enterprise extensions, unfortunately it got too convoluted and I reverted.

I plan to resolve this today, but I might need your help since you implemented the extension stuff!

jimis:
trigger build

jimis:
My 2nd reshuffling attempt also failed miserably... I'll merge this as is, if it passes, should be good enough to avoid the duplicated Makefile snippet.

PR #2241, opened 2015-06-18
Add: Test for ability to double dereference variables from variablesm…

kacfengine:
…atching

Ref: https://dev.cfengine.com/issues/7254


kacfengine:
trigger build

kacfengine:
Damn, the test crashes before it has time to reach the soft failure part. I'm afraid we can't merge this with the way the test framework currently works.

PR #2242, opened 2015-06-18
Redmine #7193: Fix SIGTERM waking up the loop instead of ending it.

kacfengine:
After all the sleeping we do, we need to make sure that we check if we
should terminate, otherwise the signal will simply spin the loop,
which triggers cf-agent, which spawns a new cf-execd, and we are back
where we started. At least by checking the flag, this probability is
drastically reduced.

Note that this does not fully fix the problem, since the signal
arrival is inherently a race condition, and this cannot be avoided. So
additional changes to the service scripts are still needed in order to
make sure that all daemons are stopped.

Also make sure that we do not enter yet another sleep after the first
has been interrupted. Again, this is a race condition, but it should
still improve things.

### DIFF
--- a/cf-execd/cf-execd.c
+++ b/cf-execd/cf-execd.c
@@ -313,6 +313,21 @@ void ThisAgentInit(void)
 
 /*****************************************************************************/
 
+// msg should include exactly one reference to unsigned int.
+static unsigned int MaybeSleepLog(LogLevel level, const char *msg, unsigned int seconds)
+{
+    if (IsPendingTermination())
+    {
+        return seconds;
+    }
+
+    Log(level, msg, seconds);
+
+    return sleep(seconds);
+}
+
+/*****************************************************************************/
+
 /* Might be called back from NovaWin_StartExecService */
 void StartServer(EvalContext *ctx, Policy *policy, GenericAgentConfig *config, ExecdConfig **execd_config, ExecConfig **exec_config)
 {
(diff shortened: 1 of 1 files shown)


kacfengine:
@jimis: Needs review.

kacfengine:
trigger build

jimis:
+1 for the initiative

You think it's related to https://dev.cfengine.com/issues/6659 ?

libpromises/cf3.defs.h
```diff
@@ -322,8 +322,6 @@ struct GidList_
 
 #define CF_DONEPASSES  4
 
-#define CFPULSETIME 60
```
jimis: Eddy has used CFPULSETIME in nova.

kacfengine:
Yes, that's exactly it. I should add that ticket to the commit as well.

libpromises/cf3.defs.h
kacfengine: Ah, good catch, I thought I had searched through everything, but I missed nova.

jimis:
+1

kacfengine:
trigger build

cf-execd/cf-execd.c
```diff
@@ -371,8 +386,14 @@ void StartServer(EvalContext *ctx, Policy *policy, GenericAgentConfig *config, E
...
+                MaybeSleepLog(LOG_LEVEL_VERBOSE, "Sleeping for splaytime %u seconds", (*execd_config)->splay_time);
+
+                // We are sleeping both above and inside ScheduleRun(), so make
+                // sure a terminating signal did not arrive during that time.
+                if (IsPendingTermination())
+                {
+                    break;
+                }
 
                 if (!LocalExecInThread(*exec_config))
```
jimis: Wow, cf-execd is multithreaded! That was unexpected...

kacfengine:
This was cherry-picked to both 3.6.x and 3.7.x.

PR #2244, opened 2015-06-18
Redmine#7244: Do not allow to kill self.

jimis:


libpromises/process_lib.h
```diff
@@ -27,7 +27,12 @@
 
 #include <platform.h>
 
-#define PROCESS_START_TIME_UNKNOWN ((time_t)0)
+
+/* TODO move to libutils, once windows implementation is merged. */
+
```
jimis: @kacfengine this is an example of what I was describing. One function implemented in libcfenterprise (`GracefulTerminate()`) keeps all these files (`process_*.[ch]`) in libpromises, instead of libutils where they belong.

jimis:
trigger build

libpromises/process_lib.h
kacfengine: Right, I see what you mean.

kacfengine:
Cherry-picked to 3.7.x (the enterprise one too).

PR #2245, opened 2015-06-18
Parse /proc/net/route and allow dumping all JSON vars

tzz:
See https://dev.cfengine.com/issues/7295

There are several commits here:

* builds on top of #2477
* add IPv4 parsing in the big-endian format `0A0B0C0D` used in `/proc/net/route`
* add IPv6 parsing in the big-endian format (32 hex digits) used in `/proc/net/tcp6` etc.
* add port parsing and output
* parse `/proc/net/route` and build a JSON object called `sys.proc_routes` with all the route flags and info.  Note this is not a "classic" array but will behave like it for things like `getindices` so it's nice.  Also create `ipv4_gw_a_b_c_d` hard classes for every gateway thus found.
* parse `/proc/net/ipv6_routes` and build a JSON object called `sys.proc_ipv6_routes` with all the IPv6 route flags and info.  Note this is not a "classic" array but will behave like it for things like `getindices` so it's nice.
* parse `/proc/net/if_inet6` and build a JSON object called `sys.ipv6_addresses` with all the IPv6 interfaces and their addresses.  Note this is not a "classic" array but will behave like it for things like `getindices` so it's nice.
* parse `/proc/net/netstat` and build a JSON object called `sys.proc_stats` with all the netstats.  Note this is not a "classic" array but will behave like it for things like `getindices` so it's nice.
* parse `/proc/net/{tcp,tcp6,udp,udp6}` and build a JSON object called `sys.proc_ports_NAME` with all the port connection info.  Note this is not a "classic" array but will behave like it for things like `getindices` so it's nice.
* parse `/proc/net/dev` and build a JSON object called `sys.proc_dev` with all the device stats, **KEYED BY DEVICE NAME**.  Note this is not a "classic" array but will behave like it for things like `getindices` so it's nice.
* set `sys.default_gateway` from routes with flags up + gateway and destination `0.0.0.0`
* everything is refactored several times to use a few general parsing functions and I extract to JSON integers and IP strings whenever possible

TODOs:

* Linux-only.  I don't think it will be too different on other platforms, though.
* I assume the flags are in hex format, I can't find an example though.  Lots of TODOs that need to be confirmed.
* I assume `/proc/net/netstat` always follows the headers of a section with the section data
* the code definitely needs cleanup
* verify list of port states in `GetPortStateString`
* JSON integers overflow when I parse things like number of octets `72886446156` so I left many counters as strings

Sample output:

```console
% sudo /var/cfengine/bin/cf-promises --show-vars-all|grep proc_route

default:sys.proc_routes                  [{"dest":"0.0.0.0","flags":["up","net","gateway"],"gateway":"192.168.2.1","interface":"eth0","mask":"0.0.0.0"},{"dest":"169.254.0.0","flags":["up","net","local"],"gateway":"0.0.0.0","interface":"eth0","mask":"255.255.0.0"},{"dest":"172.17.0.0","flags":["up","net","local"],"gateway":"0.0.0.0","interface":"docker0","mask":"255.255.0.0"},{"dest":"192.168.2.0","flags":["up","net","local"],"gateway":"0.0.0.0","interface":"eth0","mask":"255.255.255.0"}] source=agent

% sudo /var/cfengine/bin/cf-promises --show-vars-all|grep proc_stats

default:sys.proc_stats                   {"IpExt":{"InBcastOctets":"29635203","InBcastPkts":"529095","InCEPkts":"6710","InCsumErrors":"0","InECT0Pkts":"6017","InECT1Pkts":"67","InMcastOctets":"76888189","InMcastPkts":"239199","InNoECTPkts":"67713439","InNoRoutes":"0","InOctets":"67651131214","InTruncatedPkts":"0","OutBcastOctets":"0","OutBcastPkts":"0","OutMcastOctets":"692006","OutMcastPkts":"5579","OutOctets":"54165901838"},"TcpExt":{"ArpFilter":"0","BusyPollRxPackets":"0","DelayedACKLocked":"224","DelayedACKLost":"20947","DelayedACKs":"514102","EmbryonicRsts":"2575","IPReversePathFilter":"0","ListenDrops":"89","ListenOverflows":"0","LockDroppedIcmps":"0","OfoPruned":"0","OutOfWindowIcmps":"31","PAWSActive":"0","PAWSEstab":"404","PAWSPassive":"0","PruneCalled":"0","RcvPruned":"0","SyncookiesFailed":"0","SyncookiesRecv":"0","SyncookiesSent":"0","TCPAbortFailed":"0","TCPAbortOnClose":"8436","TCPAbortOnData":"58955","TCPAbortOnLinger":"0","TCPAbortOnMemory":"0","TCPAbortOnTimeout":"1168","TCPAutoCorking":"75223","TCPBacklogDrop":"0","TCPChallengeACK":"1766","TCPDSACKIgnoredNoUndo":"6106","TCPDSACKIgnoredOld":"168","TCPDSACKOfoRecv":"82","TCPDSACKOfoSent":"540","TCPDSACKOldSent":"24164","TCPDSACKRecv":"13273","TCPDSACKUndo":"329","TCPDeferAcceptDrop":"0","TCPDirectCopyFromBacklog":"2473697","TCPDirectCopyFromPrequeue":"108895753","TCPFACKReorder":"9","TCPFastOpenActive":"0","TCPFastOpenActiveFail":"0","TCPFastOpenCookieReqd":"0","TCPFastOpenListenOverflow":"0","TCPFastOpenPassive":"0","TCPFastOpenPassiveFail":"0","TCPFastRetrans":"44675","TCPForwardRetrans":"2634","TCPFromZeroWindowAdv":"5","TCPFullUndo":"654","TCPHPAcks":"2509398","TCPHPHits":"13927595","TCPHPHitsToUser":"27396","TCPHystartDelayCwnd":"11019","TCPHystartDelayDetect":"254","TCPHystartTrainCwnd":"695","TCPHystartTrainDetect":"31","TCPLossFailures":"4094","TCPLossProbeRecovery":"5192","TCPLossProbes":"20383","TCPLossUndo":"6593","TCPLostRetransmit":"2675","TCPMD5NotFound":"0","TCPMD5Unexpected":"0","TCPMemoryPressures":"0","TCPMinTTLDrop":"0","TCPOFODrop":"0","TCPOFOMerge":"417","TCPOFOQueue":"601830","TCPOrigDataSent":"39020959","TCPPartialUndo":"122","TCPPrequeueDropped":"0","TCPPrequeued":"1544329","TCPPureAcks":"7046974","TCPRcvCoalesce":"3759616","TCPRcvCollapsed":"0","TCPRenoFailures":"16","TCPRenoRecovery":"10","TCPRenoRecoveryFail":"1","TCPRenoReorder":"1","TCPReqQFullDoCookies":"0","TCPReqQFullDrop":"0","TCPRetransFail":"0","TCPSACKDiscard":"0","TCPSACKReneging":"1","TCPSACKReorder":"37","TCPSYNChallenge":"1253","TCPSackFailures":"5080","TCPSackMerged":"88803","TCPSackRecovery":"20133","TCPSackRecoveryFail":"3073","TCPSackShiftFallback":"57948","TCPSackShifted":"37892","TCPSchedulerFailed":"2","TCPSlowStartRetrans":"16433","TCPSpuriousRTOs":"4722","TCPSpuriousRtxHostQueues":"14","TCPSynRetrans":"259394","TCPTSReorder":"165","TCPTimeWaitOverflow":"0","TCPTimeouts":"85582","TCPToZeroWindowAdv":"5","TCPWantZeroWindowAdv":"20","TW":"386769","TWKilled":"0","TWRecycled":"0"}} source=agent              

% sudo /var/cfengine/bin/cf-promises --show-vars-all|grep proc_ports

default:sys.proc_ports_tcp               [{"local":"0.0.0.0:48367","remote":"0.0.0.0:0","state":"UNKNOWN"},{"local":"192.168.2.48:59463","remote":"12.0.2.8:443","state":"ESTABLISHED"},...] source=agent                            

...etc...

default:sys.proc_ports_tcp6              [{"local":"0:0:0:0:0:0:0:0:22","remote":"0:0:0:0:0:0:0:0:0","state":"UNKNOWN"},...] source=agent                            

default:sys.proc_ports_udp               [{"local":"0.0.0.0:1900","remote":"0.0.0.0:0","state":"CLOSE"},...] source=agent                            

default:sys.proc_ports_udp6              [{"local":"0:0:0:0:0:0:0:0:43899","remote":"0:0:0:0:0:0:0:0:0","state":"CLOSE"},{"local":"0:0:0:0:0:0:0:0:5353","remote":"0:0:0:0:0:0:0:0:0","state":"CLOSE"},{"local":"0:0:0:0:0:0:0:0:47735","remote":"0:0:0:0:0:0:0:0:0","state":"CLOSE"}] source=agent                          

default:sys.proc_dev                     {
"docker0":{"device":"docker0","receive_bytes":"0","receive_compressed":"0","receive_drop":"0","receive_errors":"0","receive_fifo":"0","receive_frame":"0","receive_multicast":"0","receive_packets":"0","transmit_bytes":"0","transmit_compressed":"0","transmit_drop":"0","transmit_errors":"0","transmit_fifo":"0","transmit_frame":"0","transmit_multicast":"0","transmit_packets":"0"},
"eth0":{"device":"eth0","receive_bytes":"74859697272","receive_compressed":"0","receive_drop":"0","receive_errors":"0","receive_fifo":"0","receive_frame":"0","receive_multicast":"1113481","receive_packets":"75288382","transmit_bytes":"63116604703","transmit_compressed":"0","transmit_drop":"0","transmit_errors":"0","transmit_fifo":"0","transmit_frame":"0","transmit_multicast":"0","transmit_packets":"74735004"},
"lo":{"device":"lo","receive_bytes":"13597339","receive_compressed":"0","receive_drop":"0","receive_errors":"0","receive_fifo":"0","receive_frame":"0","receive_multicast":"0","receive_packets":"87933","transmit_bytes":"13597339","transmit_compressed":"0","transmit_drop":"0","transmit_errors":"0","transmit_fifo":"0","transmit_frame":"0","transmit_multicast":"0","transmit_packets":"87933"}} inventory,networking,source=agent       

default:sys.proc_ipv6_routes             [{"dest":"0:0:0:0:0:0:0:0","dest_prefix":"40","flags":["up","net","local"],"interface":"eth0","metric":256,"next_hop":"0:0:0:0:0:0:0:0","refcnt":0,"source_prefix":"00","use":0},{"dest":"0:0:0:0:0:0:0:0","dest_prefix":"80","flags":["up","net","local"],"interface":"lo","metric":0,"next_hop":"0:0:0:0:0:0:0:0","refcnt":1,"source_prefix":"00","use":0},{"dest":"0:0:0:0:0:0:0:0","dest_prefix":"08","flags":["up","net","local"],"interface":"eth0","metric":256,"next_hop":"0:0:0:0:0:0:0:0","refcnt":0,"source_prefix":"00","use":0}] inventory,networking,source=agent       

 ```

### CODE REVIEW

**Comment by tzz** [comment]
I added a commit that simplifies parsing significantly for some cases by delegating to PCRE regular expression extraction.

**Comment by tzz** [comment]
I added a commit to collect open ports, see PR summary...

**Comment by tzz** [comment]
The latest commit also fixes some bugs in earlier commits.  Sorry for the accumulation, I can squash if needed.

**Comment by jimis** [comment]
Thanks Ted. It might take some time to review, since it's a feature aimed for master (i.e. 3.8) and not 3.7.

**Comment by tzz** [comment]
@jimis no problem.  I'm sure I'll need to change the code a bit.  I also think some of the new functions should go elsewhere in the core, they are generally useful.

Now that I have mostly addressed the original feature request in https://dev.cfengine.com/issues/7295 I have two directions here:

* keep adding parsers for `/proc` paths (it's really easy now, with PCRE)
* provide alternatives when `/proc` is not available.  For example, we can use `netstat -rn` to approximate the route information.

Please let me know if those directions are OK or if you want to cut off the flow until you have a chance to review things so far.

**Comment by tzz** [comment]
...added `/proc/net/dev` and `/proc/net/ipv6_route` parsing, plus lots of refactoring

**Comment by tzz** [comment]
I can't think of any other `/proc/net` parsers I would find useful, suggestions are welcome.  Also verifications of the code and the conversions (especially where I parse things as hex or decimal) are very welcome.

So next is some code review, more refactoring and moving or renaming functions, and maybe providing fallbacks for the common cases where `/proc` is not available but the data may be available elsewhere.  I'll wait for @jimis for that.

**Comment by tzz** [comment]
ping?

**Comment by estenberg** [comment]
The variables need to avoid "proc" from their name, which is implementation-specific.
Also, we should review other sys variables for duplication..

**Comment by tzz** [comment]
Which one of these should I use?

* `proc.routes` (new special context)
* `sys.proc` (data container with all the parsed data inside)
* `net.routes` (new special context)
* `sys.network` (data container)
* something else?

Also I was wondering if it would be useful to check `enable_feature_proc` (default `any`) and `disable_feature_proc` (default `!any`) classes, so users can turn this off if they don't want it.

**Comment by tzz** [comment]
Note also that I plan to add process inventory here, so whatever naming convention we use, it should fit a process inventory as well.

**Comment by estenberg** [comment]
What about sys.ports_XYZ and sys.routes?
I don't think we should nest everything into one container as it will get very deep, large and hard to use.

We should not inventory volatile data (won't be useful for configuration nor reporting), so not sure if we need proc_stats?

Why not simply add JSON support to --show-vars in stead of increasing complexity with yet-another option? Is there some kind of technical limitation?

**Comment by tzz** [comment]
OK, removed `--show-vars-all` and now `show-vars` shows structured data as well.

**Comment by tzz** [comment]
I'd like to keep `proc_stats` but maybe move it to a new context for volatile statistics data.  How about `mon.network_stats`?

**Comment by tzz** [comment]
OK; removed `proc_` prefix from all the new variables except `proc_stats` pending that decision.  Also added better tags to all the variables and patched `JsonObjectGetAsString` to work on non-string primitives (so integers and reals etc. can be serialized).  Ready for another review.

**Comment by worr** [comment]
Would it be worth trying to do some abstraction around network information in libutils or something? While the easiest way to get network information in Linux is through `/proc`, the same interface doesn't apply to any other platforms.

In Solaris, it will likely still be through `/proc`, but it's far easier than Linux since we can just cast the data read to the appropriate struct (Solaris' `/proc` are binary dumps of structs).

In FreeBSD, there is no `/proc` unless it's specifically enabled, and compatibility with Linux there is fuzzy at best. `sysctl(3)` is likely the best way to get network info. Same with OpenBSD and NetBSD, but they don't have `/proc` at all, and will only support `sysctl(3)`

Idk about AIX or HP-UX.

Wrapping these is probably the easiest way to handle this while avoiding `ifdef`s in the rest of the code.

Just my 2c.

**Comment by tzz** [comment]
I'm Linux-only, so the work for other platforms would have to be done by other contributors.  Otherwise it seems like a good idea.

I see a split, though: there's this PR, which is a direct parsing of `/proc/net`.  And you're proposing a more abstract view of networking, which would not have all the detail.  So maybe it's a followup PR rather than something we have to do in this one?

**Comment by worr** [comment]
Yeah, that's exactly what I'm proposing.

The reason I mention it in this PR, is that whomever (I'd be willing to, once I'm no longer lawyer-blocked) picks up the multiplatform work here, will stick these into the same ifdefs that are littered throughout the PR. Would it not make sense to pump this either into libutils or (keep it) in libenv, but put the platform-specific implementations into their own files? We can conditionally compile these files and keep the implementation separate.

That is to say, we can take functions like `GetNetworkingRouteInfo`, and pump them into a file like `netinfo_linux.c` and conditionally compile that in.

That keeps the implementation separate, cleans up the `ifdef`s, and provides an API defined in the header for these that other implementors can implement. This also keeps platform-specific helper functions like `GetProcInfo` separate from the rest of the implementation.

It seems to make sense to do it in this PR, since you're introducing the API here. Does this make sense? Mike asked me to take a look at the PR, so I'm just offering my 2c.

**Comment by tzz** [comment]
I'm happy to support you, @worr, but I don't want to write or maintain the code for all the platforms you mentioned.  You seem to have the right ideas, so why not take my code and build on top of it as needed, or rewrite it?  I'll be more than happy to help out and close this PR.

The only things I would suggest are:

* maybe keep the parsing at the text level instead of talking directly to the kernel structs?  That's a judgment call but to me it seems easier to support incompatible structs in the long run by adjusting regexes
* maybe we don't have to support dead platforms in this PR.  If users need it, let them contribute.  But if the code uses regexes (as I suggest), it will be easier for non-expect contributors to provide the right changes.
* keep the detailed platform data avaialble instead of hiding it under a compatibility layer (I think that's your plan already)
* `libenv` is the right place for environment discovery, not `libuitils`

**Comment by worr** [comment]
I'll be happy to submit Solaris and BSD variants, once the CLA stuff is cleared up.

If this gets committed, I'll be happy to refactor/build around this. The code itself lgtm.

**Comment by jimis** [comment]
I can see that some of the first commits that I reviewed are actually overwritten by some of your later commits. So it would be better if you squashed all the relevant commits together, before review.

One general notice is that the pull request does not include tests.

**Comment by tzz** [comment]
I've rebased and squashed to a single commit, including the `header_line_size` fix.

I would appreciate either direct pointers to what needs changing for IP parsing or maybe @jimis can make a followup fix commit to adjust things as needed.

**Comment by nickanderson** [comment]
Ping @jimis

**Comment by jimis** [comment]
Un-assigning from me because of lack of time. Sorry, but any other reviewer feel free to review ...

File: cf-promises/cf-promises.c [hunk 1]
```diff
@@ -452,15 +452,25 @@ static void ShowVariablesFormatted(EvalContext *ctx)
...
-        if(StringIsPrintable(StringWriterData(w)))
+        const char *var_value;
+        if (json)
+        {
+            var_value = StringWriterData(w);
+        }
+        else if (StringIsPrintable(StringWriterData(w)))
         {
             var_value = StringWriterData(w);
         }
```
kacf: Same action for both `if` branches, I would use an `||` for this test.

But I'm questioning whether we should short-circuit with `json` at all. It seems reasonable to me that the `StringIsPrintable` test should be applied to the JSON string as well, it's a string just like everything else. What do you think?

cf-promises/cf-promises.c
```diff
@@ -469,6 +479,11 @@ static void ShowVariablesFormatted(EvalContext *ctx)
             var_value = "<non-printable>";
         }
 
+
+        StringSet *tagset = EvalContextVariableTags(ctx, v->ref);
+        Buffer *tagbuf = StringSetToBuffer(tagset, ',');
+
+        char *line;
```
kacf: I think the diff for this whole file should be moved out and into a separate pull request. It's not really related to the network discovery stuff, and it would be small and quick to get in separately.

libenv/unix_iface.c
```diff
@@ -752,4 +755,370 @@ static int aix_get_mac_addr(const char *device_name, uint8_t mac[6])
 }
 #endif /* _AIX */
 
+/*******************************************************************/
+
+// returns NULL on any failure
+JsonElement* StringCaptureData(const char* regex, const char* data)
```
kacf: Should be `static`. This goes for all functions that are not exported via the `unix_iface.h` file.

libenv/unix_iface.c
```diff
@@ -752,4 +755,370 @@ static int aix_get_mac_addr(const char *device_name, uint8_t mac[6])
...
+        SeqDestroy(s);
+        return NULL;
+    }
+
+    JsonElement *json = JsonObjectCreate(SeqLength(s)/2);
+
+    for (int i = 0; i < SeqLength(s); i+=2)
+    {
+        Buffer *key = SeqAt(s, i);
+        Buffer *value = SeqAt(s, i+1);
```
kacf: For safety I would rewrite this loop to:

```
for (int i = 1; i < SeqLength(s); i+=2)
{
    Buffer *key = SeqAt(s, i-1);
    Buffer *value = SeqAt(s, i);
```

kacf: In case you get an odd number of elements. Probably you won't, but better safe than sorry.

kacf:
This is a big pull request with lots of details! I got about halfway through it, will continue tomorrow.

cf-promises/cf-promises.c
tzz: This was legacy from before, where we had a switch to print JSON vars and normally we didn't. Fixed in separate PR.

cf-promises/cf-promises.c
tzz: OK, see #2477

tzz:
OK, this has been rebased on top of #2477 now.

tzz:
I hate C typedefs... if you know my mistake here, let me know please:

```
unix_iface.c:1112:64: warning: passing argument 5 of 'GetProcFileInfo' from incompatible pointer type [-Wincompatible-pointer-types]
     GetProcFileInfo(ctx, "/proc/net/udp6", "ports_udp6", NULL, &NetworkingPortsPostProcessInfo, ports_regex);
                                                                ^
unix_iface.c:1032:13: note: expected 'ProcPostProcessFn {aka void (*)(void *, void *)}' but argument is of type 'void (* (*)(void *, void *))(void *, void *)'
 static void GetProcFileInfo(EvalContext *ctx, const char* filename, const char* key, const char* extracted_key, ProcPostProcessFn post, const char* regex)
```

libenv/unix_iface.c
tzz: fixed

libenv/unix_iface.c
tzz: brilliant, yeah, fixed

tzz:
I'll definitely add acceptance tests (as much as possible) once this is acceptable on the C level.

libenv/unix_iface.c
```diff
@@ -752,4 +755,370 @@ static int aix_get_mac_addr(const char *device_name, uint8_t mac[6])
...
+
+    // Inter-|   Receive                                                |  Transmit
+    //  face |bytes    packets errs drop fifo frame compressed multicast|bytes    packets errs drop fifo colls carrier compressed
+    //   eth0: 74850544807 75236137    0    0    0     0          0   1108775 63111535625 74696758    0    0    0     0       0          0
+
+    GetProcFileInfo(ctx, "/proc/net/dev", "dev", "device", NULL,
+                    "^\\s*(?<device>[^:]+)\\s*:\\s*"
+                    "(?<receive_bytes>\\d+)\\s+(?<receive_packets>\\d+)\\s+(?<receive_errors>\\d+)\\s+(?<receive_drop>\\d+)\\s+(?<receive_fifo>\\d+)\\s+(?<receive_frame>\\d+)\\s+(?<receive_compressed>\\d+)\\s+(?<receive_multicast>\\d+)\\s+"
+                    "(?<transmit_bytes>\\d+)\\s+(?<transmit_packets>\\d+)\\s+(?<transmit_errors>\\d+)\\s+(?<transmit_drop>\\d+)\\s+(?<transmit_fifo>\\d+)\\s+(?<transmit_frame>\\d+)\\s+(?<transmit_compressed>\\d+)\\s+(?<transmit_multicast>\\d+)");
+}
```
kacf: +1, nice way of doing it with `GetProcFileInfo()`.

I'd break up the lines a little bit to make it easier to read, like you already did for `ipv6_routes`.

I won't even attempt to decipher those regexes. :-)

libutils/ip_address.c
```diff
@@ -105,6 +105,32 @@ static int IPV4_parser(const char *source, struct IPV4Address *address)
...
+        {
+            address->octets[3] = a;
+            address->octets[2] = b;
+            address->octets[1] = c;
+            address->octets[0] = d;
+            address->port = pport;
+            return 0;
+        }
+    }
+
```
kacf: Both of these functions (`IPV4_parser` and `IPV6_parser`) state their expected input, which is the text format. I think we should not magically accept hex format behind people's backs. Since you already know that the string is a hex string of a certain length, there is no way it can be an invalid address if it passes the `sscanf`, so I'd just make a new function, `IPAddressNewFromHex()` or similar to perform this parsing step.

kacf:
Done reviewing! I think other than the comment about IP address parsing, this is a good piece of code, so if you address that (pun intended :-P), you can immediately proceed with tests and documentation. Since this is OS probing, testing will probably be somewhat limited, but catching some very common cases would at least be nice.

estenberg:
Which performance implications does this have?
I'm a bit concerned since this is *a lot* of data, and it seems be be gathered on *every* run...

Did we do any test of changes to cf-agent runtime or memory usage? Can it be turned off in resource-constrained environments?

kacf:
I'm not that concerned about performance. This only happens once per run, as opposed to our JSON iteration code for example, which can easily reach multiple millions.

But I can do a timed acceptance test run comparison to be sure.

kacf:
Btw, I get quite a lot of warnings in this file now. Would be nice to get rid of those:

```
unix_iface.c: In function 'NetworkingRoutesPostProcessInfo':
unix_iface.c:892:1: warning: no return statement in function returning non-void [-Wreturn-type]
unix_iface.c: In function 'NetworkingIPv6RoutesPostProcessInfo':
unix_iface.c:896:18: warning: unused variable 'ctx' [-Wunused-variable]
unix_iface.c:924:1: warning: no return statement in function returning non-void [-Wreturn-type]
unix_iface.c: In function 'NetworkingPortsPostProcessInfo':
unix_iface.c:952:18: warning: unused variable 'ctx' [-Wunused-variable]
unix_iface.c:968:1: warning: no return statement in function returning non-void [-Wreturn-type]
unix_iface.c: In function 'GetNetworkingInfo':
unix_iface.c:1096:21: warning: passing argument 5 of 'GetProcFileInfo' from incompatible pointer type [enabled by default]
unix_iface.c:1032:13: note: expected 'ProcPostProcessFn' but argument is of type 'void (* (*)(void *, void *))(void *, void *)'
unix_iface.c:1105:21: warning: passing argument 5 of 'GetProcFileInfo' from incompatible pointer type [enabled by default]
unix_iface.c:1032:13: note: expected 'ProcPostProcessFn' but argument is of type 'void (* (*)(void *, void *))(void *, void *)'
unix_iface.c:1109:5: warning: passing argument 5 of 'GetProcFileInfo' from incompatible pointer type [enabled by default]
unix_iface.c:1032:13: note: expected 'ProcPostProcessFn' but argument is of type 'void (* (*)(void *, void *))(void *, void *)'
unix_iface.c:1110:5: warning: passing argument 5 of 'GetProcFileInfo' from incompatible pointer type [enabled by default]
unix_iface.c:1032:13: note: expected 'ProcPostProcessFn' but argument is of type 'void (* (*)(void *, void *))(void *, void *)'
unix_iface.c:1111:5: warning: passing argument 5 of 'GetProcFileInfo' from incompatible pointer type [enabled by default]
unix_iface.c:1032:13: note: expected 'ProcPostProcessFn' but argument is of type 'void (* (*)(void *, void *))(void *, void *)'
unix_iface.c:1112:5: warning: passing argument 5 of 'GetProcFileInfo' from incompatible pointer type [enabled by default]
unix_iface.c:1032:13: note: expected 'ProcPostProcessFn' but argument is of type 'void (* (*)(void *, void *))(void *, void *)'
```

kacf:
Well, bonus points to @estenberg: There is a noticeable performance decrease with this change. In my testing the acceptance test suite went from 504 seconds to 533 seconds on average. But the question is how significant this is in practice, we are talking about more than 1300 tests! Which means we are paying that cost 1300 times. `(533 - 504) / 1300 = 0.022` per run. I think we can live with 22ms extra startup time, considering all the extra information about the system you get. Agree, @estenberg?

libenv/unix_iface.c
tzz: OK, fixed. The regex is much clearer that way actually, but I added a comment as well.

libutils/ip_address.c
tzz: I think the value here (which is why I put it in the same function) is that now CFEngine can **generally** accept that format for parsing IP addresses. If you still don't think that's a good thing, I'll move it; otherwise I can change the function's doc string.

tzz:
@estenberg the performance decrease is across the acceptance tests, which are very small pieces of policy, so the increase in runtime is more noticeable. In practice I believe this will be unnoticeable, because most users will have much bigger policies. As @kacf said you can expect 22ms more, which is very reasonable, and the performance will scale linearly with the amount of data. We could optimize further, perhaps, but I think it would be wasted effort compared to optimizing variable passing or other parts of the code.

If you disagree, perhaps we can add a `def.json` setting to turn this off, but I hope it's not necessary.

tzz:
@kacf warnings fixed, I believe in a correct way.

tzz:
@estenberg also note that the code does nothing if there is no `/proc` filesystem.

estenberg:
@tzz @kacf It's a 6% performance decrease with a small policy though, so I wouldn't say it's completely insignificant. But I agree the feature has sufficient value to warrant this.

Can this affect memory use significantly?

libenv/unix_iface.c
```diff
@@ -752,4 +755,389 @@ static int aix_get_mac_addr(const char *device_name, uint8_t mac[6])
...
+    JsonArrayAppendString(decoded_flags, is_up ? "up":"down");
+    JsonArrayAppendString(decoded_flags, is_host ? "host":"net");
+    JsonArrayAppendString(decoded_flags, gw_type);
+    JsonObjectAppendElement(route, "flags", decoded_flags);
+
+    if (is_up && is_gw)
+    {
+        Buffer *formatter = BufferNew();
+        BufferPrintf(formatter, "ipv4_gw_%s", JsonObjectGetAsString(route, "gateway"));
+        EvalContextClassPutHard(ctx, BufferData(formatter), "inventory,networking,/proc,source=agent,procfs");
```
estenberg: add tag "attribute_name=none" to exclude this from reporting by default.

libenv/unix_iface.c
```diff
@@ -752,4 +755,389 @@ static int aix_get_mac_addr(const char *device_name, uint8_t mac[6])
...
+        BufferDestroy(formatter);
+
+        if (is_default_route)
+        {
+            EvalContextVariablePutSpecial(ctx,
+                                          SPECIAL_SCOPE_SYS, "default_gateway",
+                                          JsonObjectGetAsString(route, "gateway"), CF_DATA_TYPE_STRING,
+                                          "inventory,networking,/proc,source=agent,procfs");
```
estenberg: Useful for reporting, would add tag "attribute_name=Default gateway".

libenv/unix_iface.c
```diff
@@ -752,4 +755,389 @@ static int aix_get_mac_addr(const char *device_name, uint8_t mac[6])
...
+                RlistDestroy(info);
+                BufferDestroy(type);
+            }
+
+        }
+
+        free(header_line);
+
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, "proc_stats", stats, CF_DATA_TYPE_CONTAINER,
+                                      "inventory,networking,/proc,source=agent,procfs");
```
estenberg: attribute_name=none

libenv/unix_iface.c
```diff
@@ -752,4 +755,389 @@ static int aix_get_mac_addr(const char *device_name, uint8_t mac[6])
...
+                }
+            }
+        }
+
+        free(line);
+
+        Buffer *varname = BufferNew();
+        BufferPrintf(varname, "%s", key);
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, BufferData(varname), info, CF_DATA_TYPE_CONTAINER,
+                                      "inventory,networking,/proc,source=agent,procfs");
```
estenberg: attribute_name=none

tzz:
@estenberg all the `attribute_name` changes are done, thanks for the suggestions.

The memory usage will be tiny, except on machines with tens of thousands of connections, which is not a typical situation.

libutils/ip_address.c
kacf: If it were a common way to specify an IP address, I would agree, but I have never seen this format anywhere else. It's also much more likely to be mistaken for something it isn't, since it is just a number, no formatting.

kacf:
I will start the build here as soon as #2477 is done.

estenberg:
OK. What about the runtime on machines with tens of thousands of connections? The machine where the acceptance tests are run presumably have nearly no connections.

I think we should consider an option for turning this off, I do think there is a high chance someone will get into performance issues with this change on some highly-loaded server.

What about a body common control option that would turn off some network discovery? We have *ignore_interfaces.rx* file that is already a bit hacky but I guess the reason is that it needs to be discovered before policy is loaded...?

libutils/ip_address.c
tzz: It's used in `/proc` :) I honestly couldn't see how this could be a problem in practice, but I've pulled it out to a separate function as you requested.

tzz:
@estenberg I estimate tens of thousands of connections will take a few hundred KB of JSON storage. I can add a `def.json` option to turn off network discovery, and can add acceptance tests to make sure the "machine with 100K connections" case is handled well. I don't expect it to be a significant problem. Keep in mind that `cf-promises` does network discovery before parsing the policy, so controlling network discovery wouldn't work well as a `body common control` option. Let me know what's acceptable to proceed.

(As an aside, migrating `ignore_interfaces.rx` to `def.json` would make a lot of sense and require little code... I could probably put it in this same PR if you want to control all network discovery in a consistent way.)

kacf:
For testing the performance I suggest using a unit test instead, and calling `GetProcFileInfo()` directly, with a carefully prepared file. Then it should be simple to prepare a big one and a small one and compare the time differences. It's probably easier than using an acceptance test since you won't be able to override the filename easily there. And you also get rid of a lot of unrelated factors that can influence the timing.

Let's see what the results are for a machine with 100K fake connections before we discuss alternatives for turning it off.

tzz:
I've rebased this on top of #2477 so when that's merged, this will merge cleanly.

I've added the acceptance test with a bit over 6500 TCP connections but the test data became pretty big at that point so I didn't grow it further. I implemented `CFENGINE_TEST_OVERRIDE_PROCDIR` in order to do this. If you set it to `/foo` then the code will look in `/foo/proc`.

The acceptance test is easy to extend and modify using the `expected.json` structure.

The acceptance test will let us create more detailed timing tests, but with 200K TCP connections the **network discovery** finished in 4 seconds on my machine (the acceptance test took 4 times longer because it was copying containers and doing JSON diffs). I doubt that will be anywhere near real-world usage. I think the acceptance test is relevant because the most typical usage will copy these data structures, and we currently copy them even if it's not necessary. See https://github.com/cfengine/core/pull/2476 for code that will copy JSON data containers only when needed, which should speed things up.

A unit test as @kacf suggested should be fairly easy, since `CFENGINE_TEST_OVERRIDE_PROCDIR` is now available. Please feel free to add it after this is merged; I hope this PR is done otherwise.

estenberg:
@tzz thanks for looking into the performance.

So a total of 4 or 16 seconds added run-time overhead on a machine with 200k connections? It's a bit hard to tell from the description what the impact on a real-world scenario would be. Also, did you check memory usage?

What is the added run-time to a more common amount of connections, like for a busy web server? I suppose it would have about 30k connections.

Sorry for pestering about this, but if we add a feature that significantly slows down an already busy server and there is no way to turn it off we would be in a bad place. So it is important to be sure we don't end up in that position. :)

tzz:
I think the CFEngine developers should measure memory and performance overhead, I've done all I can do with my one development machine. Also I think at this point you're worrying about an undefined performance target. Set specific performance targets if you need the code to meet them, based on what your customers want.

As a tuning aid, I'm happy to do one of:

* add a `control->disable_network_discovery` setting to `def.json` to disable `/proc` parsing altogether (internally it would be a class, defaulting to undefined). It can't live in policy because network discovery happens before policy parsing.
* add a `control->max_{tcp,tcp6,udp,udp6}_ports` setting to `def.json` to limit the number of ports parsed and default it to 10K (internally it would be an integer). This is probably closest to what you're looking for. It can't live in policy because network discovery happens before policy parsing.

kacf:
I think the `control->disable_network_discovery` sounds good, do you agree @estenberg?

estenberg:
I am a bit reluctant to creating yet another special case in our parsing, this time it's a control attribute that only exists in def.json? With all these special cases and exceptions it will be difficult to explain how it works and maintain it.

Can we do it like the other control attributes, so that it is added to body common control somewhere and it references an attribute in def.json? Or is this a completely different problem than the one we had for the other things in def.json (e.g. mail settings)?

kacf:
I think it's too late by the time `body common control` is parsed, the proc parsing is done before that. @tzz, how difficult would it be to do this "on demand", e.g. parsing proc only when the variable is first referenced (and caching it, obviously)?

tzz:
OK, so first of all, remember we are discussing the **exception**. The default behavior will be to parse `/proc` in a few milliseconds and move on, so the vast majority of users won't care about this.

@estenberg `body common control` is parsed and evaluated much later, we'd have to move the whole block of code to a new place, separate from the other network discovery. I think it would be inconsistent. It would also be unreliable, for the same reasons that we had to parse `def.json` in C: the evaluator is not deterministic.

This is also the problem with the `ignored_interfaces.rx` special file, which is one of the very few other places where users can control the network discovery process. I propose that we move that file into `def.json` as well and establish a simple, sensible pattern of putting special controls that govern CFEngine's startup (note this applies to all agents, not just `cf-agent`) in a single, logically consistent place. That would be a good opportunity to let the user tune network discovery in general.

Doing network discovery on demand is an interesting approach, but I fear it could break in many cases. Again, considering this is the exception, I'd rather not add risky code to the evaluator, but just keep things simple and put an override in `def.json`.

@worr you mentioned you'd probably try to refactor the network discovery after this is merged. Can you comment on this discussion and how it may relate to your needs specifically?

tzz:
OK, so first of all, remember we are discussing the **exception**. The default behavior will be to parse `/proc` in a few milliseconds and move on, so the vast majority of users won't care about this.

@estenberg `body common control` is parsed and evaluated much later, we'd have to move the whole block of code to a new place, separate from the other network discovery. I think it would be inconsistent. It would also be unreliable, for the same reasons that we had to parse `def.json` in C: the evaluator is not deterministic.

This is also the problem with the `ignored_interfaces.rx` special file, which is one of the very few other places where users can control the network discovery process. I propose that we move that file into `def.json` as well and establish a simple, sensible pattern of putting special controls that govern CFEngine's startup (note this applies to all agents, not just `cf-agent`) in a single, logically consistent place. That would be a good opportunity to let the user tune network discovery in general.

Doing network discovery on demand is an interesting approach, but I fear it could break in many cases. Again, considering this is the exception, I'd rather not add risky code to the evaluator, but just keep things simple and put an override in `def.json`.

@worr you mentioned you'd probably try to refactor the network discovery after this is merged. Can you comment on this discussion and how it may relate to your needs specifically?

kacf:
After thinking about this some more, I'm wondering if we should drop the `/proc/net/{tcp,tcp6,upd,udp6}` portion of this pull request. This is information about each and every connection, it will be responsible for 99% of the added startup cost on a loaded server, and the information is not that useful IMHO. In fact, I think that on a busy server, the information will already be out of date a few seconds after cf-agent has started.

In that sense I think that the rest of the information which is slightly more static, such as the routes and device info, are much more useful in a configuration management setting.

I'd like to hear to if you have a use case for the connection information, but otherwise I think we should take that part out, which should satisfy the performance concern @estenberg has.

nickanderson:
The information from there would be useful from within cf-monitord, it could be useful to have the ability to do anomaly detection. Maybe the agent doesn't need to parse out that info, but i would like to keep the code available for future use in cf-monitord. Whatever the right way to do that would be.

nickanderson:
Regarding the idea of placing controls in def.json, I plan to slowly insrument more of masterfiles things with variables defined in def. I had imagined that I would add stuff like this:

```
{
  "vars":
     "controls" :
      { "agent": { "domain": "cfengine.com"},
        "common": { "bundlesequence": [ "bundle1", "bundle2", "bundlen" ]
    }
}
```

tzz:
@kacf OK, rebased and trimmed to exclude the ports parsing. The ports parsing will be in a followup PR.

I also noticed that the ports parsing compiled the regex for every line, so I optimized that out of the inner loop. The change is minor but added some code.

tzz:
The acceptance test was adjusted as well, and I removed the Linux requirement since I expect several other platforms to be supported eventually.

tzz:
@kacf OK, #2495 has the ports parsing function!

tzz:
Some comments:

* @kacf @estenberg the connection information is definitely useful. I implemented it through custom means for several CFEngine users. For example, it lets us find out if we're being probed on a particular port across the host population (which wouldn't necessarily be noticeable on a single host). This could be put into a Mustache template and reported, or directly reported to the hub.
* @nickanderson cf-monitord could probably use connection information to do the above automatically, but the data wouldn't be available for reports.
* @nickanderson Sorry but this will be a long digression... I think `bundlesequence` should be a top-level key in `def.json`, independent of variables and classes. Basically everything in `body common control` could be allowed, or it could go under a `common` key:
  * `bundlesequence`
  * `ignore_missing_bundles`
  * `ignore_missing_inputs`
  * `inputs` (done already)
  * `version`
  * `output_prefix`
  * `domain`
  * `syslog_host`
  * `syslog_port`
  * `cache_system_functions`
  * `protocol_version`
  * `bwlimit`
  * `tls_ciphers`
  * `tls_min_version`
  * `package_inventory`
  * `package_module`

Similarly to the above `body common control` entries, the settings from `body agent control` etc. control bodies (serverd, execd, monitord...), as seen in `mod_common.c`, could be put under top `agent`, `server`, etc. keys in `def.json` with very little work. @nickanderson if you agree and want to propose a specific behavior, please put it in https://dev.cfengine.com/issues/7453 or in a new ticket. I'll wait for you and other CFEngine folks to speak up. Sorry for digressing, I know this PR is not about `def.json` but it does touch upon it a bit.

libenv/unix_iface.c
nickanderson: Nice, this negates the need for https://github.com/cfengine/masterfiles/pull/649

tzz:
@nickanderson note this requires `/proc/net` so it doesn't work everywhere.

tzz:
I realized I had forgotten `/proc/net/if_inet6` and added a parser for it, to extract the IPv6 interface info including addresses (parallel to `sys.ipv4_addresses`).

tzz:
Prompted by https://github.com/cfengine/masterfiles/pull/648#issuecomment-185785602 I added a commit to select the default gateway based on the lowest metric value. With acceptance test as usual.

kacf:
Thanks, I will look at this ASAP!

tests/acceptance/00_basics/environment/proc-net.cf.sub
```diff
@@ -0,0 +1,38 @@
...
+  vars:
+      "testname" string => regex_replace($(this.promise_filename), "\\.sub$", "", "");
+
+  methods:
+      "check"  usebundle => dcs_check_state(test,
+                                           "$(this.promise_filename).expected.json",
+                                           $(testname));
+      "check"  usebundle => dcs_check_state(test,
+                                           "$(this.promise_filename).expected.json",
+                                           $(this.promise_filename));
```
kacf: Did you mean to do this twice? I suppose the one with `$(testname)` is the one to keep, since you're pretending to print from the top level test.

libenv/unix_iface.c
```diff
@@ -752,4 +755,444 @@ static int aix_get_mac_addr(const char *device_name, uint8_t mac[6])
...
+    case TCP_LAST_ACK:    return "LAST_ACK";
+    case TCP_LISTEN:      return "LISTEN";
+    case TCP_CLOSING:     return "CLOSING";
+    }
+
+# endif
+    return "UNKNOWN";
+}
+
+static ProcPostProcessFn NetworkingPortsPostProcessInfo(ARG_UNUSED void *passed_ctx, void *json)
```
kacf: `NetworkingPortsPostProcessInfo()` and `GetPortStateString()` should be moved to #2495. I can't see that they are needed in this pull request. This is important if we need to revert anything later.

kacf:
Nice acceptance test!

kacf:
trigger build

tests/acceptance/00_basics/environment/proc-net.cf.sub
tzz: Yes, it's intentional, so the test can be run standalone. I know it looks strange.

libenv/unix_iface.c
tzz: OK; done. I had to make a followup commit to do it, which we can squash later, and will rebase #2495

tzz:
#2495 has been rebased

libutils/regex.c
```diff
@@ -238,10 +218,84 @@ Seq *StringMatchCaptures(const char *regex, const char *str, const bool return_n
...
+    Seq *ret = StringMatchCapturesWithPrecompiledRegex(pattern, str, return_names);
     pcre_free(pattern);
     return ret;
 }
 
+/*******************************************************************/
+
+// returns NULL on any failure
+// takes either a pre-compiled pattern OR a regex (one of the two shouldn't be NULL)
+JsonElement* StringCaptureData(pcre *pattern, const char* regex, const char* data)
```
kacf: Unit tests fail because this function references `Json*` functions, and the test is built standalone, with only string and regex functions compiled in. Perhaps this should be moved to `json.c` and `json.h` instead?

tzz: OK; done with a single commit here. I rebased #2495 as well so it has this commit, so that's fixed.

tzz:
@bahamat @kacf the PR has been updated with a followup commit to:

* move `sys.routes` to `sys.inet[routes]`
* move `sys.proc_stats` to `sys.inet[stats]`
* move `sys.ipv6_routes` to `sys.inet6[routes]`
* move `sys.ipv6_addresses` to `sys.inet6[addresses]`
* introduce `sys.inet6[stats]` parsed from `/proc/net/snmp6`

#2495 will be rebased accordingly as well.

tzz:
@kacf another small update:

* on suggestion from @bahamat,  moved `sys.default_gateway` to `sys.inet[default_gateway]` and added `sys.inet[default_route]`. Acceptance test updated.

Note that there is no corresponding `sys.inet6[default_gateway]` or `sys.inet6[default_route]` yet. They can be added later.

kacf:
Alright, sounds good.

kacf:
trigger build

kacf:
trigger build

kacf:
This failed on all non-Linux platforms. I'll try to take a look at why.

tzz:
@kacf with the help of @zofrex I found that this does not fail on FreeBSD if it's rebased against `origin/master`. So I've done that here!

I'll rebase #2495 when this PR is merged, the constant rebasing is annoying, I've done it 4 times now. I hope that's all right.

tzz:
If it helps @zofrex or others, here's the Vagrantfile I used:

```
Vagrant.configure("2") do |config|
  config.vm.guest = :freebsd
  config.vm.synced_folder ".", "/vagrant", id: "vagrant-root", disabled: true
  config.vm.box = "freebsd/FreeBSD-11.0-CURRENT"
  config.ssh.shell = "sh"
  config.vm.base_mac = "080027D14C66"

  config.vm.provider :virtualbox do |vb|
    vb.customize ["modifyvm", :id, "--memory", "1024"]
    vb.customize ["modifyvm", :id, "--cpus", "1"]
    vb.customize ["modifyvm", :id, "--hwvirtex", "off"]
    vb.customize ["modifyvm", :id, "--audio", "none"]
    vb.customize ["modifyvm", :id, "--nictype1", "virtio"]
    vb.customize ["modifyvm", :id, "--nictype2", "virtio"]
  end

  config.vm.provision "shell", privileged: false, inline: <<-SHELL
      set -ex
      # cd cfengine
      sudo pkg install --yes gmake bash automake libtool tokyocabinet pcre git
      sudo ln -s /usr/local/bin/bash /bin/bash
      echo 'Host github.com' >> ~/.ssh/config && echo '  StrictHostKeyChecking no' >> ~/.ssh/config
      git clone git@github.com:cfengine/core.git
      cd core
      git fetch origin pull/2245/head && git checkout FETCH_HEAD
      ./autogen.sh --with-tokyocabinet=/usr/local --with-pcre=/usr/local
      ./configure --with-tokyocabinet=/usr/local --with-pcre=/usr/local
      make clean
      make
    SHELL
end
```

kacf:
Doh! Of course, this is ifdef'ed out on non-Linux. I believe it succeeded on all Linuxes though.

If you add a `"test_skip_unsupported" string => "!linux"`, I will run it once more for safety, but I think you can otherwise consider this approved and move to documentation.

tzz:
OK, I added that to the acceptance test. Docs coming.

kacf:
trigger build

tzz:
@kacf sorry I made one last change: renamed `sys.dev` to `sys.interfaces_data` because, when I started writing the docs, I realized the `sys.dev` name is too generic and unhelpful. I also rebased on top of master.

kacf:
Ah, good thinking! The build hasn't started yet, so it should pick up those changes as well.

tzz:
docs in https://github.com/cfengine/documentation/pull/1400

kacf:
trigger build

kacf:
trigger build

kacf:
Finally green! Squashed and merged manually together with some additional build fixes and changelog fixups.

tzz:
wooo hooo this is merged thank you so much for your help everyone!

PR #2246, opened 2015-06-19
Redmine #7075 Store absolute time with process information

kacfengine:
<pre>
commit 203dd9a7328d4557b913d489e00e84d98a1ab77a
Author: Stefan Weil <sw@weilnetz.de>
Date:   Mon Mar 30 20:35:36 2015

    Redmine #7075 Store absolute time with process information
    
    Processes in CFEngine are identified by their OS process identifier (pid)
    and their start time.
    
    Up to now, the start time was relative to the boot time, so it was not
    unique across system restarts.
    
    Using an absolute start time assures that the combination of pid and
    start time is unique. This fixes bug #7075.
    
commit fd5b4f7f91a264e1ddfa10cf4d376a75d6430dcf
Author: Stefan Weil <sw@weilnetz.de>
Date:   Mon Mar 30 20:27:08 2015

    Add new function GetBootTime
    
    The new function will be used for Linux in the following commit.
    
    The code was taken from the old implementation of GetUptimeSeconds
    which now uses GetBootTime.
    
    An implementation for Windows can be added later (when needed).
    
</pre>

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -2666,103 +2666,119 @@ static void GetCPUInfo(EvalContext *ctx)
 #ifndef __MINGW32__
 
 /**
+   Return the current boot time (number of seconds since Epoch).
+*/
+time_t GetBootTime(void)
+{
+    static time_t boot_time = 0;
+    errno = 0;
+
+    if (boot_time == 0)
+    {
+#if defined(BOOT_TIME_WITH_SYSINFO)         // Most GNU, Linux platforms
+
+        time_t now = time(NULL);
+        struct sysinfo s;
+        if (sysinfo(&s) == 0)
+        {
+           // Don't return yet, sanity checking below
+           boot_time = now - s.uptime;
+        }
+
+#elif defined(BOOT_TIME_WITH_KSTAT)         // Solaris platform
+
+        /* From command line you can get this with:
+           kstat -p unix:0:system_misc:boot_time */
+        kstat_ctl_t *kc = kstat_open();
+        if (kc != 0)
+        {
+            kstat_t *kp = kstat_lookup(kc, "unix", 0, "system_misc");
+            if (kp != 0)
+            {
+                if (kstat_read(kc, kp, NULL) != -1)
+                {
+                    kstat_named_t *knp = kstat_data_lookup(kp, "boot_time");
+                    if (knp != NULL)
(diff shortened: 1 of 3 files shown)


kacfengine:
trigger build

libenv/sysinfo.c
```diff
@@ -2666,103 +2666,119 @@ static void GetCPUInfo(EvalContext *ctx)
...
+    {
 #if defined(BOOT_TIME_WITH_SYSINFO)         // Most GNU, Linux platforms
 
-    struct sysinfo s;
-    if (sysinfo(&s) == 0)
-    {
-       // Don't return yet, sanity checking below
-       boot_time = now - s.uptime;
-    }
+        time_t now = time(NULL);
```
jimis: What if "now" changes as the system proceeds? For example battery-less embedded devices will start with unset time, then set it via NTP.

kacfengine: In that case we get the same issue as before the fix, boot time will be zero (or another constant value), and processes may get the same start time unintentionally. Only after the time is set do we get unique start times.

In theory you could defeat this mechanism by setting the time backwards, but you'd have to be *insanely* unlucky to both have the process start at the same second that the previous one started *and* get the same PID. This is practically impossible unless your intention is to trigger it and you try over and over. Certainly it is much more likely that the original issue described in https://dev.cfengine.com/issues/7075 will happen using the old method, than triggering what I just described with the new method.

jimis: You're right, but I'm not arguing this.

What I'm commenting about is that `boot_time` is *static*, which means that
a long running process will have the wrong time for all its lifetime.

kacfengine: Aah, now I see what I mean, well spotted. You're right, that may be a problem. I wonder how expensive it is to find out uptime every time. Hmm, need to think about that one...

kacfengine: ...to find out *boot time* every time...

kacfengine:
@jimis brought up a very good point that with this fix you can render CFEngine unable to kill any stale agent simply by adjusting the system time one second forwards or backwards. The reason is that the new boot time depends on current time, and so if you change it, all processes will now appear foreign to CFEngine, and it will not attempt to kill them if they have gone stale. This is not good.

Damn, I had high hopes for this fix.

PR #2247, opened 2015-06-22
libutils/json.c: support bareword keys in JSON parsing

tzz:
This code supports bareword JSON keys, e.g.

```
{
  bar : "zulu",
  foo: [
    "alpha",
    "bravo"
  ]
}
```

It also defines a new hard class `feature_json_bareword_keys` to detect if this is enabled.

I believe the code is safe and correct but appreciate some review.  Upon acceptance I can write tests and docs.

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -954,6 +954,8 @@ static void BuiltinClasses(EvalContext *ctx)
 #ifdef HAVE_LIBYAML
     CreateHardClassesFromCanonification(ctx, "feature_yaml", "source=agent");
 #endif
+
+    CreateHardClassesFromCanonification(ctx, "feature_json_bareword_keys", "source=agent");
 }
 
 /*******************************************************************/
--- a/libutils/json.c
+++ b/libutils/json.c
@@ -2061,7 +2061,30 @@ static JsonParseError JsonParseAsObject(const char **data, JsonElement **json_ou
             return JSON_PARSE_OK;
 
         default:
-            if (property_name)
+            // Note the character class excludes ':'.
+            // This will match the key from { foo : 2 } but not { -foo: 2 }
+            if (NULL == property_name &&
+                StringMatch("^\\w[-\\w]*\\s*:", *data, NULL, NULL))
+            {
+                char *colon = strchr(*data, ':');
+
+                // Step backwards until we are on the last whitespace.
+
+                // Note that this is safe because the above regex guarantees
+                // we will find at least one non-whitespace character as we
+                // go backwards.
+                char *ws = colon;
+                while (IsWhitespace(*(ws-1)))
+                {
+                    ws -= 1;
+                }
+
+                property_name = xstrndup(*data, ws - *data);
+                *data = colon;
+
+                break;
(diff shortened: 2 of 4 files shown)


neilhwatson:
++ Saving me from typing is a Good Thing.

kacfengine:
Is this part of the JSON standard? While slightly ambiguous, the specification over at http://json.org/ does seem to indicate that keys are real strings, and thus should be quoted. I'd rather not have us deviate from the spec, since it will ultimately cause confusion.

What about other parsers?

tzz:
Almost every parser supports this option, and one of the original authors of Javascript (where JSON originated) has said it was a mistake to require quoted keys.  Basically things like `return` couldn't be handled by the parser and they decided to be strict and require string keys.  I can't find the article anymore but I recall it clearly.

We can make it a compile-time or run-time option but I think there's never any harm in allowing it.

tzz:
ping?

estenberg:
Looks good.

tzz:
An idea: it could be configurable.  We could check `enable_feature_json_bareword_keys` (default any) and `disable_feature_json_bareword_keys` (default !any) classes, so users can turn this off if they don't want it.  I don't think it's worthwhile personally.

tzz:
Acceptance test added, ready for merge.

kacfengine:
> An idea: it could be configurable. We could check enable_feature_json_bareword_keys (default any) and disable_feature_json_bareword_keys (default !any) classes, so users can turn this off if they don't want it. I don't think it's worthwhile personally.

No, I don't think so either. I was just worried we might break the standard, but it seems harmless enough.

kacfengine:
trigger build

kacfengine:
Testing https://github.com/tzz/core/commit/a85b581ce92e0f21a55393faa2ac66c11a9b0756.

kacfengine:
Thanks, all good, merged manually.

kacfengine:
Oh, and I removed the ``feature_json_bareword_keys``, since it's not configurable.

tzz:
...but now users don't have a way to check if bareword keys are supported.  The idea was not to make it configurable so much as discoverable.  Can you consider bringing that feature class back?

kacfengine:
Do you think that's needed? Many features are added in new versions, and we don't add feature classes for all of them. A version check can accomplish the same thing.

Not necessarily against it though, it just seems a bit excessive.

tzz:
In this case specifically, the user won't know if it's safe to use bareword keys until they are rejected.  With the feature class, they can do `cf-promises --list-classes` and grep for that feature before triggering errors.

kacfengine:
Yes, but he can use ``cfengine_3_8::``. Don't you think that's good enough? After all we don't recommend running multi-version setups, although it does happen, of course.

tzz:
Testing for a specific version is a bit of a pain but I won't argue if you think it's the right way.

kacfengine:
Yeah, unless you have a strong feeling the other way I think it's better to leave it out. After all it's only useful in the transition from 3.7 to 3.8, after that it will be useless.

PR #2249, opened 2015-06-24
Redmine #7297: Make sure systemd is detected on Debian systems.

kacfengine:
Debian systems symlink /sbin/init to their systemd binary, but the
/proc/1/cmdline entry will contain the name of the link, not of the
target, hence our existing method will not work. So check the link
target of /sbin/init in addition.

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -989,7 +989,10 @@ static void OSClasses(EvalContext *ctx)
 /* First we check if init process is systemd, and set "systemd" hard class. */
 
     {
+        bool systemd_detected = false;
+
         char init_cmdline[CF_BUFSIZE];
+        init_cmdline[0] = '\0';
         if (ReadLine("/proc/1/cmdline", init_cmdline, sizeof(init_cmdline)))
         {
             char *p;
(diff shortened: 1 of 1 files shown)


kacfengine:
I tested by switching out the link manually (scary on a live system!), but I haven't tested on an actual Debian system. Inspired by https://github.com/cfengine/masterfiles/pull/488.

kacfengine:
trigger build

kacfengine:
@jimis: Review?

jimis:
Hmmm, wouldn't it be safer to first read `/proc/1/cmdline`, second see if it's a link and if so dereference it, and third see if the end result ends with "/systemd"?

kacfengine:
My thinking was that reading a link would be faster than opening and reading a process entry, so we start with the cheapest option. But this is guesstimation at best, and we can reverse that order if you think that's better.

``readlink`` will both tell you if it's a link and dereference it, so you only need that one call. I assume under the hood that both ``readlink`` and ``stat`` go to the inode.

kacfengine:
> ``readlink`` will both tell you if it's a link and dereference it

Which is the ``EINVAL`` error check, I should add.

jimis:
My guesstimation is that reading /proc/whatever is faster, since it has nothing to do with filesystem, inode cache, and even possibly hitting the disk.

Plus it catches the case where `/sbin/init` exists but was not chosen as PID 1.

kacfengine:
> My guesstimation is that reading /proc/whatever is faster, since it has nothing to do with filesystem, inode cache, and even possibly hitting the disk.

Ah, good point, I didn't think of that.

> Plus it catches the case where /sbin/init exists but was not chosen as PID 1.

A perverse case, but yeah, I suppose it is possible! I'll change it.

kacfengine:
Alright, moved the readlink check after the /proc check, and also reused the name obtained from ``/proc/1/cmdline`` instead of hardcoding ``/sbin/init``.

kacfengine:
trigger build

jimis:
I would nitpick on this that it's too complex, e.g. there is code duplication about detecting the last component of the path. Instead I took 10 min and wrote what I consider as simple:

https://github.com/cfengine/core/pull/2252

What do you think?

PR #2250, opened 2015-06-24
Add: Test that classes without constraints fails

nickanderson:
Classes without constraints should fail cf-promises and cause it to return
nonzero.

Ref: https://dev.cfengine.com/issues/7308


kacfengine:
Ticket was rejected.

PR #2252, opened 2015-06-25
Redmine #7297: Make sure systemd is detected if PID 1 is a symlink

jimis:
(E.g. on Debian).

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -989,17 +989,26 @@ static void OSClasses(EvalContext *ctx)
 /* First we check if init process is systemd, and set "systemd" hard class. */
 
     {
-        char init_cmdline[CF_BUFSIZE];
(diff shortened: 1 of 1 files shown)


kacfengine:
Good idea to use ``realpath``. It is simpler indeed.

jimis:
> Good idea to use realpath. It is simpler indeed

And it will follow all possible symlinks, for example you might need two dereferences to reach the goal.

kacfengine:
Indeed.

kacfengine:
trigger build

kacfengine:
Cherry-picked to both 3.6.x and 3.7.x.

jrobeson:
isn't it best to use /proc/1/comm to detect systemd ?

kacfengine:
Both are good ways I think.

jimis:
@jrobeson `/proc/1/comm` doesn't exist on old kernels.

jrobeson:
and those old kernels can't run systemd either.

PR #2253, opened 2015-06-25
Redmine#7116: getvalues() should always return a list and should reference inside data containers

tzz:
See https://dev.cfengine.com/issues/7116

 Minor style and legibility fixes.

`getvalues()` now always returns a list when called on a JSON scalar.

`getvalues("datacontainer[key]")` and `getindices("datacontainer[key]")` now work.


tzz:
@jimis @kacfengine please consider this bugfix for 3.6 and 3.7.  I'll write tests and docs if it's acceptable.

tzz:
After discussion with @nickanderson I think the patch is ready for review and the behavior is what it should be.

jimis:
I believe this will have to wait for the first patch release since the release is imminent we're lacking time to put this in at this stage. The fix is much appreciated though! :-)

tzz:
I hope this can be reviewed and merged.

estenberg:
Looks good.

tzz:
@nickanderson since you requested this and know the spec, can you write an acceptance test?  Using `dcs_check_state` it should be pretty trivial.

libpromises/evalfunction.c
```diff
@@ -1865,39 +1865,83 @@ static FnCallResult FnCallRegArray(EvalContext *ctx, ARG_UNUSED const Policy *po
...
+                VarRefDestroy(ref);
+                return FnFailure();
+            }
         }
     }
 
-    DataType type = CF_DATA_TYPE_NONE;
-    const void *value = EvalContextVariableGet(ctx, ref, &type);
+    type = CF_DATA_TYPE_NONE;
+    const void *var_value = EvalContextVariableGet(ctx, ref, &type);
```
kacfengine: You just did this before the ``if``. Can you move this assignment up there, and save one function call?

libpromises/evalfunction.c
```diff
@@ -1936,68 +1980,45 @@ static FnCallResult FnCallGetIndices(EvalContext *ctx, ARG_UNUSED const Policy *
...
```
kacfengine: Same here, you're doing this twice.

kacfengine:
Looks good, but I'm also missing an acceptance test, pinging @nickanderson, ref https://github.com/cfengine/core/pull/2253#issuecomment-128726433.

libpromises/evalfunction.c
tzz: If the variable name was not a data container, `ref` may change, so I don't think it's possible to consolidate.

libpromises/evalfunction.c
tzz: Yup, same problem: `ref` may change in the preceding block.

tzz:
I added a note in https://dev.cfengine.com/issues/7116 for @nickanderson so it's clear this is not done.

libpromises/evalfunction.c
kacfengine: Right, I see what you mean.

tzz:
This is rebased and the acceptance test for `getvalues()` has been expanded to demonstrate:

* `null`s are skipped
* data containers are crawled recursively
* maps of arrays are crawled correctly

Please merge, it's good(TM)

kacfengine:
trigger build

kacfengine:
Sorry for the delay, this looks good!

kacfengine:
Apparently Jenkins doesn't like it though, I get this:

```
Error Message

FAIL (UNEXPECTED FAILURE) ./02_classes/02_functions/getvalues_containers.cf

Stacktrace

R: arr_v a
R: arr_v 1
R: arr_v 2
R: arr_v 5.30
R: arr_v true
R: /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_x86_64_linux_ubuntu_12/core/tests/acceptance/./02_classes/02_functions/getvalues_containers.cf FAIL
```

Do you know what it might be?

tzz:
Sorry, I have no idea.  It could be a bug for sure, but I can't tell from the log you posted.

kacfengine:
This seems to fail across the board on Linux though, can you reproduce it locally?

tzz:
I just rebased this PR and tested `01_variables/02_functions/getvalues*.cf`:

```
./01_vars/02_functions/getvalues_containers.cf Pass
./01_vars/02_functions/getvalues_flattens-list-elements.cf Pass
./01_vars/02_functions/getvalues_returns_all_values_for_given_array_index_merging_strings.cf Pass
./01_vars/02_functions/getvalues_returns_all_values_for_given_datacontainer_index_merging_strings.cf Pass
./01_vars/02_functions/getvalues_returns_slist_for_array_scalar.cf Pass
./01_vars/02_functions/getvalues_returns_slist_for_datacontainer_scalar.cf Pass
```

I moved `./01_vars/02_functions/getvalues_containers.cf` to the right place (it tests variables, not classes), and refactored it using the new bundlestate testing. But it didn't segfault before that, either, I just got errors. So I'm not sure what I'm doing wrong that it's not failing for me, sorry.

nickanderson:
trigger build

kacf:
It passes now, thanks!

kacf:
Merged manually.

kacf:

PR #2254, opened 2015-06-28
contrib/cfengine.el: minor regex improvement for the macro syntax

tzz:
Minor fix to avoid indenting a `@(myvar)` like the below to the beginning of the line.  Safe to merge.

```
mybundle(
                 @(myvar)
);
```


kacfengine:
It could even be ``@if`` and ``@endif`` explicitly, but this should be quite safe as well.

tzz:
@kacfengine I considered it but this way I don't have to release a new version when we add new macros!

PR #2255, opened 2015-06-29
Redmine#7152: Allow whitespace in class expressions inside strings

tzz:
See https://dev.cfengine.com/issues/7152

This change applies **only** to contexts inside strings.  So anything passed to `if/unless/ifvarclass` will be stripped of whitespaces.  Also, quoted string contexts like the below full example will be likewise stripped.  But something like `linux | cfengine::` without quotes will not be OK.

This does not change the parser, only the treatment of context strings **after** syntax parsing and **before** expression parsing.  Thus, I believe it's a safe change.  We can remove `\r` and `\n` from the allowed characters if you think they are too much.

```
bundle agent main
{
  reports:
    'any | (cfengine | linux)'::
      "hello!";
}
```

If this is acceptable, I'll write tests and docs.

### DIFF
--- a/libpromises/eval_context.c
+++ b/libpromises/eval_context.c
@@ -441,6 +441,11 @@ static char *EvalVarRef(ARG_UNUSED const char *varname, ARG_UNUSED VarRefType ty
 
 /**********************************************************************/
 
+bool ClassCharIsWhitespace(char ch)
+{
+    return ch == ' ' || ch == '\t' || ch == '\n' || ch == '\r';
+}
+
 bool IsDefinedClass(const EvalContext *ctx, const char *context)
 {
     ParseResult res;
@@ -450,7 +455,10 @@ bool IsDefinedClass(const EvalContext *ctx, const char *context)
         return true;
     }
 
-    res = ParseExpression(context, 0, strlen(context));
+    Buffer *condensed = BufferNewFrom(context, strlen(context));
+    BufferRewrite(condensed, &ClassCharIsWhitespace, true);
+    res = ParseExpression(BufferData(condensed), 0, BufferSize(condensed));
+    BufferDestroy(condensed);
 
     if (!res.result)
     {
--- a/libpromises/eval_context.c
+++ b/libpromises/eval_context.c
@@ -47,6 +47,9 @@
 #include <logging_priv.h>
 #include <known_dirs.h>
 #include <printsize.h>
+#include <regex.h>
+
+static pcre *context_expression_whitespace_rx = NULL;
 
 static bool BundleAborted(const EvalContext *ctx);
 static void SetBundleAborted(EvalContext *ctx);
@@ -453,6 +456,23 @@ bool IsDefinedClass(const EvalContext *ctx, const char *context)
     if (!context)
     {
         return true;
+    }
+
+    if (NULL == context_expression_whitespace_rx)
+    {
+        context_expression_whitespace_rx = CompileRegex(CFENGINE_REGEX_WHITESPACE_IN_CONTEXTS);
+    }
+
+    if (NULL == context_expression_whitespace_rx)
+    {
+        Log(LOG_LEVEL_ERR, "The context expression whitespace regular expression could not be compiled, aborting.");
+        return false;
+    }
+
+    if (StringMatchFullWithPrecompiledRegex(context_expression_whitespace_rx, context))
+    {
+        Log(LOG_LEVEL_INFO, "class names can't be separated by whitespace without an intervening operator in expression '%s'", context);
+        return false;
     }
 
     Buffer *condensed = BufferNewFrom(context, strlen(context));
--- a/libutils/buffer.c
+++ b/libutils/buffer.c
@@ -445,6 +445,37 @@ void BufferSetMode(Buffer *buffer, BufferBehavior mode)
     buffer->mode = mode;
 }
 
+Buffer* BufferFilter(Buffer *buffer, BufferFilterFn filter, const bool invert)
+{
+    assert(buffer);
+
+    Buffer *filtered = BufferNew();
+    for (unsigned int i = 0; i < buffer->used; ++i)
+    {
+        bool test = (*filter)(buffer->buffer[i]);
+        if (invert)
+        {
+            test = !test;
+        }
+
+        if (test)
+        {
+            BufferAppendChar(filtered, buffer->buffer[i]);
+        }
+    }
+
+    return filtered;
+}
+
+void BufferRewrite(Buffer *buffer, BufferFilterFn filter, const bool invert)
+{
+    assert(buffer);
+
+    Buffer *rewrite = BufferFilter(buffer, filter, invert);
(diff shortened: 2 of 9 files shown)


tzz:
ping?

estenberg:
Looks good to me.

nickanderson:
yup

tzz:
Acceptance test added to existing policy and simplified.

libutils/buffer.c
```diff
@@ -445,6 +445,33 @@ void BufferSetMode(Buffer *buffer, BufferBehavior mode)
...
+Buffer* BufferFilter(Buffer *buffer, FilterFn filter, const bool invert)
+{
+    assert(buffer);
+
+    Buffer *filtered = BufferNew();
+    for (unsigned int i = 0; i < buffer->used; ++i)
+    {
+        const bool test = (*filter)(buffer->buffer[i]);
+
+        if (test || (invert && !test))
```
kacfengine: This will not invert, it will simply include every character if ``invert`` is true, in which case the expression essentially becomes ``test || !test``. A better expression would be ``invert ? !test : test``.

libutils/buffer.h
```diff
@@ -56,6 +56,7 @@ typedef struct
 } Buffer;
 
 
+typedef bool (*FilterFn)(char item);
```
kacfengine: Propose renaming this to ``BufferFilterFn``, to avoid cluttering the namespace.

kacfengine:
Agree fully with the intention, this makes class expressions much more readable. How hard would it be to do the same for non-quoted class expressions? I assume that would require changes in the parser? We can leave it for another PR, I'm just wondering.

tzz:
Non-quoted class expressions are parsed in a very different way (when quoted, they end up in an `ifvarclass` attribute).  So I don't think it will be easy to put it in this PR, but I also don't think it's very hard if we disallow carriage returns or newlines (vertical whitespace).  Make a ticket if you think it's worthwhile...

libutils/buffer.h
tzz: OK; done

libutils/buffer.c
tzz: I expanded it into several lines to be much clearer.  Too much cleverness makes computers sad.

tzz:
OK, cleaned up and logic and tests fixed.  Thank you for the help.  This is ready, I'll write docs when approved.

kacfengine:
I would argue that whitespace in non-quoted expressions are part of the original ticket as it is written. This completes it halfway I suppose.

Agree about the newlines. Although I think we wouldn't get any ambiguities in the language even if we allowed them, I see little benefit in doing so if it's difficult.

Looks good to me now. Assuming the tests pass, this is ready to go.

kacfengine:
trigger build

tzz:
I realized that it's actually really simple to implement the other half, so I did it...  see the second commit here.  It works for me and I added to the acceptance test, but I think it will need to run through the Jenkins gauntlet before it's acceptable.

It's 100% backwards compatible, I believe.  Example:

```
bundle agent main
{
  reports:
    any | (cfengine | linux)::
      "hello with contexts!";
}
```

If this is accepted, I'll have to modify the Emacs CFEngine mode, right now it's confused by this new syntax.

tzz:
I didn't allow newlines because I feel they would destabilize the parser.  You can try adding them separately after this is merged.

basvandervlies:
@tzz some time it simple, nice one

kacfengine:
Great!

kacfengine:
trigger build

basvandervlies:
I just looked and the code and the only minor thing is that this will be a valid class if we do not use '|' or '.':
 * `debian redhat::`

will set the class:
 * `debianredhat`

kacfengine:
Oh, nice catch!

tzz:
@basvandervlies I'm not sure if that's an actual problem.

Why would it happen and how should we guard against it?  Is it worth the extra code?  I can imagine scanning the context expression string for the pattern `[a-zA-Z0-9]<space>[a-zA-Z]` but it seems like an unlikely situation to me.

kacfengine:
@tzz: Just make sure the string alternates between non-operators and operators. It will probably require a more specialized function than the ``BufferFilter`` function.

basvandervlies:
@tzz What I said It is a minor thing, but it should be realized that you can endup with a valid class even if you did not expect it. I agree the chance is low, bt if you hit it then search can be painful.  When we encounter a space char a operator must be the next char.

tzz:
OK; done.

I use a regular expression because doing it with a C string scan is nasty, error-prone, and I'm lazy (in a good way :).  The regex is compiled once so it should be OK for performance (it's a one-time parse-time penalty and the data set is small in count and string size).

The acceptance tests have been adjusted accordingly.  I can't think of any other edge cases but if you do, please let me know.  Thanks @basvandervlies and @kacfengine

libpromises/eval_context.c
```diff
@@ -450,7 +455,10 @@ bool IsDefinedClass(const EvalContext *ctx, const char *context)
         return true;
     }
 
-    res = ParseExpression(context, 0, strlen(context));
+    Buffer *condensed = BufferNewFrom(context, strlen(context));
+    BufferRewrite(condensed, &ClassCharIsWhitespace, true);
+    res = ParseExpression(BufferData(condensed), 0, BufferSize(condensed));
+    BufferDestroy(condensed);
```
kacfengine: Should we do the "no operator in between" check here as well, in addition to (or perhaps instead of) above? Then the check will work for both quoted and non-quoted strings, unless I'm mistaken.

tests/acceptance/00_basics/01_compiler/string_contexts.x.cf
```diff
@@ -0,0 +1,30 @@
...
+#######################################################
+
+bundle agent test
+{
+  reports:
+    cf e ngine::
+      "hello";
+
+    cf engine::
+      "hello";
```
kacfengine: Perhaps a ``"cf engine"::`` case, as per my other comment.

tzz:
Hmm, I can't consolidate it because the `cf3lex.l` error handling is at the lexer level and aborts the whole parse with `yyerror`.  The check in `IsDefinedClass` is run-time, we don't want to abort altogether, just warn at `INFO` level and `return false`.  Otherwise data sources could feed dynamic class expressions there and break the policy.

So I added a new check in `eval_context.cf`, sharing the regex itself with the lexer.

kacfengine:
Nice! I'll run it through Jenkins a bit later, but you can consider this approved from my side. Just docs missing.

kacfengine:
trigger build

kacfengine:
Everything green. Will you provide the docs?

tzz:
docs in https://github.com/cfengine/documentation/pull/1142

tzz:
docs PR updated, thanks for the review

kacfengine:
Merged manually. Thanks for the feature!

PR #2257, opened 2015-06-29
Change: Improve git based release id

nickanderson:
Now cf-promises -T is not limited to working when you are on a checked out
branch. Now you get the commit that is checked out as the release id even if
you are checked out from a specific commit or tag.

It would be really nice if this goes into 3.7.0. I also think it should be backported to 3.6.x

Ref: https://dev.cfengine.com/issues/7332

### DIFF
--- a/libpromises/generic_agent.c
+++ b/libpromises/generic_agent.c
@@ -816,11 +816,11 @@ static bool GeneratePolicyReleaseIDFromGit(char *release_id_out,
     {
         char git_head[128];
         int scanned = fscanf(git_file, "ref: %127s", git_head);
-        fclose(git_file);
 
         if (scanned == 1)
         // Found HEAD Reference which means we are on a checked out branch
         {
+            fclose(git_file);
             snprintf(git_filename, PATH_MAX, "%s/.git/%s", policy_dir, git_head);
             git_file = fopen(git_filename, "r");
             Log(LOG_LEVEL_DEBUG, "Found a git HEAD ref");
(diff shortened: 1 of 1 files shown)


nickanderson:
trigger build

nickanderson:
@kacfengine thoughts?

atsaloli:
We look forward to having this functionality, we rely on tags to promote code from Stage through PreProd to Prod.

kacfengine:
Looks pretty good to me! I suggest the following changes for efficiency, but otherwise good: https://gist.github.com/kacfengine/f004489185447a0e4433

Agree?

nickanderson:
YES patch integrated

kacfengine:
Thanks!

kacfengine:
trigger build

jimis:
Please ignore build failure on mingw, it auto-built there before the buildslave was excluded.

kacfengine:
Cherry-picked to 3.7.x.

kacfengine:
This could go to 3.6.x as well, since it's a bug fix, but that will require a dedicated backport, since that code section has changed.

libpromises/generic_agent.c
```diff
@@ -809,21 +809,40 @@ static bool GeneratePolicyReleaseIDFromGit(char *release_id_out,
...
         {
+            fclose(git_file);
             snprintf(git_filename, PATH_MAX, "%s/.git/%s", policy_dir, git_head);
             git_file = fopen(git_filename, "r");
+            Log(LOG_LEVEL_DEBUG, "Found a git HEAD ref");
         }
         else
         {
-            git_file = NULL;
+            Log(LOG_LEVEL_DEBUG, "Unable to find HEAD ref in %s. Looking for commit", git_file);
```
jimis: s/`git_file`/`git_filename`/

I'll commit a quick fix.

kacfengine: Ah, nice catch!

atsaloli: Lovely, thanks, folks!

PR #2260, opened 2015-06-30
Minor bugs revealed by trying to build with newer MinGW version.

jimis:


jimis:
trigger build

jimis:
If the build turns out green, just merge and trigger the TEMPtesting Jenkins job, and it should build on debian 8 as well.

kacfengine:
Cherry-picked to both 3.7.x and 3.6.x.

PR #2261, opened 2015-07-01
Deal with perverse ps output

ediosyncratic:
This adds a `#if`-ed out unit test in which to document (and be ready for any brave soul who thinks they can fix the parser to handle) real-world ps output lines that our parser can't presently handle.

A second commit re-enables a known failing-on-Solaris test and adds a diagnostic log message in one place that shall *hopefully* shed some light on why the test fails, when it (seldom) does.
It's a single commit to make it easier to revert once that's reported the hideous ps line that causes the problem (and added it to the first commit's test), while someone thinks about how to fix it.

None of this is urgent but it would make sense to include in master at some point when you can endure a failing test on Solaris for a short while.

### DIFF
--- a/libpromises/processes_select.c
+++ b/libpromises/processes_select.c
@@ -339,8 +339,9 @@ static int SelectProcTimeCounterRangeMatch(char *name1, char *name2, time_t min,
 
         if (value == CF_NOINT)
         {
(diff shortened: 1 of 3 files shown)


kacfengine:
Either it fails or it doesn't. Might revert this later when we know.

ediosyncratic:
> Either it fails or it doesn't.
Indeed - good luck ;-)

> Might revert this later when we know.

It's designed with reverting in mind ...

PR #2262, opened 2015-07-01
Monitord thermal errors redmine7238

jimis:


cf-monitord/mon_temp.c
```diff
@@ -86,10 +86,17 @@ void MonTempInit(void)
 {
     struct stat statbuf;
 
-    if (stat("/sys/devices/virtual/thermal", &statbuf) != -1)
+    for (int i = 0; i < 4; i++)
```
kacfengine: Why 4?

kacfengine: Oh, I see, it only supports four of them.

jimis: yes, it's already hardcoded elsewhere. It was to much for me to change at this point. :-)

kacfengine:
Looks good. I suppose none of this stuff has an acceptance test anyway, so if you tested it locally, just merge it.

jimis:
@kacfengine great thanks for reviewing!

PR #2263, opened 2015-07-02
Redmine #7335: Fix missing environment file on Debian systems.

kacfengine:
The paths differ, and systemd doesn't allow us to do any logic inside
its files, so we need to decide at the configure stage and then make a
replacement inside the service file.


kacfengine:
trigger build

kacfengine:
@jimis: This should fix Debian 8.

jimis:
+1 for going the autoconf-way!

PR #2265, opened 2015-07-02
DO NOT MERGE YET Change: Consolidate classes bodies in failsafe

nickanderson:
@jimis I started to replace all of the classes bodies in failsafe.cf with
scoped_classes_generic.

I stopped when I hit the repaired_two classes body.

Its used when restarting the services with systemd to define a class indicating
that both execd and serverd have been restarted. Can we really tell if both
components were started correctly? I want to replace that with a single class
for cfengine_components or something.

I was able to remove all uses of the repaired classes body.

### DIFF
--- a/libpromises/failsafe.cf
+++ b/libpromises/failsafe.cf
@@ -45,7 +45,7 @@ body agent control
       skipidentify => "true";
 
       # Bootstrapping can't continue without keys
-      abortclasses => { "no_ppkeys_ABORT" };
+      abortclasses => { "no_ppkeys_report_kept" };
 }
 
 ################################################################################
@@ -60,7 +60,7 @@ bundle agent cfe_internal_checkkeys
   reports:
     !have_ppkeys::
       "No public/private key pair is loaded, please create one by running \"cf-key\""
-        classes => kept("no_ppkeys_ABORT");
+        classes => scoped_classes_generic("namespace", "no_ppkeys_report");
 }
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


libpromises/failsafe.cf
```diff
@@ -45,7 +45,7 @@ body agent control
       skipidentify => "true";
 
       # Bootstrapping can't continue without keys
-      abortclasses => { "no_ppkeys_ABORT" };
+      abortclasses => { "no_ppkeys_report_kept" };
```
jimis: Why rename the abort class? It should be clearly visible that setting this will cause an ABORT! :-)

libpromises/failsafe.cf
```diff
@@ -94,23 +94,23 @@ bundle agent cfe_internal_update
...
 
-    !windows.got_policy_failed::
+    !windows.got_policy_using_shortcut_failed::
 
       "$(sys.inputdir)"
       handle => "cfe_internal_bootstrap_update_files_sys_workdir_inputs_not_windows",
       copy_from => u_scp("$(sys.masterdir)"),
       depth_search => u_recurse("inf"),
-      classes => repaired("got_policy");
+      classes => scoped_classes_generic("bundle", "got_policy_using_direct");
```
jimis: You differentiate the shortcut and the direct case. Is it for future-proofness?

libpromises/failsafe.cf
```diff
@@ -60,7 +60,7 @@ bundle agent cfe_internal_checkkeys
   reports:
     !have_ppkeys::
       "No public/private key pair is loaded, please create one by running \"cf-key\""
-        classes => kept("no_ppkeys_ABORT");
+        classes => scoped_classes_generic("namespace", "no_ppkeys_report");
```
jimis: The `kept` body makes is easier to read here IMHO. I basically read it as "if promise is kept, then ABORT". Of course I'm the one who wrote it so I'm obviously biased...

jimis:
I'm not really sure about the `repaired_two` thing, do what you consider best. ;-)

libpromises/failsafe.cf
nickanderson: It's for more precisely knowing the state.

PR #2268, opened 2015-07-06
Redmine#2504: merge ifvarclass/if FnCall and scalar correctly

tzz:
@nickanderson @kacfengine 

This fixes bug exposed by https://github.com/cfengine/masterfiles/pull/503#issuecomment-118873667

With acceptance test.

### DIFF
--- a/libpromises/policy.c
+++ b/libpromises/policy.c
@@ -1426,14 +1426,47 @@ Constraint *PromiseAppendConstraint(Promise *pp, const char *lval, Rval rval, bo
             if (strcmp(old_cp->lval, "ifvarclass") == 0 ||
                 strcmp(old_cp->lval, "if") == 0)
             {
-                Buffer *grow = BufferNew();
-                BufferAppendF(grow, "(%s).(%s)",
-                              RvalScalarValue(old_cp->rval),
-                              RvalScalarValue(rval));
-                RvalDestroy(cp->rval);
-                rval = RvalNew(BufferData(grow), RVAL_TYPE_SCALAR);
-                BufferDestroy(grow);
-                cp->rval = rval;
+                // merge two if/ifvarclass promise attributes this
+                // only happens in a variable context when we have a
+                // scalar already in the attribute (old_cp)
+                switch (rval.type)
+                {
+                case RVAL_TYPE_FNCALL: // case 1: merge FnCall with scalar
+                {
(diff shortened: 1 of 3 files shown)


nickanderson:
I can confirm that with your code the test you supplied passes and without it it fails.

jimis:
I had a look, but I'm not comfortable with the evaluator, especially function calling. I'll leave this for @kacfengine.

@tzz what does it fix? A ProgrammingError in case we use function calls and same class as the class guard?

tzz:
Yup @jimis it fixes the case where the `ifvarclass` is a FnCall, merging it with the desired one.  The old code only handled the case where it was a string.

The synthetically constructed FnCall in the merge case is `and("stringcontext", old-fn-call(...))`.

The code is not generic, it's specifically built to resolve this case.

jimis:
Thanks @tzz. A comment in the beginning of the acceptance test would be
great! :-)

tzz:
OK; added.

libpromises/policy.c
```diff
@@ -1426,14 +1426,47 @@ Constraint *PromiseAppendConstraint(Promise *pp, const char *lval, Rval rval, bo
...
+                    RlistAppendScalar(&synthetic_args, xstrdup(RvalScalarValue(old_cp->rval)));
+
+                    // append the old Rval (a function call) under the arguments of the new one
+                    RlistAppend(&synthetic_args, rval.item, RVAL_TYPE_FNCALL);
+
+                    Rval replacement = (Rval) { FnCallNew(xstrdup("and"), synthetic_args), RVAL_TYPE_FNCALL };
+                    Log(LOG_LEVEL_DEBUG, "PromiseAppendConstraint: MERGED %s rval %s", old_cp->lval, RvalToString(replacement));
+
+                    // overwrite the old Constraint rval with its replacement
+                    cp->rval = replacement;
```
kacfengine: Don't you need to ``RvalDestroy(cp->rval)`` this before assigning, like you do in the other case branch?

kacfengine:
Added ``RvalDestroy`` and running through a private branch in order to test both fixes together, this one and https://github.com/cfengine/masterfiles/pull/503.

kacfengine:
Both changes green, so merged.

kacfengine:

tzz:
Thanks for catching that memory leak and reviewing the whole thing.

PR #2269, opened 2015-07-06
simplify code

jimis:
This simplifies code and gives more informative messages. It's needed for fixing Redmine#7358.


jimis:
Passes all local tests, I'm merging in order to get a package with the needed error message for resolving Redmine#7358.

PR #2270, opened 2015-07-07
Redmine#7358: Double backslashes in statedir cause errors on Windows.

jimis:


jimis:
I think I'll just merge if it passes my local tests.

kacfengine:
+1

PR #2271, opened 2015-07-07
Redmine #7362: Fix asprintf/vasprintf being misdetected as compliant.

kacfengine:
They were not being tested as compliant, and hence were omitted from
the build simply if they were available. We fix it not by testing for
compliance, but by reusing the result from the snprintf test, it's
almost certainly going to be the same result.

This fixes Windows builds, where this is a problem on certain MinGW
versions.


jimis:
+1

PR #2273, opened 2015-07-08
libutils/mustache.c: fix {{@}} key

tzz:
Thanks to bronto for finding this.

The old behavior, when iterating across a map or array of maps, was to abort if the key was requested with `{{@}}`.

The new behavior is to always replace `{{@}}` with either the key name or the iteration position in the array.  An error is printed if `{{@}}` is used outside of a Mustache iteration section.

The acceptance tests were adjusted accordingly.

### DIFF
--- a/libutils/mustache.c
+++ b/libutils/mustache.c
@@ -343,17 +343,17 @@ static void RenderContent(Buffer *out, const char *content, size_t len, bool htm
     }
 }
 
-static bool RenderVariablePrimitive(Buffer *out, const JsonElement *primitive, const bool escaped, const bool key_mode)
+static bool RenderVariablePrimitive(Buffer *out, const JsonElement *primitive, const bool escaped, const char* json_key)
 {
-    if (key_mode && JsonElementGetPropertyName(primitive))
+    if (NULL != json_key)
     {
         if (escaped)
         {
-            RenderHTMLContent(out, JsonElementGetPropertyName(primitive), strlen(JsonElementGetPropertyName(primitive)));
+            RenderHTMLContent(out, json_key, strlen(json_key));
         }
         else
         {
-            BufferAppendString(out, JsonElementGetPropertyName(primitive));
+            BufferAppendString(out, json_key);
         }
         return true;
     }
@@ -420,7 +420,8 @@ static bool RenderVariableContainer(Buffer *out, const JsonElement *container, b
 static bool RenderVariable(Buffer *out,
                            const char *content, size_t content_len,
                            TagType conversion,
-                           Seq *hash_stack)
+                           Seq *hash_stack,
+                           const char *json_key)
 {
     JsonElement *var = NULL;
     bool escape = conversion == TAG_TYPE_VAR;
@@ -433,10 +434,25 @@ static bool RenderVariable(Buffer *out,
     if (item_mode || key_mode)
     {
         var = SeqAt(hash_stack, SeqLength(hash_stack) - 1);
+
+        // Leave this in, it's really useful when debugging here but useless otherwise
+        // for (int i=1; i < SeqLength(hash_stack); i++)
+        // {
+        //     JsonElement *dump = SeqAt(hash_stack, i);
+        //     Writer *w = StringWriter();
+        //     JsonWrite(w, dump, 0);
+        //     Log(LOG_LEVEL_ERR, "RenderVariable: at hash_stack position %d, we found var '%s'", i, StringWriterClose(w));
+        // }
     }
     else
     {
         var = LookupVariable(hash_stack, content, content_len);
+    }
+
+    if (key_mode && NULL == json_key)
+    {
+        Log(LOG_LEVEL_WARNING, "RenderVariable: {{@}} Mustache tag must be used in a context where there's a valid key or iteration position");
+        return false;
     }
 
     if (!var)
@@ -448,17 +464,18 @@ static bool RenderVariable(Buffer *out,
     {
     case JSON_ELEMENT_TYPE_PRIMITIVE:
         // note that this also covers 'serialize' on primitives
-        return RenderVariablePrimitive(out, var, escape, key_mode);
+        return RenderVariablePrimitive(out, var, escape, key_mode ? json_key : NULL);
 
     case JSON_ELEMENT_TYPE_CONTAINER:
         if (serialize || serialize_compact)
         {
             return RenderVariableContainer(out, var, serialize_compact);
         }
-        else
+        else if (key_mode)
         {
-            assert(false);
-            return false;
+            // this will only use the JSON key property, which we know is good
+            // because we return false earlier otherwise
+            return RenderVariablePrimitive(out, var, escape, json_key);
         }
     }
 
@@ -504,6 +521,7 @@ static bool SetDelimiters(const char *content, size_t content_len,
 }
 
 static bool Render(Buffer *out, const char *start, const char *input, Seq *hash_stack,
+                   const char *json_key,
                    char *delim_start, size_t *delim_start_len,
                    char *delim_end, size_t *delim_end_len,
                    bool skip_content,
@@ -563,7 +581,7 @@ static bool Render(Buffer *out, const char *start, const char *input, Seq *hash_
             {
                 if (tag.content_len > 0)
                 {
-                    if (!RenderVariable(out, tag.content, tag.content_len, tag.type, hash_stack))
+                    if (!RenderVariable(out, tag.content, tag.content_len, tag.type, hash_stack, json_key))
                     {
                         return false;
                     }
@@ -586,7 +604,7 @@ static bool Render(Buffer *out, const char *start, const char *input, Seq *hash_
                 if (!var)
                 {
                     const char *cur_section_end = NULL;
-                    if (!Render(out, start, input, hash_stack, delim_start, delim_start_len, delim_end, delim_end_len,
+                    if (!Render(out, start, input, hash_stack, NULL, delim_start, delim_start_len, delim_end, delim_end_len,
                                 skip_content || tag.type != TAG_TYPE_INVERTED, section, &cur_section_end))
                     {
                         free(section);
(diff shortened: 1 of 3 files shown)


libutils/mustache.c
```diff
@@ -447,14 +457,54 @@ static bool RenderVariable(Buffer *out,
...
+            else
+            {
+                return RenderVariablePrimitive(out, var, escape, key_mode);
+            }
+        }
+        else
+        {
+            // note that this also covers 'serialize' on primitives
+            return RenderVariablePrimitive(out, var, escape, key_mode);
+        }
```
kacfengine: You can get rid of both else blocks here and consolidate the two ``RenderVariablePrimitive`` into one.

kacfengine:
```
if (key_mode && SeqLength(hash_stack) > 1)
{
    JsonElement *parent = SeqAt(hash_stack, SeqLength(hash_stack) - 2);
    if (JsonGetElementType(parent) == JSON_ELEMENT_TYPE_CONTAINER &&
        JsonGetContainerType(parent) == JSON_CONTAINER_TYPE_ARRAY)
    {
        BufferAppendF(out, "%jd", iteration_position);
        return true;
    }
}
return RenderVariablePrimitive(out, var, escape, key_mode);
```

libutils/mustache.c
```diff
@@ -447,14 +457,54 @@ static bool RenderVariable(Buffer *out,
...
+            }
+            else
+            {
+                assert(false);
+                return false;
+            }
+        }
         else
         {
             assert(false);
```
kacfengine: Same here, you can get rid of all else blocks, since all earlier blocks return and the remaining ones will end up at the assert and return below.

kacfengine:
The details of the JSON manipulation are a bit lost on me, but it looks good AFAICS.

I don't think this will be in 3.7.0, since we already have a package which may be the final. This is anyway a new feature, so I don't consider it that critical. It will be fixed in the first patch release.

tzz:
Thanks for the review.  I've rewritten the code to be much, much simpler by passing the iteration key from the iteration context down every time.  I also added an error if the iteration key is requested but isn't available and extended the acceptance test to check that inside a double iteration, the key will be inner one.

tzz:
I also removed an extra useless return I added originally when I did the serialization logic.

As far as the JSON logic goes, JSON elements inside containers either have a property key inside an object or they have a position inside an array (the two types of containers).  In the latter case, we construct a synthetic key from the position because it's useful and consistent with the rest of CFEngine's behavior.  Then we pass that key down to the `RenderVariable` logic iff the `{{@}}` tag is specified.

tzz:
I added the ability to iterate over the top container, which was not available before.  See the PR comment at the top.  It's not really a new feature, just a way to avoid having to construct an artificial wrapper like this:

```
"wrapped" data => mergedata('{ "topkey": mycontainer }');
```

kacfengine:
I think these should be split. ``{{#~}}`` introduces new syntax which makes it a new feature and it should be discussed first. I've posted some comments in the ticket.

kacfengine:
It's easier for us to test the bug fix if it's submitted separately.

tzz:
The bug fix and the top-level feature are separate commits.  I can make a new PR of course, but let's discuss for a second.

The new syntax is invalid currently and just aborts the Mustache rendering.  So it's not going to break anyone.  It's badly needed.  It doesn't change the way data is passed or stored; only how it's **accessed**.  So I think it's literally an improvement, not a new feature in the sense that it does not create any new and exciting opportunities for crashes or bugs and does not change the **language** syntax.  So please consider the purpose of the "no new features rule" and maybe that it doesn't apply to this case.

kacfengine:
> The new syntax is invalid currently and just aborts the Mustache rendering. So it's not going to break anyone.

It's going to break for anyone who upgrades to a new patch release for some hosts, but not all, and then starts using the new feature. The point about patch releases is that they should be a 100% drop in replacement, with no change in behavior except for fixing bugs (which may of course change behavior, but that means the old behavior was broken). Feature compatibility is not just about the parser barfing on some new keyword, but to be confident that whatever you write on any version will work on all other versions of the same series, save for bugs.

I'm sure some of the concern you have is that the feature-part of this PR will be stuck in a master branch for a year or more before you can use it, but I can assure you that feature releases *will* become more frequent. As I mentioned to you in the chat, we will blog about this soon, there's just some preparatory work, among other things making sure our enterprise contracts are in line with what we announce.

tzz:
OK, top-level iteration moved to #2274 

I still think you're being too strict about treating top-level iteration as a new feature, but understand your viewpoint nevertheless.

kacfengine:
> OK, top-level iteration moved to #2274

Thanks!

> I still think you're being too strict about treating top-level iteration as a new feature, but understand your viewpoint nevertheless.

Some customers are very conservative about upgrading, so this eases their concerns at least somewhat! :-)

kacfengine:
+1 for the code cleanup btw, really nice now!

kacfengine:
trigger build

kacfengine:
Jenkins hiccup

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
This failed quite a lot of tests (several identical ones due to many platforms):

* ./00_basics/ifelapsed_and_expireafter/timed/defaults.cf.defaults.cf
* ./05_processes/01_matching/timed/stime.cf.stime.cf
* ./10_files/templating/mustache_datastate_demo.cf.mustache_datastate_demo.cf
* ./10_files/templating/mustache_datastate_demo.cf.mustache_datastate_demo.cf
* ./10_files/templating/mustache_datastate_demo.cf.mustache_datastate_demo.cf
* ./10_files/templating/mustache_datastate_demo.cf.mustache_datastate_demo.cf
* ./05_processes/01_matching/timed/stime.cf.stime.cf
* ./00_basics/ifelapsed_and_expireafter/timed/defaults.cf.defaults.cf
* ./05_processes/01_matching/timed/stime.cf.stime.cf
* ./10_files/templating/mustache_datastate_demo.cf.mustache_datastate_demo.cf
* ./05_processes/01_matching/timed/stime.cf.stime.cf
* ./00_basics/ifelapsed_and_expireafter/timed/defaults.cf.defaults.cf
* ./10_files/templating/mustache_datastate_demo.cf.mustache_datastate_demo.cf

tzz:
You'll have to look into this, unfortunately.  The templating tests pass for me and I suspect the other failures are not necessarily from my code:

```console
% testall 10_files/templating                
...
Passed tests:  15
Failed tests:  0
Skipped tests: 2
Soft failures: 0
Total tests:   17

% lsb_release -a
Distributor ID:	Ubuntu
Description:	Ubuntu 15.04
Release:	15.04
Codename:	vivid
```

kacfengine:
I also can't reproduce it. Trying again.

kacfengine:
trigger build

kacfengine:
Nope, the problem is persistent. I'll see if I can find out more.

kacfengine:
This is the output from ``./10_files/templating/mustache_datastate_demo.cf.mustache_datastate_demo.cf``

<pre>
R: Rendering template file /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_ULTRASPARC_SOLARIS_9/core/tests/acceptance/./10_files/templating/demo.datastate.mustache to /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_ULTRASPARC_SOLARIS_9/core/tests/acceptance/workdir/__10_files_templating_mustache_datastate_demo_cf/tmp/TEST.cfengine
R: "&lt;h1>Colors&lt;/h1>

 red
 green
 blue

  <p>The list is empty.</p>

 key "0" value "4"
 key "1" value "5"
 key "2" value "6"

 key "a" value "x"
 key "b" value "y"

 key "q" value "{
  "a": "x",
  "b": "y"
}"

 key "0" value "{
  "a": "x",
  "b": "y"
}"

 key "0" value "{
  "a": "x",
  "b": "y"
}"
 key "1" value "{
  "a": "x",
  "b": "y"
}"

 nested key should be "q": "q" value "{
  "a": "x",
  "b": "y"
}"
 nested key should be "q": "q" value "{
  "a": "x",
  "b": "y"
}"
" != "&lt;h1>Colors&lt;/h1>

 red
 green
 blue

  <p>The list is empty.</p>

 key "4290707968" value "4"
 key "4294967307" value "5"
 key "8589934603" value "6"

 key "a" value "x"
 key "b" value "y"

 key "q" value "{
  "a": "x",
  "b": "y"
}"

 key "4290707968" value "{
  "a": "x",
  "b": "y"
}"

 key "4290707968" value "{
  "a": "x",
  "b": "y"
}"
 key "4294967317" value "{
  "a": "x",
  "b": "y"
}"

 nested key should be "q": "q" value "{
  "a": "x",
  "b": "y"
}"
 nested key should be "q": "q" value "{
  "a": "x",
  "b": "y"
}"
"
</pre>

Note the key values which are suddenly very high, like they have wrapped over to the negative side or something. Any clues?

libutils/mustache.c
```diff
@@ -638,15 +656,28 @@ static bool Render(Buffer *out, const char *start, const char *input, Seq *hash_
...
                                 SeqAppend(hash_stack, child_hash);
 
+                                Buffer *kstring = BufferNew();
+                                if (JSON_CONTAINER_TYPE_OBJECT == JsonGetContainerType(var))
+                                {
+                                    BufferAppendString(kstring, JsonElementGetPropertyName(child_hash));
+                                }
+                                else
+                                {
+                                    BufferAppendF(kstring, "%jd", i);
```
tzz: Note number formatting here

tzz:
It feels like it might be related to that `%jd` formatting, it's the only new code and I may have the flags wrong.  Sorry I can't see the problem on my side...

kacfengine:
Indeed, it was the formatting. Changing ``%jd`` to ``%zd`` fixed the problem. I think it's only a problem on 32-bit platforms, which is why we didn't see it locally.

kacfengine:
Merged and cherry-picked to 3.7.x. Thanks!

PR #2275, opened 2015-07-10
Fix tls options to be plain strings.

jimis:
Bug slipped in from old implementation attempts, that involved these
attributes as slists or options. However the final decision was to use
plain strings. Examples:

body common control:
```
    tls_min_version => "1.0";
    tls_ciphers     => "AES256-GCM-SHA384";
```
body server control:
```
    allowtlsversion => "1.0"
    allowciphers    => "AES256-GCM-SHA384";
```


jimis:
trigger build

jimis:
Please review, but don't merge yet, as I know for sure that things are not working as they should yet.

jimis:
trigger build

jimis:
I've added some irrelevant commits, in my usual fix-as-I-go style about trivial issues that I encounter.

I honestly don't see a risk in those, but I've no problem excluding them since they are so late, just before the release.

The only irrelevant commit that I suggest to *not* exclude is f680261, which removes lexer messages from debug output. I decided to commit it to save my eyes from the pain. It's long overdue IMHO, and the parser messages are still there, but they are fewer and much clearer. Honestly, if we need lexer debug messages in the future, lets find out *in that case* a special flag to include them.

basvandervlies:
I agree that we need a special flag for the lexer debug message if needed

libpromises/cf3.defs.h
```diff
@@ -413,7 +413,7 @@ typedef enum
     COMMON_CONTROL_TLS_CIPHERS,
     COMMON_CONTROL_TLS_MIN_VERSION,
     COMMON_CONTROL_PACKAGE_INVENTORY,
-    COMMON_CONTROL_PACKAGE_MANAGER,
+    COMMON_CONTROL_PACKAGE_MODULE,
```
kacfengine: Ah, good that someone caught that!

tests/acceptance/16_cf-serverd/serial/nondefault_ciphers_tlsversion.srv
```diff
@@ -0,0 +1,37 @@
...
+}
+
+#########################################################
+# Server config
+#########################################################
+
+body server control
+
+{
+      port => "9888";
```
kacfengine: Just out of curiosity, if we made all these port numbers unique, could we run the server tests in parallel?

kacfengine:
+1 for lots of tests and fixing memory leaks!

kacfengine:
Looks good to me, and the build is green. Is it ready to merge?

tests/acceptance/16_cf-serverd/serial/nondefault_ciphers_tlsversion.srv
jimis: On Tue, Jul 14, 2015 at 10:56 AM, Kristian Amlie <notifications@github.com>
wrote:

> Just out of curiosity, if we made all these port numbers unique, could we
> run the server tests in parallel?


Not really, mainly because the same server is sometimes needed by two or
more tests.

PR #2276, opened 2015-07-12
Redmine #7030: Don't hang when operating on fifos

worr:
Implemented using `seteuid` and `setegid` per @kacfengine's comments. Had to open new pull request due to some git fudgery on my end. See #2192 for discussion. Related to #2194.

### DIFF
--- a/libcompat/fchmodat.c
+++ b/libcompat/fchmodat.c
@@ -1,0 +1,101 @@
+/*
+   Copyright (C) CFEngine AS
+
+   This file is part of CFEngine 3 - written and maintained by CFEngine AS.
+
+   This program is free software; you can redistribute it and/or modify it
+   under the terms of the GNU General Public License as published by the
+   Free Software Foundation; version 3.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#include <platform.h>
+#include <misc_lib.h>
+#include <logging.h>
+#include <generic_at.h>
+
+#include <fcntl.h>
+#include <grp.h>
+#include <errno.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#ifndef __MINGW32__
+
+typedef struct
+{
+    const char *pathname;
+    mode_t mode;
+    int dirfd;
+    int flags;
+} fchmodat_data;
+
+static int fchmodat_inner(void *generic_data)
+{
+    struct stat buf;
+    int ret = 0;
+    uid_t olduid = 0;
+
+    fchmodat_data *data = generic_data;
+    if (data->flags & AT_SYMLINK_NOFOLLOW)
+    {
+        errno = ENOTSUP;
+        return -1;
+    }
+
+    /* {l,}stat(2) is used here since we don't want to lock on the inner
+     * function
+     */
+    if ((ret = lstat(data->pathname, &buf)))
+    {
+        return ret;
+    }
+
+    /* save old euid */
+    olduid = geteuid();
+
+    if ((ret = seteuid(buf.st_uid)))
+    {
+        return ret;
+    }
+
+    ret = chmod(data->pathname, data->mode);
+
+    if (seteuid(olduid))
+    {
+        return -1;
+    }
+
+    return ret;
+}
+
+static void cleanup(ARG_UNUSED void *generic_data)
+{
+}
+
+int fchmodat(int dirfd, const char *pathname, mode_t mode, int flags)
+{
+    fchmodat_data data = {
+        .pathname = pathname,
+        .mode = mode,
+        .flags = flags,
+        .dirfd = dirfd,
+    };
+
+    return generic_at_function(dirfd, &fchmodat_inner, &cleanup, &data);
+}
+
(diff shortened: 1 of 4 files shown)


libcompat/fchmodat.c
```diff
@@ -0,0 +1,106 @@
...
+static int fchmodat_inner(void *generic_data)
+{
+    struct stat buf;
+	struct group *gr;
+    int ret = 0;
+	gid_t gid = 0, oldgid = 0;
+	uid_t olduid = 0;
+
+    fchmodat_data *data = generic_data;
+    if ((ret = fstatat(data->dirfd, data->pathname, &buf, data->flags)))
```
kacfengine: You should (and in fact, must) use ``fstat`` here, since if the underlying platform doesn't implement ``fstatat``, then it will also use ``generic_at_function`` and get stuck when trying to relock the already held ``CHDIR_LOCK``. Using ``fstat`` is safe though, since you are already inside the correct directory at this point.

kacfengine: Err, not ``fstat``, I mean ``stat`` or ``lstat``, depending on the value of ``flags``.

libcompat/fchmodat.c
```diff
@@ -0,0 +1,106 @@
...
+	oldgid = getegid();
+
+	fputs("getgrnam\n", stderr);
+	if (! (gr = getgrnam("nobody")) && ! (gr = getgrnam("nogroup")))
+		return -1;
+	gid = gr->gr_gid;
+
+	fprintf(stderr, "%d\n", gid);
+	fputs("setegid\n", stderr);
+	if ((ret = setegid(gid))) {
```
kacfengine: Using ``setegid()`` is not necessary, only the owner can change permissions, regardless of the existing permissions for the group.

libcompat/fchmodat.c
```diff
@@ -0,0 +1,106 @@
...
+        return ret;
+
+	fputs("chmod\n", stderr);
+	ret = chmod(data->pathname, data->mode);
+
+	if (seteuid(olduid))
+		return -1;
+
+	if (setegid(oldgid))
+		return -1;
```
kacfengine: Same as above, can be removed.

libcompat/fchmodat.c
kacfengine: Damn, I'm wrong again. This is tricky stuff to get right!

What we need to know is who owns the leaf, regardless of whether that is a link or something else. So always using ``lstat`` here is the correct approach. That means that noone can sneak in a link to a sensitive file, since the link would have to be owned by them, and hence would trigger the ``seteuid()`` below.

In addition, I think you should return error if ``flags`` contains ``AT_SYMLINK_NOFOLLOW``, since you cannot possibly honor it (we don't call this anywhere from our code, but it makes it future proof).

libutils/file_lib.c
```diff
@@ -736,14 +725,27 @@ int safe_chmod(const char *path, mode_t mode)
...
+
+    char *parent_dir_alloc = xstrdup(path);
+    char *leaf_alloc = xstrdup(path);
+    char *parent_dir = dirname(parent_dir_alloc);
+    char *leaf = basename(leaf_alloc);
+
+    if ((dirfd = safe_open(parent_dir, O_RDONLY)) == -1)
+        goto bad;
+
+    if ((ret = fchmodat(dirfd, leaf, mode, AT_SYMLINK_NOFOLLOW)))
```
kacfengine: ``flags`` should be ``0``, not ``AT_SYMLINK_NOFOLLOW``, since this is ``chmod``, not ``lchmod``.

kacfengine: See also comments inside ``fchmodat_inner``.

libutils/file_lib.c
```diff
@@ -453,6 +454,13 @@ int safe_open(const char *pathname, int flags, ...)
                 return -1;
             }
 
+            // We can't safely open(2) a FIFO without blocking infinitely
+            if (S_ISFIFO(stat_before.st_mode))
+            {
+                close(currentfd);
+                return -1;
+            }
```
kacfengine: I think this should be handled elsewhere, since it is not categorically wrong to open a fifo, it's just in the particular way we have been using it from our ``chown`` implementation. In fact, is this even needed anymore? The new implementations of ``safe_chmod``, ``safe_chown`` and ``safe_lchown`` won't open files anymore.

Opening fifos in general could be desirable, for example using ``readfile`` from policy, when you know that another process is at the other end.

kacfengine:
Some issues remain, but I agree with the overall approach, this is the correct way to solve it.

The issues that must be fixed:

* The indentation is all screwed up in ``libcompat/fchmodat.c``. Should be four spaces for each level. Also the coding style requires braces around every if statement, even oneliners.
* There is a lot of debug print statements left over in the same file.
* ``fstatat`` issue in ``fchmodat_inner`` (see separate comments).
* Removing ``setegid`` (see separate comments).
* Switch ``fchmodat`` parameter ``AT_SYMLINK_NOFOLLOW`` to zero (see separate comments).
* Remove FIFO detection (see separate comments).

Thanks for working on this very complicated issue!

jimis:
This is complex issue, much more complex than it appears. +1 for diving into this hole :-)

worr:
Hey,

So since the CLA is just about worked out, I've updated this diff.

I've rebased it against the current master, as well as addressed all of the other issues. @kacfengine: Can you take another look at this? Thanks!

libutils/file_lib.c
```diff
@@ -650,6 +651,32 @@ int safe_chdir(const char *path)
...
+    char *parent_dir_alloc = xstrdup(path);
+    char *leaf_alloc = xstrdup(path);
+    char *parent_dir = dirname(parent_dir_alloc);
+    char *leaf = basename(leaf_alloc);
+
+    if ((dirfd = safe_open(parent_dir, O_RDONLY)) == -1)
+        goto bad;
+
+    if ((ret = fchownat(dirfd, leaf, owner, group, flags)))
+        goto bad;
```
kacfengine: Code style (here and several other): We want braces around all if blocks, even one liners. I have fixed this for now, but remember it next time!

kacfengine:
This is looking very good. I've fixed some minor issues and resubmitted here: https://github.com/cfengine/core/pull/2390. If that gets merged I will close this one.

kacfengine:
Fixed in #2390.

PR #2277, opened 2015-07-13
fix tls_ciphers and tls_version

jimis:
This pull request appends just one commit to pull request #2275, however I submitted it separately because it is a bit risky, at this stage. Details:

     Fix tls_ciphers and tls_version in body common control.
    
    The fix came out a bit more intrusive than I wanted, and mainly deals
    with the issue that you have to parse and load all policy, and *then*
    initialize the network library.
    
    The risk here is that the loading of the policy itself might include
    network functions, e.g. selectservers(). However such functions ought to
    be simple and nothing like copy_from promises, thus having the network
    library uninitialised should be OK.


jimis:
trigger build

kacfengine:
Agree that the extra commit is risky, it's hard to predict the effects it might have. I vote for adding it to master only, and we can include it in 3.8.

jimis:
> Agree that the extra commit is risky, it's hard to predict the effects it might have. I vote for adding it to master only, and we can include it in 3.8.

I wouldn't say it's that risky, it's just a bit late to put it in. I was thinking 3.7.1.

kacfengine:
Does it fix a bug? If so, I agree. But if it's just a cleanup, I think 3.8 is the way to go, it's more important to keep 3.7 stable than to clean it up.

jimis:
On Tue, Jul 14, 2015 at 12:44 PM, Kristian Amlie <notifications@github.com>
wrote:

> Does it fix a bug? If so, I agree. But if it's just a cleanup, I think 3.8
> is the way to go, it's more important to keep 3.7 stable than to clean it
> up.
>
>
It fixes a big bug. Basically the common control options do not work
without merging it, since they are NULL before parsing is done!

jimis:
To explain it better: `GenericAgentInitialize()` is called before policy is loaded, thus common control options are not known yet. The new `GenericAgentPostLoadInit()` is called after loading is done.

kacfengine:
Ah, I see! Then I agree, we should get this into 3.7.x as well.

kacfengine:
Just thinking, would it not make sense to add the ``GenericAgentPostLoadInit()`` to all agents? I think in theory even cf-monitord can do network by assigning to variables using network functions. And regardless the function doesn't indicate that it's network specific, perhaps better to initialize it everywhere. What do you think?

jimis:
On Tue, Jul 14, 2015 at 2:20 PM, Kristian Amlie <notifications@github.com>
wrote:

> Just thinking, would it not make sense to add the
> GenericAgentPostLoadInit() to all agents? I think in theory even
> cf-monitord can do network by assigning to variables using network
> functions. And regardless the function doesn't indicate that it's network
> specific, perhaps better to initialize it everywhere. What do you think?
>

I believe it makes sense, especially because in the future this hook can be
used for various initialisations that are now spread out in the common
control evaluator code.

But I'm not expecting it's needed, because all connections initiated from
*functions* are simple TCP connections, and don't use libcfnet. If they
*do* use CFEngine protocol, then we are in trouble, as variables are
evaluated in `LoadPolicy()`, i.e. before the PostInit hook.

kacfengine:
> But I'm not expecting it's needed, because all connections initiated from *functions* are simple TCP connections, and don't use libcfnet.

What about ``remotescalar``? That one uses CFEngine protocol, doesn't it?

jimis:
On Tue, Jul 14, 2015 at 2:53 PM, Kristian Amlie <notifications@github.com>
wrote:

> What about remotescalar? That one uses CFEngine protocol, doesn't it?
>

indeed...

jimis:
trigger build

cf-serverd/cf-serverd.c
```diff
@@ -61,7 +63,9 @@ int main(int argc, char *argv[])
         policy = LoadPolicy(ctx, config);
     }
 
+    GenericAgentPostLoadInit(ctx);
```
kacfengine: Beware! There is a separate implementation of this in ``libcfexecd/win_service_exec.c``, specifically for the Windows service framework. You need to make the change there as well.

kacfengine: In Enterprise.

jimis: Right, thanks! The code is indeed tricky, I'm not sure if anything is needed to be added there. Testing will tell...

PR #2278, opened 2015-07-14
Redmine #7149: Make sure file desc. inheritance is honored on Windows.

kacfengine:
It's quite important because Windows does not allow you to do anything
to a file that is open for writing, and if the file descriptor is
inherited by a process that is long lived, you'll never be able to
delete it.

This particular case was related to cf-execd output logs, that could
not be rotated because of cf-agent inheriting a file descriptor from
cf-execd, and cf-agent then launching an unrelated and long lived
process.

Also remove our null definition of the FD_CLOEXEC flag. Since we now
have the wrapper, we don't want people to use it by mistake.

### DIFF
--- a/libutils/file_lib.h
+++ b/libutils/file_lib.h
@@ -97,6 +97,15 @@ int safe_lchown(const char *path, uid_t owner, gid_t group);
 int safe_creat(const char *pathname, mode_t mode);
 
 /**
+ * @brief Sets whether a file descriptor should be closed on
+ *        exec()/CreateProcess().
+ * @param fd      File descriptor.
+ * @param inherit Whether to enable close-on-exec or not.
+ * @return true on success, false otherwise.
+ */
+bool SetCloseOnExec(int fd, bool enable);
+
+/**
  * @brief Deletes directory path recursively. Symlinks are not followed.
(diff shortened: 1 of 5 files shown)


kacfengine:
trigger build

kacfengine:
This goes together with https://github.com/cfengine/enterprise/pull/278. The Windows build will fail without it, but I've already verified locally that it works.

libutils/file_lib.h
```diff
@@ -97,6 +97,14 @@ int safe_lchown(const char *path, uid_t owner, gid_t group);
 int safe_creat(const char *pathname, mode_t mode);
 
 /**
+ * @brief Sets whether a file descriptor should be inherited by sub processes.
+ * @param fd      File descriptor.
+ * @param inherit Whether to enable inheritance or not.
+ * @return true on success, false otherwise.
+ */
+bool SetInheritFileDescriptor(int fd, bool inherit);
```
jimis: What's the default if I don't call this function? It's a bit confusing, because it's implied that if I don't call it, the FD is not inherited.

kacfengine:
@jimis: Merge if it looks good now. This should be cherry-picked to both 3.6.x and 3.7.x.

And don't forget the Enterprise part.

PR #2280, opened 2015-07-14
Fix issue #3964 on the 3.6.x branch

riccardomurri:
See https://dev.cfengine.com/issues/3964 for details.


kacfengine:
Not cherry-picking this for 3.6.x, since it only takes critical fixes at this point.

PR #2282, opened 2015-07-14
Fix issue #3964 on the "master" branch

riccardomurri:
See https://dev.cfengine.com/issues/3964 for details.

### DIFF
--- a/libpromises/cf3.defs.h
+++ b/libpromises/cf3.defs.h
@@ -67,6 +67,7 @@
 #define CF_BLOWFISHSIZE 16
 #define CF_MAXVARSIZE 1024
 #define CF_MAXSIDSIZE 2048      /* Windows only: Max size (bytes) of security identifiers */
+#define CF_MAXFRAGMENT 19       /* abbreviate long promise names to 2*MAXFRAGMENT+3 */
 #define CF_NONCELEN (CF_BUFSIZE/16)
 #define CF_MAXLINKSIZE 256
 #define CF_PROCCOLS 16
--- a/libpromises/eval_context.c
+++ b/libpromises/eval_context.c
@@ -1546,7 +1546,7 @@ char *EvalContextStackPath(const EvalContext *ctx)
             break;
 
         case STACK_FRAME_TYPE_PROMISE_ITERATION:
-            BufferAppendAbbreviatedStr(path, frame->data.promise_iteration.owner->promiser, 19);
+            BufferAppendAbbreviatedStr(path, frame->data.promise_iteration.owner->promiser, CF_MAXFRAGMENT);
             if (i == SeqLength(ctx->stack) - 1)
             {
                 BufferAppendF(path, "[%zd]", frame->data.promise_iteration.index);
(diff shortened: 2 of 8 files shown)


libpromises/cf3.defs.h
```diff
@@ -67,6 +67,7 @@
 #define CF_BLOWFISHSIZE 16
 #define CF_MAXVARSIZE 1024
 #define CF_MAXSIDSIZE 2048      /* Windows only: Max size (bytes) of security identifiers */
+#define CF_MAXFRAGMENT 19       /* abbreviate long promise names to 2*MAXFRAGMENT+3 */
```
kacfengine: Why ``2*MAXFRAGMENT+3``? Why not just set it to the proper value right away?

libpromises/eval_context.c
```diff
@@ -1546,26 +1546,7 @@ char *EvalContextStackPath(const EvalContext *ctx)
...
-                    BufferAppendChar(path, '.');
-                    break;
-
-                default:
-                    BufferAppendChar(path, *ch);
-                    break;
-                }
-            }
-            BufferAppendChar(path, '\'');
+            BufferAppendAbbreviatedStr(path, frame->data.promise_iteration.owner->promiser, 19);
```
kacfengine: I guess you should use ``CF_MAXFRAGMENT`` here?

kacfengine:
Generally about coding style: Braces go at the same level as the if/for/while that precedes them, and only the content is indented, like this:

```
while (true)
{
    printf("Hello world\n");
}
```

libpromises/cf3.defs.h
riccardomurri: because it needs to take MAX_FRAGMENT chars from the beginning of the promise string, and MAX_FRAGMENT from the end, and insert `...` in the middle.

I think it's better to define MAX_FRAGMENT this way than `assert((MAX_FRAGMENT - 3) / 2 > 0)` later on.

riccardomurri:
Where can I find the CFEngine coding style defs for Emacs, then?

riccardomurri:
A question about updates following code review: would you like me to keep a sensible commit history at the cost of a `git push -f`, or do I just pile up small "fix this" and "fix that" commits and never ever do a forced update?

libpromises/eval_context.c
riccardomurri: Indeed. Will fix right away.

riccardomurri:
@kacfengine Never mind, I found `cfengine-code-style.el`

kacfengine:
> would you like me to keep a sensible commit history at the cost of a git push -f, or do I just pile up small "fix this" and "fix that" commits and never ever do a forced update?

Force pushes are better, so that commits are tidy.

riccardomurri:
> Force pushes are better, so that commits are tidy.

Ok, I will consolidate commit history when we're done with the code review and you're ready to merge.

Thanks!

riccardomurri:
@kacfengine I think I have addressed all the points you raised -- can you please have another look and/or merge?

estenberg:
@riccardomurri - @kacfengine will be back in a few weeks; sorry for the delay.

libpromises/cf3.defs.h
kacfengine: I see, you're right.

kacfengine:
trigger build

kacfengine:
Sorry for the delay, some badly needed vacation time. ;-)

kacfengine:
Thanks, squashed and merged and cherry-picked to 3.7.x.

PR #2283, opened 2015-07-16
Redmine #7399: Define result classes when package is already up to date

amousset:
Ref: https://dev.cfengine.com/issues/7399

### DIFF
--- a/cf-agent/verify_packages.c
+++ b/cf-agent/verify_packages.c
@@ -1780,7 +1780,8 @@ static PromiseResult SchedulePackageOp(EvalContext *ctx, const char *name, const
             }
             else
             {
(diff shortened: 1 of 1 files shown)


estenberg:
Thanks, @kacfengine can probably review the code. 

There are several known issues (like this) with the old packages promise, though, so I'd suggest looking at 3.7 for a smoother experience in general (https://docs.cfengine.com/latest/reference-promise-types-packages.html).

kacfengine:
Looks good! Will run it through Jenkins soon.

kacfengine:
trigger build

kacfengine:
Thanks, merged manually.

PR #2285, opened 2015-07-16
Redmine#7405: cfruncommand does not work if it does not contain spaces, with the TLS protocol.

jimis:


jimis:
The fix is basically one line, all the rest are tests and messaging. I've tested locally, so I'm merging.

libpromises/conversion.c
```diff
@@ -818,7 +818,7 @@ size_t CommandArg0_bound(char *dst, const char *src, size_t dst_size)
         end_delimiter = ' ';
     }
 
-    char *end = strchr(start, end_delimiter);
+    char *end = strchrnul(start, end_delimiter);
```
jimis: this is the fix ...

PR #2289, opened 2015-07-20
Don't execute cfruncommand under shell.

jimis:


jimis:
@estenberg @maciejmrowiec just in case I opened this pull request.  ;-)

PR #2290, opened 2015-07-22
Reverse mapping from IP to interface and iprange() can look by interface

tzz:
Both of the changes in this PR relate to better support for network interface inventory.  I can split them if necessary.

The `sys.ip2iface` array (name is subject to change) maps every IP to its interface name if known.

The `iprange()` function can take N interface names after the IP range in the first parameter, and those parameters constrain the IP matching to those specific interfaces.  So for instance if you have `1.2.3.4` on `eth0`, `iprange('1.2.3.4', 'lo0')` will return false, but `iprange('1.2.3.4', 'eth0')` and `iprange('1.2.3.4')` will return true.  In order to provide this, the `classes` field in the IP address list in the system context stores the interface now, if known.  If unknown, it stores the empty string `""` as it did before.

Note that also `iprange('0.0.0.0/0', 'docker0')` will return true only if the `docker0` interface exists, so it's an indirect way to check for an interface.

I will add tests and documentation if this is acceptable.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -166,6 +166,16 @@ static FnCallResult FnReturnContext(bool result)
     return FnReturn(result ? "any" : "!any");
 }
 
+static FnCallResult FnReturnContextTrue()
+{
+    return FnReturnContext(true);
+}
+
+static FnCallResult FnReturnContextFalse()
+{
+    return FnReturnContext(false);
+}
+
 static FnCallResult FnFailure(void)
 {
     return (FnCallResult) { FNCALL_FAILURE };
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -4695,14 +4695,40 @@ static FnCallResult FnCallIPRange(EvalContext *ctx, ARG_UNUSED const Policy *pol
 
     for (const Item *ip = EvalContextGetIpAddresses(ctx); ip != NULL; ip = ip->next)
     {
-        if (FuzzySetMatch(range, VIPADDRESS) == 0 ||
-            FuzzySetMatch(range, ip->name) == 0)
+        Rlist *ifaces = finalargs->next;
+        // we match on VIPADDRESS iff no interfaces were requested
+        if (FuzzySetMatch(range, VIPADDRESS) == 0 && NULL == ifaces)
         {
-            return FnReturnContext(true);
+            Log(LOG_LEVEL_DEBUG, "%s: found range %s on hostip %s, no interface", fp->name, range, ip->name);
+            return FnReturnContextTrue();
+        }
+        else if (FuzzySetMatch(range, ip->name) == 0)
+        {
+            if (NULL == ifaces) // no interfaces requested
+            {
+                Log(LOG_LEVEL_DEBUG, "%s: found range %s on IP %s, any interface", fp->name, range, ip->name);
+                return FnReturnContextTrue();
+            }
+            else // check the specific interfaces given as arguments
+            {
+                for (; ifaces != NULL; ifaces = ifaces->next)
+                {
+                    Buffer *iface = BufferNewFrom(RlistScalarValue(ifaces), strlen(RlistScalarValue(ifaces)));
+                    BufferCanonify(iface);
+                    Log(LOG_LEVEL_DEBUG, "%s: checking range %s on IP %s, interface %s", fp->name, range, ip->name, RlistScalarValue(ifaces));
+                    if (0 == strcmp(BufferData(iface), ip->classes))
+                    {
+                        Log(LOG_LEVEL_DEBUG, "%s: found range %s on IP %s, interface %s", fp->name, range, ip->name, RlistScalarValue(ifaces));
+                        BufferDestroy(iface);
+                        return FnReturnContextTrue();
+                    }
+                    BufferDestroy(iface);
+                }
+            }
         }
     }
 
-    return FnReturnContext(false);
+    return FnReturnContextFalse();
 }
 
 /*********************************************************************/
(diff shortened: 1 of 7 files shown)


estenberg:
iprange() looks good, backward compatible.

sys.iface2ip[interface_name] also useful (complementing sys.ipv4[interface_name]).

nickanderson:
+1

tzz:
I'm not sure how to test this in a portable way.  Any suggestions on writing good tests here?  I believe the functionality is correct but can't think of a way that will work across all platforms.

kacfengine:
@pasinskim: Could you take a look at this?

tzz:
For https://dev.cfengine.com/issues/7949 the second argument could also be an IPv4 address. Those are distinct from interface names so there won't be any confusion.

kacf:
This looks good to me.

As for testing, maybe you can try testing some very wide and likely to work parameters, at least for Linux. So for example:

```
iprange("8.8.8.8/31") -> False
iprange("8.8.8.8/24") -> False
iprange("8.8.8.8/31", "eth0", "eth1", "eth2", "eth3", "wlan0") -> False
iprange("8.8.8.8/24", "eth0", "eth1", "eth2", "eth3", "wlan0") -> False
iprange("0.0.0.0/0") -> True
iprange("0.0.0.0/0", "eth0", "eth1", "eth2", "eth3", "wlan0") -> True
iprange("0.0.0.0/1") OR iprange("128.0.0.0/1") -> True
iprange("0.0.0.0/1", "eth0", "eth1", "eth2", "eth3", "wlan0") OR iprange("128.0.0.0/1", "eth0", "eth1", "eth2", "eth3", "wlan0") -> True
iprange("0.0.0.0/0", "dummy") -> False
For all iprange("$(sys.ip_addresses)/32") -> True
```

Just some suggestions.

tzz:
OK, I'll use your suggestions, thanks!

tzz:
Acceptance test added with @kacf suggestions, thanks!!!

Note also my question about https://dev.cfengine.com/issues/7949 in https://github.com/cfengine/core/pull/2290#issuecomment-184890502 that would make @nickanderson happy.

kacf:
> Note also my question about https://dev.cfengine.com/issues/7949 in #2290 (comment) that would make @nickanderson happy.

Yes, that's a good idea, and I don't think there will be any ambiguity, neither an IPv4 nor an IPv6 address should ever look like an interface. I suggest you implement that in a separate PR though.

kacf:
trigger build

kacf:
This fails in `tests/unit`:

```
08:44:42 exec-config-test.c: In function ‘run_test_in_policy’:
08:44:42 exec-config-test.c:32:5: error: too few arguments to function ‘EvalContextAddIpAddress’
08:44:42 ./../../libpromises/eval_context.h:205:6: note: declared here
08:44:42 exec-config-test.c:33:5: error: too few arguments to function ‘EvalContextAddIpAddress’
08:44:42 ./../../libpromises/eval_context.h:205:6: note: declared here
08:44:42 make[4]: *** [exec-config-test.o] Error 1
08:44:42 make[4]: *** Waiting for unfinished jobs....
08:44:43 make[3]: *** [check-am] Error 2
08:44:43 make[2]: *** [check-recursive] Error 1
08:44:43 make[1]: *** [check-recursive] Error 1
08:44:43 make: *** [check] Error 2
08:44:43 make: Leaving directory `/home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_x86_64_linux_ubuntu_12/core'
```

tzz:
OK; fixed unit test, thanks for catching that.

kacf:
trigger build

kacf:
Failed on Windows and Solaris. I'm trying to figure out why.

libpromises/evalfunction.c
```diff
@@ -4670,14 +4680,36 @@ static FnCallResult FnCallIPRange(EvalContext *ctx, ARG_UNUSED const Policy *pol
...
+            {
+                Log(LOG_LEVEL_DEBUG, "%s: found range %s on IP %s, any interface", fp->name, range, ip->name);
+                return FnReturnContextTrue();
+            }
+            else // check the specific interfaces given as arguments
+            {
+                for (; ifaces != NULL; ifaces = ifaces->next)
+                {
+                    Log(LOG_LEVEL_DEBUG, "%s: checking range %s on IP %s, interface %s", fp->name, range, ip->name, RlistScalarValue(ifaces));
+                    if (0 == strcmp(RlistScalarValue(ifaces), ip->classes))
```
kacf: The reason it fails on Solaris is because the interface is named `e1000g0:4`. However, the interface name is canonified in `GetInterfacesInfo()`, but not here, leading to a false comparison. I'd fix it myself, but I'm not sure what is the right approach.

The Most Correct (tm) way is to not canonify in the first place, since this is an interface name and not a class name. However, you're (slightly) abusing the classes field of the `Item` object here to hold the interface name in a class field. Do you think there will be any ill effects of not canonifying it? IOW are we using it in a class context anywhere? I suppose we shouldn't be, since it used to be empty.

Linux can also have `eth0:1` type interfaces, so this is important there as well.

tzz: The canonification is pre-existing behavior, I don't know the right approach either. I think it's better not to store canonified but that will probably break some Albanian code. So I've amended the commit to canonify for now, but let me know if you think of a better way.

kacf: Lol! I guess we're being a bit paranoid, but yes, that sounds like the safest option.

kacf:
trigger build

kacf:
This is green, except for the Windows part, which I've got on a separate branch, I'll get that fixed tomorrow.

kacf:
You can start the docs! :-)

tzz:
docs are in https://github.com/cfengine/documentation/pull/1343

kacf:
Merged manually, thanks a lot, Ted!

PR #2291, opened 2015-07-24
Remove: Unnecessary and confusing report

nickanderson:
Ref: https://dev.cfengine.com/issues/7417

This message is confusing and I am unsure what value it provides at this 
point.  I believe this stems from the deconflation of updating from failsafe.
At one time exec_command in body executor control was by default set to run
failsafe.cf and then update.cf. I think that this report intended to warn
users about the usage. update.cf has been in place for some time, and this
policy is confusing new users. I think its best to just remove this report
completely unless someone can think of a good reason to keep it.

### DIFF
--- a/libpromises/failsafe.cf
+++ b/libpromises/failsafe.cf
@@ -173,6 +173,19 @@ bundle agent cfe_internal_update
 
   reports:
 
+    !bootstrap_mode::
+      "Built-in failsafe policy triggered"
+        handle => "cfe_internal_bootstrap_update_reports_failsafe_notification",
+	comment => "Be sure to inform the user that the failsafe policy has
+		    been triggered. This typically indicates that the agent has
+		    recieved broken policy. It may also indicate legacy
+                    configuration in body executor control.";
+
+    bootstrap_mode::
+      "Bootstrapping from host '$(sys.policy_server):5308' via built-in policy"
+        handle => "cfe_internal_bootstrap_update_reports_bootstrap_notification",
+        comment => "Be sure to inform the user that they have triggerd a bootstrap.";
+
     bootstrap_mode.am_policy_hub::
 
       "This host assumes the role of policy server"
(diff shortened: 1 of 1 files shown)


tzz:
I'd reword it and put it in `inform_mode` instead of removing it: "Bootstrapping from host $(sys.policy_server) via built-in policy on port 5308, cf-serverd protocol"

kacfengine:
In which case does CFEngine produce this warning? If the executor could not be started it seems sensible to produce some kind of error message (not necessarily this one, though).

nickanderson:
@tzz I like the suggestion, I think it makes sense as a separate report only when in bootstrap mode. (added).

@kacfengine Before my proposed change CFEngine produces the warning when execd does not get started started and when $(sys.inputdir)/promises.cf exists.

So if execd is already running it won't get started, and would produce this report. This would be the case when there is corrupt policy on the system that triggers failsafe. That's an appropirate time for a message that you are running failsafe, but it makes no sense to inform the user they should be running the update policy, afterall under that condition failsafe is triggered automatically, and the user has no way of running update.cf instead.

nickanderson:
ping @jimis thoughts on this?

libpromises/failsafe.cf
```diff
@@ -173,6 +173,19 @@ bundle agent cfe_internal_update
...
+    !bootstrap_mode::
+      "Built-in failsafe policy triggered"
+        handle => "cfe_internal_bootstrap_update_reports_failsafe_notification",
+	comment => "Be sure to inform the user that the failsafe policy has
+		    been triggered. This typically indicates that the agent has
+		    recieved broken policy. It may also indicate legacy
+                    configuration in body executor control.";
+
+    bootstrap_mode::
+      "Bootstrapping from host '$(sys.policy_server):5308' via built-in policy"
```
jimis: Maybe skip the port completely, for now at least, until it's properly configurable. I'd also give just a bit more details by saying:

"Bootstrapping from hub "..." via built-in policy failsafe.cf."

jimis:
I think it's a good initiative! That old message is showing up when you least expect it, e.g. you bootstrap and the server is already running etc.

I would also consider restructuring failsafe.cf to do a "service restart" even if the executor or server are already running. It makes sense to start a new set of processes when rebootstrapping.

nickanderson:
@jimis I will leave the service  for another time. Perhaps we should open a redmine for the service restart on bootstrap if this looks OK lets get it merged and backported.

kacfengine:

PR #2292, opened 2015-07-24
Remove: Unnecessary and confusing report

nickanderson:
Ref: https://dev.cfengine.com/issues/7417

This message is confusing and I am unsure what value it provides at this 
point.  I believe this stems from the deconflation of updating from failsafe.
At one time exec_command in body executor control was by default set to run
failsafe.cf and then update.cf. I think that this report intended to warn
users about the usage. update.cf has been in place for some time, and this
policy is confusing new users. I think its best to just remove this report
completely unless someone can think of a good reason to keep it.

(cherry picked from commit 650509831a06cc1e9727f37219ba2d50d298d170)

### DIFF
--- a/libpromises/failsafe.cf
+++ b/libpromises/failsafe.cf
@@ -173,6 +173,19 @@ bundle agent cfe_internal_update
 
   reports:
 
+    !bootstrap_mode::
+      "Built-in failsafe policy triggered"
+        handle => "cfe_internal_bootstrap_update_reports_failsafe_notification",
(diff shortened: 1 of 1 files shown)


kacfengine:
In which case does CFEngine produce this warning? If the executor could not be started it seems sensible to produce *some* kind of error message (not necessarily this one, though).

kacfengine:
Never mind, taking discussion in master PR.

kacfengine:
Cherry-picked manually.

PR #2295, opened 2015-07-30
Allow promiser to be a function call

tzz:
Fixes https://tracker.mender.io/browse/CFE-1092

This code creates a new `promiser_attribute` promise attribute.  When the promiser is a function call, that attribute is auto-filled with the fncall Rval, and the promiser itself is set to the canonified unique string version of the fncall until the right time inside `EvalContextStackPushPromiseIterationFrame`.

In `EvalContextStackPushPromiseIterationFrame` the expanded `promiser_attribute` is written back into the promiser, if it's a scalar string.

I've tried to be very careful about how this code works but it definitely will need review and, if acceptable, good tests.

The test I used is in `00_basics/01_compiler/rval-promiser.cf` and `00_basics/01_compiler/rval-promiser.cf.expected.json`.

The debugging commit, which adds debugging output to chase the rval state of the `promiser_attribute` attribute, will not be a part of the final commit, but it's really useful to trace the chain of events.

Mailing list discussion: https://groups.google.com/d/msg/dev-cfengine/kDt0tAAcUis/VCbdhqBj4HcJ

### DIFF
--- a/cf-monitord/env_monitor.c
+++ b/cf-monitord/env_monitor.c
@@ -269,7 +269,7 @@ void MonitorStartServer(EvalContext *ctx, const Policy *policy)
         Bundle *bp = PolicyAppendBundle(monitor_cfengine_policy, NamespaceDefault(), "monitor_cfengine_bundle", "agent", NULL, NULL);
         PromiseType *tp = BundleAppendPromiseType(bp, "monitor_cfengine");
 
-        pp = PromiseTypeAppendPromise(tp, "the monitor daemon", (Rval) { NULL, RVAL_TYPE_NOPROMISEE }, NULL, NULL);
+        pp = PromiseTypeAppendPromise(tp, "the monitor daemon", RvalNULL(), NULL, NULL);
     }
     assert(pp);
 
--- a/libpromises/cf3parse.y
+++ b/libpromises/cf3parse.y
@@ -37,7 +37,7 @@
 #include "string_lib.h"
 #include "logic_expressions.h"
 #include <json-yaml.h>
-
+#include "files_names.h"
 // FIX: remove
 #include "syntax.h"
 
@@ -430,10 +430,13 @@ promisee_statement:    promiser
                                    ParseError("Missing promise type declaration");
                                }
 
-                               P.currentpromise = PromiseTypeAppendPromise(P.currentstype, P.promiser,
-                                                                           RvalCopy(P.rval),
-                                                                           P.currentclasses ? P.currentclasses : "any",
-                                                                           P.currentvarclasses);
+                               P.currentpromise = PromiseTypeAppendPromiseWithPromiserAttribute(
+                                   P.currentstype, P.promiser,
+                                   RvalCopy(P.rval),
+                                   P.currentclasses ? P.currentclasses : "any",
+                                   P.currentvarclasses,
+                                   P.promiser_attribute);
+                               P.promiser_attribute = RvalNULL();
                                P.currentpromise->offset.line = CURRENT_PROMISER_LINE;
                                P.currentpromise->offset.start = P.offsets.last_string;
                                P.currentpromise->offset.context = P.offsets.last_class_id;
@@ -458,10 +461,13 @@ promiser_statement:    promiser
                                    ParseError("Missing promise type declaration");
                                }
 
-                               P.currentpromise = PromiseTypeAppendPromise(P.currentstype, P.promiser,
-                                                                (Rval) { NULL, RVAL_TYPE_NOPROMISEE },
-                                                                           P.currentclasses ? P.currentclasses : "any",
-                                                                           P.currentvarclasses);
+                               P.currentpromise = PromiseTypeAppendPromiseWithPromiserAttribute(
+                                   P.currentstype, P.promiser,
+                                   RvalNULL(),
+                                   P.currentclasses ? P.currentclasses : "any",
+                                   P.currentvarclasses,
+                                   P.promiser_attribute);
+                               P.promiser_attribute = RvalNULL();
                                P.currentpromise->offset.line = CURRENT_PROMISER_LINE;
                                P.currentpromise->offset.start = P.offsets.last_string;
                                P.currentpromise->offset.context = P.offsets.last_class_id;
@@ -483,10 +489,29 @@ promiser:              QSTRING
                                free(P.promiser);
                            }
                            P.promiser = P.currentstring;
+                           P.promiser_attribute = RvalNULL();
+
                            P.currentstring = NULL;
                            CURRENT_PROMISER_LINE = P.line_no;
                            ParserDebug("\tP:%s:%s:%s:%s:%s promiser = %s\n", P.block, P.blocktype, P.blockid, P.currenttype, P.currentclasses ? P.currentclasses : "any", P.promiser);
                        }
+
+                     | usefunction
+                       {
+                           if (P.promiser)
+                           {
+                               free(P.promiser);
+                           }
+
+                           P.promiser_attribute = RvalNew(P.currentfncall[P.arg_nesting+1], RVAL_TYPE_FNCALL);
+                           FnCallDestroy(P.currentfncall[P.arg_nesting+1]);
+                           P.currentfncall[P.arg_nesting+1] = NULL;
+                           P.promiser = RvalToString(P.promiser_attribute);
+                           CanonifyNameInPlace(P.promiser);
+                           CURRENT_PROMISER_LINE = P.line_no;
+                           ParserDebug("\tP:%s:%s:%s:%s:%s promiser = %s\n", P.block, P.blocktype, P.blockid, P.currenttype, P.currentclasses ? P.currentclasses : "any", P.promiser);
+                       }
+
                      | error
                        {
                           INSTALL_SKIP = true;
@@ -501,7 +526,7 @@ promiser:              QSTRING
                           }
                           else
                           {
-                             ParseError("Expected promiser string, got '%s'", yytext);
+                             ParseError("Expected promiser string or rval, got '%s'", yytext);
                           }
 
                           yyclearin;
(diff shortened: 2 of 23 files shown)


tzz:
@jimis @kacfengine @nickanderson this is as discussed and requested on the mailing list.

tzz:
The first two commits clean up the code and should be acceptable without the rest of this PR.

tzz:
Acceptance test added.

kacf:
What's the status of this one now that the recursive function calls have been (mostly) merged? It looks like this is still valid, but I suppose the desired implementation might look a bit different in that context. Feel free to leave this if you don't think it's important right now, I'm just wondering if I should pick it up.

tzz:
It needs to be rebased. The code is tricky so I would rather try to get my other PRs from https://github.com/cfengine/core/pulls/tzz which will merge cleanly. We can try for this one as a bonus if the others get done :)

kacf:
Sure thing.

tzz:
I rebased this but interaction with body inheritance causes some tests to fail, so I'll fix it up and update when it's ready.

tzz:
Rebased, all tests pass.

kacf:
@jimis, maybe you could look at this?

tzz:
This is out of sync again. Please help me get this merged, rebasing is a lot of unnecessary work and this has been delayed several times already. Next time I rebase, it would really help to pull it in quickly.

jimis:
We'll let you know when to rebase then. We can proceed with the review on this out-of-date branch until the last moment.

This is big and intrusive so it's not easy to deal with.

tzz:
OK. Please keep in mind we discussed it on the mailing lists and it was suggested by the developers as the best way to resolve some core deficiencies over a year ago.

libpromises/rlist.h
```diff
@@ -45,7 +45,10 @@ JsonElement *RvalContainerValue(Rval rval);
 
 const char *RvalTypeToString(RvalType type);
 
-Rval RvalNew(const void *item, RvalType type);
+Rval RvalNULL();
```
jimis: Needs to be like the following, else it will allow arguments:
```Rval RvalNULL(void);```

**Comment by jimis** [comment]
@tzz would it be easy to point to the mailing list thread? It would help to avoid asking the same questions. :-)

File: libpromises/promises.c [hunk 2]
```diff
@@ -552,14 +554,30 @@ Promise *ExpandDeRefPromise(EvalContext *ctx, const Promise *pp, bool *excluded)
...
     *excluded = false;
 
-    Rval returnval = ExpandPrivateRval(ctx, NULL, "this", pp->promiser, RVAL_TYPE_SCALAR);
-    if (!returnval.item || (strcmp(returnval.item, CF_NULL_VALUE) == 0)) 
+    const Constraint *promiser_attribute = PromiseGetConstraint(pp, "promiser_attribute");
+
+    Rval returnval = promiser_attribute ?
+        ExpandPrivateRvalDirect(ctx, NULL, "this", promiser_attribute->rval) :
+        ExpandPrivateRval(ctx, NULL, "this", pp->promiser, RVAL_TYPE_SCALAR);
+
```
jimis: So this is the key part of this pull request AFAICT. Just taking notes...

cf-monitord/env_monitor.c
```diff
@@ -269,7 +269,7 @@ void MonitorStartServer(EvalContext *ctx, const Policy *policy)
         Bundle *bp = PolicyAppendBundle(monitor_cfengine_policy, NamespaceDefault(), "monitor_cfengine_bundle", "agent", NULL, NULL);
         PromiseType *tp = BundleAppendPromiseType(bp, "monitor_cfengine");
 
-        pp = PromiseTypeAppendPromise(tp, "the monitor daemon", (Rval) { NULL, RVAL_TYPE_NOPROMISEE }, NULL, NULL);
+        pp = PromiseTypeAppendPromise(tp, "the monitor daemon", RvalNULL(), NULL, NULL);
```
jimis: All these could go in a preparatory commit, together with other stylistic stuff you have like whitespace, introducing new functions and using them in various places etc. That preparatory commit should would not change the actual  functionality of CFEngine, but will make the commit adding the actual feature really really small, which is nice. :-)

libpromises/cf3parse.y
```diff
@@ -483,10 +489,31 @@ promiser:              QSTRING
...
+                     | usefunction
+                       {
+                           if (P.promiser)
+                           {
+                               free(P.promiser);
+                           }
+
+                           Log(LOG_LEVEL_DEBUG, "Chasing promiser_attribute: fncall %s, nesting %d", P.currentfncall[P.arg_nesting+1]->name, P.arg_nesting+1);
+                           P.promiser_attribute = RvalNew(P.currentfncall[P.arg_nesting+1], RVAL_TYPE_FNCALL);
+                           Log(LOG_LEVEL_DEBUG, "Chasing promiser_attribute: resulting rval %s", RvalToString(P.promiser_attribute));
```
jimis: This `RvalToString()` leaks. The same is true for a bunch of occurences of this function. What I would suggest would be to remove the "chasing" debug messages, they are confusing.

libpromises/eval_context.c
```diff
@@ -1384,6 +1384,27 @@ Promise *EvalContextStackPushPromiseIterationFrame(EvalContext *ctx, size_t iter
 
     EvalContextStackPushFrame(ctx, StackFrameNewPromiseIteration(pexp, iter_ctx, iteration_index));
 
+    for (size_t i = 0; i < SeqLength(pexp->conlist); i++)
+    {
```
jimis: This for loop make me skeptical, it's a hot path, and it will cause a slow down for sure. Still thinking of a better approach.

libpromises/expand.c
```diff
@@ -577,6 +581,13 @@ Rlist *ExpandList(EvalContext *ctx,
 
 /*********************************************************************/
 
+Rval ExpandPrivateRvalDirect(EvalContext *ctx,
```
jimis: The introduction of this function can also go in the preparatory commit. I would just call this `ExpandPrivateRval()` and the old one, in the rare occurences that remain still, would be called like `ExpandPrivateRval(..., (Rval) {value, type})`.

libpromises/parser_state.h
```diff
@@ -59,6 +59,7 @@ typedef struct
     bool references_body;
 
     char *promiser;
+    Rval promiser_attribute;
```
jimis: Comments missing. I don't like the "promiser_attribute" name, "promiser" is opposite of "attribute" in my head. 

A better name would be "promiser_as_rval", but then it must not be a new pseudo-constraint in the syntax, just an internal field of our Promiser struct.

libpromises/mod_common.c
```diff
@@ -492,6 +492,7 @@ const ConstraintSyntax CF_COMMON_BODIES[] =
     ConstraintSyntaxNewString("if", "", "Extended classes ANDed with context", SYNTAX_STATUS_NORMAL),
     ConstraintSyntaxNewString("unless", "", "Negated 'if'", SYNTAX_STATUS_NORMAL),
     ConstraintSyntaxNewStringList("meta", "", "User-data associated with policy, e.g. key=value strings", SYNTAX_STATUS_NORMAL),
+    ConstraintSyntaxNewString("promiser_attribute", "", "Specify the promiser as an attribute", SYNTAX_STATUS_NORMAL),
```
jimis: This I don't like at all, too big of a hack. :-)

I think the pull request would work fine if that was not like that, right?

cf-monitord/env_monitor.c
tzz: I had a prep commit but rebasing it was really painful and I squashed it. I can split it out.

libpromises/promises.c
tzz: Yeah, this is the magic :)

libpromises/rlist.h
tzz: OK, thanks.

libpromises/cf3parse.y
tzz: All right, will do.

libpromises/eval_context.c
tzz: Typically this is less than 30 items so I think it will be OK.

libpromises/mod_common.c
tzz: I don't think it would work without being an attribute because of the way variable and function resolution works. You can take my PR and work from that, I guess... I just don't know all the places it could break. I'll fix it up as you suggested and you can see what you think.

libpromises/parser_state.h
tzz: There are no comments on anything else there!!! Why start now?!?! :)

The name can be adjusted with a global search and replace, but I'm not eager to change the implementation, see my other comment... too much risk IMO.

tzz:
All right, rebased and split into a prep commit plus a followup.

The second commit has two minor fixups for the prep commit (I didn't want to bother rebasing again) but otherwise it's small as you requested.

jimis:
All I can say is that the prep commit is good. :-)

nickanderson:
FYI this was discussed in the community advisory board meeting.

```
Eystein’s first reaction is that it makes sense in strings (e.g. reports: and vars:)

    Would expect it to error if you tried to use a function in place of a body name.
    Attributes that take a bundle or body should not be allowed to be replaced with a function.
    Potential future extension could cover all non-menu strings
```

So this can move forward as far as PM is concerned at the feature level.

https://github.com/cfengine/cfengine-community-advisory-board/blob/master/minutes/2016-08-25.org

tzz:
This merges cleanly right now, so from my side I have nothing to add and will wait for @jimis or another reviewer.

tzz:
I'm closing this in favor of the `with` attribute we discussed.

jimis:
Thanks Ted.
​

PR #2297, opened 2015-08-06
Redmine#7453: parse def.json vars, classes, and inputs in C

tzz:
See https://dev.cfengine.com/issues/7453

This work does the following:

* define the hard class `feature_augments` so in the masterfiles we can see if we need to load `def.json` or not (which means this can go into 3.7.1 seamlessly)
* look for `def.json` in several locations as specified in the ticket
* use the current format of `def.json`
* define a new special `def` system scope and use it
* verify and load variables, converting primitives to CFEngine strings; simple string lists to CFEngine slists; and any other data types to CFEngine data containers.  Note that all keys and values are expanded (like `data_expand` would do it).
* verify and load classes, accepting either context or classmatch regex expressions
* load inputs into `def.loaded_inputs` which is then used in `loading.c` as if it was part of the `body common control` inputs, but loaded beforehand (duplicates are OK, we already use a hash map to prevent loading them)
* move `ClassesMatching` to `eval_context.c` and change the signature a little
* move `JsonPrimitiveToString` to `json.c`
* move `DataExpandElement` to `eval_context.c` and rename it to `JsonExpandElement`
* expose `ContainerToRlist` from `iteration.h`

The behavior, after all of this, is almost the same as the current `def.json` parsing, except for the new locations and that variables and inputs are evaluated correctly.

Acceptance tests and docs upon acceptance.

### DIFF
--- a/libpromises/generic_agent.c
+++ b/libpromises/generic_agent.c
@@ -64,6 +64,7 @@
 #include <ornaments.h>
 #include <cf-windows-functions.h>
 #include <loading.h>
+#include <iteration.h>
 
 static pthread_once_t pid_cleanup_once = PTHREAD_ONCE_INIT; /* GLOBAL_T */
 
@@ -85,6 +86,27 @@ static bool MissingInputFile(const char *input_file);
 #if !defined(__MINGW32__)
 static void OpenLog(int facility);
 #endif
+
+static JsonElement *ReadJsonFile(const char *filename, LogLevel log_level)
+{
+    struct stat sb;
+    if (stat(filename, &sb) == -1)
+    {
+        Log(log_level, "Could not open JSON file %s", filename);
+        return NULL;
+    }
+
+    JsonElement *doc = NULL;
+    JsonParseError err = JsonParseFile(filename, 4096, &doc);
+
+    if (err != JSON_PARSE_OK
+        || NULL == doc)
+    {
+        Log(log_level, "Could not parse JSON file %s: %s", filename, JsonParseErrorToString(err));
+    }
+
+    return doc;
+}
 
 /*****************************************************************************/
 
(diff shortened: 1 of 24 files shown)


nickanderson:
You note that cfengine variables are dereferenced. But as this is being done by C very early what are the implications with what can be expanded?

Should I expect that only system variables and hard classes are expanded? Or only variables and classes defined in pre-eval (what I could see from cf-promises --show-vars or --show-classes)?

tzz:
> You note that cfengine variables are dereferenced. But as this is being done by C very early what are the implications with what can be expanded?

That things you define in policy won't be expanded.

> Should I expect that only system variables and hard classes are expanded? Or only variables and classes defined in pre-eval (what I could see from cf-promises --show-vars or --show-classes)?

Right, only the things available in the generic agent just after system discovery and before bootstrap checks (see the source code for the exact location).  Basically `sys`, `def`, `const`, and the hard or persistent classes, plus the time classes... exactly what `cf-promises --show-vars` and `--show-classes` do.

If we do it later, inputs and other system settings won't work well.

I actually think there will need to be a preliminary pass before anything else happens, to load things like the syslog facility if we make that controllable in `def.json`.

estenberg:
I think this will make CFEngine evaluation even harder to explain.. not sure if this is a good idea.

tzz:
Since this removes a huge amount of customization from the user side and simplifies the masterfiles (removing many lines from `def.cf`), I think it will be an improvement.  I think it doesn't complicate evaluation in any way, only adds a step between system discovery and policy evaluation.

estenberg:
We already have 6 steps documented as part of evaluation (https://docs.cfengine.com/latest/guide-language-concepts-normal-ordering.html)..  More things to manage generally makes things more complex.

tzz:
@estenberg I'll respond in the Redmine ticket.

tzz:
OK, rebased and adjusted to reflect discussion in https://dev.cfengine.com/issues/7615

* only `$(sys.inputdir)/def.json` is loaded.  The rest are commented out for now.
* load the variable scope, so the verbose message will say we loaded `def.xyz` for instance
* rename `def.loaded_inputs` to `def.augment_inputs`

Ready for review.  I'll add the tests and docs when it's acceptable.

libpromises/generic_agent.c
```diff
@@ -116,6 +138,236 @@ void MarkAsPolicyServer(EvalContext *ctx)
...
+                                const char *vkey;
+                                while ((vkey = JsonIteratorNextKey(&iter)))
+                                {
+                                    JsonElement *data = JsonObjectGet(vars, vkey);
+                                    if (JsonGetElementType(data) == JSON_ELEMENT_TYPE_PRIMITIVE)
+                                    {
+                                        char *value = JsonPrimitiveToString(data);
+                                        Log(LOG_LEVEL_VERBOSE, "Installing augments variable '%s.%s=%s' from file '%s'",
+                                            SPECIAL_SCOPE_DEF, vkey, value, BufferData(expbuf));
+                                        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_DEF, vkey, value, CF_DATA_TYPE_STRING, "source=augments");
```
kacfengine: I think this should be ``source=augments_file`` in order to follow the same naming as in masterfiles.

kacfengine: There are many more of these, my comment applies to all of them.

libpromises/generic_agent.c
```diff
@@ -116,6 +138,236 @@ void MarkAsPolicyServer(EvalContext *ctx)
...
+{
+    Buffer *filebuf = BufferNewFrom(filename, strlen(filename));
+    Buffer *expbuf = BufferNew();
+    ExpandScalar(ctx, NULL, "this", BufferData(filebuf), expbuf);
+    if (strstr(BufferData(expbuf), "/.json"))
+    {
+        Log(LOG_LEVEL_DEBUG, "Skipping augments file '%s' because it failed to expand the base filename, resulting in '%s'",
+            BufferData(filebuf),
+            BufferData(expbuf));
+    }
```
kacfengine: This function nests incredibly deeply, and is hard to read because of code running off to the right. Many of them are due to code paths not taken if there is an error, and I think this is one of the cases where it is justified to use ``goto`` in order to quickly escape from the error condition and down to the cleanup. I suggest adding a ``goto`` to this block and then letting the next block continue outside of the ``else`` block.

This goes for several of the error conditions in the file; most of them can have their ``else`` block replaced by ``goto`` down to the end. Just make sure you goto to the correct cleanup location.

Here is an example: https://github.com/cfengine/core/blob/master/libpromises/locks.c#L954-L961. A lot of ``else`` blocks are avoided that way.

kacfengine:
Code is looking promising. A few things that need to be resolved, in addition to the code comments:

1. This *must* have acceptance tests. Possibly the test might live separately in https://github.com/cfengine/masterfiles, if it needs the framework there.
2. There needs to be a corresponding PR in masterfiles to use the ``feature_augments`` to avoid loading ``def.json`` (side question: would this be bad?).
3. If this is going to have any chance of making it into 3.7, it must be a 100% drop-in replacement. IOW, CFEngine must continue functioning even if the user installs the package without changing anything in masterfiles. This relates mostly to the previous point, that nothing bad must happen if ``def.json`` is loaded both here and through masterfiles.

libpromises/generic_agent.c
tzz: OK; fixed.

tzz:
Regardng (1) I'll write the acceptance tests on the core side, it's no problem, after this is accepted as code.

I think @nickanderson should take a look at the masterfiles behavior and acceptance testing, but otherwise I agree with point (2).

I also would like this to Just Work :)

tzz:
Ready for another review.  I hope the nesting is not too bad now.

nickanderson:
I can start looking at the masterfiles stuff when I have a package with the feature. So a 3.8 package would be the first step I think.

nickanderson:
Can we get this in to master so that we can start looking at the masterfiles changes?

kacfengine:
Thanks @tzz, the nesting is much better.

@nickanderson: Do you need this in master to get started on tests/masterfiles changes? Normally we should merge everything in one go, but we can merge this early if it makes it easier to work with.

tzz:
@kacfengine are you waiting for me? I'm waiting for your approval of the code before writing acceptance tests, as usual. Then after final approval I'll write the docs and you can merge.

kacfengine:
Oh! I guess I am. I was waiting for Nick, thinking that maybe he would write the acceptance tests in masterfiles. But if you have what is needed for the tests, then this branch is ready as far as I'm concerned. It needs to be rebased, but don't worry about it now, we can bring it up to date when the tests are ready.

tzz:
I think since the point of this PR is to parse `def.json` in C, the acceptance tests should be in core.git.  So I'll add some tests here.

tzz:
OK, tests added as follows under `tests/acceptance/00_basics/def.json/`:

* `bad.cf`: verifies bad JSON will not be parsed and an error will be produced
* `classes.cf`: verifies class loading
* `inputs.cf`: verifies input augments
* `vars.cf`: verifies variable loading
* `state.cf`: loads complex variables from `def.json` and compares to an expected state.

I added a change to `testall` which, when looking at `TEST`, will also check for `TEST.def.json` and if found, will copy it to `WORKDIR/inputs/def.json`. Thus we can craft future tests on `def.json` with less fuss.

These tests pass on Linux for me. If all of the above is acceptable, I can proceed to documentation.  Let me know.

nickanderson:
Oh goody I can't wait for this.

kacfengine:
trigger build

kacfengine:
Nice, great stuff! I have sent the branch through the testing pipeline.

Any idea how we should handle masterfiles? I assume that we should detect somehow that we have already loaded the relevant sections from C, and avoid doing that in masterfiles. Ordinarily I would use the `@if` syntax, but *if* this becomes a bugfix, we may need to detect it some other way (3.7 does not support `minimum_version` with patch level, but 3.8 does).
Maybe define a class inside `def.json` that is always present and use that for detection in the masterfiles. Or we could define the class on the C level already. What do you think?

kacfengine:
@nickanderson: If you want to experiment with this, the custom build should be available at:

* Build result: http://jenkins.usdc.cfengine.com/job/testing-enterprise-pr_core/286/
* Package: http://buildcache.usdc.cfengine.com/packages/testing-enterprise-pr_core/jenkins-bootstrap-enterprise-pr_core-294/output/ (assuming it passes)

tzz:
@nickanderson @kacfengine I think the easiest way to integrate with the masterfiles is to define `feature_augments` and that would let us turn off the parsing of `def.json` **in masterfiles policy** instead of requiring `@if`. If you agree, I'll add that to the C code, it's trivial.

tzz:
Eh, I did it anyway, see `feature.c` for the trivial change. I also rebased to `origin/master`.

kacf:
Fully agree. The only comment is about the name, I think "augments" is very non-obvious to those not involved. Maybe `feature_def_json_preparsed`? Will make it easier to understand the flow in masterfiles later.

tzz:
OK, renamed and rebased.

kacf:
trigger build

kacf:
trigger build

kacf:
trigger build

kacf:
Alright, this is passing in the test farm and the code looks good. I've also had a talk with @nickanderson to analyze all angles in which it could fail if this was added to a patch release, and our conclusion is it should be pretty safe, since if `def.json` is parsed in both C and in masterfiles, it will simply be evaluated twice, but will still end up with the same value.

So I think this is good to go. The next steps are documentation and a corresponding change in masterfiles to disable policy-based `def.json` loading when `feature_def_json_preparse` is defined.

tzz:
Documentation added in https://github.com/cfengine/documentation/pull/1267 for your review.

masterfiles change is up to @nickanderson

nickanderson:
trigger build

tzz:
You don't have to redo the build, I just rebased on the master branch while testing the documentation.

nickanderson:
I did because I wanted our build system to produce a package (previous one had already aged out).
So I have just started to play with this, one thing I am wondering. 

Would it make more sense to look at augments file relative to the policy entry?

So if I do cf-agent -KIf ./test.cf it would look for def.json at ./def.json. For normal runs it would of course be sys.inputdir. 

That would be more in line with how its implemented currently in policy woudln't it?

tzz:
I think of `def.json` as a consistent replacement of `def.cf` so I don't think it should be loaded outside of the `sys.inputdir`. But it could be overridden with a universal command-line argument like `--defs=/tmp/foo.json` if you think that's useful. I can't think of any use cases for it.

kacf:
I would avoid automatic loading of such a def file outside the normal inputs directory, that's too magic and a potential security risk (if someone evaluates `/tmp/myfile.cf`, someone else can insert policy into that by making `/tmp/def.json`, assuming the original user didn't already do so).

kacf:
This is just waiting for @nickanderson to make a masterfiles change now.

tzz:
Updated to remove `feature_augments` which I left by accident, and to load relative to the policy (which means we now load `def.json` at the very end of the agent initialization, after the input file is finally set).

kacf:
Great! I'm running this through the test farm in a private branch together with the masterfiles changes.

tzz:
@kacf after community discussion the relative path won the vote. Please consider that `def.cf` used a relative path as well from the start. I hope you and the rest of the developers agree.

kacf:
Yes, that's ok. The concern I raised is not *that* important, and I think the other arguments weigh heavier.

The branch seems mostly ok in the test farm, but one of the tests exposed a completely unrelated bug, see https://github.com/cfengine/core/pull/2461. I will rerun this after I have merged that one.

kacf:
Alright the beast is in, I have merged this manually!

kacf:
Cherry-picked to 3.8.x as well, but waiting with 3.7.x as described in https://dev.cfengine.com/issues/7615#note-27.

kacf:
Thanks for the work, Ted!

tzz:
Thank you for all the review and hard work too, this was an important milestone!

PR #2299, opened 2015-08-12
log error the reason for failed connect to make it easier to diagnose why Zendesk #2126

estenberg:
Not sure if this one has unintendend consequences that will result in syslog spamming in some cases?
Was there a reason for inform mode previously (it should either be error or verbose, not inform, by the way)?

### DIFF
--- a/libcfnet/net.c
+++ b/libcfnet/net.c
@@ -381,7 +381,7 @@ int SocketConnect(const char *host, const char *port,
     int ret = getaddrinfo(host, port, &query, &response);
     if (ret != 0)
     {
-        Log(LOG_LEVEL_INFO,
+        Log(LOG_LEVEL_ERR,
               "Unable to find host '%s' service '%s' (%s)",
               host, port, gai_strerror(ret));
         if (response != NULL)


kacfengine:
I'd like @jimis to look at this one, I'm not sure under which circumstances it happens.

jimis:
I don't see the connection with zendesk#2126.

jimis:
`SocketConnect()` is a very generic function used in many contexts. Whether failure to connect is an ERROR or not depends on the context. 

I believe that the proper course of action is to find the specific caller and log ERROR message in case the function returns -1. Then the user can run with increased verbosity and find out what went wrong.

jimis:
> it should either be error or verbose, not inform, by the way

Despite not being an error in some contexts (e.g. when called from `selectservers`) it is still a failure that the user should know about. Why do you think VERBOSE is more appropriate?

estenberg:
Formally, CFEngine reserves "inform" to "changes to the system" (see cf-agent --help). So if it's not inform, it's either error or verbose. :)

As long as we don't start to spam syslog for existing installations when there are non-critical errors, let's do error. But your case with selectservers() proves that error is not the right level..

So I will close this request.

PR #2302, opened 2015-08-21
Add commands arglist attribute, augmenting args attribute

tzz:
As proposed in mailing list.  This acts **exactly** like `args` and is simply appended to the command line right after it.  An example is easiest.  See how the final command is `/bin/echo 123 a b a`.

100% backwards compatible and requires no evaluation or behavior changes.

If it's acceptable, I will write acceptance tests and docs.

```
bundle agent main
{
  commands:
      "/bin/echo" args => "123", arglist => { "a", "b", "c" };
}
```

Output:

```console
% cf-agent -KI -f test_arglist.cf
    info: Executing 'no timeout' ... '/bin/echo 123 a b c'
  notice: Q: ".../bin/echo 123 a": 123 a b c
    info: Last 1 quoted lines were generated by promiser '/bin/echo 123 a b c'
    info: Completed execution of '/bin/echo 123 a b c'
```


tzz:
This would be very convenient for users and low risk to include.  Please consider it for 3.8.

kacf:
Oh well, I guess time ran out on us regarding 3.8, but looks useful for 3.9! :-)

libpromises/locks.c
```diff
@@ -431,6 +431,11 @@ static char *BodyName(const Promise *pp)
                 continue;
             }
 
+            if (strcmp(cp->lval, "arglist") == 0)      /* Exception for arglist, by symmetry, for locking */
+            {
+                continue;
+            }
+
```
kacf: I don't quite understand why this needs to be here, but it seems preexisting, so better to follow the previous intent, I guess.

kacf:
Looks good to me! Acceptance test and docs would be nice!

tzz:
Rebased and added acceptance test

tzz:
OK, docs in https://github.com/cfengine/documentation/pull/1401

kacf:
Thanks!

kacf:
trigger build

kacf:
All green, merged manually!

tzz:
woo hoo!

PR #2303, opened 2015-08-24
action_policy warn produces errors

amousset:
The ``action_policy => warn`` allows to warn about non-kept promises without attempting to repair them, but a lot of log messages about un-kept promises in this case are ERROR messages (and sometimes INFO or VERBOSE), which can be surprising as we asked for a warning, and as it is not really an error.

This PR changes the loglevels of messages related to warn action_policy to WARNING.

### DIFF
--- a/cf-agent/acl_posix.c
+++ b/cf-agent/acl_posix.c
@@ -331,7 +331,7 @@ static int CheckPosixLinuxACEs(EvalContext *ctx, Rlist *aces, AclMethod method, 
         {
         case cfa_warn:
 
-            cfPS(ctx, LOG_LEVEL_ERR, PROMISE_RESULT_WARN, pp, a, "%s ACL on file '%s' needs to be updated", acl_type_str, file_path);
+            cfPS(ctx, LOG_LEVEL_WARNING, PROMISE_RESULT_WARN, pp, a, "%s ACL on file '%s' needs to be updated", acl_type_str, file_path);
             *result = PromiseResultUpdate(*result, PROMISE_RESULT_WARN);
             break;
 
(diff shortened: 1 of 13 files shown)


nickanderson:
Nice

jooooooon:
:+1: 

I actually think this should be considered a bug fix - using CFEngine with warn_only policy (or dry-run mode) currently spits out ERRORs, which is rather worrying, especially since what is actually meant by "ERROR" here is "We would have a made a change but actually did nothing" - more of a "did nothing, don't worry" status, really.

libpromises/verify_reports.c
```diff
@@ -92,7 +92,7 @@ PromiseResult VerifyReportPromise(EvalContext *ctx, const Promise *pp)
 
     if (a.transaction.action == cfa_warn)
     {
-        cfPS(ctx, LOG_LEVEL_VERBOSE, PROMISE_RESULT_WARN, pp, a, "Need to repair reports promise: %s", pp->promiser);
+        cfPS(ctx, LOG_LEVEL_WARNING, PROMISE_RESULT_WARN, pp, a, "Need to repair reports promise: %s", pp->promiser);
```
jooooooon: Dunno if it really makes sense to print a warning on a reports promise. For consistency, it would make sense, but from a usefulness point of view, not really.

jooooooon:
Heya, just checking in to see if there's any feedback on this? It's a pretty LOUD bug when trying to use dry-run mode... @jimis @kacfengine ?

libpromises/verify_reports.c
kacfengine: Yes, I tend to agree. The whole concept is kinda strange when applied to reports promises, but your argument makes sense.

kacfengine:
Definitely agree that it's good to clean this up. I think the one for reports promises should be left alone as @jooooooon points out, but otherwise it looks good.

amousset:
Updated the PR to remove the change on report promises.

kacfengine:
Thanks! Will run it through Jenkins shortly (busy with release builds ATM).

kacfengine:
trigger build

kacfengine:
Thanks, merged manually.

jooooooon:
Great, thanks @kacfengine!

Any chance this will be backported to 3.6/3.7? As I said above, this is a bugfix IMHO.

kacfengine:
See the ticket: https://dev.cfengine.com/issues/7274. Most likely we will, but will wait until the current patch release is out.

PR #2307, opened 2015-08-28
Add a default body action

amousset:
This PR allows to define a default body action, useful to avoid having to define explicitly an action for each promise to alter the general behavior of the agent.

For each promise:
* if it has a defined action, use it
* if not, test if there is an existing *promise_type***_default_action** body
  * if it exists, use it
  * if not, use the default values

It is similar to the generic package_method mechanism.

### DIFF
--- a/libpromises/attributes.c
+++ b/libpromises/attributes.c
@@ -30,6 +30,7 @@
 #include <logging.h>
 #include <chflags.h>
 #include <audit.h>
+#include <string_lib.h> /* StringConcatenate */
 
 #define CF_DEFINECLASSES "classes"
 #define CF_TRANSACTION   "action"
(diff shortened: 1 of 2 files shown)


amousset:
PR updated with a default body action by promise type, called *promise_type***_default_action**.

It allows for example to reproduce the behavior of the *-n* option, with all promises in warn_only except "internal" promises like reports or methods, which is useful to get all warnings about unkept promises. It also allows switching specific promise types to warn_only.

tzz:
I think this is very nice, especially in combination with the new `inherit_from` functionality.

I want to suggest, however, that it could be even more powerful if **all** body types could be used.  So the name would be `promisetype_bodytype_default` for instance.  That would support `action` bodies as `promisetype_action_default` and any other bodies.  Another example: the default `contain` body for the `commands` promise would be `commands_contain_default`.

In addition, if `promisetype` was `any`, the body could apply to all promises.  So the body `any_action_default` would provide a default action to any promise type.

jimis:
I won't comment on the policy side, @nickanderson is a better judge.

On the technical side, please include the full description in the commit message, and also a "Changelog:" entry (see HACKING.md for details).

libpromises/attributes.c
```diff
@@ -607,6 +608,22 @@ TransactionContext GetTransactionConstraints(const EvalContext *ctx, const Promi
...
     char *value;
 
+    bool has_action =
+            PromiseBundleOrBodyConstraintExists(ctx, "action", pp);
+
+    if (!has_action)
+    {
+        /* Check if we have default action. */
+        const Policy *policy = PolicyFromPromise(pp);
+        char* default_body_name = StringConcatenate(2, pp->parent_promise_type->name, "_default_action");
```
jimis: IIRC this function allocates. Is it freed at some point?

nickanderson:
This is neat, I can imagine some cases especially for auditing and testing that it could be useful.

I also like Teds suggested changes to be able to support individual body types

ncharles:
As discussed on IRC: I don't think we should have default for all body type. For instance body perms: if we are to define a default body perm, all file editions and copy will change the permissions of the file, unless stated  otherwise. I forsee a lot of headscratching around this.

@tzz had a great idea: have a specific namespace for these body default, so that they won't be created by accident, but really on purpose

peckpeck:
version with any body type default https://github.com/cfengine/core/pull/2342

PR #2309, opened 2015-08-31
Fix POSIX shell detection on Solaris.

kacfengine:
The old approach has been found not to work under some circumstances,
however this approach has been tested and found to work.


kacfengine:
trigger build

PR #2310, opened 2015-08-31
Redmine #7381: Add changelog generator script.

kacfengine:
Not putting the test in tests/unit, because it depends on Git. It
should not need to be run very often anyway.


tzz:
Sorry, I didn't see this while I was on vacation.  Good stuff, though I don't know Python well enough to comment on the code :)

PR #2312, opened 2015-09-01
Report failure to close the socket.

jimis:
I'm suspecting some CLOSE_WAIT problems, especially on windows, could be related to not properly closed sockets. So with this pull request, such failure is logged in verbose.

### DIFF
--- a/libcfnet/misc.c
+++ b/libcfnet/misc.c
@@ -25,6 +25,7 @@
 #include <cfnet.h>
 
 #include <misc_lib.h>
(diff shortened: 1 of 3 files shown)


jimis:
Do not merge, lets trigger a test build after the release is out.

kacfengine:
Do you want to merge this now? It should be quite safe, even if it turns out not to be the problem.

jimis:
trigger build

jimis:
Cherry-picking to 3.7.x for better debugging.

PR #2314, opened 2015-09-03
Add: Test case for using a list with select_class in common bundles

nickanderson:
Ref: https://dev.cfengine.com/issues/7482

### DIFF
--- a/tests/acceptance/02_classes/01_basic/common_bundle_select_class_list.cf
+++ b/tests/acceptance/02_classes/01_basic/common_bundle_select_class_list.cf
@@ -1,0 +1,32 @@
+body common control {
+      inputs => { "../../default.cf.sub" };
+      bundlesequence  => { default("$(this.promise_filename)") };
+      version => "1.0";
+}
+
+bundle common test 
+{
+  meta:
+      "test_soft_fail" string => "any",
+        meta => { "redmine7482" };
+
+  vars:
+    any::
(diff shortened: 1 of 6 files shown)


nickanderson:
Note this has to do with how pre-eval works. More details in the ticket: https://dev.cfengine.com/issues/7482

tests/acceptance/02_classes/01_basic/common_bundle_select_class_list.cf
```diff
@@ -0,0 +1,35 @@
...
+      "common_classes" slist => { "common_1", "common_2" };
+
+    common_class_selected::	
+      "class_selected"
+        string => "$(common_classes)",
+        if => "$(common_classes)";
+  
+  classes:
+      # During pre-eval classes are checked first. During that initial pass the
+      # common_classes list is undefined, and so the promise should be skipped.
```
pasinskim: This is not quite true unfortunately. It won't be evaluated during first pre-evaluation round but there are multiple and should be evaluated in subsequent.

nickanderson: So you want me to adjust to "during the first pre-eval pass classes are checked first"?

Or are you saying the test is invalid?

kacfengine:
Covered in https://github.com/cfengine/core/pull/2327.

PR #2315, opened 2015-09-03
Add: Test case for using a list with select_class in common bundles

nickanderson:
Ref: https://dev.cfengine.com/issues/7482

(cherry picked from commit f8680fb97af3acae20acfc372336c043fe72a362)


nickanderson:
Note this has to do with how pre-eval works. More details in the ticket: https://dev.cfengine.com/issues/7482

PR #2316, opened 2015-09-09
Redmine#7346: implement regex_replace

tzz:
See https://dev.cfengine.com/issues/7346

Implements `regex_replace` with no external dependencies.

Example:

```
bundle agent main
{
  vars:
      "test" string => "abcdefghij";
      "void" string => regex_replace("", $(test), "", "g");
      "void1" string => regex_replace(".", $(test), "", "g");
      "void2" string => regex_replace("", $(test), "x", "g");
      "allofitwithTEST" string => regex_replace(".*", $(test), "TEST", "");
      "nomatch" string => regex_replace("nonesuch", $(test), "TEST", "g");
      "dots_everywhere" string => regex_replace(".", $(test), "-DOT-", "g");
      "cap123" string => regex_replace("(...)", $(test), "[cap=$1]", "g");
      "simple_nocase" string => regex_replace("ABC", $(test), "ABC", "gi"); # case-insensitive
      "simple" string => regex_replace("abc", $(test), "ABC", "g");

  methods:
      "" usebundle => expected();
      "" usebundle => report();
}

bundle agent expected
{
  vars:
      "void" string => $(main.test);
      "void1" string => "";
      "void2" string => "xaxbxcxdxexfxgxhxixjx";
      "allofitwithTEST" string => "TEST";
      "nomatch" string => $(main.test);
      "dots_everywhere" string => "-DOT--DOT--DOT--DOT--DOT--DOT--DOT--DOT--DOT-";
      "cap123" string => "[cap=abc][cap=def][cap=ghi]j";
      "simple" string => "ABCdefghij";
      "simple_nocase" string => "ABCdefghij";

  methods:
      "" usebundle => report();
}

bundle agent report
{
  vars:
      "state" data => bundlestate("main");
      "state_string" string => storejson(state);

      "expected" data => bundlestate("expected");
      "expected_string" string => storejson(expected);

  reports:
      "ACTUAL: $(state_string)";
      "EXPECTED: $(expected_string)";
}
```

output:

```
R: ACTUAL: {
  "allofitwithTEST": "TEST",
  "cap123": "[cap=abc][cap=def][cap=ghi]j",
  "dots_everywhere": "-DOT--DOT--DOT--DOT--DOT--DOT--DOT--DOT--DOT--DOT-",
  "nomatch": "abcdefghij",
  "simple": "ABCdefghij",
  "simple_nocase": "ABCdefghij",
  "test": "abcdefghij",
  "void": "abcdefghij",
  "void1": "",
  "void2": "xaxbxcxdxexfxgxhxixjx"
}
R: EXPECTED: {
  "allofitwithTEST": "TEST",
  "cap123": "[cap=abc][cap=def][cap=ghi]j",
  "dots_everywhere": "-DOT--DOT--DOT--DOT--DOT--DOT--DOT--DOT--DOT-",
  "nomatch": "abcdefghij",
  "simple": "ABCdefghij",
  "simple_nocase": "ABCdefghij",
  "void": "abcdefghij",
  "void1": "",
  "void2": "xaxbxcxdxexfxgxhxixjx"
}
```

### DIFF
--- a/libutils/pcre_wrap.c
+++ b/libutils/pcre_wrap.c
@@ -1,0 +1,501 @@
+/*
+
+ * The original pcrs code came from ftp://ftp.csx.cam.ac.uk/pub/software/programming/pcre/Contrib/pcrs-0.0.3-src.tar.gz
+
+ * This is a rewrite of PCRS with the same API and behavior.
+
+ */
+
+
+
+#include <platform.h>
+
+#include <pcre_wrap.h>
+
+static int pcre_wrap_parse_perl_options(const char *optstring, int *flags);
+
+static pcre_wrap_substitute *pcre_wrap_compile_replacement(const char *replacement, int trivialflag,
+                                                           int capturecount, int *errptr);
+
+// Retain the same error messages as PCRS
+const char *pcre_wrap_strerror(const int error)
+{
+    if (error < 0)
+    {
+        switch (error)
+        {
+            /* Passed-through PCRE error: */
+        case PCRE_ERROR_NOMEMORY:     return "(pcre:) No memory";
+
+            /* Shouldn't happen unless PCRE or PCRE_WRAP bug, or user messed with compiled job: */
+        case PCRE_ERROR_NULL:         return "(pcre:) NULL code or subject or ovector";
+        case PCRE_ERROR_BADOPTION:    return "(pcre:) Unrecognized option bit";
+        case PCRE_ERROR_BADMAGIC:     return "(pcre:) Bad magic number in code";
+        case PCRE_ERROR_UNKNOWN_NODE: return "(pcre:) Bad node in pattern";
+
+            /* Can't happen / not passed: */
+        case PCRE_ERROR_NOSUBSTRING:  return "(pcre:) Fire in power supply"; 
+        case PCRE_ERROR_NOMATCH:      return "(pcre:) Water in power supply";
+
+            /* PCRE_WRAP errors: */
+        case PCRE_WRAP_ERR_NOMEM:          return "(pcre_wrap:) No memory";
+        case PCRE_WRAP_ERR_CMDSYNTAX:      return "(pcre_wrap:) Syntax error while parsing command";
+        case PCRE_WRAP_ERR_STUDY:          return "(pcre_wrap:) PCRE error while studying the pattern";
+        case PCRE_WRAP_ERR_BADJOB:         return "(pcre_wrap:) Bad job - NULL job, pattern or substitute";
+        case PCRE_WRAP_WARN_BADREF:        return "(pcre_wrap:) Backreference out of range";
+
+            /* What's that? */
+        default:  return "Unknown error";
+        }
+    }
+    /* error >= 0: No error */
+    return "(pcre_wrap:) Everything's just fine. Thanks for asking.";
+
+}
+
+static int pcre_wrap_parse_perl_options(const char *options, int *flags)
+{
+    int rc = 0;
+    *flags = 0;
+
+    if (NULL == options)
+    {
+        return 0;
+    }
+
+    size_t max = strlen(options);
+    for (size_t i = 0; i < max; i++)
+    {
+        switch(options[i])
+        {
+        case 'e': break; /* Nein! Jamais! Nunca! Well, maybe if you ask nicely... */
+        case 'g': *flags |= PCRE_WRAP_GLOBAL; break;
+        case 'i': rc |= PCRE_CASELESS; break;
+        case 'm': rc |= PCRE_MULTILINE; break;
+        case 'o': break;
+        case 's': rc |= PCRE_DOTALL; break;
+        case 'x': rc |= PCRE_EXTENDED; break;
+        case 'U': rc |= PCRE_UNGREEDY; break;
+        case 'T': *flags |= PCRE_WRAP_TRIVIAL; break;
+        default: break;
+        }
+    }
+    return rc;
+
+}
+
+static pcre_wrap_substitute *pcre_wrap_compile_replacement(const char *replacement, int trivialflag, int capturecount, int *errptr)
+{
+    int refnum = 0;
+    char *text;
+    pcre_wrap_substitute *rstruct;
+
+    /*
+     * Sanity check
+     */
+    if (NULL == replacement)
+    {
+        replacement = "";
+    }
+
+    size_t length = strlen(replacement);
+
+    rstruct = calloc(1, sizeof(pcre_wrap_substitute));
+    if (NULL == rstruct)
+    {
+        *errptr = PCRE_WRAP_ERR_NOMEM;
+        return NULL;
+    }
+
+    text = calloc(1, length + 1);
+    if (NULL == text)
+    {
+        free(rstruct);
+        *errptr = PCRE_WRAP_ERR_NOMEM;
+        return NULL;
+    }
+
+    size_t textpos = 0;
+
+    if (trivialflag)
+    {
+        text = memcpy(text, replacement, length + 1);
+        textpos = length;
+    }
+    else // non-trivial search and replace
+    {
+        int quoted = 0;
+        size_t sourcepos = 0;
+
+        while (sourcepos < length)
+        {
+            /* Quoting */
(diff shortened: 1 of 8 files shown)


nickanderson:
@kacfengine @jimis @estenberg anything in particular holding this up?

libutils/pcre_wrap.c
```diff
@@ -0,0 +1,495 @@
...
+    pcre_wrap_substitute *rstruct;
+
+    /*
+     * Sanity check
+     */
+    if (NULL == replacement)
+    {
+        replacement = "";
+    }
+    int length = (int)strlen(replacement); // intentional
```
jimis: Intentional, but what's the intent? The comment should state it. Otherwise `size_t` is preferred.

libutils/pcre_wrap.c
```diff
@@ -0,0 +1,495 @@
...
+
+    rstruct = (pcre_wrap_substitute *)calloc(1, sizeof(pcre_wrap_substitute));
```
jimis: Cast not needed.

libutils/pcre_wrap.c
```diff
@@ -0,0 +1,495 @@
...
+    if (NULL == rstruct)
+    {
+        *errptr = PCRE_WRAP_ERR_NOMEM;
+        return NULL;
+    }
+
+    text = (char *)calloc(1, length + 1);
```
jimis: Same here with unnecessary casting of `(void *)` return value and a couple of other places as well.

libutils/pcre_wrap.c
```diff
@@ -0,0 +1,495 @@
...
+        free(rstruct);
+        *errptr = PCRE_WRAP_ERR_NOMEM;
+        return NULL;
+    }
+
+    int textpos = 0;
+
+    if (trivialflag)
+    {
+        text = strncpy(text, replacement, length + 1);
```
jimis: `s/strncpy/memcpy/`

jimis:
The code generally looks good, but I can't possibly follow with all the logic in here! :-) I would just trust the code if it came with a bunch of unit tests, trying a bunch of corner cases like empty strings, backreferences, quoting etc.

libutils/pcre_wrap.c
tzz: OK, explained that we want to cap it to MAXINT

libutils/pcre_wrap.c
tzz: OK; fixed.

libutils/pcre_wrap.c
tzz: I think the original code I used for guidance did this for compatibility reasons.  Can we leave it or do you feel it must be fixed?

libutils/pcre_wrap.c
tzz: OK; changed.

tzz:
I think the acceptance test is pretty thorough.  Please suggest more tests to add there.

tzz:
PR rebased and pushed.

libutils/pcre_wrap.c
```diff
@@ -0,0 +1,495 @@
...
+
+ * This is a rewrite of PCRS with the same API and behavior.
+
+ */
+
+

+
+#include <string.h>
+#include <ctype.h>
```
jimis: Always include our `<platform.h>` on top for portability reasons, which BTW makes the standard C library includes here redundant, so you can remove them.

libutils/pcre_wrap.h
```diff
@@ -0,0 +1,95 @@
+#ifndef PCRE_WRAP_H_INCLUDED
+#define PCRE_WRAP_H_INCLUDED
+
+
+
```
jimis: Include `<platform.h>` on top also here.

jimis:
> I think the acceptance test is pretty thorough. Please suggest more tests to add there.

A couple of quoting (backslash) cases.

I still feel like unit testing would be best for `BufferSearchAndReplace()` and you can reuse `buffer_test.c` for that so it should be quite easy. But it's not a blocker since you have an acceptance test.

libutils/pcre_wrap.c
jimis: It's not a style issue, [this cast is dangerous](http://c-faq.com/malloc/mallocnocast.html). Its roots are in the pre-ANSI C era, but today it has a negative effect. So yes, please do a quick search for "alloc" and remove the casts.

libutils/pcre_wrap.c
```diff
@@ -0,0 +1,495 @@
...
+    pcre_wrap_substitute *rstruct;
+
+    /*
+     * Sanity check
+     */
+    if (NULL == replacement)
+    {
+        replacement = "";
+    }
+    int length = (int)strlen(replacement); // intentional: we want to use at most MAXINT bytes
```
jimis: If you want to use MAXINT bytes at most, cast is not the way to go. Because if `strlen()` returns greater number (i.e. if string size is 2GB) then `length` becomes negative. Extreme case, I know, but possible.

IMHO just change `length`, `textpos` and `sourcepos` to size_t and you are good to know. Except if you have something else in mind that I don't see.

tests/acceptance/01_vars/02_functions/regex_replace.cf
```diff
@@ -0,0 +1,51 @@
...
+      # no matches, original intact
+      "nomatch" string => regex_replace("nonesuch", $(test), "TEST", "g");
+      # replace any character with -DOT-
+      "dots_everywhere" string => regex_replace(".", $(test), "-DOT-", "g");
+      # capture any three characters and replace with a backreference
+      "cap123" string => regex_replace("(...)", $(test), "[cap=$1]", "g");
+      # match ABC case-insensitive and replace with ABC
+      "simple_nocase" string => regex_replace("ABC", $(test), "ABC", "gi"); # case-insensitive
+      # match abc and replace with ABC
+      "simple" string => regex_replace("abc", $(test), "ABC", "g");
```
jimis: Plus one test for source string empty. Plus one test for both source and pattern string empty.

jimis: What about `\1 \2` style backreferencing?

libutils/pcre_wrap.c
tzz: OK; done.

libutils/pcre_wrap.c
tzz: OK; done.

libutils/pcre_wrap.c
tzz: OK; done.

libutils/pcre_wrap.h
tzz: OK; done.

tests/acceptance/01_vars/02_functions/regex_replace.cf
tzz: Empty string tests added.

The `\1` backreference style was not supported, only `$1`.  I added it in a followup commit.  It has its own tests.  When a numeric backreference is requested and no such capture was made, the code replaces it with an empty string.

tzz:
I would prefer to let the CFEngine developers work on the C unit tests, I don't know that part of the core.  Please feel free to add the unit tests as you suggested after this is merged.

jimis:
This looks very good, thanks @tzz!

jimis:
trigger build

jimis:
Does it pass "make check" in your platform? It failed all over the place in our CI:
```
CCLD   str_test
./.libs/libstr.a(buffer.o): In function `BufferSearchAndReplace':
../../libutils/buffer.c:471: undefined reference to `pcre_wrap_compile'
```
You probably need to modify the unit test makefiles.

tzz:
Could you please do this as a followup commit?  If you prefer I do it, I won't be able to do it until the weekend.  Sorry, I'm just booked solid.

jimis:
> Could you please do this as a followup commit? If you prefer I do it, I won't be able to do it until the weekend. Sorry, I'm just booked solid.

Allright, I'll try to steal some time!

jimis:
Fix amended in pull request #2392. Closing!

PR #2319, opened 2015-09-16
jq integration

tzz:
see https://dev.cfengine.com/issues/7435

Here's a minimal example to test things:

```
bundle agent main
{
  vars:
      "items" data => '{"user":"stedolan","titles":["JQ Primer", "More JQ"]}';
      "items2" data => mapdata("jq", "{user, title: .titles[]}", items);
      "items2_str" string => format("%S", items2);

  reports:
      "we have item $(items)";
      "we have items2 $(items2_str)";
}
```

output:

```
R: we have item stedolan
R: we have items2 [{"title":"JQ Primer","user":"stedolan"},{"title":"More JQ","user":"stedolan"}]
```


tzz:
@jimis I'm not sure how to proceed, can you take a look?

jimis:
Argh, linking statically is not an easy job. Both CFEngine and jq are using yacc and lex it seems:
```
3rdparty/jq-1.5/parser.y:131: multiple definition of `yyerror'
jq-1.5/parser.y:145: multiple definition of `yylex'
jq-1.5/parser.c:2066: multiple definition of `yyparse'
```

In this case I don't think it's worth making them cooperate.

If I abandon this path, the only other way is to link dynamically to `libjq.so`, which means that we should also install `libjq.so`, which also means that we should prefer system's `libjq.so` if available, as I commented on my other comment.

jimis:
@tzz you think you can try patching CFEngine in order to change the lexer/parser keywords, for example ``yyerror()`` to ``cf_yyerror()``. I'm quite sure this is supported as simple as a configuration option to our lex and yacc files. If this proves easy, I can help you even further in the static build route.

In addition, can you see if there is a way to pass down options to the `./configure` execution in jq's directory? The following options shall be needed `--disable-shared --with-pic` to force static linking, and `--disable-maintainer-mode` in order to not require bison.


Dynamic linking seems to work for me, and "make install" installs both libpromises and libjq. But I'm skeptical of conflicts with systems that already have libjq installed.

tzz:
I think static linking is better for CFEngine specifically.  We don't want to lose core functionality by surprise.

tzz:
I imported the actual Git contents of jq 1.5 and get the `./configure` error:

```
configure: WARNING: no configuration information is in 3rdparty/jq-1.5
```

I had to run `setup.sh` manually inside `3rdparty/jq-1.5`.  That built the library and the executable.

I was then able to immediately build `evalfunction.c` and the rest of CFEngine without changing the configure options or doing anything else special.  I don't know what Autoconf did internally but it seemed to work: the `jv_*` functions were called properly.  So this will almost certainly require more work from @jimis but I'm unblocked to work on the actual jq integration now.

tzz:
The example at the beginning was updated to show a successful invocation.  This is very exciting!

The current data passing mechanism is inefficient, dumping the data to JSON back and forth.  I'd love for someone else to hack on that.

The function always returns an array.  I think that's OK.  The real `jq` can return multiple JSON documents so it doesn't really map here.

Still missing: acceptance tests and docs.  If the current state of the code is acceptable I can write those next.

jimis:
As I commented in the ticket, I'm not so sure we want such a huge library imported in our source tree. It is a big bunch of code to import unreviewed, and it's not really used everywhere to just trust it.

tzz:
@jimis I need your approval of the pipe approach here or in the ticket; then I will resubmit this PR and we can proceed.

libpromises/evalfunction.c
```diff
@@ -2775,6 +2847,22 @@ static FnCallResult FnCallMapData(EvalContext *ctx, ARG_UNUSED const Policy *pol
         return FnFailure();
     }
 
+    if (mapdatamode && jqmode)
+    {
+        // TODO: make this customizable
```
tzz: I need advice on how to make this customizable.

tzz:
@jimis @kacf @estenberg I didn't hear back for months, so I've rewritten this using pipes anyway.

The change is to:

* move all the pipe-related code to `pipes_unix.c` from `package_module.c` and change options and function names accordingly (@pasinskim please check). NB: this also needs whitespace cleanup badly, lots of trailing space. I did some minor code cleanups here but it was mostly just moving things.

* write an implementation that runs `/usr/bin/jq` (which should be customizable, advice welcome) with the right parameters to make the simple test above pass. It uses the pipe code mentioned above.

I can split the pipe code move from the rest of the commit if you want.

Note that this is now a very small change as far as `jq` is concerned, just a new function option. All the 3rdparty stuff is gone. So reviewing and maintaining this should be much easier.

libpromises/evalfunction.c
tzz: Maybe the `jq` options should also be customizable.

tzz: Another proposal I made in the ticket https://dev.cfengine.com/issues/7435 was to support the syntax `mapdata("|/usr/bin/jq", ...)` which would make this moot. The user decides what to pipe.

kacf:
I have asked @jimis to look at this again.

cf-agent/package_module.c
```diff
@@ -201,219 +201,6 @@ static void ParseAndLogErrorMessage(const Rlist *data)
...
-            "Package module returned with non zero return code: %d",
-            close);
-        RlistDestroy(res);
-        return -1;
-    }
-
-    *response = res;
-    return 0;
-}
-
```
jimis: @kacf you might want to review these package module changes since it is your domain. The functionality was moved to generic functions in `pipes_unix.c`.

tzz: I can break that into a separate commit if it helps.

libpromises/evalfunction.c
```diff
@@ -2724,6 +2724,74 @@ bool EvalContextVariablePutSpecialEscaped(EvalContext *ctx, SpecialScope scope,
     return EvalContextVariablePutSpecial(ctx, scope, lval, value, type, tags);
 }
 
+/*********************************************************************/
+
+static JsonElement* ExecJQ(ARG_UNUSED EvalContext *ctx, const char *basecmd, const char *script, JsonElement *container)
```
jimis: Why pass EvalContext (first arg) if it's completely unused?

tzz: An earlier version of this needed it. I will remove it after your review.

libpromises/evalfunction.c
```diff
@@ -2775,6 +2844,22 @@ static FnCallResult FnCallMapData(EvalContext *ctx, ARG_UNUSED const Policy *pol
...
     }
 
+    if (mapdatamode && jqmode)
+    {
+        // TODO: make this customizable
+        JsonElement *returnjq = ExecJQ(ctx, "/usr/bin/jq", arg_map, container);
+
+        if (NULL == returnjq)
+        {
+            Log(LOG_LEVEL_ERR, "Function '%s' did not get a valid container back from jq after passing variable '%s' and arguments '%s', sorry", fp->name, varname, arg_map);
```
jimis: This message will be seen too often, I believe, mainly because of missing `jq` executable. So maybe a simpler message, for example "Function %s failed to get output from 'jq' execution".

libpromises/evalfunction.c
jimis: I'm not sure what kind of customizations you would like to have in the `jq` command line. For start, this looks OK. The pipe thing you brought up on the ticket is obviously the most extensible, but also the ugliest, it's up to policy writers to decide ... ;-)

tzz: Well, some people may not have it in `/usr/bin` :)

Maybe it can be in `def.jq` so you can override it from `def.json`.

tzz:
I fixed those two minor issues (unused context and too-verbose log mesage).

libpromises/evalfunction.c
jimis: Hmmm you are right. Actually it would be preferable if we were only executing "jq" and having the system look it up in `PATH`. Unfortunately `cf_popen_full_duplex()` uses `execv()`, what I think would make sense would be to use `execvp()` instead. Quoting the man page:

> The execlp(), execvp(), and execvpe() functions duplicate the actions of the shell in searching for an executable file if the specified filename does not contain a slash (/) character.

Maybe change `cf_popen_full_duplex()` or duplicate it in an alternative? Do you think this would cover most customization issues?

tzz: Yeah, it should be enough for now. I just changed `cf_popen_full_duplex()` in a backwards-compatible way.

tzz:
Changed to just run `jq` instead of `/usr/bin/jq` as suggested by @jimis

libpromises/pipes_unix.c
```diff
@@ -267,11 +269,16 @@ IOData cf_popen_full_duplex(const char *command, bool capture_stderr)
         CloseChildrenFD();
         
         char **argv  = ArgSplitCommand(command);
-        if (execv(argv[0], argv) == -1)
+        if (require_full_path && execv(argv[0], argv) == -1)
```
jimis: Please keep the logic clear, something like the following or similar:
```c
if (require_full_path)
  res=execv();
else
  res=execvp();
if (res == -1)
  Log("err (exec: %s)");
```

tzz:
OK, C logic changed as @jimis suggested.

libpromises/pipes_unix.c
tzz: OK; done.

jimis:
This looks good, thanks @tzz!

jimis:
trigger build

tzz:
As discussed in Redmine and on the mailing list, I'll change this to use a `json_pipe` mode and provide a default `def.jq` variable.

As soon as #2297 is merged, I'll update this PR to set `def.jq` (because that PR carries `SPECIAL_SCOPE_DEF`). The change is ready but if I update it now, it will not build. I'll also add tests and docs then after approval.

tzz:
As discussed in Redmine and on the mailing list, I changed this to use a `json_pipe` mode and provide a default `def.jq` variabl (in a followup commit which requires #2297).

jimis:
Un-assigning from me because of lack of time. Sorry, but any other reviewer feel free to review ...

tzz:
Rebased, ready for review. Note that this was already checked by @jimis several times.

kacf:
Sorry, I haven't looked at this one. I'm trying to go from the bottom of the list unless a PR is specifically requested.

nickanderson:
Ok, lets please try to get this one in before 3.9 freeze

libpromises/evalfunction.c
```diff
@@ -2699,6 +2699,70 @@ bool EvalContextVariablePutSpecialEscaped(EvalContext *ctx, SpecialScope scope,
...
+/*********************************************************************/
+
+static JsonElement* ExecJSON_Pipe(const char *cmd, JsonElement *container)
+{
+    IOData io = cf_popen_full_duplex(cmd, false, false);
+
+    if (io.write_fd == -1 || io.read_fd == -1)
+    {
+        Log(LOG_LEVEL_INFO, "An error occurred while communicating with '%s'", cmd);
+        return NULL;
```
kacf: You need to close whichever one is non-zero, if any.

kacf: Err, not -1, I mean.

libpromises/evalfunction.c
```diff
@@ -2699,6 +2699,70 @@ bool EvalContextVariablePutSpecialEscaped(EvalContext *ctx, SpecialScope scope,
...
+    // write the container to a string
+    Writer *w = StringWriter();
+    JsonWrite(w, container, 0);
+    char *container_str = StringWriterClose(w);
+
+    if (PipeWrite(&io, container_str) != strlen(container_str))
+    {
+        Log(LOG_LEVEL_VERBOSE, "Couldn't send whole container data to '%s'.", cmd);
+        free(container_str);
+        return NULL;
```
kacf: You still need to close `io`.

libpromises/evalfunction.c
```diff
@@ -2699,6 +2699,70 @@ bool EvalContextVariablePutSpecialEscaped(EvalContext *ctx, SpecialScope scope,
...
+            cmd, close);
+    }
+
+    JsonElement *returnjq = JsonArrayCreate(5);
+
+    for (const Rlist *rp = returnlist; rp != NULL; rp = rp->next)
+    {
+        const char *data = RlistScalarValue(rp);
+        JsonElement *parsed = NULL;
+        if (JsonParse(&data, &parsed) == JSON_PARSE_OK)
```
kacf: Is this guaranteed to always work? If I understand correctly, what you get is a list of lines, but who's to say that the JSON text can't be split over several lines, leaving each line individually as incomplete/invalid JSON. Any thoughts on this?

libpromises/evalfunction.c
```diff
@@ -2745,6 +2810,21 @@ static FnCallResult FnCallMapData(EvalContext *ctx, ARG_UNUSED const Policy *pol
...
+    {
+        JsonElement *returnjson_pipe = ExecJSON_Pipe(arg_map, container);
+
+        if (NULL == returnjson_pipe)
+        {
+            Log(LOG_LEVEL_ERR, "Function %s failed to get output from 'json_pipe' execution", fp->name);
+            return FnFailure();
+        }
+
+        JsonDestroy(container);
```
kacf: The `VarNameOrInlineToJson()` semantics have changed since this was made, you'll need to use the `JsonDestroyMaybe()` here.

libpromises/evalfunction.c
tzz: Please look at `PipeWriteData()` which doesn't. This code is copied from there, so if this needs fixing, then `PipeWriteData()` needs fixing too.

libpromises/evalfunction.c
tzz: Fixed.

libpromises/evalfunction.c
tzz: The invocation command must handle this, ensuring a single-line return (`jq` does that by default). If it doesn't parse, we exit. I don't think it makes sense to do extra work to handle pretty-printed JSON.

libpromises/evalfunction.c
tzz: Oooof, yeah, thanks for catching that!!!

tzz:
Added two fixup commits and rebased.

libpromises/evalfunction.c
kacf: You're right, both are guaranteed to be closed by `cf_popen_full_duplex`, so no check is necessary.

libpromises/evalfunction.c
kacf: This seems very unexpected and unintuitive behavior to me: A tool that expects JSON from a pipe, but barfs if the JSON spans more than one line? This makes no sense, it's extremely common that JSON spans many lines. Since this functionality has become generic, not catering only to `jq`, I consider this a must.

tzz: You're right, the JSON standard specifies that whitespace includes newlines and carriage returns and we should accept them. I'll add that.

tzz:
Added multiple lines fix, acceptance test, and rebased. Ready for review before docs.

tzz:
The acceptance test will fail if `jq` is not installed. I'm not sure if you want the test to soft-fail in that case on some platforms, or to soft-fail any time `jq` is not installed. It depends on your testing setup.

tzz:
@kacf regarding the multiple lines fix, I found that there is a one-line limit of `CF_BUFSIZE` to `PipeReadData()` which can possibly corrupt single-line JSON data. But the proper fix is beyond the scope of this PR, I think, and can be done independently later. If you agree, let me know and I can open a ticket or you can go ahead and do it... thanks!

kacf:
> regarding the multiple lines fix, I found that there is a one-line limit of CF_BUFSIZE to PipeReadData() which can possibly corrupt single-line JSON data.

Hmm, can it? Isn't `buff`, which is `CF_BUFSIZE` long, added to `data`, which has no limit? I only skimmed through it quickly so I might have missed something, but that's what it looks like.

If it is a problem as you say, then I think the call should at least fail if it exceeds the `CF_BUFSIZE`, so that you don't get a proper error about exceeded capacity and not a JSON parsing error. Most likely this was never considered in the original `PipeReadData` implementation because it was only used by the package modules which are typically short.

> The acceptance test will fail if jq is not installed. I'm not sure if you want the test to soft-fail in that case on some platforms, or to soft-fail any time jq is not installed. It depends on your testing setup.

I have a simple remedy for this: Preprocess everything you need from `jq`, and then use a small shell script which simply checks that the input is what it expects and outputs the already processed `jq` output. Then you can test it anywhere, without needing `jq`, but you still get to test it as if it was present.

tzz:
> Hmm, can it? Isn't buff, which is CF_BUFSIZE long, added to data, which has no limit? I only skimmed through it quickly so I might have missed something, but that's what it looks like.

Yeah, you're right. My mistake!

> I have a simple remedy for this: Preprocess everything you need from jq, and then use a small shell script which simply checks that the input is what it expects and outputs the already processed jq output. Then you can test it anywhere, without needing jq, but you still get to test it as if it was present.

Nice, OK. I did that and pushed the test again. It now has three cases:

* passthrough: makes sure the items are passed in
* jqinput: simulates a jq invocation
* input: does several multiline inputs to test the reassembly logic

So I think this is ready.

kacf:
Perfect, thanks!

kacf:
trigger build

kacf:
`00_basics/def.json/state.cf.state.cf` fails with:

```
R: DEBUG file_make: creating /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_i386_linux_debian_4/core/tests/acceptance/workdir/__00_basics_def_json_state_cf/tmp/TESTDIR.cfengine/template with contents "{{%state}}"
R: DEBUG file_make: creating /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_i386_linux_debian_4/core/tests/acceptance/workdir/__00_basics_def_json_state_cf/tmp/TESTDIR.cfengine/actual with contents ""
R: FILES DIFFER BUT SHOULD BE THE SAME
R: CONTENTS OF /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_i386_linux_debian_4/core/tests/acceptance/workdir/__00_basics_def_json_state_cf/tmp/TESTDIR.cfengine/actual:
{
  "acl": [
    ".*mydomain note this will remain unexpanded!!!"
  ],
  "domain": "mydomain",
  "input_name_patterns": [
    ".*\\.cf",
    ".*\\.dat",
    ".*\\.txt",
    ".*\\.conf",
    ".*\\.mustache",
    ".*\\.sh",
    ".*\\.pl",
    ".*\\.py",
    ".*\\.rb",
    "cf_promises_release_id",
    ".*\\.json",
    ".*\\.yaml",
    ".*\\.js"
  ],
  "jq": "jq --compact-output --monochrome-output --ascii-output --unbuffered --sort-keys"
}

R: CONTENTS OF /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_i386_linux_debian_4/core/tests/acceptance/./00_basics/def.json/state.cf.expected.json:
{
  "acl": [
    ".*mydomain note this will remain unexpanded!!!"
  ],
  "domain": "mydomain",
  "input_name_patterns": [
    ".*\\.cf",
    ".*\\.dat",
    ".*\\.txt",
    ".*\\.conf",
    ".*\\.mustache",
    ".*\\.sh",
    ".*\\.pl",
    ".*\\.py",
    ".*\\.rb",
    "cf_promises_release_id",
    ".*\\.json",
    ".*\\.yaml",
    ".*\\.js"
  ]
}

R: --- /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_i386_linux_debian_4/core/tests/acceptance/./00_basics/def.json/state.cf.expected.json	2016-04-04 09:13:33.000000000 +0200
+++ /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_i386_linux_debian_4/core/tests/acceptance/workdir/__00_basics_def_json_state_cf/tmp/TESTDIR.cfengine/actual	2016-04-04 09:27:03.000000000 +0200
@@ -17,5 +17,6 @@
     ".*\\.json",
     ".*\\.yaml",
     ".*\\.js"
-  ]
+  ],
+  "jq": "jq --compact-output --monochrome-output --ascii-output --unbuffered --sort-keys"
 }
R: /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_i386_linux_debian_4/core/tests/acceptance/./00_basics/def.json/state.cf FAIL
R: augments policy = /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_i386_linux_debian_4/core/tests/acceptance/./00_basics/def.json/state.cf
R: domain = mydomain
```

I assume this is because `jq` was added into that namespace.

In addition, `01_vars/02_functions/mapdata_json_pipe.cf` is also failing on older Linuxes (RHEL 4 and Debian 4) as well as exotic Unixes, but I cannot tell why. There is no message at all, maybe it crashes? But then again I would have expected to see that in the output. Any ideas?

tzz:
I fixed up the last commit to include the `state.cf` fix and rebased.

I don't know why this is failing on exotics, sorry. If you can't replicate it, I'll try to spin up one of those and test there, but it will take me a while.

kacf:
I'll see if I can find anything on the problem.

kacf:
This is getting stranger by the minute: It only fails if run in parallel with other tests, not on its own. It's quite consistently failing though.

kacf:
At least I'm getting some output now:

```
----------------------------------------------------------------------
./01_vars/02_functions/mapdata_json_pipe.cf
----------------------------------------------------------------------
Broken Pipe
   error: Policy failed validation with command '"/home/jenkins/workspace/testin
g-kristian-privateBranch/label/PACKAGES_ULTRASPARC_SOLARIS_9/core/tests/acceptan
ce/workdir/__01_vars_02_functions_mapdata_json_pipe_cf/bin/cf-promises" -c "./01
_vars/02_functions/mapdata_json_pipe.cf"'
   error: CFEngine was not able to get confirmation of promises from cf-promises
, so going to failsafe
   error: CFEngine failsafe.cf: /home/jenkins/workspace/testing-kristian-private
Branch/label/PACKAGES_ULTRASPARC_SOLARIS_9/core/tests/acceptance/workdir/__01_va
rs_02_functions_mapdata_json_pipe_cf/inputs /home/jenkins/workspace/testing-kris
tian-privateBranch/label/PACKAGES_ULTRASPARC_SOLARIS_9/core/tests/acceptance/wor
kdir/__01_vars_02_functions_mapdata_json_pipe_cf/inputs/failsafe.cf
R: No public/private key pair is loaded, please create one by running "cf-key"
   error: Fatal CFEngine error: cf-agent aborted on defined class 'no_ppkeys_ABO
RT_kept'

Return code is 1.

  ==> FAIL (UNEXPECTED FAILURE)
```

*Any* idea is welcome here....

kacf:
It's clear that cf-promises is crashing somehow, but the core file is corrupted, so I'm not able to a backtrace.

jimis:
I see the `Broken pipe` message, and I recall that we are doing quite a bit of magic with stdin/out in our acceptance tests. Maybe this messes up the `jq` subprocess somehow?

tzz:
The acceptance test doesn't use `jq` itself. It just does `cat test.json` to simulate it.

kacf:
It certainly can be stdin/stdout related, but it's more likely it happens in CFEngine's own handling of those descriptors, since this test in particular makes use of them (by calling `cat`).

kacf:
I figured out the problem, see #2571 and #2572.

tzz:
Nice, thank you for working on this and finding the issue. I really appreciate it.

kacf:
Superseded by https://github.com/cfengine/core/pull/2572.

PR #2323, opened 2015-09-18
Add: Test that modules can define empty lists

nickanderson:


nickanderson:
trigger build

kacfengine:
trigger build

tests/acceptance/08_commands/01_modules/module_allows_empty_lists.cf
```diff
@@ -0,0 +1,74 @@
...
+      bundlesequence  => { default("$(this.promise_filename)") };
+      version => "1.0";
+}
+
+#######################################################
+
+bundle agent test
+{
+  meta:
+      "test_suppress_fail" string => "any",
```
kacfengine: Should use `test_soft_fail`, otherwise all our builds turn yellow. `test_suppress_fail` are for regressions and more serious bugs that we need to fix very fast, but still don't want to block the build.

I fixed this before merging.

kacfengine:
Merged manually.

PR #2327, opened 2015-09-22
Redmine #7482: select_class not working in common bundle

pasinskim:
Before evaluating select_class check if list is expandable.
If list used for selecting class is not evaluated don't set
select_class and wait for subsequent evaluation pass.


pasinskim:
@jimis @kacfengine can one of you take a look at this one, please?

libpromises/verify_classes.c
```diff
@@ -147,20 +147,34 @@ static bool SelectClass(EvalContext *ctx, const Rlist *list, const Promise *pp)
...
         return false;
     }
-    assert(list);
+    else if (count == 1) /* There is no point of computation if there is only 
+                          * one item on a list. */
+    {
+        if (IsVarList(RlistScalarValue(list)))
+        {
+            Log(LOG_LEVEL_VERBOSE,
+                "select_class: Can not expand list '%s' for setting class.",
```
nickanderson: So will this result in a message that indicates the promise is skipped because the list cannot be expanded? (I don't know if this will trigger some other log message that specifically says skipped). I think we should call it out specifically.

jimis:
Acceptance test, and changelog entry in commit message are missing.

jimis:
The logic looks good to me.

But acceptance tests for the following corner cases would make sense. `select_class` having:
* empty list
* one simple class
* more than one classes
* variable not expanding
* variable expanding to empty (is that even possible?)
* variable expanding to one class
* variable expanding to more than one class

nickanderson:
I have an acceptance test here: #2315 and #2314

nickanderson:
@jimis I added more tests: https://github.com/cfengine/core/pull/2314/files
- Test that it is skipped when it has no classes given ( select_class => { }  )
- Test that it is skipped when it is given only an empty list ( list => {}; select_class { @(list) };)
- Test that select_class works when given a single class (select_class => { "myclass" };)
- Test that select_class works when given a single class as a variable (var => "myclass"; select_class => {"$(var)"}
- Test that select_class works when iterating over a list of classes (list => { "one", "two" }; select_class => {  $(list) };)

libpromises/verify_classes.c
pasinskim: This is handled exactly the same way as you have empty list to choose from (couple lines up in the same file). There won't be explicit message that evaluation will be skipped but we are not having this for any EvalExpression case so the behavior is consistent with what we are already having.

pasinskim:
Thanks @jimis. Changelog entry added to commit message. I guess acceptance tests part is covered by @nickanderson.

jimis:
Thanks @nickanderson for the thorough acceptance tests!

@pasinskim maybe cherry-pick the commits from pull request #2314, change the soft-fail flag, and verify that they are all passing? Then we can merge this in one go, being sure that it works.

pasinskim:
@jimis @nick tests added from #2315 and #2314. Everything tested locally with some extra additions of checking if final value to be used as select_class is expanded before it is being set.

Please take a look at this once more.

jimis:
trigger build

jimis:
Green build, new changes look good, merging!

estenberg:
Please backport to 3.7 at minimum.

PR #2328, opened 2015-09-22
Fix passopenfile_test that was failing on HP-UX.

jimis:
The registered atexit() call was being called for each test, at the
exit() of the child process, and was removing the Unix domain socket at
/tmp. This happened on every child, no matter if it had setup a socket
or not, because DIALEDUP was being turned on around the beginning of the
parent process and never turned off, despite the file being deleted by
the child.

The solution is to clean up the Unix domain socket file after every
test, both for child (via atexit()) and parent (manually) processes. The
cleanup is accompanied by setting DIALEDUP=false, thus the next spawned
child has correct state and won't try to remove a file that hasn't
created itself.

### DIFF
--- a/tests/unit/passopenfile_test.c
+++ b/tests/unit/passopenfile_test.c
@@ -31,18 +31,46 @@ static bool DIALEDUP = false;
  * connect()ed to by the same user; likewise by one in its group.
  * Probably put that test in some other file ! */
 
+
+/* Ensure no stray child is running. */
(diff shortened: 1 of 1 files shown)


jimis:
trigger build

jimis:
Triggered build is green, merging so that testing-enterprise-master is green by the morning.

kacfengine:
Nice!

PR #2329, opened 2015-09-24
Don't try to call readlink() on regular files as part of files_select

estenberg:
Redmine #6996 (also has acceptance test).

### DIFF
--- a/cf-agent/files_select.c
+++ b/cf-agent/files_select.c
@@ -489,9 +489,9 @@ static int SelectIsSymLinkTo(EvalContext *ctx, char *filename, Rlist *crit)
         {
             // Do nothing.
         }
-        else if (S_ISDIR(statbuf.st_mode))
+        else if (!S_ISLNK(statbuf.st_mode))
         {
-            Log(LOG_LEVEL_DEBUG, "Skipping readlink() on directory %s", filename);
+            Log(LOG_LEVEL_DEBUG, "Skipping readlink() on non-symlink %s", filename);
             return false;
         }
 
--- a/tests/acceptance/10_files/zendesk_1776/zendesk_1776.cf
+++ b/tests/acceptance/10_files/zendesk_1776/zendesk_1776.cf
@@ -22,10 +22,6 @@ body common control
 bundle agent check
 {
   meta:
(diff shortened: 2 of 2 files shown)


cf-agent/files_select.c
```diff
@@ -494,6 +494,11 @@ static int SelectIsSymLinkTo(EvalContext *ctx, char *filename, Rlist *crit)
             Log(LOG_LEVEL_DEBUG, "Skipping readlink() on directory %s", filename);
             return false;
         }
+        else if (S_ISREG(statbuf.st_mode))
```
kacfengine: I would rather change consolidate both of these into this:

```
else if (!S_ISLNK(statbuf.st_mode))
{
    Log(LOG_LEVEL_DEBUG, "Skipping readlink() on non-symlink %s", filename);
    return false;
}
```

Then you catch other error cases such as device files, sockets and fifos. We've been getting several fixes lately for these cases by @worr, for example.

kacfengine:
You're missing a Changelog entry for this. I strongly urge you to run ``core/misc/githooks/install-hooks`` since it will install Git hooks that will remind you.

When amending this commit you'll have to add it manually though, since the commit message is preexisting. It takes this form:

```
Don't try to call readlink() on regular files as part of files_select
Redmine #6996.

Changelog: <message>
```

Remember that changelog messages are user facing, so you should describe how the behavior has changed, not what you did to fix it.

estenberg:
OK, thanks, updated.

estenberg:
Acceptance test: #2114

kacfengine:
Ah, does this fix the acceptance test? Then you should remove the ``"test_soft_fail"`` portion of that test in the same commit, because it inverts the pass/fail relationship, so if it starts passing, the result is fail. :-) This is so that we learn of tests that start passing and mark them appropriately.

estenberg:
Alright, trying again.

kacfengine:
Thanks!

kacfengine:
trigger build

kacfengine:
Thanks, merged manually.

kacfengine:
Cherry-picked to 3.6.x and 3.7.x as well.

PR #2330, opened 2015-09-24
Change: Soften obsoletion message for old package promises

nickanderson:
Ref: https://dev.cfengine.com/issues/7603

### DIFF
--- a/cf-agent/verify_packages.c
+++ b/cf-agent/verify_packages.c
@@ -201,7 +201,7 @@ PromiseResult VerifyPackagesPromise(EvalContext *ctx, const Promise *pp)
             break;
         case PACKAGE_PROMISE_TYPE_OLD:
             Log(LOG_LEVEL_VERBOSE,
-                "Using old package promise. Please note that this will be obsolete in 3.8");
+                "Using old package promise. Please note that the old implementation is considered obsolete as of 3.8. The old implementation will continue to work, but forward development will be directed toward the new implementation.");
 
             result = HandleOldPackagePromiseType(ctx, pp, a);
         


nickanderson:
Should be backported to 3.7.x if accepted.

cf-agent/verify_packages.c
```diff
@@ -201,7 +201,7 @@ PromiseResult VerifyPackagesPromise(EvalContext *ctx, const Promise *pp)
             break;
         case PACKAGE_PROMISE_TYPE_OLD:
             Log(LOG_LEVEL_VERBOSE,
-                "Using old package promise. Please note that this will be obsolete in 3.8");
+                "Using old package promise. Please note that the old implementation is considered obsolete as of 3.8. The old implementation will continue to work, but forward development will be directed toward the new implementation.");
```
kacfengine: Very long line, should be split.

Also, don't say that it will continue to work, say that it is deprecated and may be removed in a future version, since that is what we ultimately want to do, and we don't want to give people the impression it will be around forever.

kacfengine: OTOH, the number of unsupported package managers by the new package promise does make a convincing case to keep it around a bit longer. Maybe disregard my comment for now. We should make it clear once we have a deprecation path though.

estenberg: Yes, the thing is that packages is a really core functionality of config management, so we should expect that there will be *lots* of problems caused if we remove it too fast.

basvandervlies: The nice thing about the old package method is that we do need external modules for the package installation.  With the external modules you create dependencies on the language it is written in. For now it's python and there can also be problems with version python version 2 and 3.  The executable python can be installed in different locations.  

I agree with @kacfengine with the new package module there is only support for yum/apt.  My colleague that asked the question about windows was worried about the statement that his package statements were deprecated.

kacfengine: Alright, I simply split the line and merged as is.

kacfengine:

PR #2331, opened 2015-09-25
this.callers and this.callers_promisers

tzz:
This PR takes #2301 (rebased against master) and also adds `this.callers` which contains the full call stack.  I haven't heard from @amousset in a while on #2301 so I think it's justified to continue the work in their absence.

I am positive this is useful when debugging policies.  In will also help create reusable bundles that can sniff out information about their callers.  I think it will also enable other neat use cases for frameworks and libraries.

Demo:

```
body common control
{
        inputs => { "$(sys.libdir)/stdlib.cf" };
}

bundle agent main
{
  vars:
      "callers" data => mergedata("this.callers");
      "promisers" slist => { @(this.callers_promisers) };

  methods:
      "" usebundle => one(x,y,z);
      "" usebundle => report();
}

bundle agent one(a,b,c)
{
  vars:
      "callers" data => mergedata("this.callers");
      "promisers" slist => { @(this.callers_promisers) };

  methods:
      "two" usebundle => two;
}

bundle agent two
{
  vars:
      "callers" data => mergedata("this.callers");
      "promisers" slist => { @(this.callers_promisers) };
}

bundle agent report
{
  vars:
      "state" data => datastate();
      "state2" data => mergedata('{ "main": state[vars][main] }','{ "one": state[vars][one] }', '{ "two": state[vars][two] }');

  methods:
      "" usebundle => file_make("$(this.promise_filename).mustache", "{{%.}}");

  files:
      "/tmp/dump.json"
        edit_template => "$(this.promise_filename).mustache",
        create => "true",
        template_data => mergedata(state2),
        template_method => "mustache";
}
```

Output (note it's truncated! the full stack would be better dumped in a file):

```
{
  "main": {
    "callers": [
      {
        "bundle": {
          "arguments": [],
          "bundleType": "agent",
          "line": 6,
          "name": "main",
          "namespace": "default",
          "promiseTypes": [
            {
              "contexts": [
                {
                  "name": "any",
                  "promises": [
                    {
                      "attributes": [
                        {
                          "line": 9,
                          "lval": "data",
                          "rval": {
                            "arguments": [
                              {
                                "type": "string",
                                "value": "this.callers"
                              }
                            ],
                            "name": "mergedata",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 9,
                      "promiser": "callers"
                    },
                    {
                      "attributes": [
                        {
                          "line": 10,
                          "lval": "slist",
                          "rval": {
                            "type": "list",
                            "value": [
                              {
                                "type": "string",
                                "value": "@(this.callers_promisers)"
                              }
                            ]
                          }
                        }
                      ],
                      "line": 10,
                      "promiser": "promisers"
                    }
                  ]
                }
              ],
              "line": 8,
              "name": "vars"
            },
            {
              "contexts": [
                {
                  "name": "any",
                  "promises": [
                    {
                      "attributes": [
                        {
                          "line": 13,
                          "lval": "usebundle",
                          "rval": {
                            "arguments": [
                              {
                                "type": "string",
                                "value": "x"
                              },
                              {
                                "type": "string",
                                "value": "y"
                              },
                              {
                                "type": "string",
                                "value": "z"
                              }
                            ],
                            "name": "one",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 13,
                      "promiser": ""
                    },
                    {
                      "attributes": [
                        {
                          "line": 14,
                          "lval": "usebundle",
                          "rval": {
                            "arguments": [],
                            "name": "report",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 14,
                      "promiser": ""
                    }
                  ]
                }
              ],
              "line": 12,
              "name": "methods"
            }
          ],
          "sourcePath": "/home/tzz/sync/cf/test/test_callers.cf"
        },
        "depth": 0,
        "frame": 3,
        "type": "bundle"
      },
      {
        "depth": 1,
        "frame": 2,
        "promise_type": "vars",
        "type": "promise_type"
      },
      {
        "depth": 2,
        "frame": 1,
        "promise_classes": "any",
        "promise_comment": "",
        "promise_type": "vars",
        "promiser": "callers",
        "type": "promise"
      }
    ],
    "promisers": [
      "@(this.callers_promisers)"
    ]
  },
  "one": {
    "callers": [
      {
        "bundle": {
          "arguments": [],
          "bundleType": "agent",
          "line": 6,
          "name": "main",
          "namespace": "default",
          "promiseTypes": [
            {
              "contexts": [
                {
                  "name": "any",
                  "promises": [
                    {
                      "attributes": [
                        {
                          "line": 9,
                          "lval": "data",
                          "rval": {
                            "arguments": [
                              {
                                "type": "string",
                                "value": "this.callers"
                              }
                            ],
                            "name": "mergedata",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 9,
                      "promiser": "callers"
                    },
                    {
                      "attributes": [
                        {
                          "line": 10,
                          "lval": "slist",
                          "rval": {
                            "type": "list",
                            "value": [
                              {
                                "type": "string",
                                "value": "@(this.callers_promisers)"
                              }
                            ]
                          }
                        }
                      ],
                      "line": 10,
                      "promiser": "promisers"
                    }
                  ]
                }
              ],
              "line": 8,
              "name": "vars"
            },
            {
              "contexts": [
                {
                  "name": "any",
                  "promises": [
                    {
                      "attributes": [
                        {
                          "line": 13,
                          "lval": "usebundle",
                          "rval": {
                            "arguments": [
                              {
                                "type": "string",
                                "value": "x"
                              },
                              {
                                "type": "string",
                                "value": "y"
                              },
                              {
                                "type": "string",
                                "value": "z"
                              }
                            ],
                            "name": "one",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 13,
                      "promiser": ""
                    },
                    {
                      "attributes": [
                        {
                          "line": 14,
                          "lval": "usebundle",
                          "rval": {
                            "arguments": [],
                            "name": "report",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 14,
                      "promiser": ""
                    }
                  ]
                }
              ],
              "line": 12,
              "name": "methods"
            }
          ],
          "sourcePath": "/home/tzz/sync/cf/test/test_callers.cf"
        },
        "depth": 0,
        "frame": 7,
        "type": "bundle"
      },
      {
        "depth": 1,
        "frame": 6,
        "promise_type": "methods",
        "type": "promise_type"
      },
      {
        "depth": 2,
        "frame": 5,
        "promise_classes": "any",
        "promise_comment": "",
        "promise_type": "methods",
        "promiser": "",
        "type": "promise"
      },
      {
        "depth": 3,
        "frame": 4,
        "iteration_index": 0,
        "type": "iteration"
      },
      {
        "bundle": {
          "arguments": [
            "a",
            "b",
            "c"
          ],
          "bundleType": "agent",
          "line": 17,
          "name": "one",
          "namespace": "default",
          "promiseTypes": [
            {
              "contexts": [
                {
                  "name": "any",
                  "promises": [
                    {
                      "attributes": [
                        {
                          "line": 20,
                          "lval": "data",
                          "rval": {
                            "arguments": [
                              {
                                "type": "string",
                                "value": "this.callers"
                              }
                            ],
                            "name": "mergedata",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 20,
                      "promiser": "callers"
                    },
                    {
                      "attributes": [
                        {
                          "line": 21,
                          "lval": "slist",
                          "rval": {
                            "type": "list",
                            "value": [
                              {
                                "type": "string",
                                "value": "@(this.callers_promisers)"
                              }
                            ]
                          }
                        }
                      ],
                      "line": 21,
                      "promiser": "promisers"
                    }
                  ]
                }
              ],
              "line": 19,
              "name": "vars"
            },
            {
              "contexts": [
                {
                  "name": "any",
                  "promises": [
                    {
                      "attributes": [
                        {
                          "line": 24,
                          "lval": "usebundle",
                          "rval": {
                            "type": "symbol",
                            "value": "two"
                          }
                        }
                      ],
                      "line": 24,
                      "promiser": "two"
                    }
                  ]
                }
              ],
              "line": 23,
              "name": "methods"
            }
          ],
          "sourcePath": "/home/tzz/sync/cf/test/test_callers.cf"
        },
        "depth": 4,
        "frame": 3,
        "type": "bundle"
      },
      {
        "depth": 5,
        "frame": 2,
        "promise_type": "vars",
        "type": "promise_type"
      },
      {
        "depth": 6,
        "frame": 1,
        "promise_classes": "any",
        "promise_comment": "",
        "promise_type": "vars",
        "promiser": "callers",
        "type": "promise"
      }
    ],
    "promisers": [
      ""
    ]
  },
  "two": {
    "callers": [
      {
        "bundle": {
          "arguments": [],
          "bundleType": "agent",
          "line": 6,
          "name": "main",
          "namespace": "default",
          "promiseTypes": [
            {
              "contexts": [
                {
                  "name": "any",
                  "promises": [
                    {
                      "attributes": [
                        {
                          "line": 9,
                          "lval": "data",
                          "rval": {
                            "arguments": [
                              {
                                "type": "string",
                                "value": "this.callers"
                              }
                            ],
                            "name": "mergedata",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 9,
                      "promiser": "callers"
                    },
                    {
                      "attributes": [
                        {
                          "line": 10,
                          "lval": "slist",
                          "rval": {
                            "type": "list",
                            "value": [
                              {
                                "type": "string",
                                "value": "@(this.callers_promisers)"
                              }
                            ]
                          }
                        }
                      ],
                      "line": 10,
                      "promiser": "promisers"
                    }
                  ]
                }
              ],
              "line": 8,
              "name": "vars"
            },
            {
              "contexts": [
                {
                  "name": "any",
                  "promises": [
                    {
                      "attributes": [
                        {
                          "line": 13,
                          "lval": "usebundle",
                          "rval": {
                            "arguments": [
                              {
                                "type": "string",
                                "value": "x"
                              },
                              {
                                "type": "string",
                                "value": "y"
                              },
                              {
                                "type": "string",
                                "value": "z"
                              }
                            ],
                            "name": "one",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 13,
                      "promiser": ""
                    },
                    {
                      "attributes": [
                        {
                          "line": 14,
                          "lval": "usebundle",
                          "rval": {
                            "arguments": [],
                            "name": "report",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 14,
                      "promiser": ""
                    }
                  ]
                }
              ],
              "line": 12,
              "name": "methods"
            }
          ],
          "sourcePath": "/home/tzz/sync/cf/test/test_callers.cf"
        },
        "depth": 0,
        "frame": 11,
        "type": "bundle"
      },
      {
        "depth": 1,
        "frame": 10,
        "promise_type": "methods",
        "type": "promise_type"
      },
      {
        "depth": 2,
        "frame": 9,
        "promise_classes": "any",
        "promise_comment": "",
        "promise_type": "methods",
        "promiser": "",
        "type": "promise"
      },
      {
        "depth": 3,
        "frame": 8,
        "iteration_index": 0,
        "type": "iteration"
      },
      {
        "bundle": {
          "arguments": [
            "a",
            "b",
            "c"
          ],
          "bundleType": "agent",
          "line": 17,
          "name": "one",
          "namespace": "default",
          "promiseTypes": [
            {
              "contexts": [
                {
                  "name": "any",
                  "promises": [
                    {
                      "attributes": [
                        {
                          "line": 20,
                          "lval": "data",
                          "rval": {
                            "arguments": [
                              {
                                "type": "string",
                                "value": "this.callers"
                              }
                            ],
                            "name": "mergedata",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 20,
                      "promiser": "callers"
                    },
                    {
                      "attributes": [
                        {
                          "line": 21,
                          "lval": "slist",
                          "rval": {
                            "type": "list",
                            "value": [
                              {
                                "type": "string",
                                "value": "@(this.callers_promisers)"
                              }
                            ]
                          }
                        }
                      ],
                      "line": 21,
                      "promiser": "promisers"
                    }
                  ]
                }
              ],
              "line": 19,
              "name": "vars"
            },
            {
              "contexts": [
                {
                  "name": "any",
                  "promises": [
                    {
                      "attributes": [
                        {
                          "line": 24,
                          "lval": "usebundle",
                          "rval": {
                            "type": "symbol",
                            "value": "two"
                          }
                        }
                      ],
                      "line": 24,
                      "promiser": "two"
                    }
                  ]
                }
              ],
              "line": 23,
              "name": "methods"
            }
          ],
          "sourcePath": "/home/tzz/sync/cf/test/test_callers.cf"
        },
        "depth": 4,
        "frame": 7,
        "type": "bundle"
      },
      {
        "depth": 5,
        "frame": 6,
        "promise_type": "methods",
        "type": "promise_type"
      },
      {
        "depth": 6,
        "frame": 5,
        "promise_classes": "any",
        "promise_comment": "",
        "promise_type": "methods",
        "promiser": "two",
        "type": "promise"
      },
      {
        "depth": 7,
        "frame": 4,
        "iteration_index": 0,
        "type": "iteration"
      },
      {
        "bundle": {
          "arguments": [],
          "bundleType": "agent",
          "line": 27,
          "name": "two",
          "namespace": "default",
          "promiseTypes": [
            {
              "contexts": [
                {
                  "name": "any",
                  "promises": [
                    {
                      "attributes": [
                        {
                          "line": 30,
                          "lval": "data",
                          "rval": {
                            "arguments": [
                              {
                                "type": "string",
                                "value": "this.callers"
                              }
                            ],
                            "name": "mergedata",
                            "type": "functionCall"
                          }
                        }
                      ],
                      "line": 30,
                      "promiser": "callers"
                    },
                    {
                      "attributes": [
                        {
                          "line": 31,
                          "lval": "slist",
                          "rval": {
                            "type": "list",
                            "value": [
                              {
                                "type": "string",
                                "value": "@(this.callers_promisers)"
                              }
                            ]
                          }
                        }
                      ],
                      "line": 31,
                      "promiser": "promisers"
                    }
                  ]
                }
              ],
              "line": 29,
              "name": "vars"
            }
          ],
          "sourcePath": "/home/tzz/sync/cf/test/test_callers.cf"
        },
        "depth": 8,
        "frame": 3,
        "type": "bundle"
      },
      {
        "depth": 9,
        "frame": 2,
        "promise_type": "vars",
        "type": "promise_type"
      },
      {
        "depth": 10,
        "frame": 1,
        "promise_classes": "any",
        "promise_comment": "",
        "promise_type": "vars",
        "promiser": "callers",
        "type": "promise"
      }
    ],
    "promisers": [
      "",
      "two"
    ]
  }
}
```

The entries all have a `type` key and a `frame` counter.  For different frames along the stack frame path, we have different data in the entries:

* bodies: under key `body` the entry has a full dump of the body policy as JSON, same as what `cf-promises -p json` would produce, using `BodyToJson`
* bundles: under key `bundle` the entry has a full dump of the bundle policy as JSON, same as what `cf-promises -p json` would produce, using `BundleToJson`
* promise iteration: the `iteration_index` is recorded
* promises: the `promise_type`, `promiser`, `promise_classes`, and `promise_comment` are recorded
* promise sections (types): the `promise_type` is recorded

Clearly this is a lot of data that we don't want to record for every promise.  I have several suggestions:

* make it a common control setting
* trigger it with a class (note it can be deactivated as well that way)
* trigger it with a promise `meta` tag (most awkward but most controllable as well)
* make it a function call instead of a `this.callers` variable (`this.caller_promisers` would also be moved to a function).  The function could even take an optional number to indicate a specific frame desired.  This is my preferred approach.


kacfengine:
> make it a function call instead of a this.callers variable (this.caller_promisers would also be moved to a function). The function could even take an optional number to indicate a specific frame desired. This is my preferred approach.

Agree.

```
{
        "depth": 1,
        "frame": 2,
        "promise_type": "vars",
        "type": "promise_type"
      },
      {
        "depth": 2,
        "frame": 1,
        "promise_classes": "any",
        "promise_comment": "",
        "promise_type": "
```
(from the bottom of the example)

I'm don't quite understand how these fit into the picture. Does the mere promise type itself count as a frame? And the context as well? Perhaps you can explain. And maybe post a little more of the example as well, enough to get to the bottom of that caller list, at least.

tzz:
I've update the example with a full JSON dump.  The stack is accurate even though it's verbose.  I think that's the right approach even though it's a lot of data.  Promises, promise iterations, bodies, etc. are all stack frames in the way the CFEngine evaluation works.

We can add parameters to the functions to limit them to only specific types of frames (e.g. show only bundle calls).

tzz:
I've added a new commit that creates the functions `cfengine_callers()` and `cfengine_promisers()` with the same functionality as `this.callers` and `this.callers_promisers` previously proposed.

The updated example doesn't change much:

```
body common control
{
        inputs => { "$(sys.libdir)/stdlib.cf" };
}

bundle agent main
{
  vars:
      "callers" data => mergedata("this.callers");
      "promisers" slist => { @(this.callers_promisers) };

  methods:
      "" usebundle => one(x,y,z);
      "" usebundle => report();
}

bundle agent one(a,b,c)
{
  vars:
      "callers" data => cfengine_callers();
      "promisers" slist => cfengine_promisers();

  methods:
      "two" usebundle => two;
}

bundle agent two
{
  vars:
      "callers" data => cfengine_callers();
      "promisers" slist => cfengine_promisers();
}

bundle agent report
{
  vars:
      "state" data => datastate();
      "state2" data => mergedata('{ "main": state[vars][main] }','{ "one": state[vars][one] }', '{ "two": state[vars][two] }');

  methods:
      "" usebundle => file_make("$(this.promise_filename).mustache", "{{%.}}");

  files:
      "/tmp/dump.json"
        edit_template => "$(this.promise_filename).mustache",
        create => "true",
        template_data => mergedata(state2),
        template_method => "mustache";
}
```

@jooooooon @amousset do the functions work for you?  I ask since the original design was perhaps more convenient for something Normation is doing?

jooooooon:
Wow, nice work @tzz!

Pinging @peckpeck to test this on Normation's use case of this.

peckpeck:
@jooooooon @tzz the problem with functions is that we cannot build retrocompatible code with them since they will be parsed as an error by cf-promises.
That's why we chose the variable based solution, is allow us to write one code instead of having one file per cfengine version.

jooooooon:
Oh year, I hadn't noticed that. I'd definitely prefer this to be a variable, in ```this``` context. Functions don't really make sense for this usage, IMHO (ie, no arguments, just return data, that changes on each function call).

tzz:
We can have both the function and the variables.  But please note that the variable has to be constructed regardless of whether it's needed, so it will add to the runtime.  I think it needs to be optional in some way; I am not eager to walk up the stack for every promise just in case someone ever needs that variable.  The CFEngine developers probably share my reluctance.

IMHO the easiest way would be with a `def.json` setting if that file was parsed in C before anything else (see #2297).  I've given several more suggestions above.

Another way to consider would be to get the value of the variable only when it's requested, but that's a pretty big evaluation change.

peckpeck:
For other problems of this kind you should add annotations. Let's call this one '@future'
This would allow you to write 
  @future
  "variable" string => unknown_function();

And the function_call would be ignored by cf-promises.
Was it implemented in older cfengine versions, we would not have this discussion.

peckpeck:
And for the current case, we cannot use the function version until we deprecate cfengine 3.7 in our product, so I'd really prefer having a variable version available. 
To control the variable behavior, something more cfenginy than the json file would be within a body common control.

tzz:
@peckpeck the CFEngine `@if` macro is for the purpose of future compatibility.  It was added in 3.7 and would do what you're suggesting with `@future` annotations.  As far as what could have been done in the past and what's possible in the future, that's a great discussion for the mailing lists; see my proposal for a 4.0 version etc.

If you want to maintain compatibility with 3.7, the promise or body `meta` tag approach is the best way.  To start collecting `this.callers_promisers` in a body or in a promise you'd write:

```
"whatever" meta => { "cfengine:callers_promisers" }, ...
```
and
```
body x y
{
  meta => { "cfengine:callers_promisers" };
  ...
}
```

I think that would be all right, personally, in addition to the functions.

peckpeck:
Thank you for the @if
First I'd like to maintain compatibility with 3.6
And I don't understand how your example works. How would retrieve the "callers_promisers" list in a variable ?

tzz:
@peckpeck Promises and bodies with that `meta` tag would get the `this.callers_promisers` variable created.  So in my examples, you'd just use that variable.  If CFEngine doesn't support it, it won't resolve.  If a promise or body calls another promise or body, the variable may not be available (needs testing, I'm not sure).

If you want to be compatible with 3.6, then you can't use the `meta` attribute in bodies.

peckpeck:
Even if it could be nice, I don't need this information in bodies for the moment.
But I'm not sure of what you are proposing.

Adding the meta everywhere we use the variable would be a bit cumbersome.
That's why I proposed something global like a common control.

Adding something in addition to functions to maintain compatibility is a good idea.

tzz:
@peckpeck I don't think I'm explaining it clearly.  Please understand you can't have 3.6 compatibility if you introduce a new body control setting.  My vote is against a global body control setting because I think my proposal is much better.  Such a setting also has to get approved by CFEngine developers and it will slow the agent down significantly.  It will traverse the call stack on **every** promise iteration, collecting data that no one needs during normal operation.

My proposal is:

* add the new functions (I think we all agree on that, so we can proceed with this PR)
* collect the `this.callers` and `this.callers_promisers` variables only when the promise or body `meta` includes a special tag (for bodies, this is not compatible with 3.6 because the `meta` tag for bodies was introduced in 3.7, but IIUC that's irrelevant to you)

If the second part of my proposal is not acceptable to you, please put forth an alternate proposal.  Either way, I would move that discussion to Redmine as a new ticket because we're clearly not reaching consensus on the second part while we agree on the first part.

peckpeck:
Well I'm OK with meta, I think we can handle just one line added to places where we use the variable.

tzz:
@peckpeck @amousset @jooooooon I created https://dev.cfengine.com/issues/7611 for the `meta` proposal.  Please comment there and watch the issue to indicate your support.

@kacfengine I think this PR is ready for review.

kacfengine:
FYI, #2344 mentions that it might merge with this one, so I'm not doing anything here until I hear if that has been resolved.

tzz:
I think this is again ready for review.

tzz:
@kacfengine #2344 is closed so this PR is the one to review and merge

kacf:
Sorry, this got left behind for a while. Some new comments:

* Codewise, this looks really nice.
* Tests are lacking for the function variant of this feature.
* I think we should not include the variable variant, because 3.6 is EOL in just two months. Let's rather keep it simple and stick to the function variant.

tzz:
I think @amousset or @jooooooon may have reasons to keep the variable variant, let's see.

From my side it's OK to remove the variable variant. It removes complexity. But I don't hate it :)

After rebasing this against today's master, the acceptance test fails, so it will need some love. Let me know the decision on the variants please, because I'll write a new test and drop the current ones if the variable variant is dropped.

kacf:
Yeah, let's drop the variable variant. The function variant will work on all 3.7+ versions, thanks to the `@if` macro, and by the time this gets released, 3.6 will be unsupported.

jooooooon:
Hi! Good to see this moving again :)

@peckpeck is the one who know about this feature here, but he's on vacation this week. Any chance you could hold off merging until he gets back to give us an overview of which use case prompted the variable variant?

kacf:
My understanding is that the function variant cannot be used if there are any 3.6 agents, because the infamous syntax error problem when encountering new keywords. I think this is the only reason for the variable variant, but I could be mistaken. We'll continue testing, but I can hold off the merge until next week.

jooooooon:
Ah, if that is indeed the case, I think it's pretty important that we 
keep the variable variant then. We still have to write code that is 
compatible for both 3.6 and 3.7 agents - that's the cost of writing a 
framework for many users :)

kacf:
If you really need this for 3.6 as well, I propose that you use the dynamic loading trick (`inputs => { @(list) };` where `list` depends on version), and then put the new function in there. There are *lots* of new features in 3.7, 3.8 and 3.9 that are all incompatible with 3.6 at a syntax level, this is just one of many, so I don't see why we should make an exception for this case when there are workarounds. The price if we do it is that we get an inconsistency in the language (a variable sort of behaving like a function) that we have to live with forever.

jooooooon:
Kristian, I understand. However, as I said, I don't have visibility on 
this topic, @peckpeck does. Please be patient and let him explain his 
reasons for implementing this in the first place - I'm just offering 
some context, that may or may not be relevant.

kacf:
Sure! :-)

peckpeck:
@kacf The goal is to use this in all methods, so the inputs trick is equivalent to duplicating all our code which is a bad idea.
Moreover we cannot use the function version since the @if macros have not been backported.

So this feature is useless for us if we don't have the variable variant (or or a backported @if)

kacf:
Well, how long are you planning to keep 3.6 around?

A very quick count reveals that there is at least 38 forwards-incompatible changes to the syntax introduced in 3.9, and that's compared to 3.8. Compared to 3.6 the number will be much higher. We're talking about one such incompatibility here, what about the 38 others that already exist, should we make extra variables for all of those?

If this was an incompatibility with 3.7 I might be convinced, but not for 3.6, which is basically EOL.

kacf:
Btw, I assume that you won't have anyone simultaneously running 3.6 and 3.9 in the same installation? Your problem is that you want to maintain one code base, but offer it to customers using both 3.6 and 3.9, am I right?

In that context it would not be difficult to preprocess your policy files for 3.6 installations, by simply removing sections enclosed in the `@if` block. This would give you one unified code base, but a slightly more complicated 3.6 build process. Just a suggestion.

tzz:
Aside from the discussion of variables vs. functions and 3.6 vs. newer: #1413 brought up that it would be useful to have the real and effective UID and GID, plus the current working directory, stored in the call stack. Let me know what you think and I can amend this PR.

jooooooon:
Hey Kristian,

Let's take a litte bit of context into account here:

The reason we care is that we develop ncf (http://www.ncf.io), a CFEngine framework that provides simple bundles to "do one thing but do it well". It abstracts a lot of the complexity of CFEngine, and provides workarounds and tested methods for simple tasks that are sometimes hard to achieve if you're not well versed in CFEngine. It sees a lot of downloads (several hundreds per week), but like most open source projects, I don't know how many actual users it has. Based on bug reporters I know for certain at least dozens, I assume hundreds, maybe thousands.

ncf makes a simple promise: it will work and provide the same functionality on all currently supported versions of CFEngine (one exception is that some new features only work on CFEngine versions where those features are available, eg reading/writing JSON files). It is, in many ways, similar to an API: it doesn't matter where (aka what version, OS, etc) it's running, the same call will have the same results. Therefore, as authors of ncf, we must provide a consistent experience across all currently supported versions of CFEngine (so 3.6, 3.7 and 3.8 as of today, according to https://cfengine.com/company/blog-detail/cfengine-3-7-3-3-6-7-and-3-8-2-released/).

A significant proportion of ncf users have multiple agent versions in production, that include 3.6 and 3.7, sometimes 3.8 (but rarely, I assume due to it's "non-LTS" status). We see a clear majority of 3.7 agents nowadays, but there is often a minority of 3.6 agents too. Therefore, ncf needs to provide "pure" CFEngine code that will work on 3.6 *and* 3.7 (*and* 3.9, yes).

(side note: we do have a mechanism in ncf to only load some methods if agent version is >= X.Y, which is how we support new features in CFEngine that are not backwards compatible, like new functions)

Why do these users not upgrade their 3.6 agents to 3.7 (LTS) or the latest versions? Several reasons that I know of, maybe some I don't know of, but most can be traced back to "red tape issues". There are several kinds of these, again: 1) mission-critical servers that are often in a "freeze" period where no upgrades are allowed, and non-freeze time sees other priorities than upgrading an agent that is still officially supported, 2) remote, embedded devices, where any upgrade is risky, and IT prefer to validate the new version for 6-12 months before rolling it out, 3) lack of time, etc.

Put simply: they will, they just haven't yet, and they see no reason to rush as 3.6 is still officially supported. I certainly can't do anything to speed it up, apart from providing a version of ncf that's compatible with both old- and new-versions, which is the object of this discussion in the first place.

Sorry for very long comment - I hope this helps clarify why it's important to have such features for cross-version compatibility.

kacf:
I also replied to this on communityadvisoryboard list with a more detailed response. I won't repeat that here.

For this PR, you say yourself that 3.8 is rare. 3.9 will probably be rare as well, since it is also non-LTS. By the time the first LTS version, 3.10, comes around at Christmas, 3.6 will have been unsupported for 5 months already. Like I said in my list reply, we can give you advice and help if you want to keep 3.6 on life support and backport the `@if` feature, but it will be up to you to do it.

kacf:
3.9 will be closed for new features soon, so we should get this moving. The variable variant won't be added, sorry.

I have added some extra modifications here: #2585 (see last commit).

One question I have: I was not able to add an acceptance test for the `callers` variant of the functions, because the variable overflows the well known 4096 limit (once converted to a string, that is). In fact, it seems almost impossible to use without exceeding this limit. Is it even worth adding at this point? To me it borders on useless, but maybe I'm overlooking an alternative way of using it. Needless to say, the 4K issue should be solved, but this is a different problem altogether, and we may reconsider the `callers` function later if it gets solved. What do you think?

jooooooon:
Hi. Thanks for the work in #2585, @kacf. Can you give us a couple days to test this with our promise base please?

As for the 4096 limit, I don't see how you hit it so easily. The whole list of callers can't be that long, in our examples it is something like "call main bundle/run technique one/check conditions with gm2", which is << 4096. Maybe it's just your test framework adding in a lot of callers?

kacf:
@jooooooon: To be clear, I'm talking about "callers" which is the full JSON structure, with the whole stack frame and lots of information like line numbers, file names, every attribute, etc, basically the long JSON output that @tzz posted at the top. It looks like you are talking about the "promisers" variant, which is indeed quite short, consisting only of the names of the bundles leading to the calling bundle. "promisers" is in great shape, and will be added for sure.

kacf:
> Can you give us a couple days to test this with our promise base please?

The feature merge window will close quite soon, we plan for a beta in the beginning of May, and we have to close the branch a bit before that. Is there anything you expect to change? Note that bug fixes can still go in after the window closes, as long as we agree on the basic properties of the feature.

jooooooon:
Oh, you are absolutely right @kacf. I'll drink another coffee then try to write intelligent things on the internet :)

jooooooon:
> The feature merge window will close quite soon, we plan for a beta in the beginning of May, and we have to close the branch a bit before that. Is there anything you expect to change? Note that bug fixes can still go in after the window closes, as long as we agree on the basic properties of the feature.

@peckpeck is the one who can answer that, and unfortunately, he's at a customer's until Thursday and probably won't be able to look at it before then. I don't think anything needs changing specifically, I just would have liked his confirmation.

kacf:
Coffee (TM), powering the internet since the 90'ies!

> @peckpeck is the one who can answer that

Ok, let's see if @tzz has an opinion on the use of the full JSON structure, it very well be that I've missed something. If it goes in as well, I assume @peckpeck will not have a problem, or he would probably have mentioned it already.

tzz:
@kacf your changes look great. I would worry about an acceptance test on `callers` because it's so complex and not really a user feature. But if I did, it would be something like `great_expectations` that is more structured. Maybe we should just open a ticket for that acceptance test, add me as watcher, and then we can work on it during the test period for 3.9.

kacf:
Ok, I suppose you're right, it's more of a debugging feature (but why do I get the feeling that @jooooooon and @peckpeck will find some way of using this in NCF :-) ).

In any case, later yesterday I got to thinking that this is definitely useful if you ask for specific parts of the structure by indexing (so `$(callers[bundle][bundleType])` for instance). This avoids the whole 4K issue. Not sure why I didn't think of that sooner.

kacf:
Btw, what do you think about renaming the functions to `callstack_callers` and `callstack_promisers` instead of `cfengine_*`?

jooooooon:
> Ok, I suppose you're right, it's more of a debugging feature (but why do I get the feeling that @jooooooon and @peckpeck will find some way of using this in NCF :-) ).

Absolutely not a debugging feature from where we stand, indeed! :)

You can see how we intend to use this here: https://github.com/Normation/ncf/blob/master/tree/30_generic_methods/file_create.cf#L31. In short, we'll use the list of callers to build a guaranteed unique, but predictable, class prefix for all result classes methods make.

kacf:
I made a simplified version of the acceptance test for the "callers" variant, just testing a single sub element. Not at all fullproof, but the best we can do for now. I blame the 4K limit!

tzz:
@kacf renaming is fine, but these are very specialized introspection functions so I would indicate in some way that they are not general-purpose. How about `cfengine_callstack_promisers` and `cfengine_callstack`? It doesn't matter much, in the end. Your suggestion is fine too.

kacf:
I'd like to avoid the cfengine name, since it doesn't really say anything about the function, and I thought maybe "callstack" was "introspection-y" enough. I don't really have a better idea, so I think I'll stick with it.

tzz:
Cool, I love it!

kacf:
Alright, this (or #2585 to be exact) is green. Do you mind writing some docs for this, @tzz?

tzz:
@kacf docs ready in https://github.com/cfengine/documentation/pull/1452

kacf:
Feature merged in #2585.

PR #2332, opened 2015-09-28
[3.7.x] Redmine #6957: Add test to ensure runlog can be disabled.

kacfengine:
It is disabled by symlinking it to /dev/null.


kacfengine:
Ok, updated according to discussion in ticket, runlog can be disabled with a symlink.

PR #2333, opened 2015-09-28
Redmine #6957: Move /var/cfengine/cf3.<host>.runlog log to debug output.

kacfengine:
It contains detailed locking messages which is very developer oriented
and not useful as a separate log.

Changelog: Title

### DIFF
--- a/cf-agent/files_editline.c
+++ b/cf-agent/files_editline.c
@@ -266,7 +266,7 @@ Bundle *MakeTemporaryBundleFromTemplate(EvalContext *ctx, Policy *policy, Attrib
                 }
 
                 int nl = StripTrailingNewline(promiser, size);
-                assert(nl != -1);
+                CF_ASSERT(nl != -1, "StripTrailingNewline failure");
 
                 np = PromiseTypeAppendPromise(tp, promiser, (Rval) { NULL, RVAL_TYPE_NOPROMISEE }, context, NULL);
                 np->offset.line = lineno;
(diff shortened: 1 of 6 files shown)


neilhwatson:
> It contains detailed locking messages which is very developer oriented

This could be useful if there were better developer level docs. I'm sure the user community could be more helpful debugging problems if we could reference low level documentation.

jimis:
I don't like this solution, see my comments in [Redmine#6957](https://dev.cfengine.com/issues/6957).

kacfengine:
> This could be useful if there were better developer level docs. I'm sure the user community could be more helpful debugging problems if we could reference low level documentation.

True, but I think the correct way forward here is to make the locking system simpler, so it is easily understandable, instead of documenting the complexity we have now.

kacfengine:
> Personally I'm avoiding this kind of cast, we have ARG_UNUSED etc when we need such functionality.

They are not all arguments though. Will it work on local variables? It would look a bit funny with ARG, perhaps...

> In addition it gives the wrong hint to the compiler in this case, because we are in fact using the return value as we should, even if we are only doing it in DEBUG mode. In the future the cast can be forgotten after some code refactoring and create trouble.

Fully agree, but I'm not aware of any way to avoid that without executing the code in release mode.

> What do you think about changing assert() to CF_ASSERT() for these cases?

It will mean the code runs in release mode as well, but I suppose that is very minor in this case. Unless you have a better suggestion I'll go ahead and do that.

jimis:
> > What do you think about changing assert() to CF_ASSERT() for these cases?

> It will mean the code runs in release mode as well, but I suppose that is very minor in this case. Unless you have a better suggestion I'll go ahead and do that.

+1. `CF_ASSERT()` spits out an UnexpectedError message in case the check fails, which I believe is correct in these cases.

kacfengine:
After discussion in the ticket, we decided to remove the log entirely, since the same information is already present in the debug output. Will update this soon.

kacfengine:
trigger build

jimis:
+1 for the `log` directory. -1 for removing a potentially useful log. :-)

PR #2335, opened 2015-10-07
Add: Test mustache does not html escape outside vars

nickanderson:
Ref: https://dev.cfengine.com/issues/7620


nickanderson:
trigger build

kacfengine:

PR #2337, opened 2015-10-09
Redmine#7624: Reduce malloc() thread contention

jimis:
Changelog: Reduce malloc() thread contention on heavily loaded
cf-serverd, by not exiting early in the logging function, if no message
is to be printed. (Redmine #7624)


kacfengine:
+1000!

I thought this was always the case actually. Now I feel guilty for doing all those debug prints!

kacfengine:
trigger build

kacfengine:
Cherry-picked to 3.6.x and 3.7.x.

jimis:
Oh, thanks for doing the cherry-pickery!

PR #2338, opened 2015-10-14
tests/unit: Remove non-portable unneeded type casts

stweil:
Casting between pointers and int results in compiler warnings.
It is not needed here, so simply remove it.


kacfengine:
trigger build

kacfengine:
Thanks! Cherry-picked to 3.6.x and 3.7.x as well.

PR #2339, opened 2015-10-15
Add @if feature() syntax

peckpeck:
@if feature work like @if minimum_version but allow to distinguish
between features choosen at compile time.

### DIFF
--- a/libpromises/feature.c
+++ b/libpromises/feature.c
@@ -1,26 +1,23 @@
 #include <syntax.h>
-
 #include <string.h>
+#include <cfnet.h>
+#include <sysinfo.h>
 
 static const char* features[] = {
-// the parser leaves a trailing ')' 
 #ifdef HAVE_LIBYAML
-  "yaml)",
+  "yaml",
 #endif
 #ifdef HAVE_LIBXML2
-  "xml2)",
+  "xml2",
 #endif
 
-  ")" // This terminates the array and declares empty as an always existing feature
+  NULL
 };
 
 int KnownFeature(const char *feature)
 {
-  int feature_count = sizeof(features)/sizeof(const char*);
-  char copy[200];
-
   // dumb algorithm, but still effective for a small number of features
-  for(int i=0 ; i<feature_count ; i++) {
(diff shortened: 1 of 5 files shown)


tzz:
I looked at this and overall it's good.  Some notes:

* Maybe there's a way to combine the two `@if` sections so code is not duplicated?  I am worried about the duplication of some tricky code.  Also we agreed not to nest `@if` on version, but now we need to agree on nesting for mixed `@if` macros.  My vote is to still disallow nesting but it may make life harder to library maintainers.

* The extra `feature.c` facility makes a lot of sense, but then `sysinfo.c:BuiltInClasses()` should be adjusted to walk through that feature list instead of duplicating the class names and logic.

peckpeck:
The code looks similar but it's not duplicated. It's only the depth check that is duplicated.
I don't really think we need to nest @if macro.

I will update the PR with the static class creation.

peckpeck:
@tzz PR updated

tzz:
You should target master, not 3.7.x.  The developers do backporting.

libpromises/feature.c
```diff
@@ -0,0 +1,38 @@
...
+    if(r==0) {
+      return 1;
+    }
+  }
+  return 0;
+}
+
+void CreateHardClassesFromfeatures(EvalContext *ctx, char *tags)
+{
+  char vbuff[CF_BUFSIZE];
```
tzz: Please use a `Buffer` here, it's just as easy and doesn't have a size limit.

tzz:
Looks OK to me, passing on to the developers...

tzz:
The `@if` changes should get acceptance tests.

I would prefer if this PR focused on the feature piece and then you make a new PR to add the patch level to the version `@if`.  It makes review easier, but it's just my suggestion.  The developers can review them both here.  Either way, for that second change, you need to update the existing `@if` acceptance tests that currently only check major and minor.  It's not trivial because you have to test failure modes, not just success.  Sorry I have to make your life difficult :)

peckpeck:
PR on master with tests: #2340

kacfengine:
Closing in favor of #2340.

PR #2340, opened 2015-10-15
Add @if feature() syntax

peckpeck:
@if feature work like @if minimum_version but allow to distinguish
between features choosen at compile time.

Original PR: #2339

### DIFF
--- a/libpromises/cf3lex.l
+++ b/libpromises/cf3lex.l
@@ -35,6 +35,7 @@
 #include <parser_state.h>
 #include <file_lib.h>
 #include <regex.h>
+#include <feature.h>
 
 /* yyinput/input are generated and unused */
 
@@ -83,7 +84,8 @@ line       ^.*$
 
 comment    #[^\n]*
 
-macro_if    ^@if\ minimum_version\([0-9]{1,10}\.[0-9]{1,10}\)
+macro_if_version    ^@if\ minimum_version\([0-9]{1,10}\.[0-9]{1,10}\)
+macro_if_feature    ^@if\ feature\([a-zA-Z0-9_]+\)
 macro_endif ^@endif
 
 promises   bundle
@@ -143,7 +145,7 @@ promise_type   [a-zA-Z_]+:
                           P.line_pos = 0;
                       }
 
-{macro_if}            {
+{macro_if_version}    {
                         const char* version_text = yytext+20;
                         ParserDebug("\tL:macro @if %d:version=%s\n", P.line_pos, version_text);
                         {
@@ -187,6 +189,31 @@ promise_type   [a-zA-Z_]+:
                           {
                             yyerror("fatal: Version() was unparseable");
                             return 0;
+                          }
+                        }
+                      }
+
+{macro_if_feature}    {
+                        const char* feature_text = yytext+12;
+                        ParserDebug("\tL:macro @if %d:feature=%s\n", P.line_pos, feature_text);
+                        {
+                          if (P.if_depth > 0)
+                          {
+                            yyerror("fatal: nested @if macros are not allowed");
+                            return 0;
+                          }
+
+                          P.if_depth++;
+
+                          if (KnownFeature(feature_text))
+                          {
+                            ParserDebug("\tL:macro @if %d:accepted to next @endif\n", P.line_pos);
+                          }
+                          else
+                          {
+                            /* ignore to the next @endif */
+                            ParserDebug("\tL:macro @if %d:ignoring to next @endif or EOF\n", P.line_pos);
+                            BEGIN(if_ignore_state);
                           }
                         }
                       }
--- a/libpromises/cf3lex.l
+++ b/libpromises/cf3lex.l
@@ -167,7 +167,7 @@ promise_type   [a-zA-Z_]+:
 
                           if (3 == sscanf(Version(), "%d.%d.%d", &major, &minor, &patch))
                           {
-                            request_level=sscanf(version_text, "%d.%d.%d", &request_major, &request_minor, &request_patch);
+                            request_level = sscanf(version_text, "%d.%d.%d", &request_major, &request_minor, &request_patch);
                             if (request_level >= 1)
                             {
                               if (request_major > major) {
--- a/libpromises/cf3lex.l
+++ b/libpromises/cf3lex.l
@@ -84,7 +84,7 @@ line       ^.*$
 
 comment    #[^\n]*
 
-macro_if_version    ^@if\ minimum_version\([0-9]{1,10}\.[0-9]{1,10}\)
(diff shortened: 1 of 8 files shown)


tzz:
The tests should also check nesting of the two different `@if` macros..

tests/acceptance/00_basics/macros/if_feature.cf
```diff
@@ -0,0 +1,43 @@
...
+@if feature(unknown_123)
+  classes:
+      "not_expected" expression => "any";
+@endif
+}
+
+bundle agent check
+{
+  methods:
+      "" usebundle => dcs_passif_expected("expected,expected2,expected2",
```
tzz: It will still work, but you have `expected2` twice.

libpromises/feature.c
```diff
@@ -0,0 +1,39 @@
...
+  for(int i=0 ; features[i]!=NULL ; i++) {
+    int r = strcmp(feature, features[i]);
+    if(r==0) {
+      return 1;
+    }
+  }
+  return 0;
+}
+
+void CreateHardClassesFromfeatures(EvalContext *ctx, char *tags)
```
kacfengine: Minor: CamelCase error: should be ``FromFeatures``.

kacfengine:
A note about indentation: In CFEngine we use 4 spaces per level. I can see that ``cf3lex.l`` is already wrongly indented in many places, I'm not sure why, but we'd still like to enforce proper indentation at least for new code. So if you could fix the indentation in ``feature.c`` to 4 spaces, at least.

libpromises/cf3lex.l
```diff
@@ -163,18 +167,28 @@ promise_type   [a-zA-Z_]+:
 
                           if (3 == sscanf(Version(), "%d.%d.%d", &major, &minor, &patch))
                           {
-                            if (2 == sscanf(version_text, "%d.%d", &request_major, &request_minor))
+                            request_level=sscanf(version_text, "%d.%d.%d", &request_major, &request_minor, &request_patch);
```
kacfengine: Coding style: Please leave some space between variable, ``=`` and ``sscanf``.

libpromises/cf3lex.l
```diff
@@ -163,18 +167,28 @@ promise_type   [a-zA-Z_]+:
...
-                                (request_major == major && request_minor > minor))
-                              {
-                                /* ignore to the next @endif */
+                              if (request_major > major) {
                                 ParserDebug("\tL:macro @if %d:ignoring to next @endif or EOF\n", P.line_pos);
                                 BEGIN(if_ignore_state);
-                              }
-                              else /* we're OK with the version */
-                              {
+                              } else if (request_major < major || request_level == 1) {
```
kacfengine: Won't ``request_level`` always be at least 2? It doesn't look like the ``version_text`` can ever contain less than the major and minor version number, because of the regular expression in ``macro_if_version``.

tests/acceptance/00_basics/macros/if_triple.cf
```diff
@@ -0,0 +1,48 @@
...
+  classes:
+      "expected2" expression => "any";
+@endif
+
+@if minimum_version(2.100.0)
+  classes:
+      "expected_2_100" expression => "any";
+@endif
+
+@if minimum_version(3.8.300)
```
kacfengine: Heh, I have the feeling we'll have to change this every time we make a new feature release. But at the same time I can't think of a better way to test it either. :-)

kacfengine:
Other than the mentioned details, looks like the core of the PR is good, and :+1: for the tests.

kacfengine:
This will also require some updated documentation, ala https://github.com/cfengine/documentation/pull/1021.

libpromises/cf3lex.l
peckpeck: Indeed.
But writing everything in the same format makes it easier to read and debug than having special cases. 
And it makes it easy to change the regex if needed

tests/acceptance/00_basics/macros/if_triple.cf
peckpeck: Yes I don't have a better solution.
At least it's limited to a single file.

peckpeck:
I don't understand how testing nested component could work since it breaks the parser.

peckpeck:
PR updated

kacfengine:
> I don't understand how testing nested component could work since it breaks the parser.

Make it a separate test and name the test so that it ends with ``.x.cf``. Those are tests that are meant specifically to test for invalid policy.

libpromises/cf3lex.l
kacfengine: Ok, makes sense.

peckpeck:
Documentation in https://github.com/cfengine/documentation/pull/1188

tests/acceptance/00_basics/macros/if_triple.cf
tzz: You can write a secondary `"$(G.testfile).cf"` file with the right values inside, using `file_make()` in the `init` stage, and then run it from the primary test, looking for the right output.  See `tests/acceptance/10_files/09_insert_lines/nothing_to_repeat_error.cf` for an example of running a secondary test file.

kacfengine:
trigger build

kacfengine:
The test doesn't pass in our test farm:

```
Error Message

FAIL (UNEXPECTED FAILURE) ./00_basics/macros/if_feature.cf

Stacktrace

R: /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_i386_linux_debian_4/core/tests/acceptance/./00_basics/macros/if_feature.cf FAIL
R: dcs_passif: failing based on class "__passif"
R: dcs_passif_expected: bad: expected was NOT defined
```

Does it pass for you?

peckpeck:
Yes it passes for me. But the test is made for when you build with libxml and libyaml included.

libpromises/feature.c
```diff
@@ -0,0 +1,39 @@
...
+#include <cfnet.h>
+#include <sysinfo.h>
+#include <buffer.h>
+
+static const char* features[] = {
+#ifdef HAVE_LIBYAML
+    "yaml",
+#endif
+#ifdef HAVE_LIBXML2
+    "xml2",
```
kacfengine: Found the error: This should be ``"xml"``. If you can fix that real quick I don't have to make a new pull request and can run it through the analyzer again.

peckpeck:
@kacfengine fixed

kacfengine:
Thanks!

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
All tests passed in Jenkins!

kacfengine:
Thanks @peckpeck! Merged manually after including "curl" in feature selections.

PR #2342, opened 2015-10-16
Add default body for all promise type

peckpeck:
If there is a body named files_action_default
it is used as the default action for files promises

### DIFF
--- a/libpromises/promises.c
+++ b/libpromises/promises.c
@@ -451,13 +451,11 @@ static void AddDefaultBodiesToPromise(EvalContext *ctx, Promise *promise, const 
             if(!PromiseBundleOrBodyConstraintExists(ctx, constraint_type, promise)) {
                 const Policy *policy = PolicyFromPromise(promise);
                 // default format is <promise_type>_<body_type>_default
-                char* default_body_name = StringConcatenate(4, promise->parent_promise_type->name, "_", constraint_type, "_default");
-                Seq *bodies_and_args = EvalContextResolveBodyExpression(ctx, policy, default_body_name, constraint_type);
-                // If a default has been defined
-                if (bodies_and_args && SeqLength(bodies_and_args) > 0) {
-                    printf("Found %s\n", default_body_name);
-                    const Body *bp = SeqAt(bodies_and_args, 0);
-                    CopyBodyConstraintsToPromise(ctx, promise, bp);
+                char* default_body_name = StringConcatenate(4, "default:", promise->parent_promise_type->name, "_", constraint_type);
+                const Body *bp = EvalContextFindFirstMatchingBody(policy, constraint_type, "default", default_body_name);
+                if(bp) {
+                  printf("Found %s\n", default_body_name);
+                  CopyBodyConstraintsToPromise(ctx, promise, bp);
                 }
             }
         }
--- a/libpromises/promises.c
+++ b/libpromises/promises.c
@@ -38,6 +38,7 @@
 #include <audit.h>
 
 static void DereferenceComment(Promise *pp);
+static void AddDefaultBodiesToPromise(EvalContext *ctx, Promise *promise, const PromiseTypeSyntax *syntax);
 
 void CopyBodyConstraintsToPromise(EvalContext *ctx, Promise *pp,
                                   const Body *bp)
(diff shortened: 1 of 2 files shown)


tzz:
As @ncharles and I were proposing, maybe these could be in a new namespace so it's very hard to create them accidentally.

The code looks good to me, but definitely will need acceptance testing.

Thank you very much for doing all this work.  It's really great.

peckpeck:
Here is the same PR with a default namespace

libpromises/promises.c
```diff
@@ -424,9 +425,43 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
...
+    for (int i = 0; syntax->constraints[i].lval; i++)
+    {
+        // of type body
+        if(syntax->constraints[i].dtype == CF_DATA_TYPE_BODY) {
+            const char *constraint_type = syntax->constraints[i].lval;
+            // if there is no matching body in this promise
+            if(!PromiseBundleOrBodyConstraintExists(ctx, constraint_type, promise)) {
+                const Policy *policy = PolicyFromPromise(promise);
+                // default format is <promise_type>_<body_type>_default
+                char* default_body_name = StringConcatenate(4, "default:", promise->parent_promise_type->name, "_", constraint_type);
```
tzz: Nice and simple, I like it.

libpromises/promises.c
```diff
@@ -424,9 +425,43 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
...
+                const Body *bp = EvalContextFindFirstMatchingBody(policy, constraint_type, "default", default_body_name);
+                if(bp) {
+                  printf("Found %s\n", default_body_name);
```
tzz: Please use the standard logging functions.  This is probably `VERBOSE` level.

libpromises/promises.c
```diff
@@ -424,9 +425,43 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
         SeqDestroy(bodies_and_args);
     }
 
+    // Add default body for promise body types that are not present
+    char *bundle_type = pcopy->parent_promise_type->parent_bundle->type;
+    char *promise_type = pcopy->parent_promise_type->name;
+    const PromiseTypeSyntax *syntax = PromiseTypeSyntaxGet(bundle_type, promise_type);
+    AddDefaultBodiesToPromise(ctx, pcopy, syntax);
+
+    // Add default body for global body types that are not present
```
tzz: Can you explain why this is needed?

libpromises/promises.c
peckpeck: Sorry, this was just a print line I used during debug to be sure I was guessing  the right function usage.
I'm not sure we need log at all here.

libpromises/promises.c
peckpeck: Global body type include all body types that are common to all bundles, like action. There body types are not returned bu the previous funcyion call.

libpromises/promises.c
tzz: You should log in `VERBOSE` mode so users know what's happening.  @ncharles had concerns about this implicit behavior so we need to tell the user that something happened here.

libpromises/promises.c
tzz: OK, cool.  So there are the body types that apply to every promise.

peckpeck:
OK, updated with logs

libpromises/promises.c
```diff
@@ -424,9 +425,43 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
...
+    for (int i = 0; syntax->constraints[i].lval; i++)
+    {
+        // of type body
+        if(syntax->constraints[i].dtype == CF_DATA_TYPE_BODY) {
+            const char *constraint_type = syntax->constraints[i].lval;
+            // if there is no matching body in this promise
+            if(!PromiseBundleOrBodyConstraintExists(ctx, constraint_type, promise)) {
+                const Policy *policy = PolicyFromPromise(promise);
+                // default format is default:<promise_type>_<body_type>
+                char* default_body_name = StringConcatenate(4, "default:", promise->parent_promise_type->name, "_", constraint_type);
```
tzz: Ugh, I just realized `default` is in fact the default namespace.  Sorry.  Can you change it to `bodydefault` or something else?

peckpeck: This seems to conflict with the fact that if I create a body without a namespace it's not taken into account by the "default" namespace search.

tzz: OK then.  If it works... :)

kacfengine: ``StringConcatenate`` allocates, it needs to be freed.

libpromises/promises.c
```diff
@@ -424,9 +425,43 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
...
+                const Body *bp = EvalContextFindFirstMatchingBody(policy, constraint_type, "default", default_body_name);
+                if(bp) {
+                  Log(LOG_LEVEL_VERBOSE, "Using the default body: %60s", default_body_name);
+                  CopyBodyConstraintsToPromise(ctx, promise, bp);
```
kacfengine: Small indentation error: Should be four extra spaces.

libpromises/promises.c
kacfengine: Also, I'm not quite clear on the body names here: Concatenating ``constraint_type`` indicates to me that the body name will need to be the name of the constraint (or attribute as is the more common name). So for example to make a default for a ``body perms`` and the attribute ``groups``, you'd have to call it ``body perms files_default_groups``. Is that correct? And if so, wouldn't it be be more natural to call it ``files_default_perms`` and then have all the default attributes in one body?

kacfengine:
This should have an acceptance test as well. You can probably take the one from https://github.com/cfengine/core/pull/2307, and perhaps add a second promise of a different type, for example ``commands``, so that we test at least two different types.

kacfengine:
Another thing we should measure here is performance. This is in a quite central place and might be an expensive lookup to do on every promise. Hard to say how much of an impact it will have, but the execution time of the test suite might be a good indicator.

libpromises/promises.c
peckpeck: With current code, the body would be named: body default:file_perms

kacf:
So is anything happening here?

kacf:
Superseded by #2469.

PR #2346, opened 2015-10-19
Redmine #6027: Directories randomly changed to files.

pasinskim:
This is quite nasty and border case scenario. For recursive file
copying if connection was timeouted recourse copying was continued.
In some cases timeouted package was sent by the server (usually under
heavy load) and was wrongly interpreted as response for next request.

Usually (if not always) directories were created first and then files,
so most often directory was replaced with a file and content of the file
was corrupted protocol leftovers (t 74OK: 0 432 0 0 20010 1302 140197032)
which is CFEngine copy file protocol transaction header and data.

The fix is to drop connection immediately after experiencing timeouts in
all cases where files and/or directories are copied recursively in loop.

Changelog: Redmine #6027 Directories should no more be changed randomly
into files.

### DIFF
--- a/cf-agent/verify_files_utils.c
+++ b/cf-agent/verify_files_utils.c
@@ -823,7 +823,16 @@ static PromiseResult SourceSearchAndCopy(EvalContext *ctx, const char *from, cha
         /* This sends 1st STAT command. */
         if (!ConsiderAbstractFile(dirp->d_name, from, attr.copy, conn))
         {
-            continue;
+            if (conn->conn_info->is_broken)
+            {
+                cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_INTERRUPTED, pp,
+                     attr, "connection timeout");
+                return PROMISE_RESULT_INTERRUPTED;
+            }
+            else
+            {
+                continue;
+            }
         }
 
         if (attr.copy.purge)    /* Purge this file */
@@ -854,7 +863,16 @@ static PromiseResult SourceSearchAndCopy(EvalContext *ctx, const char *from, cha
             if (cf_stat(newfrom, &sb, attr.copy, conn) == -1)
             {
                 Log(LOG_LEVEL_VERBOSE, "Can't stat '%s'. (cf_stat: %s)", newfrom, GetErrorStr());
-                continue;                                       /* TODO FAIL? */
+                if (conn->conn_info->is_broken)
+                {
+                    cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_INTERRUPTED, pp,
+                         attr, "connection timeout");
+                    return PROMISE_RESULT_INTERRUPTED;
+                }
+                else
+                {
+                    continue;
+                }
             }
         }
         else
@@ -862,7 +880,17 @@ static PromiseResult SourceSearchAndCopy(EvalContext *ctx, const char *from, cha
             if (cf_lstat(newfrom, &sb, attr.copy, conn) == -1)
             {
                 Log(LOG_LEVEL_VERBOSE, "Can't stat '%s'. (cf_stat: %s)", newfrom, GetErrorStr());
-                continue;                                       /* TODO FAIL? */
+                if (conn->conn_info->is_broken)
+                {
+                    cfPS(ctx, LOG_LEVEL_INFO,
+                         PROMISE_RESULT_INTERRUPTED, pp, attr,
+                         "connection timeout");
+                    return PROMISE_RESULT_INTERRUPTED;
+                }
+                else
+                {
+                    continue;
+                }
             }
         }
 
@@ -1032,7 +1060,16 @@ static PromiseResult VerifyCopy(EvalContext *ctx,
             if (!ConsiderAbstractFile(dirp->d_name, sourcedir,
                                       attr.copy, conn))
             {
-                continue;
+                if (conn->conn_info->is_broken)
+                {
+                    cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_INTERRUPTED,
+                         pp, attr, "connection timeout");
+                    return PROMISE_RESULT_INTERRUPTED;
+                }
+                else
+                {
+                    continue;
+                }
             }
 
             strcpy(sourcefile, sourcedir);
--- a/libcfnet/net.c
+++ b/libcfnet/net.c
@@ -123,7 +123,7 @@ int SendTransaction(const ConnectionInfo *conn_info,
  *  @return 0 in case of socket closed, -1 in case of other error, or
  *          >0 the number of bytes read.
  */
-int ReceiveTransaction(const ConnectionInfo *conn_info, char *buffer, int *more)
+int ReceiveTransaction(ConnectionInfo *conn_info, char *buffer, int *more)
 {
     char proto[CF_INBAND_OFFSET + 1] = { 0 };
     int ret;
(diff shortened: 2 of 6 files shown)


pasinskim:
@jimis @kristian can you take a look at this one?

pasinskim:
trigger build!

kristian:
Wrong @, sorry. I have no connection to Redmine.



—

cf-agent/verify_files_utils.c
```diff
@@ -854,7 +863,16 @@ static PromiseResult SourceSearchAndCopy(EvalContext *ctx, const char *from, cha
             if (cf_stat(newfrom, &sb, attr.copy, conn) == -1)
             {
                 Log(LOG_LEVEL_VERBOSE, "Can't stat '%s'. (cf_stat: %s)", newfrom, GetErrorStr());
-                continue;                                       /* TODO FAIL? */
+                if (conn->conn_info->is_timeouted)
+                {
+                    cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_INTERRUPTED, pp,
+                         attr, "connection timeout");
+                    return PROMISE_RESULT_INTERRUPTED;
```
jimis: Why do you need to return error only in the case of `is_timeouted` and not on **every** failure? Wouldn't it be much simpler?

cf-serverd/server_tls.c
```diff
@@ -429,7 +429,7 @@ int ServerTLSSessionEstablish(ServerConnectionState *conn)
         int ret = SSL_accept(ssl);
         if (ret <= 0)
         {
-            TLSLogError(ssl, LOG_LEVEL_ERR,
+            (void)TLSLogError(ssl, LOG_LEVEL_ERR,
```
jimis: This kind of cast has never been used in all of Core, lets not introduce it now without compelling reason. Did you get any compiler warnings? If so, which compiler?

libcfnet/net.c
```diff
@@ -123,7 +123,7 @@ int SendTransaction(const ConnectionInfo *conn_info,
  *  @return 0 in case of socket closed, -1 in case of other error, or
  *          >0 the number of bytes read.
  */
-int ReceiveTransaction(const ConnectionInfo *conn_info, char *buffer, int *more)
+int ReceiveTransaction(ConnectionInfo *conn_info, char *buffer, int *more)
```
jimis: The de-constification is a bit worrying. Do you really have to carry the error information in the `ConnectionInfo`?

cf-agent/verify_files_utils.c
pasinskim: There might be some other cases of this to fail. In some of those we can continue to copy next files in queue instead of dropping connection.

cf-serverd/server_tls.c
pasinskim: No errors. Just good practice. IMHO it's never to late to start doing something better. And the good reason is that you are explicitly mentioning in the code that you are ignoring the result of function which is really useful for future maintenance.

libcfnet/net.c
pasinskim: We can experiment with adding new arguments or something. In fact ConnectionInfo is modified in lot of places and almost only in this one it was const. This was quite convenient way to modify the ConnectionInfo itself as we are having access to it from all parts of code and it is easy to set/get flag needed.

cf-serverd/server.c
```diff
@@ -241,7 +241,7 @@ static void SpawnConnection(EvalContext *ctx, const char *ipaddr, ConnectionInfo
         Log(LOG_LEVEL_ERR,
             "Unable to spawn worker thread. (pthread_create: %s)",
             GetErrorStr());
-        goto cleanup;
+        /* Execute cleanup: which is the next instruction. */
```
kacfengine: Why change this? To the compiler it is essentially equivalent, and the old way would prevent someone from making a mistake if they add code under this block.

libcfnet/classic.c
```diff
@@ -58,15 +58,12 @@ static bool LastRecvTimedOut(void)
...
 {
     int already, got;
 
-    if (toget > CF_BUFSIZE - 1 || toget <= 0)
-    {
-        Log(LOG_LEVEL_ERR, "Bad software request to receive %d bytes", toget);
-        return -1;
-    }
+    assert(toget > 0);
+    assert(toget < CF_BUFSIZE - 1);
```
kacfengine: Sure you don't wanna leave this in? In release mode you will skip right over this check and go unprotected into the next block.

kacfengine:
It'll be nice to finally squash this bug!

cf-serverd/server.c
pasinskim: Make sense. Will revert this. I've just seen some unnecessary code which could be removed. In fact this whole function can be refactored easily to get rid of goto.

libcfnet/classic.c
pasinskim: Right. I will change this wan. In fact we have quite a lot of similar asserts in other places which are skipped so maybe I will fix those in separate commit added to this PR.

cf-agent/verify_files_utils.c
jimis: > There might be some other cases of this to fail.

True, `cf_stat()` can fail in a bunch of ways. But one possible cause might be network error; in that case, you don't need to differentiate with timeout: **any** network failure should terminate the promise, not only timeout.

This might help untangling this mess, since the whole "return the network timeout from deep inside the network path" issue won't even be needed to be resolved.

cf-serverd/server_tls.c
jimis: So the reason is for documentation towards human programmers who read the code; Why not add a comment?
Answer: Because if we start commenting on each and every function it will become a mess. Same if we start typecasting each and every call that we don't check its value. Too many functions like that, ugly mess.

`TLSLogError()` might not be necessary to check, so skip all casts - and I say /might/ because I haven't validated this refactoring yet. However `strlcpy()` should **almost always** be checked. If not checked, then a `(void)` cast is **not enough** documentation; I want to read a comment saying **why** truncation is not checked. So I don't see a point in `(void)` casting at all.

C is an ugly language, but very powerful. Type casts are one of the most powerful and most dangerous features of C: they nullify the type system, they shut the compiler up from any warning. Lets use them only where we *must*. From HACKING.md:

> Avoid using type casts, unless absolutely necessary. Usually a compiler warning is better satisfied with correct code rather than using a type cast.

libcfnet/net.c
```diff
@@ -133,9 +133,25 @@ int ReceiveTransaction(const ConnectionInfo *conn_info, char *buffer, int *more)
...
+             * This might lead to packages being not delivered in correct
+             * order and unexpected issues like directories being replaced
+             * with files. 
+             * In order to make sure that file transfer is reliable we have to
+             * close connection to avoid broken packages being received. */
+            conn_info->is_timeouted = true;
+        }
         break;
     case CF_PROTOCOL_TLS:
         ret = TLSRecv(conn_info->ssl, proto, CF_INBAND_OFFSET);
```
jimis: Shouldn't **any** failure of `TLSRecv()` be propagated upwards, and cause the promise to fail?

cf-agent/verify_files_utils.c
pasinskim: @jimis this is not true unfortunately. I agree we should close connection if there is any problem with it detected so I'll change this accordingly.
Unfortunately this won't eliminate the mess. The connection problems are discovered "deep inside network path" and this stuff is called by many functions in many places. In order to fix whole call stuck you will have to probably rewrite fair amount of functions and analyze which are supposed to close the connection and which are not.
Setting a flag and checking it when needed it IMHO much cleaner approach and in fact less risky. We will set flag in correct place and check it exactly where we are deciding if we should continue or not.

cf-serverd/server_tls.c
pasinskim: @jimis I disagree and in fact this is just a comment to ensure that error codes are handled unless you are explicitly ignoring those. Some static analyzers will warn about this and this is not a good practice to skip checking return codes. 
In some cases you don't care but IMHO you should write this explicitly. 

This code will be removed anyway so I wont argue more here...

libcfnet/net.c
pasinskim: Makes a lot of sense. Fixing!

jimis:
This has now conflicts because of my earlier pull requests: #2350 and #2351. But the changes here are good.

So @pasinskim I kept your commit, rebased, and amended a couple of mine in #2356. Closing this one, but please have a look at #2356.

PR #2348, opened 2015-10-21
3.7.x

pasinskim:


pasinskim:
@kacfengine @jimis this is stuff ported from master so should be merged easily.

PR #2350, opened 2015-10-22
Redmine#6027: Fix possible corruption in case of socket recv() timeout

jimis:
Make sure a timed-out connection gets closed at the lowest level, so
that further calls of recv() do not return the delayed server reply.

Complementing pull request #2346 from @pasinskim, this pull request ensures that the socket will never be reused in case of socket timeout.

### DIFF
--- a/libcfnet/classic.c
+++ b/libcfnet/classic.c
@@ -53,10 +53,11 @@ static bool LastRecvTimedOut(void)
  * @param buffer Buffer into which to read data
  * @param toget Number of bytes to read; a '\0' shall be written after
  *        the data; buffer must have space for that.
-
- * @return -1 on error; or number of bytes received. It should return less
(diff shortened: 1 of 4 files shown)


jimis:
trigger build

jimis:
trigger build

jimis:
Unfortunately there are still paths that the protocol can take and become out-of-sync, this pull request covers only the timeout path.

Fortunately AFAICT the other paths require a malicious or buggy server, that sends invalid sizes as request replies. In that case, yes, the client's protocol might become out of sync and write garbage again.

jimis:
@pasinskim @kacfengine I merged it after it passed all tests, but I still would appreciate if you had a quick look. 

To me it looks like this closes all paths for the protocol to become out-of-sync in case of timeout.

kacfengine:
Looks good to me!

PR #2351, opened 2015-10-22
stricter transaction protocol

jimis:
A small change that makes the transaction protocol stricter, plus lots of comments.

### DIFF
--- a/libcfnet/classic.c
+++ b/libcfnet/classic.c
@@ -54,10 +54,11 @@ static bool LastRecvTimedOut(void)
  * @param toget Number of bytes to read; a '\0' shall be written after
  *        the data; buffer must have space for that.
  *
- * @return number of bytes actually received, might be less than #toget
- *         <toget  when connection has been gracefully closed while we
- *             were expecting more data.
+ * @return number of bytes actually received, must be equal to #toget
  *         -1  in case of timeout or error - socket is unusable
+ *         -1  also in case of early proper connection close
+ *         0       NEVER
+ *         <toget  NEVER
  */
 int RecvSocketStream(int sd, char buffer[CF_BUFSIZE], int toget)
 {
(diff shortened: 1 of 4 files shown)


libcfnet/tls_generic.c
```diff
@@ -668,14 +668,16 @@ int TLSSend(SSL *ssl, const char *buffer, int length)
  * @param buffer Buffer, of size at least #toget + 1 to store received data.
  * @param toget Length of the data to receive, must be < CF_BUFSIZE.
  *
- * @return The length of the received data, which could be smaller or equal
- *         than the requested amount.
+ * @return The length of the received data, which should be equal or less
+ *         to the requested amount.
```
pasinskim: s /to/than/

libcfnet/classic.c
```diff
@@ -100,12 +101,17 @@ int RecvSocketStream(int sd, char buffer[CF_BUFSIZE], int toget)
...
+            Log(LOG_LEVEL_ERR,
+                "Peer closed connection early (received=%dB, expecting=%dB)",
+                already, toget);
+            buffer[already] = '\0';
+            /* TODO Why return 0? Early close is an error so -1 is better. */
+            return 0;
         }
     }
-    assert(already <= toget);
+    assert(already == toget);
```
pasinskim: Maybe it would be safer here to add some condition like if (already!=toget) {} to make sure that this is checked in release not only debug.

jimis:
trigger build

libcfnet/tls_generic.c
jimis: You're right, no idea why I edited this. :-)

libcfnet/classic.c
jimis: It is a programming error if this is the case, that's why I'm using assert. If it was dependant on any external factor, it would be vital to be checked in prod code as well, but it's not.

jimis:
@pasinskim After our discussion, I added TODO comments in all the possible points that the protocol to be out of sync. It's hard to handle them though since it would require refactoring as we discussed.

The good thing it that this out-of-sync case is not easy to happen any more; it requires a malicious server that sends bad replies.

PR #2353, opened 2015-10-23
Redmine #7634: Create failsafe.cf when not exists.

pasinskim:
Before setting input file to be failsafe.cf first we are checking if
file exists. If failsafe.cf is not accessible we are creating it
the same way as during bootstrap.

Changelog: failsafe.cf will be created when needed. (Redmine #7634)

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -238,10 +238,19 @@ int main(int argc, char *argv[])
     {
         Log(LOG_LEVEL_ERR, "CFEngine was not able to get confirmation of promises from cf-promises, so going to failsafe");
         EvalContextClassPutHard(ctx, "failsafe_fallback", "attribute_name=Errors,source=agent");
-        GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
-        policy = LoadPolicy(ctx, config);
+        
+        if (CheckAndGenerateFailsafe(GetInputDir(), "failsafe.cf"))
+        {
+            GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
+            policy = LoadPolicy(ctx, config);
+        }
     }
-    assert(policy);
+    
+    if (!policy)
+    {
+        Log(LOG_LEVEL_ERR, "Error reading CFEngine policy. Exiting...");
+        exit(EXIT_FAILURE);
+    }
 
     GenericAgentPostLoadInit(ctx);
     ThisAgentInit();
--- a/cf-serverd/cf-serverd.c
+++ b/cf-serverd/cf-serverd.c
@@ -59,8 +59,17 @@ int main(int argc, char *argv[])
     {
         Log(LOG_LEVEL_ERR, "CFEngine was not able to get confirmation of promises from cf-promises, so going to failsafe");
         EvalContextClassPutHard(ctx, "failsafe_fallback", "attribute_name=Errors,source=agent");
-        GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
-        policy = LoadPolicy(ctx, config);
+        if (CheckAndGenerateFailsafe(GetInputDir(), "failsafe.cf"))
+        {
+            GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
+            policy = LoadPolicy(ctx, config);
+        }
+    }
+    
+    if (!policy)
+    {
+        Log(LOG_LEVEL_ERR, "Error reading CFEngine policy. Exiting...");
+        exit(EXIT_FAILURE);
     }
 
     GenericAgentPostLoadInit(ctx);
--- a/libpromises/generic_agent.c
+++ b/libpromises/generic_agent.c
@@ -161,6 +161,7 @@ void GenericAgentDiscoverContext(EvalContext *ctx, GenericAgentConfig *config)
                 "Error writing builtin failsafe to inputs prior to bootstrap");
             exit(EXIT_FAILURE);
         }
+        GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
 
         char canonified_ipaddr[strlen(bootstrap_arg) + 1];
         StringCanonify(canonified_ipaddr, bootstrap_arg);
@@ -759,20 +760,6 @@ void GenericAgentInitialize(EvalContext *ctx, GenericAgentConfig *config)
     }
 
     setlinebuf(stdout);
-
-    if (config->agent_specific.agent.bootstrap_policy_server)
-    {
-        snprintf(vbuff, CF_BUFSIZE, "%s%cfailsafe.cf", GetInputDir(), FILE_SEPARATOR);
-
-        if (stat(vbuff, &statbuf) == -1)
-        {
-            GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
-        }
-        else
-        {
-            GenericAgentConfigSetInputFile(config, GetInputDir(), vbuff);
-        }
-    }
 }
 
 void GenericAgentFinalize(EvalContext *ctx, GenericAgentConfig *config)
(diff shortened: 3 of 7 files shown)


pasinskim:
@kacfengine yet another PR to handle.

pasinskim:
trigger build!

cf-agent/cf-agent.c
```diff
@@ -238,11 +238,21 @@ int main(int argc, char *argv[])
...
+        if (CheckAndGenerateFailsafe(GetInputDir(), "failsafe.cf"))
+        {
+            GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
+            policy = LoadPolicy(ctx, config);
+        }
     }
     assert(policy);
-
+    
+    if (!policy)
```
jimis: Having both this check and the `assert()` is redundant. I'd say that the `assert()` is all you need, since `LoadPolicy()` can never return NULL, and it is a programming error if it does.

libpromises/generic_agent.c
```diff
@@ -762,16 +762,10 @@ void GenericAgentInitialize(EvalContext *ctx, GenericAgentConfig *config)
...
     if (config->agent_specific.agent.bootstrap_policy_server)
     {
-        snprintf(vbuff, CF_BUFSIZE, "%s%cfailsafe.cf", GetInputDir(), FILE_SEPARATOR);
-
-        if (stat(vbuff, &statbuf) == -1)
+        if (CheckAndGenerateFailsafe(GetInputDir(), "failsafe.cf"))
         {
             GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
         }
-        else
```
jimis: `exit(EXIT_FAILURE)` would be a good thing, if

jimis: BTW, since you are now doing `failsafe.cf` generation when bootstrapping here, maybe you want to remove the one from `GenericAgentDiscoverContext()`.

jimis: Hmm, after a second look, I believe this part of code changes existing behaviour: Currently, we *always* regenerate `failsafe.cf` when bootstrapping, either it exists or not.

libpromises/generic_agent.c
```diff
@@ -1636,6 +1630,30 @@ void GenericAgentConfigApply(EvalContext *ctx, const GenericAgentConfig *config)
...
+    char failsafe_path[CF_BUFSIZE];
+    
+    if (strlen(inputdir) + strlen(input_file) > sizeof(failsafe_path) - 2)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Unable to generate path for %s/%s file. Path too long.",
+            inputdir, input_file);
+        /* We could create dynamically allocated buffer able to hold the 
+           whole content of the path but this should be unlikely that we
+           will end up here. */
```
jimis: Just FYI, what you mention in your comment can easily be done using `xasprintf()`, no need for error checking then.

tests/acceptance/00_basics/validation/validate_failsafe_creating.cf
```diff
@@ -0,0 +1,35 @@
...
+bundle agent test
+{
+  commands:
+    "$(sys.cf_agent) -D AUTO -f $(this.promise_filename).sub";
+}
+
+bundle agent check
+{
+  classes:
+    "failsafe_before_after_exists" expression => fileexists("$(sys.inputdir)/failsafe.cf");
```
jimis: typo `s/failsafe_before_after_exists/failsafe_after_test_exists/`

tests/acceptance/00_basics/validation/validate_failsafe_creating.cf.sub
```diff
@@ -0,0 +1,15 @@
...
+
+body common control
+{
+  inputs => { "../../default.cf.sub" };
+  bundlesequence => { default("$(this.promise_filename)") };
+}
+
+bundle agent test
+{
+  my_non_existing_promise_type:
```
jimis: policy syntax error ;-)

jimis:
* +1, my comments are mostly nitpicks
* +2 for the acceptance test
* +3 if you add an almost identical acceptance test (i.e. with a syntax-error main policy), but you have a pre-existing custom `failsafe.cf` that touches a file, and you check after the test that `failsafe.cf` was not overwritten, and that the custom `failsafe.cf` did actually run and touched the file.

cf-agent/cf-agent.c
pasinskim: So assert() should be redundant as well. I think it is possible to return NULL by LoadPolicy() anyway but let's not argue. I will leave assert() as was before.

libpromises/generic_agent.c
pasinskim: But shouldn't we use existing failsafe.cf it it is already there? Does it make sense to generate it all the time?

libpromises/generic_agent.c
pasinskim: Yup. I don't see a need of doing this. IMHO we shouldn't allocate memory dynamically because we can. This should be very unlikely and in fact in all other places it is like that or even worse. We are not checking if buffer is big enough to handle string to be copied there.

tests/acceptance/00_basics/validation/validate_failsafe_creating.cf
pasinskim: Thanks. Fixed.

tests/acceptance/00_basics/validation/validate_failsafe_creating.cf.sub
pasinskim: How's that possible :)

libpromises/generic_agent.c
pasinskim: Actually the previous behavior was wrong. If failsafe.cf doesn't exist (stat returns -1) we are setting failsafe.cf to be exactly the same file which was checked to be not there.
What is more, in GenericAgentDiscoverContext() in case of bootstrap we are wiping out whole inputs directory (RemoveAllExistingPolicyInInputs()) and then we are generating failsafe.cf (GenericAgentDiscoverContext()). So I will right now set failsafe.cf to be policy input file there. 
Thanks for noticing this one.

cf-agent/cf-agent.c
jimis: > So assert() should be redundant as well

Yes, all asserts are redundant. It should **never** trigger. However we put them there to ensure we don't forget our own coding assumptions.

This is the way I'm deciding for using asserts:
* Does the check look redundant? Use `assert()`, it won't cost in runtime for non-debug builds anyway.
* Does the check look necessary? I.e. is there one in a million it will trigger? Then do proper error-handling.

cf-agent/cf-agent.c
```diff
@@ -238,11 +238,15 @@ int main(int argc, char *argv[])
...
-        GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
-        policy = LoadPolicy(ctx, config);
+        
+        if (CheckAndGenerateFailsafe(GetInputDir(), "failsafe.cf"))
+        {
+            GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
+            policy = LoadPolicy(ctx, config);
+        }
     }
     assert(policy);
```
jimis: @pasinskim actually I was wrong that `assert()` is better suited here. `policy` **can** indeed be NULL if `CheckAndGenerateFailsafe()` fails, so your earlier check is better, and `assert()` should be removed.

cf-serverd/cf-serverd.c
```diff
@@ -59,7 +59,11 @@ int main(int argc, char *argv[])
     {
         Log(LOG_LEVEL_ERR, "CFEngine was not able to get confirmation of promises from cf-promises, so going to failsafe");
         EvalContextClassPutHard(ctx, "failsafe_fallback", "attribute_name=Errors,source=agent");
-        GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
+        if (CheckAndGenerateFailsafe(GetInputDir(), "failsafe.cf"))
+        {
+            GenericAgentConfigSetInputFile(config, GetInputDir(), "failsafe.cf");
+            policy = LoadPolicy(ctx, config);
+        }
         policy = LoadPolicy(ctx, config);
```
pasinskim: I have noticed this while fixing other thing. I guess there is no need to load policy twice.

pasinskim:
@jimis all issues fixed. Should be ready to merge now.

jimis:
trigger build

jimis:
I wrote:
> +3 if you add an almost identical acceptance test (i.e. with a syntax-error main policy), but you have a pre-existing custom failsafe.cf that touches a file, and you check after the test that failsafe.cf was not overwritten, and that the custom failsafe.cf did actually run and touched the file.

I have created this acceptance test in pull request #2363. I'll merge that one first, and then re-test this one to make sure your failsafe changes do not break existing behaviour.

jimis:
trigger build

jimis:
All tests pass!

pasinskim:
Thanks @jimis for creating missing acceptance test.

pasinskim:
@jimis is this cherry-picked to 3.7 (and 3.6)?

jimis:
> @jimis is this cherry-picked to 3.7 (and 3.6)?

Not yet.

PR #2358, opened 2015-10-23
Redmine#7643: Hard-code root uid to 0.

jimis:
This is a quick'n'dirty fix for the performance problems that thousands
of getpwnam() calls per second are causing.

Changelog: Improve cf-serverd's lock contention because of getpwnam()
           call. (Redmine #7643)

### DIFF
--- a/cf-serverd/server_common.c
+++ b/cf-serverd/server_common.c
@@ -1670,9 +1670,16 @@ void SetConnIdentity(ServerConnectionState *conn, const char *username)
         memcpy(conn->username, username, username_len + 1);
     }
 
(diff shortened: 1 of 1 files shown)


jimis:
trigger build

jimis:
@kacfengine what do you think about hard-coding root UID to 0?

kacfengine:
@jimis I think it's quite safe. I have not seen a Unix platform yet where root is not 0. Of course someone could change the name, but I can't think of a reason why. The whole identification of user is useless anyway, I mean, anyone can identify as any user, so it doesn't provide any security.

PR #2359, opened 2015-10-23
Add: Acceptance tests for abortbundle classes

nickanderson:
Test that the agent aborts the current bundle when an exact abortclass
is matched and that the agent continues on to the next bundle.

Test that the agent aborts the current bundle when a regular expression
is matched and that the agent continues on to the next bundle.

Test that the agent aborts if a classexpression is matched and that the
agent continutes on to the next bundle.

Test that the agent does *NOT* abort if a classexpression does not
match.

Ref: https://dev.cfengine.com/issues/3611

### DIFF
--- a/tests/acceptance/15_control/02_agent/abortbundle_classes.cf
+++ b/tests/acceptance/15_control/02_agent/abortbundle_classes.cf
@@ -1,0 +1,28 @@
+##########################################################################
+#
+# Test that bundles will abort when an abortbundleclass is found to match
+# exactly. Only the one bundle should be aborted, not the entire agent
+# execution.
+#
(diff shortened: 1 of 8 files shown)


nickanderson:
@kacfengine here it is without the soft fails.

kacfengine:
Ok, thanks! Should I close the Redmine ticket after this is merged?

kacfengine:
trigger build

nickanderson:
@kacfengine yeah I think so. @estenberg  last update to the ticket noted that he updated the docs to reflect its  usage is limited to agent bundles.

kacfengine:

PR #2360, opened 2015-10-26
Revert #6027 fixes due to segfaults in CI

kacfengine:
Check out for example:

* tests/acceptance/10_files/01_create/copy_and_purge.cf


jimis:
Thanks, no idea how the segfault slipped through the CI builds that I triggered. I've opened #2361 that fixes the problem.

PR #2366, opened 2015-10-28
Tty removal

pasinskim:


pasinskim:
@jimis you can take a look at this one. I've changed some code to unify the behavior of policy loading and enabling unit testing. 
@kacfengine I will need some suggestions especially for the second commit. Changing a way we are discovering tty_interactive is causing some tests are failing because of open stdin.

kacfengine:
Aha! That's quite a creative way to test the tty detection, nice!

From commit message:
> Policy loading is now unified doe agent, server and monitor daemon.

"doe"? I don't get the meaning of the sentence.

No code comments, I think it looks nice! Feel free to trigger the build, or let jimis look first.

pasinskim:
trigger build

tests/acceptance/testall
```diff
@@ -53,6 +53,11 @@ if [ -f /usr/xpg4/bin/grep ]; then
 fi
 
 #
+# Close stdin file descriptor to avoid tty_interactive check in cf-agent being success.
+#
+exec <&-
```
jimis: What if we need to use the stdin? Example is ```./testall --gdb``` and I'm not sure if there is more.

kacfengine: Oh, that's true, I didn't think about that. Maybe skip it in that case. You'll have to do the closing further down then, after scanning arguments.

pasinskim:
@kacfengine should be "for" instead of "doe" - fixed. @jimis thanks for pointing this out. Now for gdb and valgring I am not closing stdin so that this is working correctly. Being honest I didn't even know about possibility of using gdb and valgrind with test tool. Nice to know now!

tests/acceptance/testall
```diff
@@ -684,6 +684,13 @@ while true; do
   shift
 done
 
+#
+# Close stdin file descriptor to avoid tty_interactive check in cf-agent being success.
+#
+if ["$GDB" != 1] || ["$USE_VALGRIND" != 1]; then
+  exec <&-
+fi
```
kacfengine: I think POSIX compliance demands spaces between the ``[``, ``]`` and the rest. This might fail on the exotic Unixes without it.

PR #2370, opened 2015-10-29
returnszero() should not output the output of a command

worr:
`returnzero` is a common command used to generate classes based on the return value of a particular command. Users (at least here at LinkedIn) only care about the exit code of the command, and do not want the output mixed in with the output of CFEngine. This ends up being super confusing, and creating a lot of spurious output to wade through.

This has caused our CFEngine developers to pass `useshell` to all of their `returnszero` calls, introducing a more significant cost as well introducing the potential for issues related to shell parsing of commands and influential env vars.

This turns the behavior on its head, making no output the default, and allowing the user to drop to a shell if they want to capture the output somehow. This might be controversial.

### DIFF
--- a/libpromises/unix.c
+++ b/libpromises/unix.c
@@ -24,6 +24,7 @@
 
 #include <unix.h>
 #include <exec_tools.h>
+#include <file_lib.h>
 
 #ifdef HAVE_SYS_UIO_H
 # include <sys/uio.h>
@@ -202,6 +203,24 @@ bool ShellCommandReturnsZero(const char *command, ShellType shell)
         else
         {
             char **argv = ArgSplitCommand(command);
+            int devnull;
+
+            if (LogGetGlobalLevel() < LOG_LEVEL_INFO)
+            {
+                if ((devnull = safe_open("/dev/null", O_WRONLY)) == -1)
+                {
+                    Log(LOG_LEVEL_ERR, "Command '%s' failed. (open: %s)", command, GetErrorStr());
+                    exit(EXIT_FAILURE);
+                }
+
+                if (dup2(devnull, STDOUT_FILENO) == -1 || dup2(devnull, STDERR_FILENO) == -1)
+                {
+                    Log(LOG_LEVEL_ERR, "Command '%s' failed. (dup2: %s)", command, GetErrorStr());
+                    exit(EXIT_FAILURE);
+                }
+
+                close(devnull);
+            }
 
             if (execv(argv[0], argv) == -1)
             {
--- a/tests/acceptance/02_classes/02_functions/silent_returnszero.cf.sub
+++ b/tests/acceptance/02_classes/02_functions/silent_returnszero.cf.sub
@@ -1,0 +1,11 @@
+body common control
+{
+  inputs => { "../../default.cf.sub" };
+  bundlesequence => { "main" };
+  version => "1.0";
+}
+
+bundle agent main {
+  classes:
+	"test" expression => returnszero("$(G.echo) Pass", "noshell");
+}
--- a/tests/acceptance/02_classes/02_functions/silent_returnszero.cf
+++ b/tests/acceptance/02_classes/02_functions/silent_returnszero.cf
@@ -1,0 +1,18 @@
+body common control
+{
+  inputs => { "../../default.cf.sub" };
+  bundlesequence  => { default("$(this.promise_filename)") };
+  version => "1.0";
+}
+
+bundle agent check {
+  vars:
+    "command" string => "$(sys.cf_agent) -KIf $(this.promise_filename).sub -DAUTO";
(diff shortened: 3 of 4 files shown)


kacfengine:
This is indeed quite controversial! :-) I don't think we can change the default behavior, it would be a too big behavior change, and there also would be no way to restore the original behavior.

I have a different suggestion: What about making another ``shell`` option: ``noshell_nooutput`` or something similar (bonus points if you can think of a better name), and then treating it exactly like you do in your change?

libpromises/unix.c
```diff
@@ -202,6 +203,19 @@ bool ShellCommandReturnsZero(const char *command, ShellType shell)
         else
         {
             char **argv = ArgSplitCommand(command);
+            int devnull;
+
+            if ((devnull = safe_open("/dev/null", O_WRONLY)) == -1)
```
kacfengine: Another detail: This is not Windows safe. Put a constant in ``platform.h``, for example ``PLATFORM_NULL``, make it ``nul`` for ``#ifdef _WIN32`` and ``/dev/null`` for all others, and then use the constant here.

worr:
Is it a change that we can make in a minor release? We could also make the output available if a user turns on verbose execution or something.

kacfengine:
By minor you mean 3.8 -> 3.9 (as opposed to patch release 3.8.0 -> 3.8.1)? Hmmm. Tricky question. Introducing an extra `shell` choice, yes definitely. But changing the default behavior... It seems like it would be quite confusing if output suddenly started disappearing when the policy hasn't changed, even if you did upgrade.

I like the idea of making it available in verbose/info mode though, this is perhaps more in line with promise output works elsewhere. I'm still a bit on the fence on this one.

kacfengine:
@nickanderson, @estenberg: What do you think? The idea is to make `returnszero` completely silent unless you either:

1. Turn on verbose or info mode (`-v`/`-I`)
2. Use a shell argument to `returnszero`, like `noshell_output` (name still TBD).

This is generally how most promises work with regards to output, unless they encounter an error.

eysteins:
Does returnszero() produce output during a normal run (i.e. if you just run cf-agent -K)? If so, I think it makes sense to lower the log level.

However, there must be a way to get the output from the returnszero() command -- verbose & debug mode is fine. This is critical in debugging (all kinds of weird environment / shell issues may make this hard to debug otherwise).

libpromises/unix.c
kacfengine: Btw, disregard this, apparently there is a separate Windows implementation of this function.

kacfengine:
Agree. So this is conclusion then:

1. `returnszero` is by default silent.
2. Info mode or higher verbosity turns it on again.
3. (sounds like we drop the `noshell_output` variants, info mode is enough)

Sounds good, @worr?

worr:
Works for me. I'll rework the patch.

worr:
Updated the patch to output at higher verbosities.

Feel free to merge when you get back.

I have no idea how to handle a Windows implementation of this.

kacfengine:
Thanks! I'll check out the Windows part.

kacfengine:
Looks good! The Windows part appears to close all file handles already, so I don't think we need to change anything there.

Can you provide a test for this?

kacfengine:
trigger build

kacf:
The change causes the test `10_files/zendesk_1779/zendesk_1779.cf` to unexpectedly pass (it is expected to fail because of a different bug). However, it doesn't look this change should fix that issue, so it's more likely that the test is passing erroneously. Could you look into that?

The soft fail stuff is a bit confusing, so just to be clear, the expected outcome is "Soft fail", indicating that the test failed (as expected).

worr:
I've taken a look, and it was as simple as adding a `-I` to the invocation of `zendesk_1779.cf.sub`. Since the test relied on the output of the `echo` in the `file_select`, it was being silenced after this patch. By adding `-I`, this allows the output of `echo` to make it to the test.

I've also removed a memory leak.

worr:
Updated commit message and added test.

kacf:
> it was as simple as adding a `-I` to the invocation of `zendesk_1779.cf.sub`.

Ah of course!

tests/acceptance/02_classes/02_functions/silent_returnszero.cf.sub
```diff
@@ -0,0 +1,4 @@
+bundle agent main {
+  classes:
+	"test" expression => returnszero("/bin/echo Pass", "noshell");
```
kacf: Don't forget `$(G.echo)` instead of `/bin/echo` (important for Windows)!

kacf:
trigger build

worr:
Bah, I always forget about Windows.

Fixed.

kacf:
Yeah, it's an easy mistake. Mostly it's just `s,/bin/whatever/,$(G.whatever),`.

Thanks for the effort, pull request merged manually!

PR #2371, opened 2015-10-30
Allow multiple -D or -N flags to be specified on the command line

worr:
Currently, if multiple -D or -N flags are specified on the command-line, this causes all but the last to be ignored. These commits fixes this such that multiple flags can be used, adding more classes to be raised and lowered.

https://dev.cfengine.com/issues/7191

### DIFF
--- a/cf-runagent/cf-runagent.c
+++ b/cf-runagent/cf-runagent.c
@@ -279,9 +279,7 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
             break;
 
         case 's':
-            strlcpy(SENDCLASSES, optarg, CF_MAXVARSIZE);
-
-            if (strlen(optarg) > CF_MAXVARSIZE)
+            if (strlcat(SENDCLASSES, optarg, sizeof(SENDCLASSES)) >= sizeof(SENDCLASSES))
             {
                 Log(LOG_LEVEL_ERR, "Argument too long (-s)");
                 exit(EXIT_FAILURE);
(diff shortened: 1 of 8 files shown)


libutils/set.c
```diff
@@ -91,6 +91,17 @@ void *SetIteratorNext(SetIterator *i)
     return kv ? kv->key : NULL;
 }
 
+void SetJoin(Set *set, Set *otherset)
```
jimis: Can you add a few tests in set_test.c?

cf-runagent/cf-runagent.c
```diff
@@ -279,9 +279,7 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
             break;
 
         case 's':
-            strlcpy(SENDCLASSES, optarg, CF_MAXVARSIZE);
-
-            if (strlen(optarg) > CF_MAXVARSIZE)
+            if (strlcat(SENDCLASSES, optarg, CF_MAXVARSIZE) >= sizeof(SENDCLASSES))
```
jimis: We prefer to use `sizeof(var)` where possible, instead of `CF_MAXVARSIZE` directly. I know there is a lot of legacy code doing the opposite, but we're trying to be consistent in the new code.

jimis:
+1 thanks for taking care of the WAZA TODO comments! :-)

Your patch is nice and straightforward, It should be ready to merge once the minor issues are addressed.

worr:
I've added tests and taken care of the `sizeof` issue

jimis:
trigger build

cf-runagent/cf-runagent.c
```diff
@@ -279,9 +279,7 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
             break;
 
         case 's':
-            strlcpy(SENDCLASSES, optarg, CF_MAXVARSIZE);
-
-            if (strlen(optarg) > CF_MAXVARSIZE)
+            if (strlcat(SENDCLASSES, optarg, sizeof(DEFINECLASSES)) >= sizeof(SENDCLASSES))
```
jimis: `s/DEFINECLASSES/SENDCLASSES/`

worr:
Oh, whoops, sorry. Fixed.

jimis:
@worr: one last thing, can you reference the redmine ticket and add a "Changelog:" tag in one of your commits? See `HACKING.md` for more info and recent commits for examples.

worr:
All set. I've installed the git hooks on my dev boxes, so this should be taken care of in the future.

jimis:
Great, thanks!

PR #2372, opened 2015-10-30
EvalContextClassPutSoft ought to be namespace-aware

worr:
`EvalContextClassPutSoft` really ought to be namespace aware. This fixes bugs like not being able to specify a namespaced class on the command-line.


kacf:
I put a slightly optimized version in #2463.

worr:
wfm, I should have done that in the first place. :smile: Thanks!

PR #2373, opened 2015-10-30
Fix init_script_test_helper not being cleaned properly.

kacfengine:
Instead of making a separate clean target, list it in PROGRAMS
instead, which is better automake style. This implicitly invokes
libtool as well, so adapt the test to account for that.


kacfengine:
trigger build

PR #2386, opened 2015-11-03
redmine 7137: log variable definitions in verbose output

worr:
This adds logging for mutating any variables in verbose output.

I originally dropped this into VerifyVarPromise, but the output was surprisingly confusing.

https://dev.cfengine.com/issues/7137

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -1426,6 +1426,59 @@ static PromiseResult DefaultVarPromise(EvalContext *ctx, const Promise *pp)
     return VerifyVarPromise(ctx, pp, true);
 }
 
+static void LogVariableValue(const EvalContext *ctx, const Promise *pp)
+{
+    VarRef *ref = VarRefParseFromBundle(pp->promiser, PromiseGetBundle(pp));
+    DataType type;
+    char *out = NULL;
+    size_t siz = CF_BUFSIZE;
+    size_t len = 0;
+    Writer *w;
+
+    const void *var = EvalContextVariableGet(ctx, ref, &type);
+    switch (type)
+    {
+        case CF_DATA_TYPE_INT:
+        case CF_DATA_TYPE_REAL:
+        case CF_DATA_TYPE_STRING:
+            out = xstrdup((char *)var);
+            break;
+        case CF_DATA_TYPE_INT_LIST:
+        case CF_DATA_TYPE_REAL_LIST:
+        case CF_DATA_TYPE_STRING_LIST:
+            out = xcalloc(1, CF_BUFSIZE);
+
+            for (Rlist *rp = (Rlist *)var; rp != NULL; rp = rp->next)
+            {
+                if (strlen((char *)rp->val.item) + len + 3  >= siz)  // ", " + NULL
+                {
+                    out = xrealloc(out, siz + CF_BUFSIZE);
+                    siz += CF_BUFSIZE;
+                }
+
+                if (len)
+                {
+                    len += strlcat(out, ", ", siz);
+                }
+
+                len += strlcat(out, (char *)rp->val.item, siz);
+            }
+            break;
+        case CF_DATA_TYPE_CONTAINER:
+            w = StringWriter();
+            JsonWriteCompact(w, (JsonElement *)var);
+            out = StringWriterClose(w);
+            break;
+        default:
+            assert(false && "Unknown variable type");
+            out = xstrdup("NYI");
+            break;
+    }
+
+    Log(LOG_LEVEL_VERBOSE, "V:     \"%s\" set to \"%s\"", pp->promiser, out);
+    free(out);
+}
+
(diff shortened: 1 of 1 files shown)


atsaloli:
Awesome sauce!  Thanks, Will!!

nickanderson:
YES +1

nickanderson:
MMMM someone has to play devils advocate ....

What if I have a secret in a variable? Perhaps a password that I read in from an encrypted file.

atsaloli:
Nick, the logging of variables is in verbose mode only.  I don't see a security concern?  In normal operation, variable values will not get logged.

Thank you for playing devil's advocate.

cf-agent/cf-agent.c
```diff
@@ -1435,6 +1435,58 @@ static PromiseResult DefaultVarPromise(EvalContext *ctx, const Promise *pp)
...
+                len += strlcat(out, (char *)rp->val.item, siz);
+            }
+            break;
+        case CF_DATA_TYPE_CONTAINER:
+            w = StringWriter();
+            JsonWriteCompact(w, (JsonElement *)var);
+            out = StringWriterClose(w);
+            break;
+        default:
+            out = xstrdup("NYI");
```
kacf: I guess this shouldn't happen, or? Might be a good place for an `assert()` so that we can catch it in debug mode.

cf-agent/cf-agent.c
```diff
@@ -1447,6 +1499,7 @@ static PromiseResult KeepAgentPromise(EvalContext *ctx, const Promise *pp, ARG_U
         if (result != PROMISE_RESULT_FAIL)
         {
             Log(LOG_LEVEL_VERBOSE, "V:     Computing value of \"%s\"", pp->promiser);
+            LogVariableValue(ctx, pp);
```
kacf: I would put a check for `LogGetGlobalLevel() >= LOG_LEVEL_VERBOSE` here to avoid the runtime cost of constructing the output string when it's not going to be printed.

kacf:
Sorry for picking this up so late. I think this should go in, with the small modifications suggested.

worr:
updated

kacf:
trigger build

kacf:
Thanks! However, this causes the test `./10_files/templating/timed/expired_edit_line_locks.cf` to fail. It's not immediately clear to me why. It doesn't look related, but it fails consistently everywhere.

Can you take a look at that?

kacf:
Any progress here?

worr:
I'll try and take a look. I've been super busy moving house and at work, so I haven't had a lot of time for FOSS as of late.

cf-agent/cf-agent.c
```diff
@@ -1426,6 +1426,59 @@ static PromiseResult DefaultVarPromise(EvalContext *ctx, const Promise *pp)
...
+                len += strlcat(out, (char *)rp->val.item, siz);
+            }
+            break;
+        case CF_DATA_TYPE_CONTAINER:
+            w = StringWriter();
+            JsonWriteCompact(w, (JsonElement *)var);
+            out = StringWriterClose(w);
+            break;
+        default:
+            assert(false && "Unknown variable type");
```
jimis: segfault here. BTW try to use our own debug instrumentation, for example `CF_ASSERT` would have been able to give some info here, like the value of `type`.

jimis:
+1 for the idea, much needed. The security concern is valid though, /maybe/ printing every variables should be DEBUG or even a special verbosity flag. Not sure anyway.

Kristian, is it OK if I take this one? I actually cherry-picked this because I needed it, and now I'm fixing some implementation issues it has.

cf-agent/cf-agent.c
jimis: Huh, after the `assert` call you just continue the logic. What's the intent here? What does "NYI" mean?

jimis: I've cherry-picked your commit so I changed the log line, and here is what I get:
```
cf-agent.c:1490: Unexpected Error - this is a BUG, please report it: Unknown variable type: 16
```

kacf:
> Kristian, is it OK if I take this one? I actually cherry-picked this because I needed it, and now I'm fixing some implementation issues it has.

The code looks good, but this failed in Jenkins. See the comment about `./10_files/templating/timed/expired_edit_line_locks.cf` above. It doesn't look related, but it fails consistently, and on every platform, so it is related *somehow*. If you fix it, then I'm all for it! :-)

cf-agent/cf-agent.c
jimis: It seems that `CF_DATA_TYPE_NONE` is the type of meta variables like this:
```
    meta: 
        "test_skip_unsupported" string => "!linux";
```
I think this is a separate bug.

jimis:
I've taken it, and will submit a new pull request soon.

Two things concern me currently, and I'd appreciate opinions:
1. security implications
2. huge variables are very messy when logged: should we just limit the length and use "..." at the end of the output?

jimis:
Opened #2522.

PR #2388, opened 2015-11-03
Cherries 36x

jimis:


jimis:
All tests pass here, merging.

Still please have a quick look because cherry-picking was full of conflicts and tedious!

PR #2389, opened 2015-11-03
Add: Test that missed variables are empty strings in mustache

nickanderson:
Ref: Redmine #7684


nickanderson:
This test is passing, just testing to be sure we do not lose the behaviour.

PR #2390, opened 2015-11-04
Redmine #7030: No longer hang when changing permissions/ownership on fifos

kacfengine:
Supersedes https://github.com/cfengine/core/pull/2276.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1023,10 +1023,16 @@ AC_REPLACE_FUNCS(pthread_attr_setstacksize)
 AC_CHECK_DECLS(pthread_sigmask, [], [], [[#include <signal.h>]])
 AC_REPLACE_FUNCS(pthread_sigmask)
 
-AC_CHECK_DECLS([openat], [], [], [[#include <fcntl.h>]])
-AC_CHECK_DECLS([fstatat], [], [], [[#include <sys/stat.h>]])
-AC_CHECK_DECLS([fchownat], [], [], [[#include <unistd.h>]])
-AC_REPLACE_FUNCS(openat fstatat fchownat)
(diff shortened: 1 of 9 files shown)


kacfengine:
trigger build

kacfengine:
Thanks to the tests I discovered two subtle security problems with the original approach.

1. The EUID approach would only be effective on platforms which use the wrapper. Therefore I moved that part into the ``safe_chmod_impl`` instead, to be cross platform. The only difference from the original behavior is that we are no longer able to differentiate between an unsafe link and an operation that is denied for other reasons, and hence we will return ``EPERM`` in both cases.
2. ``fchownat``, when called without the ``AT_SYMLINK_NOFOLLOW`` flag, will happily follow a link that has a different owner than its target, allowing users to change owners of arbitrary files on the system if CFEngine is operating in a directory of a non-trusted user. The normal approach has been to open the file with ``safe_open`` first, but we cannot do that here because it would defeat the original purpose of being able to operate on FIFOs. So instead we follow links manually until we reach the first non-link target, and then we switch on the ``AT_SYMLINK_NOFOLLOW`` flag to do the final chown operation. The link can still be replaced in this little window of time, but thanks to the flag the operation will not change an arbitrary target file, but only the link itself, which poses no security risk, since we already know that the link that led there has the same owner.

@worr: Since we don't open any files, I believe this still solves your original problem, but it would be nice if you could take a look at the code.

kacfengine:
Thanks for reviewing!

kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
Alright, finally passes everywhere.

kacfengine:
This took a while for various reasons, but finally we fixed it! Thanks for the perseverence!

PR #2391, opened 2015-11-05
process_test: Try to work around slow buildslaves by giving more time.

kacfengine:
They must be swapping or something, because the logs say it can take
up to ten seconds for the first process to even start.


kacfengine:
@jimis: Unsure if this will work, but it's worth a shot.

jimis:
Makes sense. We should also have a look again in the load of our storage array, now that @Enemico is gaining access there.

kacfengine:
trigger build

PR #2397, opened 2015-11-06
Redmine#7346: add regex_replace example

tzz:
@jimis this example will be used in the docs


jimis:
trigger build

jimis:
Thanks!

PR #2398, opened 2015-11-06
Minor

jimis:


jimis:
@tzz I changed slightly your test in commit 6e272cb so please review!

jimis:
trigger build

tzz:
Seems good to me.

PR #2400, opened 2015-11-11
evalfunction.c: file-size arg of readjson() and readyaml() should be optional

soumyadipdm:
A policy writer should never have to worry about low-level details like file-size while using readjson() or readyaml(), not even "inf". This pull request makes the 2nd argument of readjson() and readyaml() optional. If the 2nd argument is not specified, file size will default to "inf".

This will help a lot while parsing dynamically generated json/yaml files which can grow or shrink over the time. And policy writer does not need to "remember" to specify size arg as "inf".


basvandervlies:
I completely agree. i completely switched to readdata for this reason

nickanderson:
trigger_build

nickanderson:
I am not in love with the idea, but I appreciate efforts to make the language easier to use. I think this would improve usability so +1

nickanderson:
Thinking about this more.  I wonder if it would be better to have a default maxbytes control option.

libpromises/evalfunction.c
```diff
@@ -6021,7 +6021,11 @@ static FnCallResult FnCallReadData(ARG_UNUSED EvalContext *ctx,
     }
     else
     {
-        size_max = args->next ? IntFromString(RlistScalarValue(args->next)) : 0;
+        if (RlistLen(args) == 0)
+        {
+            FatalError(ctx, "Function '%s' requires at least one argument", fp->name);
```
kacfengine: Maybe ``FatalError`` is a bit brutal here, it will terminate the agent. Since this will only be triggered at runtime, after the policy has been parsed, you may verify the policy on the hub, thinking that everything is fine, but then the agent terminates on your clients because of different conditions. Perhaps returning ``FnFailure()`` is better. What do you think?

kacfengine:
Looks good otherwise.

libpromises/evalfunction.c
soumyadipdm: Of course you're right, we would not want cf-agent to die. I will close this one and submit a different PR, since I deleted my fork's branch thinking nobody likes the idea :) 

new PR: https://github.com/cfengine/core/pull/2417

soumyadipdm:
closing this one in favor of https://github.com/cfengine/core/pull/2417

PR #2401, opened 2015-11-12
Runagent acceptance tests

jimis:


jimis:
trigger build

jimis:
Pull requests in masterfiles and enterprise add `--cfrunagent` parameter to the `testall` invocation in their acceptance tests.

jimis:
It failed all over the place in Jenkins, investigating.

jimis:
Strange, the new tests are passing after all, it's an "rsync" command failing at the end of the Jenkins job.

jimis:
trigger build

jimis:
trigger build

jimis:
Hopefully my last commit will fix tests execution, even before merging the `testall` changes in enterprise and masterfiles.

jimis:
The last commit is full of whitespace changes, you might want to skip it if you're working on `testall`, as it will cause lots of conflicts.

jimis:
trigger build

jimis:
trigger build

jimis:
Looks like it's passing now. @kacfengine lots of changes here, but you might want to do a quick review.

jimis:
@kacfengine still want to have a quick look?

kacfengine:
Just a quick look as promised, but in short I love it! Great to have this tested!

This can probably be for later, but if you want to optimize this, you may want to look at this particular section: https://github.com/cfengine/core/blob/master/tests/acceptance/README#L120-L130. I'm not sure if it applies here (I see you have several different server setups), but perhaps running time could be reduced by grouping tests by which server instance they use, and then executing those in parallel. Not urgent, but just so you know it's possible! :-)

jimis:
Interesting, but in short I disagree! :-)

> Serial tests enforce strictly sorted order when they are encountered, so you can
> use them to initialize some precondition for a set a of tests by sorting it
> lexicographically before the others and giving it a name containing "serial".
> The other tests can then be executed in parallel, unless they need to be serial
> for other reasons. For example, you can use two serial tests, one to set up a
> cf-serverd, and one to tear it down, and in between you put normal tests.

This would create interdependencies between the tests that are not clearly distinguished. And running `testall` with a single test would not guarantee the proper setup.

In order to parallelize cf-serverd tests, I'd rather have a `_init.cf.sub` in the tests directory that is executed in advance and sets up the servers, and a `_destroy.cf.sub` that is executed in the end. `testall` should know about it and the tests Makefile should enforce the dependencies when these files are detected, ofcourse.

However these cf-runagent tests are a bit trickier than that: they alter the **server's** environment, not the runagent's environment. As a result, the final "check" bundle must be run inside the cf-serverd execution environment. And obviously many parallel connections to the same cf-serverd would skew the results.

So I don't have a way in mind to parallelize the cf-runagent tests.

PR #2403, opened 2015-11-13
Redmine #7460: Unable to edit the same value in multiple regions.

pasinskim:
The issue was that even though we've been editing value in selected
region the whole file was checked to see if value was already edited.
This was broken by commit f8c7dca55ed8d99212564ea130666de4bd5e0931 when
check if line exists in region was added.

As a bonus some optimizations are added. Instead of iterating
through whole file we can now iterate only over selected region as
any operations will be done inside the region (no need to check whole
file).

Changelog: It is possible to edit the same value in multiple regions
of one file. (Redmine #7460)

### DIFF
--- a/cf-agent/files_editline.c
+++ b/cf-agent/files_editline.c
@@ -888,7 +888,10 @@ static int InsertMultipleLinesToRegion(EvalContext *ctx, Item **start, Item *beg
 
     if (a.location.before_after == EDIT_ORDER_BEFORE)
     {
-        for (ip = *start; ip != NULL; ip = ip->next)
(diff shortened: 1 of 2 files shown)


pasinskim:
trigger build!

pasinskim:
trigger build

pasinskim:
@kacfengine @jimis build is green. All tests are passing and the suppressed test is working now. Speak now if you think I shouldn't merge this one.

jimis:
@pasinskim go ahead, I now proclaim you our new edit_lines expert!

PR #2406, opened 2015-11-16
Refactor common.sh and stage script for better code reuse

mikeweilgart:
Make code re-use easier by putting function defs from masterfiles-stage.sh
into the sourced code library common.sh

Also moved sourcing of $PARAMS from common.sh to masterfiles-stage.sh
so common.sh can be sourced from stage script implementations which
don't rely on $PARAMS variable.


mikeweilgart:
This should make it easier for CFEngine customers to do their own implementations or customizations of multiple policy channels, without reinventing the staging script code base here - instead they can just source `common.sh` and use the functions contained therein.

contrib/masterfiles-stage/masterfiles-stage.sh
```diff
@@ -111,93 +31,7 @@ git_tag_or_commit_masterstage() {
...
-                rsync -CrltDE -c --delete-after --chmod=u+rwX,go-rwx "${TMP}/" "${MASTERDIR}/" && \
-                rm -rf ${TMP}/.svn && \
-                echo "Successfully staged a policy release on $(date)"
-	    )
-	fi
-    else
-       error_exit "The staged policies in ${TMP} could not be validated, aborting."
-    fi
-}
+  STAGING_DIR="${ROOT}/_tmp${TRANSLATED_PARAMS}"
```
nickanderson: agreed STAGING_DIR is a better name for the var.

But, I think that TRANSLATED_PARAMS needs to be set inside the vcs implementation.
I don't think the path to the params will be sufficent protection in all cases. For example I think if you changed from one repo to another you could have bad effects, so for a git implementation it might make more sense to include the git server and repo in that string that gets translated.

maybe instead of translated params a better name would be UPSTREAM_STAGING_DIR?

I am bad with names

nickanderson:
These changes look pretty good to me. I don't consider the TRANSLATED_PARAMS thing necessary for merging this, but I need to test this myself first. Will try to get to it next week. But I think this is looking good, thanks!

mikeweilgart:
Great!

I actually made considerably more changes in the process of testing; I'll put those on my fork early next week.

Also if the STAGING_DIR should be set differently depending on VCS_TYPE, it could be set in the `select` switch in masterfiles-stage.sh, allowing for easier adjustment (by the customer) of exactly how it is set without requiring modification to the upstream common.sh.

And, for our own purposes at DirecTV I'll actually be integrating more closely with git.  (This is not implemented yet, but—)  The git fetch et. al. takes 7-9 seconds, and currently that's *per channel*.  I'll be changing things on our end to use a *single* .git clone to back *all* channels' staging dirs, so there will only be a single expensive `git fetch` operation no matter how many channels we have.  (It does require limiting us to a single associated git repo per hub, but that's what *we* are using anyway.)

nickanderson:
Yes that optimization sounds interesting.  It won't work for everyone but I'm interested in having the ability to leverage that feature.

It may be time to add getopts

nickanderson:
Superceeded by #2429 
Your changes are included there. Thanks!

nickanderson:
This PR should be closed.

nickanderson:
ping @jimis or @kacf can you close this PR, the changes were merged in a different commit.

PR #2407, opened 2015-11-19
Disable block buffering on stdout on Windows.

kacfengine:
Changelog: CFEngine on Windows no longer truncates log messages if the
program in question is killed halfway through.

### DIFF
--- a/libcompat/setlinebuf.c
+++ b/libcompat/setlinebuf.c
@@ -32,7 +32,19 @@
 void setlinebuf(FILE *stream);
 #endif
 
+#ifdef _WIN32
+/*
+  Line buffered mode doesn't work on Windows, as documented here:
+  https://msdn.microsoft.com/en-us/library/86cebhfs.aspx (setvbuf)
+  It will fall back to block buffering, so in that case it is better to select
+  no buffering.
+*/
+# define IO_MODE _IONBF
+#else
+# define IO_MODE _IOLBF
+#endif
+
 void setlinebuf(FILE *stream)
 {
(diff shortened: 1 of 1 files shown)


kacfengine:
@estenberg, @nickanderson, @jimis: FYI. This should help a lot when looking at Windows crash logs.

kacfengine:
trigger build

jimis:
I guess you tested it and works as expected? What about performance? Is there a big difference in runtime before and after? Cancelling all buffering might be a bit risky.

In addition you might want to change one line in `generic_agent.c` to `setlinebuf(stderr)` together with `stdout`.

kacfengine:
@jimis: Yep, I tested intentionally crashing the agent midway, and with this change I get all log output. I'm not so worried about performance, running in verbose or debug mode is not the standard mode of operation, the standard way is for the agent to report to the Windows system event log, which is not affected by this. But for what it's worth, I couldn't detect any change in performance while I was running it. Note that ``Log()`` always writes complete lines anyway, so whether you use line buffering or not is not so important, in practice it will be the same performance.

kacfengine:
> In addition you might want to change one line in ``generic_agent.c`` to ``setlinebuf(stderr)`` together with ``stdout``.

I found this line: https://github.com/cfengine/core/blob/master/libutils/logging.c#L156. I'm thinking it's better to leave it alone, cf-agent doesn't print anything there, AFAIK.

kacfengine:
Passed in Jenkins.

kacfengine:

PR #2408, opened 2015-11-19
Add extensive debugging output to cf_lock operations.

kacfengine:
The motivation is to try to track down really hard-to-find bugs where
CFEngine hangs on a lock.

The reason for not putting the code directly in the DB operations
themselves is that we want the cleartext keys for the locks, and they
are only available higher up in the callstack.

### DIFF
--- a/libpromises/locks.c
+++ b/libpromises/locks.c
@@ -45,6 +45,13 @@
 
 #define CF_CRITIAL_SECTION "CF_CRITICAL_SECTION"
 
+#define LOG_LOCK_ENTRY(__lock, __lock_sum, __lock_data)         \
+    log_lock("Entering", __FUNCTION__, __lock, __lock_sum, __lock_data)
+#define LOG_LOCK_EXIT(__lock, __lock_sum, __lock_data)          \
+    log_lock("Exiting", __FUNCTION__, __lock, __lock_sum, __lock_data)
+#define LOG_LOCK_OP(__lock, __lock_sum, __lock_data)            \
+    log_lock("Performing", __FUNCTION__, __lock, __lock_sum, __lock_data)
+
 typedef struct CfLockStack_ {
     char lock[CF_BUFSIZE];
     char last[CF_BUFSIZE];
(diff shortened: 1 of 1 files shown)


kacfengine:
@estenberg: What do you think about this one? I'm thinking about including this in 3.7 in order to try to debug some of the very elusive hanging bugs we've seen on Windows lately (ref: https://cfengine.zendesk.com/agent/tickets/2611).

@jimis: You might want to have a look too, maybe you have other ideas for debugging this as well.

kacfengine:
trigger build

kacfengine:
trigger build

jimis:
Didn't look it yet, but maybe grep'ing the DEBUG logs for Acquire/Yield gives extensive lock info, from before hashing.

kacfengine:
I thought about it too, but ``AcquireLock``/``YieldLock`` do a lot of things, and I'd like to catch exactly where it happens, because I suspect it may be a very particular key. And it also makes it easier if we get a lock file from the customer to try to use *exactly* the same key he did.

libpromises/locks.c
```diff
@@ -45,6 +45,13 @@
 
 #define CF_CRITIAL_SECTION "CF_CRITICAL_SECTION"
 
+#define LOG_LOCK_ENTRY(__lock, __lock_sum, __lock_data)         \
+    log_lock("Entering", __FUNCTION__, __lock, __lock_sum, __lock_data)
+#define LOG_LOCK_EXIT(__lock, __lock_sum, __lock_data)          \
+    log_lock("Exiting", __FUNCTION__, __lock, __lock_sum, __lock_data)
+#define LOG_LOCK_OP(__lock, __lock_sum, __lock_data)            \
+    log_lock("Performing", __FUNCTION__, __lock, __lock_sum, __lock_data)
+
```
jimis: Nice and very detailed :-)

jimis:
I'm looking forward to seeing the logs from this one. :-)

estenberg:
I can only comment on the intention; and it's great. Lock-hanging is a very common and ugly problem.

kacfengine:

PR #2409, opened 2015-11-19
Change: Soften legacy package promise implementation warning

nickanderson:
Ref: https://dev.cfengine.com/issues/7603

### DIFF
--- a/cf-agent/verify_packages.c
+++ b/cf-agent/verify_packages.c
@@ -201,8 +201,8 @@ PromiseResult VerifyPackagesPromise(EvalContext *ctx, const Promise *pp)
             break;
         case PACKAGE_PROMISE_TYPE_OLD:
             Log(LOG_LEVEL_VERBOSE,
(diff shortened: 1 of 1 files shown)


nickanderson:
I think this should go into 3.8 and also be picked into 3.7.x

nickanderson:
ping @kacfengine @estenberg

estenberg:
Looks good. Only thing is if we should set a release here so people have time to migrate. I propose "Removed from 3.11 and later versions".

kacfengine:
> Only thing is if we should set a release here so people have time to migrate. I propose "Removed from 3.11 and later versions".

In principle I agree, but I don't think we can do that unless we have a clear plan for when some of the more exotic package managers will be replaced. Inventory still depends on the old package promise for many platforms, for example.

kacfengine:
Cherry-picked to 3.8.x as well.

kacfengine:
Oh, and 3.7.x too (I didn't think the message existed there).

PR #2412, opened 2015-11-20
Change: Improve example

nickanderson:
The comment regex matches multiline, and the example did not illustrate
that behaviour.

### DIFF
--- a/examples/data_readstringarray.cf
+++ b/examples/data_readstringarray.cf
@@ -23,8 +23,10 @@
 #+begin_src prep
 #@ ```
 #@ echo a,b,c > /tmp/cfe_array
+#@ echo "# This is a comment" >> /tmp/cfe_array
 #@ echo d,e,f >> /tmp/cfe_array
 #@ echo g,h,i >> /tmp/cfe_array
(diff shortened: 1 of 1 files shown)


nickanderson:
Related to fixing documentation as described in https://dev.cfengine.com/issues/6478

nickanderson:
trigger build

examples/data_readstringarray.cf
```diff
@@ -38,8 +40,12 @@ body common control
 bundle agent example
 {
   vars:
-      "bykey" data => data_readstringarray("/tmp/cfe_array","#.*",",",10,400);
-      "byint" data => data_readstringarrayidx("/tmp/cfe_array","#.*",",",10,400);
+      # The comment regex warrents an explination:
```
kacfengine: ...*warrants* an *explanation*...

kacfengine:
FYI, the PR analyzer doesn't work for non-master branches.

In general it's better to submit just one pull request for master where you mention which branches it should go into, and the cherry-pick will happen by the merger.

nickanderson:
ok thanks @kacf

PR #2413, opened 2015-11-20
Change: Improve example

nickanderson:
The comment regex matches multiline, and the example did not illustrate
that behaviour.

Related to fixing documentation as described in https://dev.cfengine.com/issues/6478


nickanderson:
trigger build

PR #2418, opened 2015-11-24
bugfix PromiseHash(): RvalHash() should use hash yielded by StringHash() as its seed

soumyadipdm:
Currently line# 166 is useless in `PromiseHash()`.
`RvalHash()` in line# 167 should be using hash generated by `StringHash()` as its seed.


kacfengine:
Agreed. Since this will change the hash of every single promise, it might have a large "inflation" effect on the lock database, therefore I think this should be committed to the next feature release only. Technically it is a bug fix, but the risk is significant compared to the gain.

Will run this also through the build machinery soon, but letting the resources be used for the release builds right now.

kacfengine:
trigger build

kacfengine:
Thanks!

PR #2419, opened 2015-11-25
Use new EXEC code for classic protocol, remove old.

jimis:


jimis:
trigger build

jimis:
While Jenkins is going on, I'm manually testing with 3.6.x cf-runagent, just to make sure that the new cf-serverd code is compatible with the old client.

cf-serverd/server_classic.c
```diff
@@ -1142,36 +1142,28 @@ int BusyWithClassicConnection(EvalContext *ctx, ServerConnectionState *conn)
...
 
-        if (!MatchClasses(ctx, conn))
-        {
-            Log(LOG_LEVEL_INFO,
-                "REFUSAL due to failed class/context match");
-            Terminate(conn->conn_info);
-            return false;
-        }
+        bool b = DoExec2(ctx, conn, args,
+                         sendbuffer, sizeof(sendbuffer));
```
kacfengine: Should this be called `DoExec` now? :-)

kacfengine:
+1 for code removal! Looks good!

cf-serverd/server_classic.c
jimis: Indeed. I think I'll leave it like that to honour the legacy, and hopefully we won't ever see a "3" version... ;-)

jimis:
Manual tests OK, auto tests in Jenkins mostly green, so I'll merge as soon as all platforms are finished. Thanks for reviewing!

PR #2421, opened 2015-11-26
Redmine#7765: Log INFO messages to Mission Portal even if not logging…

jimis:
… to console nor syslog.

### DIFF
--- a/libpromises/eval_context.c
+++ b/libpromises/eval_context.c
@@ -283,20 +283,6 @@ const char *EvalContextStackToString(EvalContext *ctx)
     return "";
 }
 
-static char *LogHook(LoggingPrivContext *pctx, LogLevel level, const char *message)
-{
-    const EvalContext *ctx = pctx->param;
-
-    StackFrame *last_frame = LastStackFrame(ctx, 0);
-    if (last_frame
-        && last_frame->type == STACK_FRAME_TYPE_PROMISE_ITERATION
-        && level <= LOG_LEVEL_INFO)
-    {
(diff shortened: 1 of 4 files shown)


jimis:
This is not good:
1. it fails in `logging_timestamp_test.c` and I can't figure out why!
2. I'm using an ugly `#ifdef` to hide enterprise-specific code. I have no clue if this is correct, but it's surely ugly. Do you know any valid way to distinguish inside libpromises whether the enterprise plugin is loaded?

pasinskim:
@jimis go home. One way of doing this migt be moving this stuff to separate
function and create enterprise stub. Then in core you can do nothing and
implement needed thing in enterprise.
I don't believe I am replaying at 12:30...

jimis:
This is ridiculous; quick, somebody erase the timestamps!

kacfengine:
@pasinskim's suggestion is the correct approach. You can't use an ifdef, since Community and Enterprise are the same binary.

kacfengine:
But isn't it much easier to simply avoid the optimization for INFO messages? I mean, *by far* the greatest bottleneck here are the DEBUG messages, and both those and VERBOSE messages will still be covered by the optimization.

jimis:
Kristian, I knew the ifdef wouldn't work, but I couldn't find some global variable either when the enterprise plugin is loaded. Are you aware of any?

I'll probably go with moving this code in Enterprise, but it's more overhead than I had hoped for.

> But isn't it much easier to simply avoid the optimization for INFO messages?

I can't just arbitrarily do "if INFO then always run hook", this is a generic logging library.

jimis:
trigger build

PR #2422, opened 2015-11-27
Fix unexpanded policy_server variable in failsafe.

pasinskim:


pasinskim:
There is no $sys.policy_server variable. We are having only $sys.policy_hub.
See this: ./libpromises/bootstrap.c:        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, "policy_hub", new_policy_server, CF_DATA_TYPE_STRING, "source=bootstrap");

kacfengine:
Cherry-picked to both 3.7.x and 3.8.x as well. Didn't apply on 3.6.x.

PR #2425, opened 2015-12-03
Add "canonify" mode to mapdata().

tzz:
This change includes acceptance tests, showing usage and effect.

It addresses a very common need: canonify a list without intermediate arrays.  See https://dev.cfengine.com/issues/2377 for one of many requests for it.

Syntax:
```
"canonified" data => mapdata("canonify", "any use of $(this.k) and $(this.v)", any_list_or_container_var);
```

This is 100% backwards compatible and very safe.  Please consider merging to 3.7 and 3.8 as well, even though it's a feature.

In addition to implementing the feature, this PR:

* adds a general `BufferCanonify` function to simplify code
* adds a note to `CanonifyName` (which has a `CF_BUFSIZE` limitation) that it should be replaced with `BufferCanonify` or `CanonifyNameInPlace`
* moves `CanonifyNameInPlace` to `string_lib.[ch]` because it's more appropriate there instead of `files_names.[ch]`
* fixed the doc of `BufferRewrite` because it doesn't return anything

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -2708,7 +2708,8 @@ static FnCallResult FnCallMapData(EvalContext *ctx, ARG_UNUSED const Policy *pol
         varname = RlistScalarValue(finalargs->next);
     }
 
-    bool jsonmode = 0 == strcmp(conversion, "json");
+    bool jsonmode = (0 == strcmp(conversion, "json"));
+    bool canonifymode = (0 == strcmp(conversion, "canonify"));
 
     VarRef *ref = ResolveAndQualifyVarName(fp, varname);
     if (!ref)
@@ -2773,6 +2774,11 @@ static FnCallResult FnCallMapData(EvalContext *ctx, ARG_UNUSED const Policy *pol
                 return FnFailure();
             }
 
+            if (canonifymode)
+            {
+                BufferCanonify(expbuf);
+            }
+
             RlistAppendScalar(&returnlist, BufferData(expbuf));
             EvalContextVariableRemoveSpecial(ctx, SPECIAL_SCOPE_THIS, "v");
 
@@ -2816,6 +2822,11 @@ static FnCallResult FnCallMapData(EvalContext *ctx, ARG_UNUSED const Policy *pol
                     BufferDestroy(expbuf);
                     JsonDestroy(container);
                     return FnFailure();
+                }
+
+                if (canonifymode)
+                {
+                    BufferCanonify(expbuf);
                 }
 
                 RlistAppendScalarIdemp(&returnlist, BufferData(expbuf));
@@ -7916,7 +7927,7 @@ static const FnCallArg MAPARRAY_ARGS[] =
 
 static const FnCallArg MAPDATA_ARGS[] =
 {
-    {"none,json", CF_DATA_TYPE_OPTION, "Conversion to apply to the mapped string"},
+    {"none,canonify,json", CF_DATA_TYPE_OPTION, "Conversion to apply to the mapped string"},
     {CF_ANYSTRING, CF_DATA_TYPE_STRING, "Pattern based on $(this.k) and $(this.v) as original text"},
     {CF_IDRANGE, CF_DATA_TYPE_STRING, "CFEngine array or data container identifier, the array variable to map"},
     {NULL, CF_DATA_TYPE_NONE, NULL}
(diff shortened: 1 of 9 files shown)


nickanderson:
Oh nice, so I just ran into a case where I needed to escape a list as well, maybe escape would be an additional option in place of the canonify. If not appropriate its OK I was able to accomplish what I needed with intermediary stuff, kind of hacky but hey, it worked.

    # We need to do similar for the entries that we want to purge
    "len" int => length(entries_to_purge);
    "idx" slist => expandrange("[0-$(len)]", 1);
    "escaped[$(idx)]" string => escape( nth( entries_to_purge, $(idx)));
    "clean_unwanted" slist => getvalues(escaped);

tzz:
@nickanderson as I was thinking about your suggestion I realized there are many other such cases: upcase, downcase, escape, dirname, basename, math, host2ip, ip2host, file_hash, hash_md5, hash_sha1, and so on.

What if I changed this into a generic `eval` mode as follows:

* `mapdata(eval, "canonify($(this.v))", mycontainer)`
* it only supports a single function call on the outside in the transformation spec, so you can't for example say `escape(canonify(...))` (this will make parsing easier)
* it supports any function that takes a scalar and returns a scalar (so e.g. `file_hash` would be supported implicitly but we don't have to write special code for it)
* to nest function calls, you need to make multiple calls to `mapdata`

On the C side I don't think this will be too hard. For me it would be nice, and I've needed it many times, but it would be great to get feedback from other community members.

@nickanderson if you agree about my proposal, please open a Redmine ticket because it's a fairly big change. This PR can probably remain because the `canonify` mapdata mode doesn't conflict with my proposal and can much more easily be allowed in the next patch 3.7 and 3.8 releases. The decision is up to the developers.

tzz:
@nickanderson ping about the `eval` option.

@kacf @estenberg please consider this, it's an incremental safe change.

kacf:
Indeed this looks very useful, and the `eval` stuff even more so.

Function calls directly inside strings have been discussed before in the topic of inline string concatenation, and were turned down. I'd like us to stick to the same principle here. Using the proposed method there is no way to insert the actual text `canonify(...)`, for example. I know, it's not very likely to be used in this case, but it's a slippery slope when you mix the two, and before you know it the syntax has become very ambiguous.

The best solution for this would most likely be the same kind of syntax that has been discussed for inline string concatenation and function calls, something like `mapdata("eval", canonify("$(this.k)") + escape("$(this.v)"), mycontainer)` (note where the quotes are), but this is very long term, probably difficult right now, and it's better to implement something simpler. My proposal is this:

`mapdata(evalfunction, "$(this.v)", mycontainer, "canonify")`

* Put function name outside and at the end, to avoid ambiguity.
* Rename `eval` to `evalfunction` to be more clear on what it does. This is also good if we later choose to implement true generic `eval` functionality with inline function calls.
* The two above changes hopefully also make it more implicit that this is a single level function call facility.

What do you think about this?

I'm not opposed to merging to merging this as is, but if you plan to implement `eval` functionality anyway, we might as well wait for that and get it all in one serving. Note that neither change is eligible for 3.7 or 3.8. Sorry, but we are strict about maintaining backwards and forwards syntax compatibility between patch releases.

tzz:
@kacf I'll pass on the `eval` stuff for now, please evaluate this PR just for `canonify`.

I asked @nickanderson to give feedback on the `eval` but it should go in a Redmine ticket for discussion. 

FWIW, I think your proposal is clean from the programmer's side, but will only be useful with `canonify` and maybe 1-2 other single-parameter functions so it's not very valuable to the policy author. So maybe we'll just wait for a true `map()` function that takes a `FnCall` in the transformation slot, instead of hacking more into `mapdata()`. That may actually be fairly easy to write, as long as we agree on the feature in Redmine.

kacf:
> FWIW, I think your proposal is clean from the programmer's side, but will only be useful with canonify and maybe 1-2 other single-parameter functions so it's not very valuable to the policy author. So maybe we'll just wait for a true map() function that takes a FnCall in the transformation slot, instead of hacking more into mapdata(). That may actually be fairly easy to write, as long as we agree on the feature in Redmine.

You're right, I didn't think about that, it would need an even more generic approach.

> I'll pass on the `eval` stuff for now, please evaluate this PR just for `canonify`.

Alright!

kacf:
trigger build

kacf:
This is passing everything, just docs remaining!

tzz:
OK, docs in https://github.com/cfengine/documentation/pull/1271

tzz:
@nickanderson we should also update the `examples/mapdata.cf` example with the new mode.

kacf:
Great work as always, thanks!

PR #2427, opened 2015-12-04
Replace select(2) with poll(2) in a few places

worr:
This is my first batch of `select(2)` -> `poll(2)` changes. This includes a Windows implementation of `poll(2)`, taken from [libressl-portable/portable](https://github.com/libressl-portable/portable/blob/a7f031ba55ac4a69263000357eb7f6d7fb88101a/apps/openssl/compat/poll_win.c).

This passed tests on my dev box, but I don't have a Windows box to test on.

### DIFF
--- a/cf-agent/package_module.c
+++ b/cf-agent/package_module.c
@@ -203,35 +203,36 @@ static void ParseAndLogErrorMessage(const Rlist *data)
 
 static int IsReadWriteReady(const IOData *io, int timeout_sec)
 {
-    fd_set  rset;
-    FD_ZERO(&rset);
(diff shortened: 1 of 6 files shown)


kacf:
Hmm, this adds quite a bit of complexity due to the Windows implementation. What's the motivation behind it? Performance?

I think we need some evidence that this is an actual bottleneck before we consider this. For example I would expect the database to be a significantly worse bottleneck than this.

kacf:
Closing this, since ATM I don't think that this will be accepted. Discussion can continue though.

PR #2428, opened 2015-12-06
Refs #7808: add test for recursive permissions change

jooooooon:
Test case for the bug reported in https://dev.cfengine.com/issues/7808

In short, changing permissions recursively can fail if the new owner isn't allowed to run chmod on the files to be changed, in particular if the new owner cannot traverse the top directory.


jooooooon:
cc @kacfengine  @worr , since you guys worked on #2390, #2276, #2192 (they all ended up in the same PR, #2390) that is realted to this.

kacf:
Superseded by #2482.

PR #2429, opened 2015-12-06
Change: Refactor

nickanderson:
- Add option parsing
 - To make it more polished
 - To make it work like the original version of masterfiles.sh when
   called without parameters

- Merged in changes from Mike Weilgart
 - Moved upstream stage and deploy implementations to common.sh so that
   they could be sourced and used by other scripts
 - Removed attempt to rebase on a tag that did not work in
   GIT_TAG_OR_COMMIT


nickanderson:
@jimis @estenberg @pasinskim or @kacfengine can you please merge these changes

nickanderson:
@mikeweilgart see changes here that include yours, and adds option parsing.

jimis:
Wow lots of changes here, merging, thanks Nick!

PR #2430, opened 2015-12-07
Fixes #6961 - bad option nlwp to vzps on Proxmox

jooooooon:
See https://dev.cfengine.com/issues/6961 for rationale

If accepted, this should be backported to currently maintained versions.


zerkms:
That's fantastic news, since due to this problem we are locked on ancient cfengine versions. Thanks.

kacf:
Nice find, thanks Jonathan! Merged manually and cherry-picked to 3.6.x, 3.7.x and 3.8.x as well.

PR #2432, opened 2015-12-08
Remove duplicate entry from the 3.8 ChangeLog

jooooooon:
"Add: New `results` classes body (Redmine #7418)"

was in there twice. I removed the second occurence, because it seemed to have a broken link syntax in there too.


kacf:
Thanks, merged manually!

PR #2441, opened 2015-12-16
Fix cf-execd not starting during bootstrap

nickanderson:
Changelog: Fix cf-execd not starting during bootstrap with trigger_upgrade enabled in update.cf Redmine #7861


nickanderson:
ping @jimis or @kacf can you pick this back to 3.8 and 3.7?

kacf:
Apparently already done.

PR #2442, opened 2015-12-17
if minimum_version did not ignore the character '@' 

basvandervlies:
Changelog: if minimum_version did not ignore the character '@' (Redmne #7862)

### DIFF
--- a/libpromises/cf3lex.l
+++ b/libpromises/cf3lex.l
@@ -87,6 +87,7 @@ comment    #[^\n]*
 macro_if_version    ^@if\ minimum_version\([0-9]{1,10}\.[0-9]{1,10}(\.[0-9]{1,10})?\)
 macro_if_feature    ^@if\ feature\([a-zA-Z0-9_]+\)
 macro_endif ^@endif
+macro_line ^[^@].*$
 
 promises   bundle
 
@@ -243,6 +244,11 @@ promise_type   [a-zA-Z_]+:
                         P.if_depth--;
                       }
 
+
+<if_ignore_state>{macro_line}  {
+                                  ParserDebug("\tL:inside macro @if, ignoring text:%s\n", yytext);
+                               }
+
 <if_ignore_state>{macro_endif} {
                                  ParserDebug("\tL:macro @endif %d\n", P.line_pos);
                                  BEGIN(INITIAL);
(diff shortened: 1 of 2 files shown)


libpromises/cf3lex.l
```diff
@@ -252,11 +258,11 @@ promise_type   [a-zA-Z_]+:
...
-                               }
-
-<if_ignore_state>[^@]      {
-                               ParserDebug("\tL:inside macro @if, ignoring text:%s\n", yytext);
-                           }
+                               }\
+<if_ignore_state>.    {
+                          /* eat up al unknown chars when line starts with @*/
+                          ParserDebug("\tL:inside macro @if, ignoring text:%s\n", yytext);
+                      }
```
kacf: @tzz: You recently made a change to add this '[^@]'. Will it still work with Bas' change?

tests/acceptance/00_basics/macros/if_ignore_with_addsign.cf
```diff
@@ -0,0 +1,30 @@
+######################################################
+#
+#  Test that @if works for greater versions
```
kacf: I guess this was copy/pasted. Should be `Test that @if also ignores lines starting with '@'`.

kacf:
Great stuff!

basvandervlies:
Just for the record. Is there a test that seed if cf-promises parse all lines and there is nothing on stdout/stderr? without this patch all '@' chars are outputted. 

thanks @kacf

tests/acceptance/00_basics/macros/if_ignore_with_addsign.cf
basvandervlies: Your right. I just copied a another acceptance test

libpromises/cf3lex.l
tzz: The original ate up to the next `@` so I'm not sure how the new version will work.

basvandervlies: This is one reads the whole line instead per char but skips line starting with @ then it switches to per char read and parse all chars and seea the @endiff. When I turned on the Lex debugger I was surprised by the one char parsing when in ignore mode. Was difficult to  follow 😀

kacf:
> Just for the record. Is there a test that seed if cf-promises parse all lines and there is nothing on stdout/stderr? without this patch all '@' chars are outputted.

No, not ATM I believe. But we have recently been talking about adding a generic check inside the `testall` script which catches unexpected text, particularly `error:` and `warning:` messages, since we have been burned by these several times. Currently the test framework only checks output in tests that explicitly look for it. I suppose that would catch this case as well.

kacf:
trigger build

kacf:
Thanks Bas, fixed slightly incorrect commit message (referring to some unrelated ticket) and merged manually. Cherry-picked to 3.7.x and 3.8.x as well.

PR #2445, opened 2015-12-18
Added example policy for motd that has been baked during 2015 pre-xmas training by enrico and nick.

Enemico:


jimis:
@enemico congrats for your first commit!

PR #2450, opened 2015-12-22
redmine#7732: cf-agent -K is now allowed even for remote execution

jimis:
cf-agent is now allowed all arguments, even when the "cfruncommand"
class is set. Additionally cf-serverd no longer appends
"-I -Dcfruncommand" to cfruncommand, this has to be done manually in
masterfiles body server control. This way configuration is fully
up to the sysadmin, as there are no hard-coded options any more.

Security is still enforced because arbitrary options can't be passed via
the EXEC command, they have to be set in policy body server control.

Changelog: cf-serverd no longer appends "-I -Dcfruncommand" to
           cfruncommand, this has to be done manually in masterfiles
           body server control.


jimis:
trigger build

PR #2454, opened 2015-12-23
Allow some functions to take inline JSON: mapdata, mergedata, and others

tzz:
Related to https://dev.cfengine.com/issues/7871 (when it's implemented, all CFEngine functions that take lists or data containers will be able to use this functionality).

With this patch, many functions will be able to take inline JSON instead of a list name or data container name. We simply try to parse the varname as JSON; since neither `@(listname)` nor `containername` are valid JSON, we end up with a reliable way to parse JSON.

`mergedata()` was a little more complicated because it supports symbolic wrapping, but it's a pretty trivial change otherwise. It also reduces line count, which is a good thing...

Example:
```
bundle agent test
{
  vars:
      "user3" data => data_expand('["$(sys.os)"]');
      "user2" data => mapdata("none", '$(this.v)', '["x"]');
      "user" data => mergedata('[]', '["x"]');
}
```

Result bundle state:

```json
{
  "user": [
    "x"
  ],
  "user2": [
    "x"
  ],
  "user3": [
    "linux"
  ]
}
```

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -320,6 +320,36 @@ static JsonElement* VarRefValueToJson(EvalContext *ctx, const FnCall *fp, const 
     return convert;
 }
 
+static JsonElement* VarNameOrInlineToJson(EvalContext *ctx, const FnCall *fp, const char *data)
+{
+    JsonElement *inline_data = NULL;
+    JsonParseError res = JsonParse(&data, &inline_data);
+
+    if (res != JSON_PARSE_OK)
+    {
+    }
+    else if (JsonGetElementType(inline_data) == JSON_ELEMENT_TYPE_PRIMITIVE)
+    {
+        JsonDestroy(inline_data);
+        inline_data = NULL;
+    }
+    else
+    {
+        return inline_data;
+    }
+
+    VarRef *ref = ResolveAndQualifyVarName(fp, data);
+    if (!ref)
+    {
+        return NULL;
+    }
+
+    JsonElement *vardata = VarRefValueToJson(ctx, fp, ref, NULL, 0);
+    VarRefDestroy(ref);
+
+    return vardata;
+}
+
 static Rlist *GetHostsFromLastseenDB(Item *addresses, time_t horizon, bool return_address, bool return_recent)
 {
     Rlist *recent = NULL, *aged = NULL;
@@ -1687,14 +1717,8 @@ static FnCallResult FnCallUrlGet(ARG_UNUSED EvalContext *ctx,
 #ifdef HAVE_LIBCURL
 
     char *url = RlistScalarValue(finalargs);
-    VarRef *ref = ResolveAndQualifyVarName(fp, RlistScalarValue(finalargs->next));
-    if (!ref)
-    {
-        return FnFailure();
-    }
+    JsonElement *options = VarNameOrInlineToJson(ctx, fp, RlistScalarValue(finalargs->next));
 
-    JsonElement *options = VarRefValueToJson(ctx, fp, ref, NULL, 0);
-    VarRefDestroy(ref);
     if (NULL == options)
     {
         return FnFailure();
@@ -2646,13 +2670,7 @@ static FnCallResult FnCallMapData(EvalContext *ctx, ARG_UNUSED const Policy *pol
     bool jsonmode = (0 == strcmp(conversion, "json"));
     bool canonifymode = (0 == strcmp(conversion, "canonify"));
 
-    VarRef *ref = ResolveAndQualifyVarName(fp, varname);
-    if (!ref)
-    {
-        return FnFailure();
-    }
-
-    JsonElement *container = VarRefValueToJson(ctx, fp, ref, NULL, 0);
+    JsonElement *container = VarNameOrInlineToJson(ctx, fp, varname);
 
     if (NULL == container)
     {
@@ -2976,61 +2994,62 @@ static FnCallResult FnCallMergeData(EvalContext *ctx, ARG_UNUSED const Policy *p
         Buffer *wrap_map_key = NULL;
         Buffer *name = NULL;
 
-        if (name_len > 2 && name_str[0] == '[')
-        {
-            Seq *s = StringMatchCaptures("^\\[ *([^ ]+) *\\]$", name_str, false);
+        // try to load directly
+        JsonElement *json = VarNameOrInlineToJson(ctx, fp, name_str);
 
-            if (s && SeqLength(s) == 2)
+        // if that failed, try the wrappers
+        if (NULL == json)
+        {
+            if (name_len > 2 && name_str[0] == '[')
             {
-                wrap_array_mode = true;
-                name = BufferCopy((const Buffer*) SeqAt(s, 1));
+                Seq *s = StringMatchCaptures("^\\[ *([^ ]+) *\\]$", name_str, false);
+
+                if (s && SeqLength(s) == 2)
+                {
+                    wrap_array_mode = true;
+                    name = BufferCopy((const Buffer*) SeqAt(s, 1));
+                }
+
+                SeqDestroy(s);
+            }
+            else if (name_len > 0 && name_str[0] == '{' && name_str[name_len-1] == '}')
+            {
+                Seq *s = StringMatchCaptures("^\\{ *\"([^\"]+)\" *: *([^ ]+) *\\}$", name_str, false);
+
+                if (s && SeqLength(s) == 3)
+                {
+                    wrap_map_key = BufferCopy((const Buffer*) SeqAt(s, 1));
+                    name = BufferCopy((const Buffer*) SeqAt(s, 2));
+                }
+
+                SeqDestroy(s);
+            }
+            else
+            {
+                name = BufferNewFrom(name_str, name_len);
+            }
+        }
+
+        // try to use `name` a second time
+        if (NULL == json && NULL != name)
+        {
+            json = VarNameOrInlineToJson(ctx, fp, BufferData(name));
+        }
+
+        // we failed to produce a valid JsonElement, so give up
+        if (NULL == json)
+        {
+            if (NULL != wrap_map_key)
+            {
+                BufferDestroy(wrap_map_key);
             }
 
-            SeqDestroy(s);
-        }
-        else if (name_len > 0 && name_str[0] == '{' && name_str[name_len-1] == '}')
-        {
-            Seq *s = StringMatchCaptures("^\\{ *\"([^\"]+)\" *: *([^ ]+) *\\}$", name_str, false);
-
-            if (s && SeqLength(s) == 3)
+            if (NULL != name)
             {
-                wrap_map_key = BufferCopy((const Buffer*) SeqAt(s, 1));
-                name = BufferCopy((const Buffer*) SeqAt(s, 2));
+                BufferDestroy(name);
             }
 
-            SeqDestroy(s);
-        }
-        else
-        {
-            name = BufferNewFrom(name_str, name_len);
-        }
-
-        VarRef *ref = NULL;
-
-        if (NULL != name)
-        {
-            ref = ResolveAndQualifyVarName(fp, BufferData(name));
-            BufferDestroy(name);
-        }
-
-        if (!ref)
-        {
             SeqDestroy(containers);
-
-            if (NULL != wrap_map_key) BufferDestroy(wrap_map_key);
-
-            return FnFailure();
-        }
-
-        JsonElement *json = VarRefValueToJson(ctx, fp, ref, NULL, 0);
-
-        VarRefDestroy(ref);
-
-        if (!json)
-        {
-            SeqDestroy(containers);
-
-            if (NULL != wrap_map_key) BufferDestroy(wrap_map_key);
 
             return FnFailure();
         }
@@ -4199,14 +4218,7 @@ static FnCallResult FnCallFold(EvalContext *ctx,
     bool max_mode = strcmp(fp->name, "max") == 0;
     bool min_mode = strcmp(fp->name, "min") == 0;
 
-    VarRef *ref = ResolveAndQualifyVarName(fp, name);
-    if (!ref)
-    {
-        return FnFailure();
-    }
-
-    JsonElement *json = VarRefValueToJson(ctx, fp, ref, NULL, 0);
-    VarRefDestroy(ref);
+    JsonElement *json = VarNameOrInlineToJson(ctx, fp, name);
 
     if (!json)
     {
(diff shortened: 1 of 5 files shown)


tzz:
The second commit here introduces the ability to say

```
      "foo" string => "bar";
      "userfoo" data => data_expand('[foo]');

      "user3" data => data_expand('["$(sys.os)"]');
      "user3a" data => data_expand('[test.user3,]');
```

in other words, to refer to a variable by name inside JSON and have it imported. Unfortunately this breaks at least 5 tests (`01_vars/02_functions/url_get_local.cf` `01_vars/02_functions/network/url_get.cf` `01_vars/02_functions/mergedata_with_list_iteration.cf` `01_vars/02_functions/maparray_multi_index.cf` `01_vars/02_functions/data_readstringarrayidx.cf`)
so I need some feedback:

* is it useful? @nickanderson ?
* do the test failures look fixable?
* is anyone interested in helping out with the work? @worr @kacf @jimis ?

If the second commit is not useful, I'll drop it.

tzz:
Full test I used for the second commit, all the references were expanded correctly:
```
bundle agent main
{
  methods:
      "test";

  vars:
      "test_state" data => bundlestate(test);
      "test_string" string => storejson(test_state);

  reports:
      "$(this.bundle): state of things = $(test_string)";
}

bundle agent test
{
  vars:
      "foo" string => "bar";
      "userA" data => data_expand('{ "x": test.user3 }');
      "user9" data => data_expand('{ "x": user3 }');
      "user8" data => data_expand('[user3]');
      "user7" data => data_expand('{foo:user3}');
      "user6" data => data_expand('[foo]');
      "user5" data => data_expand('[test.user3,]');
      "user4" data => data_expand('[test.user3]');
      "user3" data => data_expand('["$(sys.os)"]');
      "user2" data => mapdata("none", '$(this.v)', '["x"]');
      "user" data => mergedata('[]', '["x"]');
}
```

kacf:
In general I think this is very good idea, we just need to be careful we don't introduce any ambiguity.

> in other words, to refer to a variable by name inside JSON and have it imported.

I'm not quite clear on this; how do you differentiate between a variable to be expanded and a simple string. Is it just that it's unquoted?

How will it work with function returns? Can a function returning a container be used as an argument to a different function? I suppose they would be different types, but this might not be that hard to account for.

I haven't looked at the test failures yet. Do you see any recurring pattern?

tzz:
> I'm not quite clear on this; how do you differentiate between a variable to be expanded and a simple string. Is it just that it's unquoted?

At the policy level, if a function expects a container name, we also allow strings that start with `{` or `[` because those can't be container names.

Inside the JSON, currently we reject bare words unless they are keys. So we'll now allow bare words such as `myvar` or `mybundle.myvar`. Except for bareword keys, this is fully consistent.

> How will it work with function returns? Can a function returning a container be used as an argument to a different function? I suppose they would be different types, but this might not be that hard to account for.

I don't think it helps in that case currently, but we could hack it later (as the same new function to handle var refs OR inline JSON we introduce here becomes standard as in https://dev.cfengine.com/issues/7871, it will be easier to add new behavior in a single place).

> I haven't looked at the test failures yet. Do you see any recurring pattern?

I think it's either a bug, or something with bareword keys. I'll keep digging if it's not visible to anyone else...

nickanderson:
I can see how it could be useful in some cases, but I can't think of how I would use it right now.

tzz:
Hacking around `json.c` makes me think of how much easier life would be if we used libyaml to parse JSON.

tzz:
OK, all the tests pass, but I had to use a regex to capture the possible variations of `varname`, `bundle.varname`, and `varname[lookup]` which will slow down this specific code path a little bit. I wrote a new test to ensure all of that is checked.

This is ready for review...

kacf:
> Inside the JSON, currently we reject bare words unless they are keys. So we'll now allow bare words such as `myvar` or `mybundle.myvar`. Except for bareword keys, this is fully consistent.

Consistent perhaps, but not intuitive nor easy to remember. This means that a bareword in a key and a value will have very different and conflicting meanings. I'm wondering if variable support is demanding a bit much of this feature. I think we at least need to enforce "variable syntax" on it to make sure it stays readable for novice policy readers. So perhaps we could stay with barewords as you originally proposed, but we require `$(var)` around the name, like in regular policy. What do you think?

Another, more technical concern is how this works out if we indeed choose to use libyaml to parse JSON one day. Will it accept barewords in JSON values at all? It might make it more difficult to integrate with libyaml later.

And what about typing YAML directly into `mergedata`?

worr:
Should we accept barewords at all? They aren't allowed by the JSON spec, so imo allowing them at all anywhere in any JSON data container is a bug.

Like @kacf mentioned, this will likely make it harder to switch to a different parser later.

tzz:
> Consistent perhaps, but not intuitive nor easy to remember. This means that a bareword in a key and a value will have very different and conflicting meanings. I'm wondering if variable support is demanding a bit much of this feature. I think we at least need to enforce "variable syntax" on it to make sure it stays readable for novice policy readers. So perhaps we could stay with barewords as you originally proposed, but we require $(var) around the name, like in regular policy. What do you think?

I don't think it would benefit CFEngine to grow its own way to specify inline JSON, and this PR is all about ease of use. I'd rather stay consistent with Javascript and use barewords.

Remember that these are strings, so the evaluator would see `$(var)` and do nasty things to it. We'd need true JSON markup at the parser level to avoid that (and that would make it clear that these are not just strings). Maybe a special marker like `inlinejson:{...}` instead of putting it in a string?

> Another, more technical concern is how this works out if we indeed choose to use libyaml to parse JSON one day. Will it accept barewords in JSON values at all? It might make it more difficult to integrate with libyaml later.

True. Actually researching into libyaml I found it's not good at parsing pure JSON anyway, it's too strict in every way. We'll probably want libjson or something like it.

> And what about typing YAML directly into mergedata?

Because YAML, unlike JSON, is an ambiguous format, I wouldn't allow it.

> Should we accept barewords at all? They aren't allowed by the JSON spec, so imo allowing them at all anywhere in any JSON data container is a bug.

CFEngine and Javascript allow bareword keys in JSON already for key values, see https://github.com/cfengine/core/pull/2247 and especially https://github.com/cfengine/core/pull/2247#issuecomment-116605655

kacf:
> > Consistent perhaps, but not intuitive nor easy to remember. This means that a bareword in a key and a value will have very different and conflicting meanings. I'm wondering if variable support is demanding a bit much of this feature. I think we at least need to enforce "variable syntax" on it to make sure it stays readable for novice policy readers. So perhaps we could stay with barewords as you originally proposed, but we require $(var) around the name, like in regular policy. What do you think?
> 
> I don't think it would benefit CFEngine to grow its own way to specify inline JSON, and this PR is all about ease of use. I'd rather stay consistent with Javascript and use barewords.
> 
> Remember that these are strings, so the evaluator would see `$(var)` and do nasty things to it. We'd need true JSON markup at the parser level to avoid that (and that would make it clear that these are not just strings). Maybe a special marker like `inlinejson:{...}` instead of putting it in a string?

Isn't using barewords in JSON values *already* growing a custom way of parsing JSON? I don't think CFEngine would benefit from that either. Perhaps you should ping me in the chat later. I'm not sure I fully understood this.

> Actually researching into libyaml I found it's not good at parsing pure JSON anyway, it's too strict in every way. We'll probably want libjson or something like it.

I'd be interested in knowing more details about this. Were there valid JSON files that would not be parsed by libyaml? Or are we already too relaxed in our own parser that we have introduced incompatibilities?

> > And what about typing YAML directly into mergedata?
> 
> Because YAML, unlike JSON, is an ambiguous format, I wouldn't allow it.

Gotcha.

> > Should we accept barewords at all? They aren't allowed by the JSON spec, so imo allowing them at all anywhere in any JSON data container is a bug.
> 
> CFEngine and Javascript allow bareword keys in JSON already for key values, see https://github.com/cfengine/core/pull/2247 and especially https://github.com/cfengine/core/pull/2247#issuecomment-116605655

Indeed, incompatible with the full strictness of the spec, but "reasonable" in the sense that it seems to be de facto supported.

tzz:
OK, rebased against latest master.

kacf:
Thanks!

I won't post the whole log, but the summary from the IRC discussion is that bareword values as variables are already supported when using inline JSON in `mergedata`, so there is little point in debating whether we should use them now. Better to stay consistent and extend the feature here.

The point about libyaml/libjson is unlikely to be relevant, because libyaml is *too*strict for JSON parsing, and libjson doesn't give us much benefit over what we already have.

I'll take a more detailed look at the code shortly.

libutils/json.h
```diff
@@ -354,13 +354,19 @@ JsonElement *JsonArrayGet(const JsonElement *array, size_t index);
...
+  @param ctx [in] Evaluation context for variable lookups
   @param data [in] Pointer to the string to parse
   @param json_out Resulting JSON object
   @returns See JsonParseError and JsonParseErrorToString
+
+  The ctx type is void so we don't have to include eval_context.h from
+  libpromises into libutil
   */
-JsonParseError JsonParse(const char **data, JsonElement **json_out);
+JsonParseError JsonParse(void *ctx, JsonLookup *lookup, const char **data, JsonElement **json_out);
```
kacf: Can you please leave the `JsonParse` function as is, and make a new function `JsonParseWithLookup`? The former can call the latter with null arguments. Then we don't have to change it everywhere, and the interface stays clean.

Also, slightly more cosmetic, but I think `ctx` should be renamed to `map`, since in principle it has no relation to `ctx`, it can be anything. Same thing for the `JsonLookup` function above. The comment about `ctx` being used as the map can then be moved to libpromises, at the place where it is actually passed in, which is where this information belongs. The information about the `map` argument can simply be something like `map which the JsonLookup function will use to look up references`, or something like that.

In general there shouldn't be any references to ctx in libutils, neither comments nor otherwise, since it is supposed to be independent from it.

libpromises/evalfunction.c
```diff
@@ -195,8 +196,9 @@ static VarRef* ResolveAndQualifyVarName(const FnCall *fp, const char *varname)
     return ref;
 }
 
-static JsonElement* VarRefValueToJson(EvalContext *ctx, const FnCall *fp, const VarRef *ref,
-                                      const DataType disallowed_datatypes[], size_t disallowed_count)
+static JsonElement* VarRefValueToJsonAllowingScalars(EvalContext *ctx, const FnCall *fp, const VarRef *ref,
+                                                     const DataType disallowed_datatypes[], size_t disallowed_count,
+                                                     bool allow_scalars)
```
kacf: Why can't `disallowed_datatypes` be used here instead of introducing yet another argument `allow_scalars`? Can't you just put the scalar type as disallowed in the appropriate places?

libpromises/evalfunction.c
```diff
@@ -313,13 +328,85 @@ static JsonElement* VarRefValueToJson(EvalContext *ctx, const FnCall *fp, const
...
+    }
+    else if (JsonGetElementType(inline_data) == JSON_ELEMENT_TYPE_PRIMITIVE)
+    {
+        JsonDestroy(inline_data);
+        inline_data = NULL;
+    }
+    else
+    {
+        return inline_data;
+    }
```
kacf: Weird code flow with empty if. I propose to change it to the following:

```
    if (res == JSON_PARSE_OK)
    {
        if (JsonGetElementType(inline_data) == JSON_ELEMENT_TYPE_PRIMITIVE)
        {
            JsonDestroy(inline_data);
            inline_data = NULL;
        }
        else
        {
            return inline_data;
        }
    }
```

libpromises/evalfunction.c
```diff
@@ -2971,91 +3046,20 @@ static FnCallResult FnCallMergeData(EvalContext *ctx, ARG_UNUSED const Policy *p
...
-        if (!ref)
-        {
-            SeqDestroy(containers);
-
-            if (NULL != wrap_map_key) BufferDestroy(wrap_map_key);
-
-            return FnFailure();
-        }
-
-        JsonElement *json = VarRefValueToJson(ctx, fp, ref, NULL, 0);
```
kacf: +1. Nice to have this gathered in one place!

libpromises/evalfunction.c
```diff
@@ -251,7 +264,8 @@ static JsonElement* VarRefValueToJson(EvalContext *ctx, const FnCall *fp, const
                     }
                     else if (var->ref->num_indices - index_offset > 1)
                     {
-                        Log(LOG_LEVEL_DEBUG, "%s: got ref with starting depth %zd and index count %zd", fp->name, index_offset, var->ref->num_indices);
+                        Log(LOG_LEVEL_DEBUG, "%s: got ref with starting depth %zd and index count %zd",
+                            fp ? fp->name : "VarRefValueToJson", index_offset, var->ref->num_indices);
```
kacf: Can you make this `fp` check at the top of the function instead, and store the name in a different variable. Then you don't need to check this every place where it's used.

Same goes for several uses below this one.

Alternatively, since `fp` is only used to fetch the name AFAICS, perhaps you shouldn't pass in a null pointer at all, but you should change it to a string and pass in `"LookupVarRefToJson"` from the one function that now passes null. It still won't be fully verbose in the sense that the real calling function from the policy is masked, but it would take quite a bit of data passing to get that information all the way here.

libutils/json.h
tzz: OK; fixed. I renamed `ctx` to `lookup_context` because I think it's the best name for it, and `lookup` is now `lookup_function`. I hope that's acceptable, and you're right that the change is much smaller and cleaner that way.

libpromises/evalfunction.c
tzz: Because we disallow scalars by default, unfortunately.

libpromises/evalfunction.c
tzz: I made a variable `fp_name` because, the way we use `fp`, I think it may be needed in the future so just passing a string is not future-proof. I hope that's OK.

libpromises/evalfunction.c
tzz: Thanks for catching that. Fixed.

tzz:
Thanks for the review. I've fixed all but the "disallow scalar" issue, and rebased against master. The change is much smaller thanks to the `JsonParseWithLookup` suggestion.

libutils/json.h
kacf: Yep, that's good!

libpromises/evalfunction.c
kacf: Ah ok, I see.

libpromises/evalfunction.c
kacf: Sure!

kacf:
trigger build

kacf:
Build is green for 5cce98b! I assume that this should have a mention in the docs as well. Something similar to https://github.com/cfengine/documentation/commit/ef825dd14b57b61a1a17928bbd2e353164eaf76f perhaps?

tzz:
docs in https://github.com/cfengine/documentation/pull/1306 (with some notes)

kacf:
Thanks! Both merged, this one manually.

tzz:
CHEERS

PR #2455, opened 2015-12-29
testall: Expose VERBOSE var through --verbose option

worr:
In testing out my changes that fail promises based on variables not expanding, I've been finding verbose output invaluable. This is *kind of* exposed through the `$VERBOSE` env var, but it's much more convenient to set this through a command-line option to `testall`.


kacf:
YES! I have wanted this for a long time. So simple, so simple...

PR #2460, opened 2015-12-29
Provide new -w argument to override the workdir for testing

tzz:
As requested by @nickanderson and others: provide `-w` option to cf-promises and cf-agent to override the workdir for easier testing.

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -79,6 +79,7 @@
 #include <conn_cache.h>                 /* ConnCache_Init,ConnCache_Destroy */
 #include <net.h>
 #include <package_module.h>
+#include <string_lib.h>
 
 #include <mod_common.h>
 
@@ -134,7 +135,7 @@ static void ThisAgentInit(void);
 static GenericAgentConfig *CheckOpts(int argc, char **argv);
 static char **TranslateOldBootstrapOptionsSeparate(int *argc_new, char **argv);
 static char **TranslateOldBootstrapOptionsConcatenated(int argc, char **argv);
-static void FreeStringArray(int size, char **array);
+static void FreeFixedStringArray(int size, char **array);
 static void CheckAgentAccess(const Rlist *list, const Policy *policy);
 static void KeepControlPromises(EvalContext *ctx, const Policy *policy);
 static PromiseResult KeepAgentPromise(EvalContext *ctx, const Promise *pp, void *param);
@@ -168,6 +169,7 @@ static const struct option OPTIONS[] =
 {
     {"bootstrap", required_argument, 0, 'B'},
     {"bundlesequence", required_argument, 0, 'b'},
+    {"workdir", required_argument, 0, 'w'},
     {"debug", no_argument, 0, 'd'},
     {"define", required_argument, 0, 'D'},
     {"self-diagnostics", optional_argument, 0, 'x'},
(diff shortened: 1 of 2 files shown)


nickanderson:
Indeed, +1 very useful for testing.
@brontolinux @ncharles I believe you mentioned this recently.

brontolinux:
@nickanderson I did :-)

cf-agent/cf-agent.c
```diff
@@ -308,6 +310,11 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
             TIMING = true;
             break;
 
+        case 'w':
+            Log(LOG_LEVEL_INFO, "Setting workdir to '%s'", optarg);
+            setenv("CFENGINE_TEST_OVERRIDE_WORKDIR", optarg, 1);
```
kacf: `setenv` is not a POSIX function (can you believe it??). You'll have to use `putenv`, which will require you to allocate the memory you give to it. This will leak if you later give it a different string, but I think we don't need to worry about that. This will happen only here, and only once.

cf-agent/cf-agent.c
```diff
@@ -191,6 +192,7 @@ static const char *const HINTS[] =
 {
     "Bootstrap CFEngine to the given policy server IP, hostname or :avahi (automatic detection)",
     "Set or override bundlesequence from command line",
+    "Override the work directory for testing (same as setting CFENGINE_TEST_OVERRIDE_WORKDIR)",
```
kacf: I would change this to `Override the default /var/cfengine work directory...`, just so that there is no doubt what is being talked about.

tzz:
Both suggestions done, thanks for the review.

kacf:
trigger build

kacf:
trigger build

cf-agent/cf-agent.c
```diff
@@ -308,6 +310,11 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
             TIMING = true;
             break;
 
+        case 'w':
+            Log(LOG_LEVEL_INFO, "Setting workdir to '%s'", optarg);
+            putenv("CFENGINE_TEST_OVERRIDE_WORKDIR", xstrdup(optarg), 1);
```
kacf: `putenv` takes only one argument, a string of the form `CFENGINE_TEST_OVERRIDE_WORKDIR=/var/cfengine`, and this complete string is what you need to `strdup`. If `CFENGINE_TEST_OVERRIDE_WORKDIR` exists in the environment already, it will replace the existing string.

tzz: Oh, sorry. I didn't push the latest change. I've done so now, amended to use `string_lib.h` (a local function in `cf-agent.c` had to be renamed).

cf-promises/cf-promises.c
```diff
@@ -235,6 +237,11 @@ GenericAgentConfig *CheckOpts(int argc, char **argv)
             config->agent_specific.common.show_variables = true;
             break;
 
+        case 'w':
+            Log(LOG_LEVEL_INFO, "Setting workdir to '%s'", optarg);
+            setenv("CFENGINE_TEST_OVERRIDE_WORKDIR", optarg, 1);
```
kacf: There is another one here. I opened #2484 to fix it.

kacf:
Superseded by #2484.

tzz:
Thank you for dealing with this!!!

PR #2461, opened 2015-12-30
Fix HP-UX specific bug that caused a lot of log output to disappear.

kacf:
Buffered output is discarded if you switch buffering mode without
flushing the buffered output first. This will happen anyway when
switching modes, so no performance is lost.

Also move it earlier in the initialization sequence, so that we start
line buffering sooner. This can come in handy later, since if there is
an early crash in cf-agent, the log output may be truncated unless we
have switched to line buffered mode.

Changelog: Title


kacf:
trigger build

kacf:
Cherry-picked to 3.6.x, 3.7.x and 3.8.x as well.

PR #2462, opened 2016-01-01
Add: Test that ifelse can take isvariable as param

nickanderson:
Ref:https://dev.cfengine.com/issues/7878


kacf:
Careful with this, this turned the build red.

The test passes, but the `test_soft_fail` indicates that it should fail. There is also a very recent ticket open for this, so I assume that the test itself is wrong, and it should in fact fail. I have reverted this for now, so you can investigate it.

PR #2465, opened 2016-01-07
Masterfiles stage rewrite

mikeweilgart:
`masterfiles-stage.sh` and its supporting files `common.sh` and `options.sh` have been completely refactored and largely rewritten.  Many potential bugs and edge cases have been handled, including removal of insecure use of `eval` in option parsing.

The script is now 100% a drop-in replacement for the version of the code that ships with CFEngine.

This overhaul was actually done to allow incorporating the `GIT_POLICY_CHANNELS` vcs_type, and this has been done.

Some major functional changes should be noted:

`GIT_TAG_OR_COMMIT` and `GIT` vcs_types now use a mirrored git repository rather than a normal git clone, thus avoiding many edge cases related to git repos getting out of sync.  These two `VCS_TYPE`s are now in fact identical except for variable names (see the main case switch in `masterfiles-stage.sh` to see how the difference is implemented).

The rollout method for policy has been changed from `rsync` to `mv` as an atomic operation for the git-based `VCS_TYPE`s.  (The rollout method for the `svn_branch` function hasn't been changed.)


mikeweilgart:
There is one undocumented feature which was present before and is still present: if you set a value for `MASTERDIR` in your `params.sh` file, it will override the value set on the command line.  I found this at the last moment during testing, and actually found it useful in testing.

We should keep the default setting for `MASTERDIR` so the script can still be a drop-in replacement for the version which ships with CFEngine.  And we should allow `MASTERDIR` to be set from the command line, so anyone currently using the script can go on using it that way.  But I would recommend for most users that they set `MASTERDIR` directly in the params file for simplicity.  ;)

mikeweilgart:
@nickanderson here is the pull request!

nickanderson:
ping @tzz if you don't mind taking a look, you did much of the original.

contrib/masterfiles-stage/README.org
```diff
@@ -86,7 +90,7 @@ The following parameters are in use, be sure to avoid their collision
 when developing new staging methods.
 
 - VCS_TYPE - The staging method to use
-  - Currently supported: (GIT|GIT_TAG_OR_COMMIT|SVN)
+  - Currently supported: (GIT|GIT_TAG_OR_COMMIT|GIT_POLICY_CHANNELS|SVN)
```
tzz: I'd add a note here that `GIT_TAG_OR_COMMIT` is legacy and equivalent to the other one.

contrib/masterfiles-stage/README.org
```diff
@@ -101,6 +105,15 @@ when developing new staging methods.
     - 88335d36b48c8808b12b48667a463182dc8d0338
     - mytag
 
+- channel_config_file - The path to the config file for multiple policy channels
+  - Example value:
+    - /var/cfengine/policy_channels/channel_to_source.txt
```
tzz: Could this be done with bash arrays in the params or options instead of a new separate file?

contrib/masterfiles-stage/common.sh
```diff
@@ -12,146 +12,217 @@ if [ ! "$debug_mode" == "true" ]; then
...
-  fi
+  STAGING_DIR="${ROOT}/_tmp$(echo "$PARAMS" | tr [./] _)"
+}
 
-  if /var/cfengine/bin/cf-promises -T "${STAGING_DIR}"; then
-    if ! /var/cfengine/bin/cf-promises -cf "${STAGING_DIR}/update.cf"; then
-        error_exit "Update policy staged in ${STAGING_DIR} could not be validated, aborting."
-    fi
+check_git_installed() {
+  git --version >/dev/null 2>&1 || error_exit "git not found on path: '${PATH}'"
```
tzz: Maybe also check the Git version, older ones may not handle all the options? I'm not sure.

tzz:
I think this is a great rewrite. The only thing I'd consider changing is to use arrays inside the options instead of separate channels files, but that may not be possible if we want compatibility with older shells (I don't!).

contrib/masterfiles-stage/masterfiles-stage.sh
```diff
@@ -1,45 +1,43 @@
...
-
-  # We probably want a different temporary location for each remote repository
-  # so that we can avoid conflicts and potential confusion.
-  # Example: 
-  # ROOT="/opt/cfengine/masterfiles_staging"
-  # PARAMS="/var/cfengine/policychannel/production_1.sh"
-  # TMP=/opt/cfengine/masterfiles/staging/_tmp_var_cfengine_policychannel_production_1_sh
-
-  TRANSLATED_PARAMS=$(echo $PARAMS | tr [./] _)
-  STAGING_DIR="${ROOT}/_tmp${TRANSLATED_PARAMS}"
```
mikeweilgart: @nickanderson, you mentioned elsewhere that putting things in `${ROOT}` broke integration with Design Center.  This is how the code was (and is) in the codebase before my merge request.  I simply moved these lines of code from here to their own function in `common.sh`.

contrib/masterfiles-stage/common.sh
```diff
@@ -12,146 +12,217 @@ if [ ! "$debug_mode" == "true" ]; then
...
-        echo "Cloning git repository '${GIT_URL}' '${GIT_BRANCH}' branch into '${STAGING_DIR}'"
-      fi
-      git clone --no-hardlinks "${GIT_URL}" "${STAGING_DIR}"
-
-      if [ "$debug_mode" == "true" ]; then
-        echo "Checking out '${GIT_BRANCH}' branch."
-      fi
-      cd "${STAGING_DIR}" && git checkout "${GIT_BRANCH}"
-  fi
+  STAGING_DIR="${ROOT}/_tmp$(echo "$PARAMS" | tr [./] _)"
```
mikeweilgart: @nickanderson—for easy reference; this is the new location for the code chunk mentioned in my last comment.

contrib/masterfiles-stage/common.sh
mikeweilgart: I did my testing with git version 1.7.1, which is roughly 6 years old and considered an "old version" according to [Wikipedia](https://en.wikipedia.org/wiki/Git_(software)#History).  So I think we're okay.

nickanderson:
One important difference in the way it currently works from the way it previously worked.

The git repo information is not available when the staged policy is tagged for release (cf-promises -T)

The agent opportunistically uses .git/HEAD contents to populate cf_promises_release_id. If that file does not exist, then the agent caclulates a sha based on the tree hash. We definately need to have at least .git/HEAD populated correctly before the staged directory is tagged or you cannot easily correlate the release ID back to the proper git commit.

contrib/masterfiles-stage/masterfiles-stage.sh
nickanderson: The version that ships with enterprise currently has:

```
[root@hub dc-scripts]# grep ROOT masterfiles-stage.sh 
  TMP="${ROOT}_tmp"
```

Now I made a similar modification to your proposed changes and I still have some breakage in Design Center that I am trying to track down. With the script we ship everything works fine, but I need to make sure the new and improved version will continue to work. It really seems like it should be as easy using paths that are next to ROOT instead of inside of it. So I am still confused as to why with that modification it still breaks design center.

contrib/masterfiles-stage/README.org
mikeweilgart: Interesting idea.  I think yes, it could.  (It would be in params, not options, as it's VCS_TYPE specific.)  I think simplest would be just one array channel_config which would be expected to contain at least two elements, and if it had an odd number of elements the last would be ignored (perhaps with a warning).

tzz: That's fine, yeah. I was actually thinking of bash associative arrays but those may not be available in older shells:

```
declare -A channels
channels[vcs]="git://..."
...
```

So any implementation is OK with me, as long as it doesn't require separate config files.

contrib/masterfiles-stage/README.org
mikeweilgart: I'll hold off on that until Nick and I come to an agreement regarding mirrored repos and how cf_promises_release_id gets populated.

contrib/masterfiles-stage/common.sh
```diff
@@ -12,146 +12,217 @@ if [ ! "$debug_mode" == "true" ]; then
...
+  # This code could be improved if there is an inexpensive way to check that
+  # a local bare repository is in fact a *mirrored* repository of a specified
+  # GIT_URL, but for now if the local_mirrored_repo is in fact a bare git
+  # repo (guaranteed by the success of the "git fetch" command) then we just
+  # assume it is a *mirrored* repository.
+  #
+  # Since the pathname is directly based on GIT_URL, there is no chance
+  # of *accidental* name collision.
+
+  local_mirrored_repo="${ROOT}/$(printf '%s' "${GIT_URL}" | sed 's/[^A-Za-z0-9._-]/_/g')"
```
mikeweilgart: @nickanderson if you work out why putting stuff in ROOT breaks Design Center I'd be happy to change this.  Otherwise I think I should just add a new parameter "staging_dirs_container" or something like that (to params.sh), and use it the way I've been using ROOT.

I understand ROOT needs to be *in* `params.sh` to integrate with Design Center; I don't think that means I have to use it at all for the staging script.  I originally used it this way only because I thought that was the *purpose* of the ROOT parameter.

contrib/masterfiles-stage/common.sh
```diff
@@ -12,146 +12,217 @@ if [ ! "$debug_mode" == "true" ]; then
...
-          git clean -fd || error_exit "Failed: git clean -fd"
-      ) || error_exit "Failed to stage '${GIT_TAG_OR_COMMIT}' in '${STAGING_DIR}' from '${GIT_URL}'" 
-
-  # If we don't have a git clone wipe the directory and create a new clone and
-  # ensure the proper branch/tag is checked out.
-  else
-      rm -rf "${STAGING_DIR}"/* "${STAGING_DIR}"/.??*
-      git clone --no-hardlinks "${GIT_URL}" "${STAGING_DIR}" && cd "${STAGING_DIR}" && git checkout "${GIT_TAG_OR_COMMIT}" || error_exit "Failed to stage '${GIT_TAG_OR_COMMIT}' from '${GIT_URL}'"
+  mkdir -p "${STAGING_DIR}" || error_exit "Failed: mkdir -p '$STAGING_DIR'"
+  git --git-dir="${local_mirrored_repo}" --work-tree="${STAGING_DIR}" checkout -q -f "$1" ||
```
mikeweilgart: @nickanderson, the obvious (and very inelegant) workaround to allow cf_promises_release_id to be populated with the contents of `.git/HEAD` when `cf-promises -T` is run (something you brought up elsewhere), would be to just copy that single file from the `${local_mirrored_repo}` in this function.  :)

Aside from that, do you just want the VCS_TYPEs GIT and GIT_TAG_OR_COMMIT to use typical `git clone`s?  (It introduces a *lot* more edge cases and potential bugs that way; I don't think the removed lines 110-131 even handle all the edge cases, just some of them.  I'm not going to be using those VCS_TYPEs personally so I don't have any forceful opinion.)

contrib/masterfiles-stage/options.sh
```diff
@@ -1,18 +1,114 @@
...
+      params+=("$param")
+      ;;
+  esac
+done
+
+set -- "${params[@]}"
+
+# Set default variable values
+# (Don't change these without updating usage.)
+MASTERDIR=/var/cfengine/masterfiles
```
mikeweilgart: It's worth noting that the assignment of `MASTERDIR` can be overwritten without using the `-d` or `--deploy-dir` flag, simply by adding a line directly to your `params.sh` file:

    MASTERDIR="/custom/path/to/deployment_dir"

mikeweilgart: @nickanderson, considering we're putting everything else in the params file—path to staging directory (or ROOT is what we use now), channel config file path (to be changed to a bash array directly in `params.sh`)—what do you think about removing the `-d` option and requiring any custom deployment dir to be set directly in `params.sh`?  (We would keep the default `MASTERDIR` value for backward compatibility.)

This would get us even closer to fully drop-in replacement, as then you could specify a custom `MASTERDIR` in your custom `params.sh` file—and put that `params.sh` file in the default location to be read in automatically.

mikeweilgart:
@nickanderson, regarding the population of `cf_promises_release_id` based on `.git/HEAD`—how does it populate it if the contents is a branch name?  I ask because the branch name would be the same in the MASTERDIR and the STAGING_DIR even if there has been another commit made to that branch since MASTERDIR was populated.

contrib/masterfiles-stage/common.sh
```diff
@@ -12,146 +12,244 @@ if [ ! "$debug_mode" == "true" ]; then
...
+  # The value of MASTERDIR that is assigned in masterfiles-stage.sh
+  # is ****IGNORED**** by this function, since there is a separate
+  # MASTERDIR for each separate policy channel.
+  #
+  # Example value for channel_config:
+  # (This is a single value, split into three lines for readability.)
+  #
+  # channel_config=()
+  # channel_config+=( "/var/cfengine/masterfiles" "my_git_tag" )
+  # channel_config+=( "/var/cfengine/policy_channels/channel_1" "my_git_branch" )
```
mikeweilgart: @tzz, I've implemented the bash array change.  Also, @nickanderson, this should fix the issue with Design Center (since I skip using `$ROOT` entirely) and also allow for the `cf_promises_release_id` to be populated correctly.

I only changed `git_stage_policy_channels`, not the `GIT` or `GIT_TAG_OR_COMMIT` vcs types.

tzz: That looks lovely to me.

Either here or in the params file you can mention that rather than hard-coding `/var/cfengine` paths, people can use `cf-promises --show-vars` to get those paths as they were compiled into the program. It could be a useful hint for people on BSD, for instance.

(@nickanderson it may be useful to add a `--show-var sys.masterdir` switch to `cf-promises`?)

nickanderson:
@mikeweilgart regarding the release id, it populates it with the git commit of the branch at that point in time, or with the commit (in the case of a checked out commit or tag). https://github.com/cfengine/core/blob/master/libpromises/generic_agent.c#L1073-L1138

I am going to take another crack at this now.

contrib/masterfiles-stage/common.sh
nickanderson: @tzz --show-var would indeed be helpful. Would need to support namespaces too so `--show-var default:sys.masterdir` should also work.

contrib/masterfiles-stage/common.sh
```diff
@@ -12,146 +12,244 @@ if [ ! "$debug_mode" == "true" ]; then
...
-  else
-      if [ "$debug_mode" == "true" ]; then
-        echo "No git repo found in '${STAGING_DIR}'. Purging directory contents to clear path for fresh clone."
-      fi
-      rm -rf "${STAGING_DIR}"/* "${STAGING_DIR}"/.??*
-      if [ "$debug_mode" == "true" ]; then
-        echo "Cloning git repository '${GIT_URL}' '${GIT_BRANCH}' branch into '${STAGING_DIR}'"
-      fi
-      git clone --no-hardlinks "${GIT_URL}" "${STAGING_DIR}"
+  STAGING_DIR="${ROOT}/_tmp$(echo "$PARAMS" | tr [./] _)"
```
nickanderson: This one is setting the staging dir inside of ROOT

contrib/masterfiles-stage/common.sh
```diff
@@ -12,146 +12,244 @@ if [ ! "$debug_mode" == "true" ]; then
...
-      #else
-          /bin/mkdir -p "${MASTERDIR}" || error_exit "Failed: Creating '${MASTERDIR}'"
-          cd "${STAGING_DIR}" && (
-          chown -R root:root "${STAGING_DIR}" && \
-          rsync -rltDE -c --delete-after --chmod=u+rwX,go-rwx "${STAGING_DIR}/" "${MASTERDIR}/" && echo "Successfully deployed branch '${GIT_BRANCH}' from '${GIT_URL}' to '${MASTERDIR}' on $(date)"
-      )
-      #fi
-  else
-    error_exit "The staged policies in ${STAGING_DIR} could not be validated, aborting."
+  local_mirrored_repo="${ROOT}/$(printf '%s' "${GIT_URL}" | sed 's/[^A-Za-z0-9._-]/_/g')"
```
nickanderson: Here the local mirrored repo is inside of ROOT

mikeweilgart: True—the only one I fixed in the recent commit is GIT_POLICY_CHANNELS.

mikeweilgart:
@nickanderson, thanks for the link to the `cf_promises_release_id` population code.  In that case the latest modification to GIT_POLICY_CHANNELS will allow git tags to populate the release_id correctly, but not git branches.  This is because I used the simple hack of copying `.git/HEAD` from the mirrored repo into the checked out directory.

I have a simple workaround I can implement to allow this to work with git branches as well: By appending `^0` to the refspec being checked out (whether tag, commit hash or branch name), the [HEAD will be detached](http://stackoverflow.com/a/13293010/5419599) so `.git/HEAD` will always end up containing a commit hash.  If you (or @tzz) have any ideas of a more elegant way I'm open to suggestions, though.

I still don't want to use a typical clone for policy channels, because I want to keep the "deploy" operation atomic (from staging dir to masterdir) and I don't see a way to avoid multiple `git fetch` operations if I use a standard clone.

contrib/masterfiles-stage/common.sh
```diff
@@ -12,146 +12,244 @@ if [ ! "$debug_mode" == "true" ]; then
...
-  if ! type "git" > /dev/null; then
-    error_exit "git not found on path: ${PATH}"
+avoid_triggering_unneeded_policy_updates() {
+  # cf_promises_validated gets updated by any run of cf-promises,
+  # but hosts use cf_promises_validated as the flag file to see
+  # if they need to update everything else (the full policy set.)
+  #
+  # cf_promises_release_id is the same for a given policy set
+  # unless changes have actually been made to the policy, so it
+  # can be used to check if we want to trigger an update.
```
mikeweilgart: @nickanderson, is there a reason why [the update policy](https://github.com/cfengine/masterfiles/blob/master/cfe_internal/update/update_policy.cf#L107-L114) uses `cf_promises_validated` rather than `cf_promises_release_id` as the flag file to trigger a full policy set download?

nickanderson: cf_promises_validated pre-dates cf_promises_release_id, this is something we can probably consider changing in the MPF in the future.

nickanderson:
After digging around a while, I am thinking that the design center breakage might be caused by the removal of sourcing the params.sh from common.sh

nickanderson:
Yep, I just added the sourcing of the params file back in and design center started working.

nickanderson:
In addition to moving the other stuff outside of ROOT.

mikeweilgart:
That's interesting...how shall we integrate that with options parsing to specify the path to the params file?

nickanderson:
Hum, perhaps we should rename options.sh to masterfiles-stage_options.sh (because there are other scripts that use that common.sh and expect to have the stuff from the params file sourced). If we rename options.sh to be more specific, then it leaves room for me to extend the other dc scripts by adding their own params. 

I am doing some quick testing to see if a minor mod on one of the other dc scripts can make having the sourcing done from common.sh unnecessry.

mikeweilgart:
@nickanderson, one idea I've had recur to me many times while modifying this script is, what about actually properly modularizing the functions rather than using global variables everywhere?  (`bash` supports a "local" keyword for variable definitions, which isn't POSIX—but we're already using `bash`-specific features such as arrays.)  The dependencies on global variable definitions and attempting to trace down where they are defined is becoming increasingly complex.

So, for example, rather than having `git_setup_local_mirrored_repo` set a global variable `$local_mirrored_repo`, it could use a local variable `local var=whatever` within the function definition, and then return that by having the last line of the function be `printf %s "$var"`.  Then when calling that function, you would do it with command substitution to capture the value: `myreponame="$(git_setup_local_mirrored_repo)"`

(It could even accept `$GIT_URL` as an argument, but that one is probably okay as a global variable.)

We don't have to do this, but now that you mention that other scripts depend on the functions in `common.sh`...the fact of everything being global variables may bite us sooner or later.

mikeweilgart:
Also, I think inlining the option parsing into `masterfiles-stage.sh` directly would be better than renaming it to `masterfiles-stage_options.sh`.

nickanderson:
OK lets in-line the option parsing if you think that's better.

With regard to modularizing functions. It would be nice if the script was POSIX compliant, but I only NEED it to work on modern shells particularly on linux since this is used by the enterprise hub.

contrib/masterfiles-stage/common.sh
```diff
@@ -12,175 +12,253 @@ if [ ! "$debug_mode" == "true" ]; then
...
+  # The '^0' at the end of the refspec
+  # populates HEAD with the SHA of the commit
+  # rather than potentially using a git branch name.
+  # Also see http://stackoverflow.com/a/13293010/5419599
+  # and https://github.com/cfengine/core/pull/2465#issuecomment-173656475
+  git --git-dir="${local_mirrored_repo}" --work-tree="${temp_stage}" checkout -q -f "${2}^0" ||
+    error_exit "Failed to checkout '$2' from '${local_mirrored_repo}'"
+  # Grab HEAD so it can be used to populate cf_promises_release_id
+  mkdir -p "${temp_stage}/.git"
+  cp "${local_mirrored_repo}/HEAD" "${temp_stage}/.git/"
```
mikeweilgart: @nickanderson, this should handle all issues with `cf_promises_release_id`.

contrib/masterfiles-stage/common.sh
```diff
@@ -12,175 +12,252 @@ if [ ! "$debug_mode" == "true" ]; then
...
+  # channel_config+=( "/var/cfengine/masterfiles" "my_git_tag" )
+  # channel_config+=( "/var/cfengine/policy_channels/channel_1" "my_git_branch" )
+
+  # Simplest validation check first
+  set -- "${channel_config[@]}"
+  [ "$#" -gt 1 ] ||
+    error_exit "The channel_config array must have at least two elements."
+
+  check_git_installed
+  git_setup_local_mirrored_repo "$dir_to_hold_mirror"
```
mikeweilgart: Note the new parameter `$dir_to_hold_mirror` to be set in the PARAMS file.  This is so the policy channels framework doesn't use ROOT in any way.

contrib/masterfiles-stage/common.sh
```diff
@@ -12,175 +12,252 @@ if [ ! "$debug_mode" == "true" ]; then
...
+  [ "$#" -eq 1 ] && error_exit "Trailing parameter found, please fix params file: '$1'"
+}
+
+git_masterstage() {
+  # Depends on $GIT_URL, $ROOT and $MASTERDIR
+  # Accepts one arg:
+  # $1 - refspec to stage
+
+  check_git_installed
+  git_setup_local_mirrored_repo "$( dirname "$ROOT" )"
```
mikeweilgart: I left GIT and GIT_TAG_OR_COMMIT using ROOT, but they don't put anything in it—they only put things (one thing, a mirrored repo) in its parent directory.

contrib/masterfiles-stage/common.sh
```diff
@@ -12,175 +12,252 @@ if [ ! "$debug_mode" == "true" ]; then
...
+  git_deploy_refspec "$MASTERDIR" "$1"
+  echo "Successfully deployed '$1' from '${GIT_URL}' to '${MASTERDIR}' on $(date)"
```
mikeweilgart: git_masterstage() is now only three function calls!  :)

contrib/masterfiles-stage/masterfiles-stage.sh
```diff
@@ -1,45 +1,157 @@
...
-  # Example: 
-  # ROOT="/opt/cfengine/masterfiles_staging"
-  # PARAMS="/var/cfengine/policychannel/production_1.sh"
-  # TMP=/opt/cfengine/masterfiles/staging/_tmp_var_cfengine_policychannel_production_1_sh
+# Load common functionality, upstream implementations
+mydirname="$(dirname "$0")";
+source "${mydirname}/common.sh" || { echo "common.sh couldn't be sourced" >&2 ; exit 17 ; }
 
-  TRANSLATED_PARAMS=$(echo $PARAMS | tr [./] _)
-  STAGING_DIR="${ROOT}/_tmp${TRANSLATED_PARAMS}"
```
mikeweilgart: I folded this [back into `svn_branch()`](https://github.com/cfengine/core/pull/2465/files#diff-623a8bfce9e4ae7029ea70bf77753723R230) because I'm not using this logic at all anymore for any git staging method.

contrib/masterfiles-stage/masterfiles-stage.sh
```diff
@@ -1,45 +1,157 @@
...
+  # VCS_TYPE (set in the PARAMS file)
+  # and additional vars set in PARAMS depending on the VCS_TYPE.
 
 case "${VCS_TYPE}" in
+    GIT_POLICY_CHANNELS)
+        git_stage_policy_channels
+        ;;
     GIT_TAG_OR_COMMIT)
-        git_tag_or_commit_masterstage
+        git_masterstage "${GIT_TAG_OR_COMMIT}"
```
mikeweilgart: (GIT_TAG_OR_COMMIT as a vcs type could be deprecated entirely in favor of GIT, since the underlying code is now completely identical.)

contrib/masterfiles-stage/common.sh
```diff
@@ -12,175 +12,252 @@ if [ ! "$debug_mode" == "true" ]; then
...
+  # Ensure absolute pathname is given
+  [ "${1:0:1}" = / ] ||
+    error_exit "You must specify absolute pathnames in channel_config: '$1'"
+  mkdir -p "$(dirname "$1")" || error_exit "Failed to mkdir -p dirname $1"
+    # We don't mkdir $1 directly, just its parent dir if that doesn't exist.
+
+  ########################## 1. CREATE EMPTY TEMP DIR
+  # Put staging dir right next to deploy dir to ensure it's on same filesystem
+  local temp_stage
+  temp_stage="$(mktemp -d --tmpdir="$(dirname "$1")" )"
```
mikeweilgart: Note use of temp directory rather than a STAGING_DIR.

nickanderson:
I have proposed some changes to the Design Center tooling to add in the direct sourcing of the PARAMS file.

Also I think we can also drop the GIT_TAG_OR_COMMIT VCS type. It was never publicly used, and git_masterstage works the same now with your refspec modifications.

Another thing that I think we should consider doing is fixing the permissions on the staging directory files before deploying to the final destination. It's not uncommon for things to get checked into git with bad permissions, and that leads to repairs on each deploy.

For example:

```
[root@hub dc-scripts]# cf-agent -KIf update.cf --define cfengine_internal_masterfiles_update
    info: Executing 'no timeout' ... '/var/cfengine/httpd/htdocs/api/dc-scripts/masterfiles-stage.sh'
    info: Command related to promiser '/var/cfengine/httpd/htdocs/api/dc-scripts/masterfiles-stage.sh' returned code defined as promise kept 0
    info: Completed execution of '/var/cfengine/httpd/htdocs/api/dc-scripts/masterfiles-stage.sh'
    info: Object '/var/cfengine/masterfiles/./sketches/sketch_template/scripts/sample.sh' had permission 0700, changed it to 0600
    info: Object '/var/cfengine/masterfiles/./sketches/sketch_template/modules/mymodule' had permission 0700, changed it to 0600
    info: Setting restart class 'no_vacuumdb' for promise '/var/cfengine/bin/vacuumdb'
[root@hub dc-scripts]# cf-agent -KIf update.cf --define cfengine_internal_masterfiles_update
    info: Executing 'no timeout' ... '/var/cfengine/httpd/htdocs/api/dc-scripts/masterfiles-stage.sh'
    info: Command related to promiser '/var/cfengine/httpd/htdocs/api/dc-scripts/masterfiles-stage.sh' returned code defined as promise kept 0
    info: Completed execution of '/var/cfengine/httpd/htdocs/api/dc-scripts/masterfiles-stage.sh'
    info: Object '/var/cfengine/masterfiles/./sketches/sketch_template/scripts/sample.sh' had permission 0700, changed it to 0600
    info: Object '/var/cfengine/masterfiles/./sketches/sketch_template/modules/mymodule' had permission 0700, changed it to 0600
    info: Setting restart class 'no_vacuumdb' for promise '/var/cfengine/bin/vacuumdb'
```

contrib/masterfiles-stage/common.sh
```diff
@@ -12,175 +12,252 @@ if [ ! "$debug_mode" == "true" ]; then
...
+  git --git-dir="${local_mirrored_repo}" --work-tree="${temp_stage}" checkout -q -f "${2}^0" ||
+    error_exit "Failed to checkout '$2' from '${local_mirrored_repo}'"
+  # Grab HEAD so it can be used to populate cf_promises_release_id
+  mkdir -p "${temp_stage}/.git"
+  cp "${local_mirrored_repo}/HEAD" "${temp_stage}/.git/"
+
+  ########################## 3. SET PERMISSIONS ON POLICY SET
+  chown -R root:root "${temp_stage}" || error_exit "Unable to chown '${temp_stage}'"
+  chmod -R go-rwx    "${temp_stage}" || error_exit "Unable to chmod '${temp_stage}'"
+
```
nickanderson: Files should be 600, and directories should be 700

```
  find "${temp_stage}" -type f  | xargs chmod 600
  find "${temp_stage}" -type d  | xargs chmod 700
```

Maybe there is a better way than using find?

mikeweilgart: Not better than using `find`, but better than using `xargs`, which will [cause issues when filenames contain whitespace](http://unix.stackexchange.com/q/131766/135943).  Instead I'll fix this to do the whole action with `find` (using the `-exec` flag).  Thanks for pointing out the permissions issue; I agree it will be good to get rid of those warning messages.

nickanderson:
@mikeweilgart my proposed changes to upstream tooling were accepted, so this will work with Design Center in future releases as a drop in replacement. Should probably note that its only drop-in replacement on 3.7.3+ and 3.8.2+ (not sure if it will make 3.8.1 or not)

nickanderson:
@mikeweilgart also thinking that we should switch to using logger so that things goto syslog instead of a custom log file.

Found this interesting:
http://urbanautomaton.com/blog/2014/09/09/redirecting-bash-script-output-to-syslog/

Thoughts?

nickanderson:
Here is a nitpick.

Staging twice in a row logs that masterfiles were successfully deployed even though nothing was changed in the deployment directory.

Perhaps instead we can log why deployment was skipped.

mikeweilgart:
Right now my code never ever skips deployment.  It only will skip triggering an update on hosts.  But yes, if we were logging to syslog, that would be excessive logging.

contrib/masterfiles-stage/README.org
```diff
@@ -64,42 +69,76 @@ sure that `masterfiles-stage.sh` is executable.
...
-
-- GIT_TAG_OR_COMMIT
+- GIT_BRANCH - The git refspec to checkout.
+  It can be a branch name, a tag name, a commit hash or a partial hash.
   - Example values:
+    - master
+    - my_tag
     - 88335d36b48c8808b12b48667a463182dc8d0338
-    - mytag
+    - cb375d0f
```
mikeweilgart: @nickanderson, is the GIT vcs type publicly used, or can I go ahead and change `GIT_BRANCH` to `GIT_REFSPEC`?

contrib/masterfiles-stage/README.org
```diff
@@ -64,42 +69,76 @@ sure that `masterfiles-stage.sh` is executable.
...
+
+- dir_to_hold_mirror - The directory where the git mirror repo will
+  be created for GIT_POLICY_CHANNELS.  (With VCS_TYPE=GIT, the repo
+  is placed alongside of ROOT.)
+  - Example value:
+    - /opt/cfengine
```
mikeweilgart: @nickanderson, maybe we could improve the name, and use it for all VCS_TYPEs?  Since this is really what ROOT was used for, but ROOT is something we should leave alone since Design Center uses it...

Maybe call it "Staging Dir" with description "The location where persistent files (such as a git repository directory, or an svn checkout) may be created and accessed by the staging script."

contrib/masterfiles-stage/README.org
```diff
@@ -172,11 +211,26 @@ bundle agent stage_masterfiles
...
+}
+#+end_src
+
+
 * Phased Rollout with Multiple Staged Masterfiles
 
 The `masterfiles-stage.sh` script can support an implementation for
 phased rollout. Phased Rollout is the process of releasing a *policy*
 change to a population in a controlled manner.
+(GIT_POLICY_CHANNELS is one such implementation.)
```
mikeweilgart: I should probably include more documentation of this...what do you think?  (The doc I cc'ed you on the other day was more intended for in-house use.)

contrib/masterfiles-stage/common.sh
```diff
@@ -12,175 +12,250 @@ if [ ! "$debug_mode" == "true" ]; then
...
+  git --git-dir="${local_mirrored_repo}" --work-tree="${temp_stage}" checkout -q -f "${2}^0" ||
+    error_exit "Failed to checkout '$2' from '${local_mirrored_repo}'"
+  # Grab HEAD so it can be used to populate cf_promises_release_id
+  mkdir -p "${temp_stage}/.git"
+  cp "${local_mirrored_repo}/HEAD" "${temp_stage}/.git/"
+
+  ########################## 3. SET PERMISSIONS ON POLICY SET
+  chown -R root:root "${temp_stage}" || error_exit "Unable to chown '${temp_stage}'"
+  find "${temp_stage}" \( -type f -exec chmod 600 {} + \) -o \
+                       \( -type d -exec chmod 700 {} + \)
```
mikeweilgart: POSIX compliant permissions fix.  :)  (I know we don't need POSIX compliance throughout the script, but it's nice to not add "GNU coreutils" to the dependency list.)

contrib/masterfiles-stage/common.sh
```diff
@@ -12,175 +12,250 @@ if [ ! "$debug_mode" == "true" ]; then
...
+      # release id is the same in stage and deploy dir
+      # so prevent triggering update on hosts by keeping old "validated" flag file
+      cp -a "${1}/cf_promises_validated" "${temp_stage}/"
+    fi
+    local third_dir
+    third_dir="$(mktemp -d --tmpdir="$(dirname "$1")" )"
+    mv "${1}" "${third_dir}"  || error_exit "Can't mv ${1} to ${third_dir}"
+    mv "${temp_stage}" "${1}"          || error_exit "Can't mv ${temp_stage} to ${1}"
+    rm -rf "${third_dir}"
   fi
```
mikeweilgart: In this `if` construct, the policy set is rolled out in all cases.  The only difference is (line 121) whether the deployment dir has to be moved out of the way first or not, and (line 125) whether the `cf_promises_validated` file should be grabbed from the deployment dir prior to the rollout, to prevent triggering full policy set updates on all bootstrapped hosts.

contrib/masterfiles-stage/README.org
nickanderson: No, GIT_BRANCH is publicly used, and can be set by issuing an api call to mission portal. So we have to keep that one the way it is for now. I would aim to clean that up if/when we formalize policy channels in the API.

contrib/masterfiles-stage/common.sh
nickanderson: yeahp

contrib/masterfiles-stage/README.org
nickanderson: I do think it would be nice to include more docs on your GIT_POLICY_CHANNELS and that doc was a good start.

contrib/masterfiles-stage/README.org
nickanderson: Yeah the name could be improved.

Perhaps mirror_dir?
And note that when using mirror_dir any repository should be mirrored into a subdirectory.

nickanderson:
I do think that it would be good if we avoided the additional churn caused by deploy fresh each time. Do you have a good idea in mind to prevent the unnecessary replacement?

contrib/masterfiles-stage/common.sh
nickanderson: So we could skip the deployment if the release id is the same to avoid the "potentially" unncessary churn. I say potentially because we only operate on partial knowledge of how different the currently deployed policy set is from the staged one (only looking at the release_id).

I dislike that assumption, but I don't have any great ideas to do further validation without it getting potentially expensive. At least move operations on the same file system are fast and not cpu intensive. A recursive diff or hash could get expensive.

contrib/masterfiles-stage/README.org
nickanderson: Oh and so is GIT as a VCS type. Its the only one supported by Mission Portal for Design Center currently.

nickanderson:
@mikeweilgart regarding population of .git/HEAD I think its good that it always contains a commit sha. The point is to be able to easily associate that back to a specific state in the repo and commit shas are what we want for that. So I think what you have done will give us what we need there.

contrib/masterfiles-stage/common.sh
mikeweilgart: I dislike that assumption also; that's why I decided to let `git` handle the checking.

We already need to do a `git fetch` to see if there's updated code, and we need to do a `git checkout` to check the latest changes.  We *could* do a `git checkout` directly into the deployment dir and thereby let `git` handle all the comparisons between the files present and the files that *should* be present...but that loses the "atomicness" of the deployment operation and opens a potentially long window (several seconds) wherein a broken policy set will be live and accessible.

So the compromise I implemented is, *pretend* we are doing a `git checkout` directly into the deployment dir, but make it atomic by dropping it right *next* to the deploy dir and then actually putting it in place with a `mv` command.  (Then we use `rm -rf` to get rid of *whatever* was present in the deploy dir before the `mv`, as opposed to having unneeded files removed by `git checkout -f` or `git clean`.)

mikeweilgart:
Re `git/HEAD`, great, glad it's useful!  :)

contrib/masterfiles-stage/common.sh
mikeweilgart: P.S.: A nice command you can run to see the temp directories getting moved into place is, assuming your `channel_config` array populates several channel directories within `/var/cfengine/policy_channels/`:

    /var/cfengine/channel_scripts/masterfiles-stage.sh & watch -n .1 'ls -li /var/cfengine/policy_channels/ | sort -n'

contrib/masterfiles-stage/common.sh
tzz: I did this in https://github.com/cfengine/core/pull/2480

nickanderson:
@mikeweilgart I've worked my way back around to this again.

I have gone through some simple tests with this code as a drop in replacement on top of 3.7.3.

- Without doing any thing special, I am able to use version control integration as expected from Mission Portal.

- Of note (probably already discussed previously in the thread) the target deployment directory gets replaced each time this runs. However the special cf_promises_validated file has special handling to preserve the previous content if cf_promises_release_id has not changed. This handling is necessary since successful validation when tagging with `cf-promises -T DIR` will always re-write the contents of cf_promises_validated.

mikeweilgart:
Perhaps I should add the policy files showing how the policy channel assignment gets pulled in by the hosts and used?

I also wrote a script to summarize the channel assignments for the hosts.  Of course, you can use policy channels just for test directories that you run directly on the Hub, so the other pieces aren't necessarily a *crucial* ingredient to the masterfiles stage script upgrade, but it seems that all the pieces put together could be quite valuable to others.

I may just create a dedicate repository and include all examples and supporting scripts there.

contrib/masterfiles-stage/common.sh
```diff
@@ -12,175 +12,255 @@ if [ ! "$debug_mode" == "true" ]; then
...
+  error_exit "Update policy staged in ${temp_stage} could not be validated, aborting."
+
+  ########################## 5. ROLL OUT POLICY SET FROM TEMP DIR TO DEPLOY DIR
+  if ! [ -d "$1" ] ; then
+    # deploy dir doesn't exist yet
+    mv "${temp_stage}" "$1" || error_exit "Failed to mv $temp_stage to $1."
+    trap -- EXIT
   else
-    error_exit "The staged policies in ${STAGING_DIR} could not be validated, aborting."
+    if /usr/bin/cmp -s "${temp_stage}/cf_promises_release_id" \
```
nickanderson: Here we find that the release id is unchanged. And we preserve cf_promises_validated to avoid unecessry client updates. But I think that it would be useful to perform a secondary check at this point to see if there are actually any differences from the staged policy release to the currently deployed policy release and then only re-deploy if a change is found. A simple check could use diff -r. The main benefit of this would be to only deploy when necessary, instead of just doing it anyway. It's the small difference between being idempotent (same action repeatedly results in same final state) vs convergence which trys to limit unnecessary changes (at the expense of additional checking).

But I won't block the merge based on this.

nickanderson:
Merged in https://github.com/cfengine/core/pull/2707

PR #2466, opened 2016-01-07
Fix incorrect failure reason when there is a typo in test outcome.

kacf:
If the test outcome is not exactly "Pass" or "FAIL" the test script
would believe that the test had passed when it should have failed (in
the case of test_soft_fail). This is obviously wrong, it should say
that the test outcome was unknown and not expected.

Readd test that was reverted earlier because of this.


kacf:
trigger build

kacf:
trigger build

PR #2467, opened 2016-01-07
Redmine #7274: Fix cases where action_policy warn still produces errors

amousset:
Fix cases missing in 8948fcaf5e376d5ee89e5b3cd933f4ed867e25e2.

### DIFF
--- a/cf-agent/package_module.c
+++ b/cf-agent/package_module.c
@@ -1125,9 +1125,9 @@ PromiseResult FileInstallPackage(const char *package_file_path,
      
     if (action == cfa_warn || DONTDO)
     {
-         Log(LOG_LEVEL_WARNING, "Should install file type package: %s",
+         Log(LOG_LEVEL_VERBOSE, "Should install file type package: %s",
              package_file_path);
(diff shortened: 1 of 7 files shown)


cf-agent/package_module.c
```diff
@@ -1125,9 +1125,9 @@ PromiseResult FileInstallPackage(const char *package_file_path,
      
     if (action == cfa_warn || DONTDO)
     {
-         Log(LOG_LEVEL_WARNING, "Should install file type package: %s",
+         Log(LOG_LEVEL_VERBOSE, "Should install file type package: %s",
```
kacf: So if I read this correctly, the reason this one and a few others are "verbose" and not "warning" is because of the `switch` statement in `verify_new_packages.c` that checks for `PROMISE_RESULT_WARN`, is that correct?

amousset: Yes, there is no need for two messages with almost the same content at warn level.

kacf:
trigger build

kacf:
Thanks Alexis! Merged manually and cherry-picked to 3.6.x, 3.7.x and 3.8.x.

PR #2468, opened 2016-01-08
Clean up ConsiderFile

amousset:
This intends to make ConsiderFile more readable, and should be functionally equivalent, except
* Remove the test relying on the verbosity level. For example, in the current version, if you copy a directory containing afile named ". . .", it will be copied only at some verbosity levels.
* Do not consider a file with an empty name
* Display information first, so that it is always displayed.

### DIFF
--- a/cf-agent/files_properties.c
+++ b/cf-agent/files_properties.c
@@ -48,12 +48,26 @@ static const char *const SKIPFILES[] =
 static bool ConsiderFile(const char *nodename, const char *path, struct stat *stat)
 {
     int i;
-    const char *sp;
+    
+    if (stat != NULL)
+    {
+        if (S_ISLNK(stat->st_mode))
+        {
+            Log(LOG_LEVEL_INFO, "'%s' is a symbolic link", nodename);
+        }
+        else if (S_ISDIR(stat->st_mode))
+        {
+            Log(LOG_LEVEL_INFO, "'%s' is a directory", nodename);
+        }
+        
+        Log(LOG_LEVEL_VERBOSE, "'%s' has size %ld and full mode %o", nodename, (unsigned long) (stat->st_size),
+              (unsigned int) (stat->st_mode));
+    }
 
     if (strlen(nodename) < 1)
     {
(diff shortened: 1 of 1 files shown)


kacf:
Completely agree! What was the logic in that old code...? :-P

kacf:
trigger build

kacf:
Thanks, merged manually. I added the text you wrote here in the pull request to the commit message itself. It's a good idea to do that in the future, so it's easier to understand when looking at the history later.

jimis:
This one caused annoying messages, traced from @nickanderson with cfengine/masterfiles#675 and #2546.

```
    info: '.' is a directory
    info: '..' is a directory
    info: 'themes' is a directory
    info: 'default' is a directory
    info: 'images' is a directory
[ ... ]
```

Before this pull request, these INFO messages used to print always after an error message, but now they print every time:
```c
    Log(LOG_LEVEL_ERR, "Suspicious looking file object '%s' masquerading as hidden file in '%s'", nodename, path);

    if (S_ISLNK(stat->st_mode))
    {
        Log(LOG_LEVEL_INFO, "'%s' is a symbolic link", nodename);
    }
    else if (S_ISDIR(stat->st_mode))
    {
        Log(LOG_LEVEL_INFO, "'%s' is a directory", nodename);
    } 
```

jimis:
Filed a bug report at https://tracker.mender.io/browse/CFE-2327.

@amousset can you have a look?

amousset:
The options I see:
* Merge #2546 (DEBUG may even be more appropriate)
* Only display these messages for suspicious files we skip (which was more or less the previous behavior, but only in the case where the file name only contained dots and spaces, which was not skipped)

I think the second is the best option as the general DEBUG/VERBOSE messages about file stat should probably be in the stat function (like cf_stat) itself.

jimis:
@amousset yes printing a message for suspicious files we skip sounds sane. Feel free to make the message more readable. :-)

amousset:
@jimis I opened #2548, finally removing these messages.

PR #2469, opened 2016-01-12
Add default body for all promise types

amousset:
Based on #2342.

* Use the `bodydefault` namespace instead of `default`
* Free `default_body_name`
* Add the test on files promises

The limitation when using a particular namespace for default bodies is that they cannot be applied to an arbitrary namespace, but only to `default`.

### DIFF
--- a/libpromises/eval_context.h
+++ b/libpromises/eval_context.h
@@ -176,6 +176,10 @@ const void  *EvalContextVariableControlCommonGet(const EvalContext *ctx, CommonC
  */
 const Bundle *EvalContextResolveBundleExpression(const EvalContext *ctx, const Policy *policy,
                                                  const char *callee_reference, const char *callee_type);
+
+const Body *EvalContextFindFirstMatchingBody(const Policy *policy, const char *type,
+                                             const char *namespace, const char *name);
+
 /**
   @brief Returns a Sequence of const Body* elements, first the body and then its parents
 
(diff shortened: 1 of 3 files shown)


kacf:
+1 for the test, and I like the idea of using the `bodydefault` namespace. Makes it much less likely that this will clash with existing body names.

kacf:
FYI, I timed the acceptance test run, since this is a very central piece of code, but I didn't see any significant change in performance.

kacf:
trigger build

kacf:
In our test farm the `tests/unit/expand_test` test is segfaulting with this change. Does it segfault for you too?

amousset:
Yes. This test defines a `dummy` promise type which has no syntax, added a check for empty syntax in `AddDefaultBodiesToPromise`.

kacf:
Thanks!

kacf:
trigger build

kacf:
Build is green! Could you add some documentation for this as well? Looks like this is the most natural place: https://github.com/cfengine/documentation/blob/master/reference/language-concepts/bodies.markdown, which corresponds to https://docs.cfengine.com/docs/master/reference-language-concepts-bodies.html.

amousset:
Added in https://github.com/cfengine/documentation/pull/1309.

kacf:
Thanks a lot, merged manually!

PR #2470, opened 2016-01-13
Redmine #7808: Do not drop privileges when changes permissions on non fifo

amousset:
Avoid dropping privileges when changing permissions on a file which is not a FIFO.

Redmine: https://dev.cfengine.com/issues/7808

### DIFF
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -831,54 +831,52 @@ int safe_chmod(const char *path, mode_t mode)
         goto cleanup;
     }
 
-    /* save old euid */
-    olduid = geteuid();
-
-#ifndef CHMOD_SETEUID_WORKS
-    if (mode & 07000 && statbuf.st_uid != 0)
+    if (S_ISFIFO(statbuf.st_mode))
     {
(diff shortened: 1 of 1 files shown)


amousset:
There is also a test for this issue in #2428.

kacf:
This looks good to me. @worr might also want to take a look.

kacf:
trigger build

kacf:
trigger build

kacf:
trigger build

worr:
lgtm

kacf:
This fix has an additional caveat, as the tests revealed: The method doesn't work if the file being operated on is a symlink to a FIFO. This is getting into rather advanced and security sensitive territory. I attempted to solve this myself as you can see in #2482, by following symlinks and opening the "true" parent directory of the file, instead of the parent directory of the symlink that points to it. Luckily the code for following symlinks and opening the true parent directory was already there, so I just refactored to use it.

Would be nice if you could take a look.

kacf:
Superseded by #2482.

PR #2471, opened 2016-01-13
Redmine#7885: add data_matching() function paralleling variablesmatching()

tzz:
As proposed in https://dev.cfengine.com/issues/7885 this is a function that parallels `variablesmatching()` in every way except it collects the value of the variables in addition to their name.

Acceptance test included.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -1079,9 +1079,9 @@ static FnCallResult FnCallClassesMatching(EvalContext *ctx, ARG_UNUSED const Pol
 }
 
 
-static StringSet *VariablesMatching(const EvalContext *ctx, VariableTableIterator *iter, const Rlist *args)
+static JsonElement *VariablesMatching(const EvalContext *ctx, const FnCall *fp, VariableTableIterator *iter, const Rlist *args, bool collect_full_data)
 {
-    StringSet *matching = StringSetNew();
+    JsonElement *matching = JsonObjectCreate(10);
 
     const char *regex = RlistScalarValue(args);
     pcre *rx = CompileRegex(regex);
@@ -1124,7 +1124,36 @@ static StringSet *VariablesMatching(const EvalContext *ctx, VariableTableIterato
 
             if (pass)
             {
-                StringSetAdd(matching, expr);
+                JsonElement *data = NULL;
+                bool allocated = false;
+                if (collect_full_data)
+                {
+                    data = VarRefValueToJson(ctx, fp, v->ref, NULL, 0, true, &allocated);
+                }
+
+                /*
+                 * When we don't collect the full variable data
+                 * (collect_full_data is false), we still create a JsonObject
+                 * with empty strings as the values. It will be destroyed soon
+                 * afterwards, but the code is cleaner if we do it this way than
+                 * if we make a JsonArray in one branch and a JsonObject in the
+                 * other branch. The empty strings provide assurance that
+                 * serializing this JsonObject (e.g. for logging) will not
+                 * create problems. The extra memory usage from the empty
+                 * strings is negligible.
+                 */
+                if (data == NULL)
+                {
+                    JsonObjectAppendString(matching, expr, "");
+                }
+                else
+                {
+                    if (!allocated)
+                    {
+                        data = JsonCopy(data);
+                    }
+                    JsonObjectAppendElement(matching, expr, data);
+                }
             }
             else
             {
@@ -1147,6 +1176,8 @@ static StringSet *VariablesMatching(const EvalContext *ctx, VariableTableIterato
 
 static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const Policy *policy, const FnCall *fp, const Rlist *finalargs)
 {
+    bool fulldata = (strcmp(fp->name, "variablesmatching_as_data") == 0);
+
     if (!finalargs)
     {
         FatalError(ctx, "Function '%s' requires at least one argument", fp->name);
@@ -1165,17 +1196,25 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
 
     {
         VariableTableIterator *iter = EvalContextVariableTableIteratorNew(ctx, NULL, NULL, NULL);
-        StringSet *global_matches = VariablesMatching(ctx, iter, finalargs);
+        JsonElement *global_matches = VariablesMatching(ctx, fp, iter, finalargs, fulldata);
+        VariableTableIteratorDestroy(iter);
 
-        StringSetIterator it = StringSetIteratorInit(global_matches);
-        const char *element = NULL;
-        while ((element = StringSetIteratorNext(&it)))
+        assert (JsonGetContainerType(global_matches) == JSON_CONTAINER_TYPE_OBJECT);
+
+        if (fulldata)
         {
-            RlistPrepend(&matches, element, RVAL_TYPE_SCALAR);
+            return (FnCallResult) { FNCALL_SUCCESS, (Rval) { global_matches, RVAL_TYPE_CONTAINER } };
         }
 
-        StringSetDestroy(global_matches);
-        VariableTableIteratorDestroy(iter);
+        JsonIterator jiter = JsonIteratorInit(global_matches);
+        const char *key;
+        while ((key = JsonIteratorNextKey(&jiter)) != NULL)
+        {
+            assert (key != NULL);
+            RlistPrepend(&matches, key, RVAL_TYPE_SCALAR);
+        }
+
+        JsonDestroy(global_matches);
     }
 
     return (FnCallResult) { FNCALL_SUCCESS, { matches, RVAL_TYPE_LIST } };
@@ -8905,7 +8944,7 @@ const FnCallType CF_FNCALL_TYPES[] =
     FnCallTypeNew("userexists", CF_DATA_TYPE_CONTEXT, USEREXISTS_ARGS, &FnCallUserExists, "True if user name or numerical id exists on this host",
                   FNCALL_OPTION_NONE, FNCALL_CATEGORY_SYSTEM, SYNTAX_STATUS_NORMAL),
     FnCallTypeNew("variablesmatching", CF_DATA_TYPE_STRING_LIST, CLASSMATCH_ARGS, &FnCallVariablesMatching, "List the variables matching regex arg1 and tag regexes arg2,arg3,...",
-                  FNCALL_OPTION_VARARG, FNCALL_CATEGORY_UTILS, SYNTAX_STATUS_NORMAL),
+                  FNCALL_OPTION_VARARG, FNCALL_CATEGORY_DATA, SYNTAX_STATUS_NORMAL),
 
     // Functions section following new naming convention
     FnCallTypeNew("string_mustache", CF_DATA_TYPE_STRING, STRING_MUSTACHE_ARGS, &FnCallStringMustache, "Expand a Mustache template from arg1 into a string using the optional data container in arg2 or datastate()",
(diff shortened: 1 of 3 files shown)


nickanderson:
Very useful. I most commonly am interested in the values of the variables that match given tags, and accessing those values especially in the case of namespaced variables can be troublesome.

tzz:
Also worth noting is that this will convert any variable to JSON, including slists and scalars. This is hard to do in policy.

tzz:
OK, this is ready for review, it's been rebased after #2454 was merged.

nickanderson:
So this is a data container that is returned with the values of all the variables matching given names/tags correct?

ping @cfrost

jpvlsmv:
Can we please not make the same mistake that PHP did with randomly-permuted and oddly-separated words in function names?  If the data matching function is meant to be parallelised to the variables matching function, skip the _ and make them datamatching() and variablesmatching()

tzz:
@jpvlsmv the old names are illegible, e.g. `parsestringarray()` and `hostsinnetgroup()` and `filesexist()` vs. `fileexists()`.

Until 3.6 the core maintainers' policy was to disallow underscores, but since then they've been allowed in function names. A good example is `splitstring()` which was kept for backwards compatibility (including broken behavior) but a better version exists as `string_split()`.

I don't think any naming policy would satisfy everyone, but personally I'd rather have legible function names than not.

tzz:
Rebased again and ready for review.

tzz:
@jimis this is a small PR with a nice benefit for the users. There are some whitespace fixes that I can split into a separate commit if you need that.

tests/acceptance/01_vars/02_functions/variablesmatching.cf
```diff
@@ -21,25 +25,15 @@ bundle agent test
       "x_vars" slist => variablesmatching("default:init.test_fbeae67f3e347b5e0032302200141131.*", "x");
       "z_vars" slist => variablesmatching("default:init.test_fbeae67f3e347b5e0032302200141131.*", "z");
 
-      "count" int => length(vars);
-      "x_count" int => length(x_vars);
-      "z_count" int => length(z_vars);
+      "fullvars" data => data_matching("default:init.test_b37ce1b03955522848f0a82d0b2b2e21.*");
```
jimis: 1. I find this confusing. What I was expecting from a function named `data_matching()` was matching the argument with the variables' contents. Instead what I see is that it matches exactly what `variablesmatching()` matches, but returns a data container instead of an slist.

2. Why is this useful? This does not address the main concern of the ticket it mentions (CFE-2265, old Redmine#7885) which is to avoid double-dereferencing the variables; you'll still need to do `getvalues()`on the returned data container.

libpromises/evalfunction.c
```diff
@@ -1147,6 +1160,8 @@ static StringSet *VariablesMatching(const EvalContext *ctx, VariableTableIterato
 
 static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const Policy *policy, const FnCall *fp, const Rlist *finalargs)
 {
+    bool fulldata = (0 == strcmp(fp->name, "data_matching"));
```
jimis: reverse operands order

libpromises/evalfunction.c
```diff
@@ -1124,7 +1124,20 @@ static StringSet *VariablesMatching(const EvalContext *ctx, VariableTableIterato
...
             if (pass)
             {
-                StringSetAdd(matching, expr);
+                JsonElement *data = NULL;
+                if (collect_full_data)
+                {
+                    data = VarRefValueToJsonAllowingScalars(ctx, fp, v->ref, NULL, 0, true);
+                }
+
+                if (NULL == data)
```
jimis: reverse operands order

nickanderson:
@jimis

> I find this confusing. What I was expecting from a function named data_matching() was matching the argument with the variables' contents. Instead what I see is that it matches exactly what variablesmatching() matches, but returns a data container instead of an slist.

I can see how you might think that by the name alone, but I have no suggestions. The name does not bother me.

> Why is this useful? This does not address the main concern of the ticket it mentions (CFE-2265, old Redmine#7885) which is to avoid double-dereferencing the variables; you'll still need to do getvalues()on the returned data container.

variablesmatching() provides a list of variable names. In order to do anything with the data I must iterate over the variables and extract the data into new variables. While this does not directly address the desire to avoid double dereferencing it does provide the data in a way is easy to extract without double dereferencing.

tests/acceptance/01_vars/02_functions/variablesmatching.cf
tzz: (1) It matches the argument with the variables' content, but because more than one variable can be found, the result is a key-value map with the variable names as keys.

(2) This is useful because everything is converted to data containers. The other way you are thinking of doesn't just double-dereference variable names, it also needs know if they are strings or lists or maps, so it's very hard to extract the data.

libpromises/evalfunction.c

libpromises/evalfunction.c

tzz:
Rebased against master with the reversed operands fixed, thanks @jimis

tests/acceptance/01_vars/02_functions/variablesmatching.cf
jimis: > It matches the argument with the variables' content

I think it matches with the variables' names. I don't see any matching against the contents.

tzz: Sorry, this was just badly phrased. The function finds variables whose name matches the argument, just like `variablesmatching()`. But instead of returning a list of names, it returns a key-value map with those names as keys, and the variable contents as values.

jimis:
>  It matches the argument with the variables' content, but because more than one variable can be found, the result is a key-value map with the variable names as keys.

That's exactly my point, the name is deceptive, it doesn't match with the data, it matches with the variable names exactly like `variablesmatching()`. I can merge it as is, if both @nickanderson and @tzz like it. Or we can brainstorm on a better name, my suggestion would be `variablesmatching_as_data()` or `variablesmatching_data()`, which shows the similarity to variablesmatching, and also turns up if someone searches in the docs for variablesmatching.

P.S. Travis tests are red BTW.

tzz:
@jimis it was using an older API for converting a VarRef to JsonElement. Fixed.

For the name, how about `variablesmatching_full()` because it's not just **presented** as data, it digs deeper into the variable contents?

Other suggestions:
* `variablecontents()`
* `variablesmatching+()` (yeah it might require parser tweaks)
* `variablesmatching:full()`
* `data_from_variablesmatching()`

jimis:
Thanks @tzz, I'd say definitely not the names with colon or plus signs, and lets see what the rest have to say.

nickanderson:
Well, to me `data_from_variablesmatching()` reads the most clearly. `variablecontents` doesnt hint in the name that its the content from multiple variables.

I also have a tendency to prefer long descriptive names, and yes, I know they get annoying to type.

jimis:
On Fri, Sep 30, 2016 at 6:54 PM, Nick Anderson <notifications@github.com>
wrote:

> Well, to me data_from_variablesmatching() reads the most clearly.
> variablecontents doesnt hint in the name that its the content from
> multiple variables.
>


what about `variablesmatching_as_data()` ?

tzz:
Renamed to `variablesmatching_as_data()` and rebased. Thanks.

tzz:
stupid tests... fixed again

tests/acceptance/01_vars/02_functions/variablesmatching.cf
```diff
@@ -21,25 +25,15 @@ bundle agent test
...
+      "x_fullvars" data => variablesmatching_as_data("default:init.test_b37ce1b03955522848f0a82d0b2b2e21.*", "x");
+      "z_fullvars" data => variablesmatching_as_data("default:init.test_b37ce1b03955522848f0a82d0b2b2e21.*", "z");
 }
 
 bundle agent check
 {
-  classes:
-      "ok" and => { strcmp("$(test.count)", "3"),
-                    strcmp("$(test.x_count)", "2"),
-                    strcmp("$(test.z_count)", "0") };
```
jimis: Why remove this? Keep both this and the new JSON comparison.

libpromises/evalfunction.c
```diff
@@ -1124,7 +1124,25 @@ static StringSet *VariablesMatching(const EvalContext *ctx, VariableTableIterato
...
+                JsonElement *data = NULL;
+                bool allocated = false;
+                if (collect_full_data)
+                {
+                    data = VarRefValueToJson(ctx, fp, v->ref, NULL, 0, true, &allocated);
+                }
+
+                if (data == NULL)
+                {
+                    JsonObjectAppendString(matching, expr, "");
```
jimis: Can you explain this one? When can it happen? And why do we need to append the empty string?

libpromises/evalfunction.c
```diff
@@ -1165,17 +1185,24 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
...
         {
-            RlistPrepend(&matches, element, RVAL_TYPE_SCALAR);
+            return (FnCallResult) { FNCALL_SUCCESS, (Rval) { global_matches, RVAL_TYPE_CONTAINER } };
         }
 
-        StringSetDestroy(global_matches);
-        VariableTableIteratorDestroy(iter);
+        JsonIterator jiter = JsonIteratorInit(global_matches);
+        const char *key;
+        while ((key = JsonIteratorNextKey(&jiter)))
```
jimis: Explicit comparison  `        while ((key = JsonIteratorNextKey(&jiter)) != NULL)`

libpromises/evalfunction.c
```diff
@@ -1165,17 +1185,24 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
...
+        {
+            RlistPrepend(&matches, key, RVAL_TYPE_SCALAR);
```
jimis: What are the odds of `key` not being a string? If zero, then `assert()`. If it's possible, then print a message and skip.

tests/acceptance/01_vars/02_functions/variablesmatching.cf
tzz: The JSON comparison has this built in: if the count is off, the JSON data will be different. The expected JSON test is much better in that you see variable count, content, and name differences.

libpromises/evalfunction.c
tzz: When we don't collect the full variable data, we create a key-value map JsonObject with empty strings as the values. It will be destroyed soon afterwards, but the code is cleaner IMHO if we do it this way than if we make a JsonArray in one branch and a JsonObject in the other branch. The empty strings provide assurance that serializing this JsonObject (e.g. for logging) will not create problems. The extra memory usage from the empty strings is negligible.

libpromises/evalfunction.c
tzz: OK; fixed

libpromises/evalfunction.c
tzz: We create the key earlier, so I think the odds are 0 that it's NULL. So I did `assert (key != NULL)` which I hope is what you wanted.

tzz:
@jimis fixed and rebased; I answered what I could.

libpromises/evalfunction.c
jimis: > When we don't collect the full variable data, we create a key-value map JsonObject with empty strings as the values. It will be destroyed soon afterwards.

This is very useful, I was hoping for a comment. :-) No big deal though, thanks for explaining.

tzz: OK, comment added!

jimis:
Thanks @tzz; looking forward for the docs! :-)

PR #2356, opened 2015-10-23
Based on Pasinskim's pull2346 "dirs changed to files".

jimis:

PR #2472, opened 2016-01-14
Improve networking errors handling

amousset:
Follows #2356.

* Fail on networking errors during file copy (this avoids continuing the copy when the
  file in error was the last evaluated in a directory)
* Free dirh in case of networking error
* Mark connection broken in SendTransaction
* ReceiveTransaction should always returns -1 in case of error (some calls to it test this value for errors)

### DIFF
--- a/cf-runagent/cf-runagent.c
+++ b/cf-runagent/cf-runagent.c
@@ -726,13 +726,7 @@ static void HailExec(AgentConnection *conn, char *peer)
     {
         memset(recvbuffer, 0, sizeof(recvbuffer));
 
-        int n_read = ReceiveTransaction(conn->conn_info, recvbuffer, NULL);
-
-        if (n_read == -1)
-        {
-            break;
-        }
-        if (n_read == 0)                               /* connection closed */
+        if (ReceiveTransaction(conn->conn_info, recvbuffer, NULL) == -1)
         {
             break;
         }
--- a/cf-serverd/server_classic.c
+++ b/cf-serverd/server_classic.c
@@ -783,14 +783,6 @@ RSA *newkey = RSA_new();
         return false;
     }
 
-    if (len_n == 0)
-    {
-        Log(LOG_LEVEL_ERR, "Authentication failure: "
-            "connection was closed while receiving public key modulus");
-        RSA_free(newkey);
-        return false;
-    }
-
     if ((newkey->n = BN_mpi2bn(recvbuffer, len_n, NULL)) == NULL)
     {
         Log(LOG_LEVEL_ERR, "Authentication failure: "
@@ -808,13 +800,6 @@ RSA *newkey = RSA_new();
     {
         Log(LOG_LEVEL_ERR, "Authentication failure: "
             "error while receiving public key exponent");
-        RSA_free(newkey);
-        return false;
-    }
-    if (len_e == 0)
-    {
-        Log(LOG_LEVEL_ERR, "Authentication failure: "
-            "connection was closed while receiving public key exponent");
         RSA_free(newkey);
         return false;
     }
@@ -929,15 +914,10 @@ RSA *newkey = RSA_new();
     int recv_len = ReceiveTransaction(conn->conn_info, recv_buf, NULL);
     if (recv_len < digestLen)
     {
-        if (recv_len < 0)
+        if (recv_len == -1)
         {
             Log(LOG_LEVEL_ERR, "Authentication failure: "
                 "error receiving counter-challenge response");
-        }
-        else if (recv_len == 0)
-        {
-            Log(LOG_LEVEL_ERR, "Authentication failure: "
-                "connection was closed while receiving counter-challenge response");
         }
         else                                      /* 0 < recv_len < expected_len */
         {
@@ -976,25 +956,18 @@ RSA *newkey = RSA_new();
         "should decrypt to %d bytes",
         keylen, session_key_size);
 
-    if ((keylen <= 0) || (keylen > CF_BUFSIZE / 2))
+    if (keylen == -1)
     {
-        if (keylen == 0)
-        {
-            Log(LOG_LEVEL_ERR, "Authentication failure: "
-                "connection was closed while receiving session key");
-        }
-        else if (keylen > CF_BUFSIZE / 2)
-        {
-            Log(LOG_LEVEL_ERR, "Authentication failure: "
-                "session key received is too long (%d bytes)",
-                keylen);
-        }
-        else                                                  /* keylen < 0 */
-        {
-            Log(LOG_LEVEL_ERR, "Authentication failure: "
-                "error receiving session key");
-        }
+        Log(LOG_LEVEL_ERR, "Authentication failure: "
+            "error receiving session key");
+        return false;
+    }
 
+    if (keylen > CF_BUFSIZE / 2)
+    {
+        Log(LOG_LEVEL_ERR, "Authentication failure: "
+            "session key received is too long (%d bytes)",
+            keylen);
         return false;
     }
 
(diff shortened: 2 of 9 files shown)


cf-agent/verify_files_utils.c
```diff
@@ -836,7 +837,7 @@ static PromiseResult SourceSearchAndCopy(EvalContext *ctx, const char *from, cha
             }
         }
 
-        if (attr.copy.purge)    /* Purge this file */
+        if (attr.copy.purge)    /* Do not purge this file */
```
kacf: Really? Later on it tests for the same condition and calls `PurgeLocalFiles()`. Seems to me like `attr.copy.purge` *does* indicate that the file should be purged. It is true that it doesn't happen here though, so perhaps a more fitting comment would be `/* Indicates that file should be purged */`.

kacf:
A note about changelog entries: It is not necessary to have changelog entries for everything, and in fact it's unneeded for some things. Sorry if this has not been clear, I have tried to make this more clear [here](https://github.com/cfengine/core/commit/c0f4276b50c698fbfb81d102534595cb669d2516). So for these commits, I think it's enough to have perhaps two changelog entries like this:

* `Changelog: Fix file descriptor leak when there are network errors.` (for the first commit)
* `Changelog: Fix bugs which handle network errors incorrectly.` (can be put in one of the others)

Make sure to include the appropriate Redmine ticket in the commit with the changelog entry as well.

kacf:
@jimis: Can you check specifically https://github.com/amousset/core/commit/f08d07ba12a7c4c503e712f0c87beb33eb6a29e3 and https://github.com/amousset/core/commit/5bec6e1868f7ab0257ec96382ef9bce47ff4bf2d? It looks good to me, but they are quite small so I hope you have the time to review those two.

cf-agent/verify_files_utils.c
amousset: The idea is that `namecache` contains the list of files not to purge, so the comment was quite misleading, but I can put it on the next line to make clearer.

kacf: Hmm, ok. So this means it will not be purged later on? Then it's even better if you can explain in the comment that the namecache is the reason for this.

jimis:
Some changes look simple and easy, but some other are tricky. As I've written in [redmine#7926](https://dev.cfengine.com/issues/7926) I would like to hear how this patch fares in real networks for a while, before merging.

cf-agent/verify_files_utils.c
amousset: I changed to comment to explain what  this line does.

kacf:
Just looking at this I cannot see any problems with it. Let's consider this a risky and non-critical bug fix, and aim it for 3.9.0.

kacf:
trigger build

jimis:
@kacf @amousset this patch changes `ReceiveTransaction()` to never return 0 on socket closed. This might or might not be a good change, but I would like to hear the rationale behind it, and the tests in real network conditions that it went through.

If there is good reason for this change it can go in, but needs to be accompanied by changes in many of `ReceiveTransaction()` callers, that check for zero return value.

Hmmm, having a quick look it seems that `ReceiveTransaction()` callers should consider an error occured if the function returns anything less than the desired amount of bytes (including 0 or -1). Still that's only after having a quick look, the matter is very subtle.

@kacf some commits look non-controversial though, e.g. 18b58b37c43cf9954e6c1e46c728e5ac0657473b and ff840df5e2bc9eda8a2b3483751c032c4e78cc01, they shouldn't be held back IMHO. @amousset can you submit them on a separate pull request?

amousset:
@jimis @kacf Opened #2493 for the simple fixes, and updated this PR.

The motivation for the `ReceiveTransaction` change is the issues that can occur because some calls to it do not consider 0 as an error, and continue execution. It leads to misleading logs, not mentioning a networking issue, like in:
* The call at https://github.com/cfengine/core/blob/master/libcfnet/client_code.c#L591, where the unitialized `buf` will likely fail to be decrypted.

Or to actual issues like:
* The call at https://github.com/cfengine/core/blob/master/libcfnet/stat_cache.c#L237. If it returns 0 and the file is a symlink, the path of the link will be empty, and the destination link will be replaced by an empty link until next execution.

Currently the possible cases are:
* `recv` returns -1. `ReceiveTransaction` returns -1
* `recv` returns 0. This is not an error at the networking level, but it is at the CFEngine protocol level. `ReceiveTransaction` returns 0.
* `recv` returns >0, but with an unexpected (at the protocol level) length. `ReceiveTransaction` returns -1.

The rationale for the `ReceiveTransaction` change is to return -1 in all error cases, and to abstract the networking or protocol states, which is currently only partially done. The networking and protocol errors are all handled and logged in `ReceiveTransaction` (currently, the networking error logging relies on the caller). Do you see a case where the caller would need to know the type of error ? (I checked the calls that currently exist).

Another option, with less impact, would be to fix the calls that do not consider 0 as an error, and maybe at least to move networking errors logging to `ReceiveTransaction`.

jimis:
@amousset Thanks for the detailed explanation! I think this change makes sense and we should try it. It certainly makes the programming interface much nicer. In addition it will give us the chance to support zero-sized transaction packets in the future (it always bothered me that return value of zero meant closed connection).

In order to merge it, we need to adjust all the callers of `ReceiveTransaction()` that check for return value of 0: first that nothing breaks, and second to remove that code. Finally change the doc string of `ReceiveTransaction()`.

Also please include your above message in the commit. It's very nice and thorough and good to have in the commit history.

cf-serverd/server_common.c
```diff
@@ -653,7 +649,7 @@ void CfEncryptGetFile(ServerFileGetState *args)
             {
                 if (SendTransaction(conn_info, out, cipherlen + finlen, CF_DONE) == -1)
                 {
-                    Log(LOG_LEVEL_VERBOSE, "Send failed in GetFile. (send: %s)", GetErrorStr());
+                    Log(LOG_LEVEL_ERR, "Protocol transaction broken off in CfEncryptGetFile. Could not send.");
```
jimis: Please try to wrap at 78 characters where possible (but avoid breaking strings at non-punctuation points, see HACKING.md). 

Also Log messages higher than VERBOSE should not be cryptic, e.g. "CFEncryptGetFile" or "protocol transaction" doesn't mean anything to end users, and this applies to most of the messages you changed.

Finally be careful when changing the level of the message: spamming syslog with errors have been many times an issue. And most times an error must have been printed down the callpath (e.g. in `TLSSend()`) before reaching this point, so no need to print ERR every time we return from a function.

I would rather if you left message changes out of this pull request, in order to be able to test/revert easily.

amousset:
@jimis Updated, removed log changes.

jimis:
Alright, this looks good, lets see how it goes. I'm still worried about the escalation of some messages from VERBOSE to ERR, that they might spam syslog, but without running this in a busy environment we won't know.

jimis:
trigger build

kacf:
Appears to fail with this message:

```
12:35:30 Making all in libcfnet
12:35:30 net.c: In function `SendTransaction':
12:35:30 net.c:114: error: assignment of read-only member `status'
12:35:30 make[2]: *** [net.lo] Error 1
```

amousset:
Sorry, I forgot to add some changes to the commit, it's now fixed.

jimis:
trigger build

jimis:
All green, lets merge and see how it goes... Thanks!

jimis:
@amousset see https://dev.cfengine.com/issues/7926 for details, but this caused a bunch of bogus log messages. I'm planning to revert the pull request soon, unless you have a suggestion for fixing the issues.

jimis:
I've tried to fix the issue without reverting, maybe it's fixed with this pull request https://github.com/cfengine/core/pull/2533

PR #2473, opened 2016-01-14
Add: Test syntax error on namespace

nickanderson:
Namespaces should not be able to contain characters that are invalid in
bundle names.


kacf:
trigger build

PR #2474, opened 2016-01-14
Namespaces allow some special characters

eysteins:
This allows creation of namespaces that cannot be reached. Redmine #7903.

Test at #2473

### DIFF
--- a/libpromises/mod_common.c
+++ b/libpromises/mod_common.c
@@ -398,7 +398,7 @@ const ConstraintSyntax CFH_CONTROLBODY[] =  /* enum cfh_control */
 
 const ConstraintSyntax file_control_constraints[] =  /* enum cfh_control */
 {
-    ConstraintSyntaxNewString("namespace", CF_IDRANGE, "Switch to a private namespace to protect current file from duplicate definitions", SYNTAX_STATUS_NORMAL),
+    ConstraintSyntaxNewString("namespace", "[a-zA-Z0-9_]+", "Switch to a private namespace to protect current file from duplicate definitions", SYNTAX_STATUS_NORMAL),
     ConstraintSyntaxNewStringList("inputs", ".*", "List of additional filenames to parse for promises", SYNTAX_STATUS_NORMAL),
     ConstraintSyntaxNewNull()
 };


nickanderson:
trigger build

nickanderson:
Lets let the PR analyser build a package to test with.

nickanderson:
trigger build

kacf:
Should probably combine this with #2473 and run them together.

kacf:
Otherwise +1 from me.

libpromises/mod_common.c
```diff
@@ -398,7 +398,7 @@ const ConstraintSyntax CFH_CONTROLBODY[] =  /* enum cfh_control */
 
 const ConstraintSyntax file_control_constraints[] =  /* enum cfh_control */
 {
-    ConstraintSyntaxNewString("namespace", CF_IDRANGE, "Switch to a private namespace to protect current file from duplicate definitions", SYNTAX_STATUS_NORMAL),
+    ConstraintSyntaxNewString("namespace", "[a-zA-Z0-9_]+", "Switch to a private namespace to protect current file from duplicate definitions", SYNTAX_STATUS_NORMAL),
```
tzz: I'd even make it stricter, requiring that it should start with a letter: `[a-zA-Z][a-zA-Z0-9_]+`

kacf: Yes, good point.

nickanderson:
Why does it have to start with a letter?
Bundles are allowed to start with a digit.

```
bundle agent main
{
methods:
  "000main";
}

bundle agent 000main
{
  reports:
    "$(this.bundle)";
}
```

kacf:
The docs make no special mention of it AFAICS. I guess we allow both variables and classes to start with numbers so it makes sense to allow namespaces as well. And since they will be quoted when used, I can't see any problem with that. Anyone object to allowing namespaces to start with numbers?

tzz:
I would prefer if we didn't allow numbers at the beginning of **namespaces**. While yes, it's possible, it creates some ambiguity if we someday support numbers in a different numeric base. See https://programmers.stackexchange.com/questions/133381/why-should-identifiers-not-begin-with-a-number

More importantly, namespaces are identifiers that separate policy at the top-most level. They should be legible and as clear as possible. I think starting them with numbers invites things like `001_foo` which make no sense in CFEngine (because there, parsing can occur in any order). It has no functional purpose otherwise.

If you still disagree, I can change the regexp or you do it, it's a one-line change.

kacf:
I honestly don't know about this one, most of the arguments made for not starting with numerals seem to go away when you have syntax that requires variables to be enclosed in prefix/suffix, like we do: `$(variable)`, since it can't be ambiguous in this case.

But it's easier to start out strict and relax the conditions later, than to try to take back a mistake we already published. So if no one else speaks up, then I agree, let's disallow numbers at the start of namespaces.

kacf:
Superseded by #2487.

PR #2475, opened 2016-01-18
Redmine #7908: Check for empty server response in RemoteDirList after decryption

amousset:
Avoid random errors when the encrypted string starts with a NULL character.


jimis:
Thanks, makes a lot of sense! I'm assuming "make check" shows no objections?

jimis:
trigger build

jimis:
trigger build

PR #2477, opened 2016-01-20
cf-promises/cf-promises.c: teach --show-vars to print JSON vars

tzz:
Simple change, picked from #2245 

@kacf should be an easy merge, docs and acceptance tests will come if the code is OK.

### DIFF
--- a/cf-promises/cf-promises.c
+++ b/cf-promises/cf-promises.c
@@ -481,15 +481,19 @@ static void ShowVariablesFormatted(EvalContext *ctx)
         char *varname = VarRefToString(v->ref, true);
 
         Writer *w = StringWriter();
-        RvalWrite(w, v->rval);
 
-        StringSet *tagset = EvalContextVariableTags(ctx, v->ref);
-        Buffer *tagbuf = StringSetToBuffer(tagset, ',');
+        switch (DataTypeToRvalType(v->type))
+        {
+        case RVAL_TYPE_CONTAINER:
+            JsonWriteCompact(w, RvalContainerValue(v->rval));
+            break;
 
-        char *line;
+        default:
+            RvalWrite(w, v->rval);
+        }
+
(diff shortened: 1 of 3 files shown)


kacf:
Regarding the changelog entry: We try to make them as user friendly as possible, including the mass of people that know nothing about the code. But it's still useful to include references to the source file like you do in your message, so in those cases it's better to make a separate entry, written from the user's perspective, like this:

```
cf-promises/cf-promises.c: teach --show-vars to print JSON vars

Changelog: cf-promises can now show JSON variables.
```

cf-promises/cf-promises.c
```diff
@@ -474,15 +474,21 @@ static void ShowVariablesFormatted(EvalContext *ctx)
         char *varname = VarRefToString(v->ref, true);
 
         Writer *w = StringWriter();
-        RvalWrite(w, v->rval);
+        bool json = false;
```
kacf: This is unused now, AFAICS.

kacf:
Looks good otherwise. I don't think we need explicit documentation for this, other than the changelog message, but an acceptance test would be nice! :-)

tzz:
Ugh, I can't get the acceptance test to work easily without https://github.com/cfengine/core/pull/2460 so I hope it's OK to delay the test until https://github.com/cfengine/core/pull/2460 is merged (I'll put it on my TODO list) and merge this now.

tzz:
Code adjusted as suggested.

kacf:
#2460 should be quite quick to get in.

kacf:
Ok, took a bit longer than I anticipated, but #2460 is in now.

tzz:
OK, this is rebased now and with a full acceptance test, ready for final review.

kacf:
Nice!

kacf:
trigger build

PR #2480, opened 2016-01-23
cf-promises: allows --show-vars and --show-classes to take an optional filter

tzz:
Effect:
- just `--show-vars` and `--show-classes` work like before
- `--show-vars=default:sys` shows only variable names that contain that filter
- `--show-classes=foo` shows only class names that contain that filter
- ~~with a filter, the header is not printed~~

### DIFF
--- a/cf-promises/cf-promises.c
+++ b/cf-promises/cf-promises.c
@@ -34,12 +34,14 @@
 #include <bootstrap.h>
 #include <string_lib.h>
 #include <loading.h>
+#include <regex.h>                                        /* CompileRegex */
+#include <match_scope.h>
 
 #include <time.h>
 
 static GenericAgentConfig *CheckOpts(int argc, char **argv);
-static void ShowContextsFormatted(EvalContext *ctx);
-static void ShowVariablesFormatted(EvalContext *ctx);
+static void ShowContextsFormatted(EvalContext *ctx, const char *regexp);
+static void ShowVariablesFormatted(EvalContext *ctx, const char *regexp);
 
 /*******************************************************************/
 /* Command line options                                            */
@@ -66,8 +68,8 @@ static const struct option OPTIONS[] =
 {
     {"workdir", required_argument, 0, 'w'},
     {"eval-functions", optional_argument, 0, OPT_EVAL_FUNCTIONS },
-    {"show-classes", no_argument, 0, OPT_SHOW_CLASSES },
-    {"show-vars", no_argument, 0, OPT_SHOW_VARS },
+    {"show-classes", optional_argument, 0, OPT_SHOW_CLASSES },
+    {"show-vars", optional_argument, 0, OPT_SHOW_VARS },
     {"help", no_argument, 0, 'h'},
     {"bundlesequence", required_argument, 0, 'b'},
     {"debug", no_argument, 0, 'd'},
@@ -95,8 +97,8 @@ static const char *const HINTS[] =
 {
     "Override the work directory for testing (same as setting CFENGINE_TEST_OVERRIDE_WORKDIR)",
     "Evaluate functions during syntax checking (may catch more run-time errors). Possible values: 'yes', 'no'. Default is 'yes'",
-    "Show discovered classes, including those defined in common bundles in policy",
-    "Show discovered variables, including those defined without dependency to user-defined classes in policy",
+    "Show discovered classes, including those defined in common bundles in policy. Optionally can take a regular expression.",
+    "Show discovered variables, including those defined without dependency to user-defined classes in policy. Optionally can take a regular expression.",
     "Print the help message",
     "Use the specified bundlesequence for verification",
     "Enable debugging output",
@@ -213,12 +215,12 @@ int main(int argc, char *argv[])
 
     if(config->agent_specific.common.show_classes)
     {
-        ShowContextsFormatted(ctx);
+        ShowContextsFormatted(ctx, config->agent_specific.common.show_classes);
     }
 
     if(config->agent_specific.common.show_variables)
     {
-        ShowVariablesFormatted(ctx);
+        ShowVariablesFormatted(ctx, config->agent_specific.common.show_variables);
     }
 
     PolicyDestroy(policy);
@@ -244,7 +246,7 @@ GenericAgentConfig *CheckOpts(int argc, char **argv)
         switch (c)
         {
         case OPT_EVAL_FUNCTIONS:
-            if (!optarg)
+            if (optarg == NULL)
             {
                 optarg = "yes";
             }
@@ -252,11 +254,19 @@ GenericAgentConfig *CheckOpts(int argc, char **argv)
             break;
 
         case OPT_SHOW_CLASSES:
-            config->agent_specific.common.show_classes = true;
+            if (optarg == NULL)
+            {
+                optarg = ".*";
+            }
+            config->agent_specific.common.show_classes = optarg;
             break;
 
         case OPT_SHOW_VARS:
-            config->agent_specific.common.show_variables = true;
+            if (optarg == NULL)
+            {
+                optarg = ".*";
+            }
+            config->agent_specific.common.show_variables = optarg;
             break;
 
         case 'w':
@@ -477,16 +487,33 @@ GenericAgentConfig *CheckOpts(int argc, char **argv)
     return config;
 }
 
-static void ShowContextsFormatted(EvalContext *ctx)
+static void ShowContextsFormatted(EvalContext *ctx, const char *regexp)
 {
+    assert(regexp != NULL);
+
     ClassTableIterator *iter = EvalContextClassTableIteratorNewGlobal(ctx, NULL, true, true);
     Class *cls = NULL;
 
     Seq *seq = SeqNew(1000, free);
 
+    pcre *rx = CompileRegex(regexp);
+
+    if (rx == NULL)
+    {
+        Log(LOG_LEVEL_ERR, "Sorry, we could not compile regular expression %s", regexp);
+        return;
+    }
+
     while ((cls = ClassTableIteratorNext(iter)))
     {
         char *class_name = ClassRefToString(cls->ns, cls->name);
+
+        if (!PartialMatch(rx, class_name))
+        {
+            free(class_name);
+            continue;
+        }
+
         StringSet *tagset = EvalContextClassTags(ctx, cls->ns, cls->name);
         Buffer *tagbuf = StringSetToBuffer(tagset, ',');
 
@@ -497,6 +524,8 @@ static void ShowContextsFormatted(EvalContext *ctx)
         BufferDestroy(tagbuf);
         free(class_name);
     }
+
+    pcre_free(rx);
 
     SeqSort(seq, (SeqItemComparator)strcmp, NULL);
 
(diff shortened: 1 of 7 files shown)


nickanderson:
Oh thats kind of handy. I am constantly piping through grep.

tzz:
@jimis small PR with no compatibility issues, when you omit the optional parameters you get the old behavior.

tzz:
@jimis no need to rebase after eval changes

cf-promises/cf-promises.c
```diff
@@ -462,17 +491,30 @@ static void ShowContextsFormatted(EvalContext *ctx)
     ClassTableIteratorDestroy(iter);
 }
 
-static void ShowVariablesFormatted(EvalContext *ctx)
+static void ShowVariablesFormatted(EvalContext *ctx, const char* regexp)
```
jimis: style and add an assert

cf-promises/cf-promises.c
```diff
@@ -447,9 +471,14 @@ static void ShowContextsFormatted(EvalContext *ctx)
         free(class_name);
     }
 
+    pcre_free(rx);
+
     SeqSort(seq, (SeqItemComparator)strcmp, NULL);
 
-    printf("%-60s %-40s\n", "Class name", "Meta tags");
+    if (0 == strcmp(regexp, ".*")) // only if we're showing everything
```
jimis: style

cf-promises/cf-promises.c
```diff
@@ -447,9 +471,14 @@ static void ShowContextsFormatted(EvalContext *ctx)
...
+    {
+        printf("%-60s %-40s\n", "Class name", "Meta tags");
```
jimis: I believe you should print the header regardless of the filter

cf-promises/cf-promises.c
```diff
@@ -426,16 +436,30 @@ GenericAgentConfig *CheckOpts(int argc, char **argv)
     return config;
 }
 
-static void ShowContextsFormatted(EvalContext *ctx)
+static void ShowContextsFormatted(EvalContext *ctx, const char* regexp)
```
jimis: Style.

Also add `assert(regexp != NULL)`

cf-promises/cf-promises.c
```diff
@@ -228,11 +230,19 @@ GenericAgentConfig *CheckOpts(int argc, char **argv)
             break;
 
         case OPT_SHOW_CLASSES:
-            config->agent_specific.common.show_classes = true;
+            if (!optarg)
```
jimis: style

cf-promises/cf-promises.c
```diff
@@ -500,9 +542,14 @@ static void ShowVariablesFormatted(EvalContext *ctx)
...
     }
 
+    pcre_free(rx);
+
     SeqSort(seq, (SeqItemComparator)strcmp, NULL);
 
-    printf("%-40s %-60s %-40s\n", "Variable name", "Variable value", "Meta tags");
+    if (0 == strcmp(regexp, ".*")) // only if we're showing everything
+    {
+        printf("%-40s %-60s %-40s\n", "Variable name", "Variable value", "Meta tags");
```
jimis: print in all cases

libpromises/match_scope.c
```diff
@@ -84,6 +84,15 @@ static int RegExMatchFullString(EvalContext *ctx, pcre *rx, const char *teststri
     }
 }
 
+/* Does not free rx! */
+bool PartialMatch(const pcre *rx, const char *teststring)
```
jimis: What's this function for? Partial means not anchored? Add a small comment please.

cf-promises/cf-promises.c
```diff
@@ -426,16 +436,30 @@ GenericAgentConfig *CheckOpts(int argc, char **argv)
...
     ClassTableIterator *iter = EvalContextClassTableIteratorNewGlobal(ctx, NULL, true, true);
     Class *cls = NULL;
 
     Seq *seq = SeqNew(1000, free);
 
+    pcre *rx = CompileRegex(regexp);
+
+    if (rx == NULL)
+    {
+        return;
```
jimis: `Log(ERR)`?

cf-promises/cf-promises.c
```diff
@@ -500,9 +542,14 @@ static void ShowVariablesFormatted(EvalContext *ctx)
         free(varname);
     }
 
+    pcre_free(rx);
+
     SeqSort(seq, (SeqItemComparator)strcmp, NULL);
```
jimis: Wow this cast is really dangerous, I believe the behaviour here is undefined, `strcmp()` is not compatible with `SeqItemComparator()`! Anyway It's old code so I ignore, and I grep'ed and it's all over the place. :-(

nickanderson:
ooooo for 3.11 :)

tzz:
Rebased and all the code fixes made. Header is always printed now.

Does this need tests?

cf-promises/cf-promises.c
```diff
@@ -244,19 +246,27 @@ GenericAgentConfig *CheckOpts(int argc, char **argv)
         switch (c)
         {
         case OPT_EVAL_FUNCTIONS:
-            if (!optarg)
+            if (optarg = NULL)
```
olehermanse: `if (optarg == NULL)`

cf-promises/cf-promises.c
```diff
@@ -244,19 +246,27 @@ GenericAgentConfig *CheckOpts(int argc, char **argv)
...
             {
                 optarg = "yes";
             }
             config->agent_specific.common.eval_functions = strcmp("yes", optarg) == 0;
             break;
 
         case OPT_SHOW_CLASSES:
-            config->agent_specific.common.show_classes = true;
+            if (optarg = NULL)
```
olehermanse: `if (optarg == NULL)`

cf-promises/cf-promises.c
```diff
@@ -244,19 +246,27 @@ GenericAgentConfig *CheckOpts(int argc, char **argv)
...
+            {
+                optarg = ".*";
+            }
+            config->agent_specific.common.show_classes = optarg;
             break;
 
         case OPT_SHOW_VARS:
-            config->agent_specific.common.show_variables = true;
+            if (!optarg)
```
olehermanse: `if (optarg == NULL)`

libpromises/match_scope.h
```diff
@@ -26,6 +26,10 @@
 #define CFENGINE_MATCH_SCOPE_H
 
 #include <cf3.defs.h>
+#include <regex.h>
```
olehermanse: Moving match_scope to libpromises is not complete. Go to `tests/unit/Makefile.am` and update the `cf-agent/match_scope` paths to `libpromises`.

@jimis Do you agree with moving `match_scope` from `cf-agent` to `libpromises` ?

cf-promises/cf-promises.c
tzz: fixed

cf-promises/cf-promises.c
tzz: fixed

cf-promises/cf-promises.c
tzz: fixed

tzz:
I believe this is ready for another review. I hope I got the `Makefile.am` changes right.

tzz:
BTW the `foo = NULL` when I mean to write `foo == NULL` is the main reason I like to write `NULL == foo`. Thanks for catching it, Ole.

nickanderson:
@kacf @jimis please review

cf-promises/cf-promises.c
```diff
@@ -34,12 +34,14 @@
...
+#include <regex.h>                                        /* CompileRegex */
+#include <match_scope.h>
 
 #include <time.h>
 
 static GenericAgentConfig *CheckOpts(int argc, char **argv);
-static void ShowContextsFormatted(EvalContext *ctx);
-static void ShowVariablesFormatted(EvalContext *ctx);
+static void ShowContextsFormatted(EvalContext *ctx, const char* regexp);
+static void ShowVariablesFormatted(EvalContext *ctx, const char* regexp);
```
olehermanse: Style. Should be `char *regexp`.

cf-promises/cf-promises.c
olehermanse: Style, still. `char *regexp`.

libpromises/generic_agent.h
```diff
@@ -77,8 +77,8 @@ typedef struct
             unsigned int parser_warnings;
             unsigned int parser_warnings_error;
             bool eval_functions;
-            bool show_classes;
-            bool show_variables;
+            char* show_classes;
+            char* show_variables;
```
olehermanse: Style:
```
char *show_classes;
char *show_variables;
```

libpromises/match_scope.h
jimis: Yeah, this file looks like it provides generic regex facilities, so go ahead and move it. Even to libutils if possible, but even libpromises sounds better than cf-agent.

olehermanse:
Apart from the small style changes, this looks good to merge. 

@jimis @kacf I started some builds to test this on all platforms in jenkins:
https://ci.cfengine.com/job/bootstrap-enterprise-pr/283/
https://ci.cfengine.com/job/testing-enterprise-pr/261/

libpromises/generic_agent.h
tzz: A style linter would be really nice to save me time for this kind of small fix :)

tzz:
Rebased with the style fixes requested by @olehermanse

tzz:
@olehermanse @jimis @kacf please don't drop this

cf-promises/cf-promises.c
```diff
@@ -213,12 +215,12 @@ int main(int argc, char *argv[])
 
     if(config->agent_specific.common.show_classes)
```
jimis: This is no longer a bool, so an explicit comparison to `NULL` would be preferred.

cf-promises/cf-promises.c
```diff
@@ -213,12 +215,12 @@ int main(int argc, char *argv[])
...
     {
-        ShowContextsFormatted(ctx);
+        ShowContextsFormatted(ctx, config->agent_specific.common.show_classes);
     }
 
     if(config->agent_specific.common.show_variables)
```
jimis: same here

libpromises/match_scope.h
```diff
@@ -26,6 +26,10 @@
...
+
+/* Does not free rx! */
+bool PartialMatch(const pcre *rx, const char *teststring);
```
jimis: I would prefer a function name that indicates this has to do with regexes, like `RegexPartialMatch()`.

jimis:
> BTW the foo = NULL when I mean to write foo == NULL is the main reason I like to write NULL == foo. Thanks for catching it, Ole.

Didn't the compiler warn? I'm quite surprised.

jimis:
Changelog tag is missing, merging manually.

jimis:
Merging manually.

jimis:
> > BTW the foo = NULL when I mean to write foo == NULL is the main reason I like to write NULL == foo. Thanks for catching it, Ole.
> 
> Didn't the compiler warn? I'm quite surprised.

I experimented a bit. The compiler warns fine when it's in an `if` statement, but *not* when it's in an `assert`, for example:

```c
assert(param = NULL);
```

This makes it *very* dangerous, creating side-effects only in DEBUG builds...

tzz:
> I experimented a bit. The compiler warns fine when it's in an if statement, but not when it's in an assert, for example:
...
> This makes it very dangerous, creating side-effects only in DEBUG builds...

Nice, thanks for checking that.

jimis:
Even further on the same subject, the assert-not-warning-for-assignments problem [is fixed in glibc-2.25 and newer](https://sourceware.org/bugzilla/show_bug.cgi?id=20386)!

PR #2481, opened 2016-01-25
In case of networking error, assume checksum is wrong

amousset:
When a checksum is done to check is it is necessary to update the file,
the copy will stop anyway because of the networking error.
When it is used to check a file after copy (before moving .cfnew)
the agent should not consider the content it has received as correct
(espacially in a networking/server error context).

### DIFF
--- a/libcfnet/client_code.c
+++ b/libcfnet/client_code.c
@@ -463,15 +463,16 @@ int CompareHashNet(const char *file1, const char *file2, bool encrypt, AgentConn
     if (SendTransaction(conn->conn_info, sendbuffer, tosend, CF_DONE) == -1)
     {
         Log(LOG_LEVEL_ERR, "Failed send. (SendTransaction: %s)", GetErrorStr());
-        return false;
+        Log(LOG_LEVEL_VERBOSE, "Networking error, assuming different checksum");
+        return true;
     }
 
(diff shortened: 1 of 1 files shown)


jimis:
Have you seen it in action? How does this bug manifest, does it cause corruption?

jimis:
trigger build

kacf:
Some pros and cons I can see here. Pros:

* With this fix, the agent will no longer report a successful file promise if the connection is broken (because the agent erroneously would think that the checksum was correct; now it won't).

Cons:

* In the worst case this might double the load on an already loaded server, since a broken connection will trigger another attempt at downloading the file. Especially for big files this could be significant.

IMHO, the best fix here is to do neither, but to have a way to report connection failure instead. Then the agent would know the promise failed but without causing a potentially expensive redownload.

kacf:
Oh, and another pro of course, which @amousset mentioned initially: When the checksum is checked *after* file copy, it will no longer treat the file as successfully downloaded, which can lead to a truncated file being accepted as valid.

jimis:
Makes sense, unfortunately I missed the results of the triggered build, so I'll trigger a new one.

jimis:
trigger build

jimis:
Merging, thanks!

jimis:
I'm not sure if it should be cherry-picked, is there a redmine ticket with reported breakage?

PR #2482, opened 2016-01-26
Redmine #7808: Do not drop privileges when changes permissions on non fifo

kacf:
The reason is that dropping privileges has some unexpected
consequences: CFEngine operates recursively on a file tree in a way
that may leave the directory being operated on last (e.g. files inside
the directory are treated before the directory itself). This means
that if we drop privileges, we may not always have the permissions to
enter the directory. This is impossible to fix for FIFOs (other than
changing the order that CFEngine does things, but that is on a
different code level and has other implications), because FIFOs will
block if we try to open a file descriptor to them, but we can fix it
for regular files by switching back to the old method of opening the
file descriptor and doing fchmod on that.

An additional level of complexity in this case is that the path may
end in a symlink *pointing* to a FIFO, hence the code that opens the
true parent directory of the leaf.

Also rename the CHMOD_SETEUID_WORKS flag to CHMOD_SETEUID_HIGHBITS_WORK
to make it clearer that it is not seteuid support which is missing,
but support for seteuid together with high bit flags in permission
modes.

Changelog: Fix a regression which would sometimes cause "Permission
denied" errors on files inside directories with very restricted
permissions.

Based on a patch by Alexis Mousset <alexis.mousset@normation.com>.

Also include acceptance test from #2428.

### DIFF
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -657,10 +657,13 @@ int safe_chdir(const char *path)
 #ifndef __MINGW32__
 
 /**
- * Implementation of safe_chown.
- * @param path Path to chown.
- * @param owner          Owner to set on path.
- * @param group          Group to set on path.
+ * Opens the true parent dir of the file in the path given. The notable
+ * difference from doing it the naive way (open(dirname(path))) is that it
+ * can follow the symlinks of the path, ending up in the true parent dir of the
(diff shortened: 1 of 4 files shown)


kacf:
trigger build

kacf:
trigger build

kacf:
trigger build

kacf:
trigger build

PR #2486, opened 2016-01-30
Fix ps options for FreeBSD to check processes only in current host an…

ddurieux:
…d not in jails

When do ps on FreeBSD, we get all process on host and of all jails. This fix will get only processes of the host, not processes of jails


kacf:
Thanks, @ddurieux! Merged manually.

kacf:

PR #2491, opened 2016-02-10
Change: Refactor failsafe for better readability

nickanderson:
Changelog: Title

ping @jimis

### DIFF
--- a/libpromises/failsafe.cf
+++ b/libpromises/failsafe.cf
@@ -30,25 +30,40 @@
 # system by fetching policies in case the standard agent run fails.
 ##############################################################################
 
-
-body common control
-{
-      bundlesequence => {
-                          "cfe_internal_checkkeys",
-                          "cfe_internal_update",
-                          "cfe_internal_call_update"
-                        };
-}
-
 body agent control
 {
-      skipidentify => "true";
-
       # Bootstrapping can't continue without keys
       abortclasses => { "no_ppkeys_ABORT" };
 }
 
-################################################################################
+##############################################################################
+
+bundle agent main
+{
+  meta:
+      "description"
+        string => "Perform bootstrap or failsafe recovery operations";
+
+  methods:
+
+      "Check Keys"
+        usebundle => cfe_internal_checkkeys,
+        comment => "Without a valid keypair we aren't going to be able
+                    to establish trust";
+
+      "Fetch Inputs"
+        usebundle => cfe_internal_update,
+        comment => "We need to fetch policy from upstream if we are
+                    bootstrapping or if we are performing failsafe
+                    recovery";
+
+      "Activate Update Policy" -> { "CFEngine Enterprise", "Mission Portal" }
+        usebundle => cfe_internal_call_update,
+        comment => "In order to speed up convergence and reporting we
+                    trigger the update policy right after initial
+                    bootstrap. This allows the first scheduled run to
+                    happen with the most complete information.";
+}
 
 bundle agent cfe_internal_checkkeys
 {
@@ -63,7 +78,7 @@ bundle agent cfe_internal_checkkeys
         classes => kept("no_ppkeys_ABORT");
 }
 
-################################################################################
+##############################################################################
 
 bundle agent cfe_internal_update
 {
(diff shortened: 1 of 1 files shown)


nickanderson:
trigger build

libpromises/failsafe.cf
```diff
@@ -30,25 +30,44 @@
...
-                          "cfe_internal_update",
-                          "cfe_internal_call_update"
-                        };
-}
-
 body agent control
 {
+      # We can't expect dns to be properly configured everywhere so we
+      # skip the dns check for bootstrap
       skipidentify => "true";
```
jimis: `skipidentify` only affects classic protocol connections, which are no longer used by default.

libpromises/failsafe.cf
```diff
@@ -30,25 +30,44 @@
...
 
       # Bootstrapping can't continue without keys
       abortclasses => { "no_ppkeys_ABORT" };
 }
 
-################################################################################
+###################################################################################################################################
```
jimis: How many hash marks are too many? :-) I would say anything over the 78th column.

libpromises/failsafe.cf
```diff
@@ -30,25 +30,44 @@
...
+
+bundle agent main
+{
+  meta:
+      "description"
+        stirng => "Perform bootstrap or failsafe recovery operations";
```
jimis: s/stirng/string/

libpromises/failsafe.cf
```diff
@@ -84,80 +109,116 @@ bundle agent cfe_internal_update
...
     windows::
 
-      "$(sys.workdir)\\bin-twin\\."
-      handle => "cfe_internal_bootstrap_update_files_sys_workdir_bin_twin_windows",
-      comment => "Make sure we maintain a clone of the binaries and libraries for updating",
-      copy_from => u_cp("$(sys.workdir)\\bin\\."),
-      depth_search => u_recurse("1");
-
+      # TODO: Remove the use of bin-twin
+      "$(sys.workdir)\\bin-twin\\." -> { "[[redmine:xxxx]]" }
```
jimis: What is this syntax?

jimis:
Nice refactoring, besides the minor comments. But does it include functionality change? I believe the checking if cf-serverd and cf-execd daemons are running, has changed, right? If so in what way? Maybe it should be separate.

libpromises/failsafe.cf
nickanderson: OK, thats fair, i was making them long enough to match the longest line

libpromises/failsafe.cf
nickanderson: oh thats some of my org stuff :)

libpromises/failsafe.cf
nickanderson: ack

libpromises/failsafe.cf
nickanderson: So we can redact the skipidentify all together then?

nickanderson:
Functionally no change, only cleanup of classes to make it more simple (at least in my mind).

The functional changes as part of this re-factor are minimal.

1) No longer warning that execd was not started based on the fact that it wasn't started. Now it only warns if it fails to restart execd. Note the subtle difference between reporting a failure when it wasn't attempted, vs reporting a failure when we know there was a failure. This handles cases where execd is already running but failsafe triggers.

2) Previously we only checked for cf-serverd if we were the policy server and we had fetched policy successfully. Now we define the class for cf-serverd process not running on all hosts. Previously the class `start_server` was only used to to start cf-serverd on non systemd hosts (but remember that this class was only set on policy servers). Now we only run the command to start cf-serverd on the policy_server when cf_serverd_not_running and its not a systemd host. Which is basically the same as what we had before since we only defined the start_server class on policy hubs to begin with.

So the only net change should be not reporting failure to start execd when we didnt bother trying (which is I think what we want).

nickanderson:
changed based on feedback and rebased.

nickanderson:
ping @jimis any obections?

jimis:
If your bootstrap and failsafe run tests work, then let me know and I'll merge! :-)

jimis:
trigger build

jimis:
hmmm, it didn't pass `make check` in Jenkins, did it pass in your local test?

```
00:05:07.764 Starting test: policy_test.c
00:05:07.764 ==================================================
00:05:07.764 test_failsafe: Starting test
00:05:07.766 FAIL: policy_test
00:05:07.860 ==================================================
```

nickanderson:
Gah your right, I had some syntax errors in there.

I found and fixed 2 issues.

```
> $ cf-promises -cf ./libpromises/failsafe.cf  
```

nickanderson:
Lets wait on this for https://github.com/cfengine/core/pull/2500/commits which fixes an issue for duplicate definition if you include the stdlib in def.json

nickanderson:
superseded by #2510

PR #2495, opened 2016-02-13
New network_connections() function

tzz:
Split off from #2245, requires it before this is merged since it incorporates it. 

* add `network_connections()` function
* export some function definitions from `unix_iface.c` so `evalfunction.c` can use them
* acceptance tests!

Implements https://dev.cfengine.com/issues/4520


tzz:
Rebased on top of #2245 again.

tzz:
Rebased on top of #2245 again. Also broke up the old format of `IP:PORT` for every entry into `{ address: "IP", port: "PORT" }`. Acceptance test updated.

tzz:
@bahamat @kacf the change makes the network connections info easier to use with IPv6, where the `:` separator can mean port or octet.

tzz:
I remembered that this implements https://dev.cfengine.com/issues/4520 so I've mentioned it in the ticket.

tzz:
This PR has been rebased after the merge of #2245 and is ready for review.

libenv/unix_iface.c
```diff
@@ -760,38 +759,50 @@ static int aix_get_mac_addr(const char *device_name, uint8_t mac[6])
...
+                JsonObjectAppendString(ip, "port", BufferData(buf));
+                JsonObjectAppendElement(element, new_key, ip);
+            }
+            else
+            {
+                JsonObjectAppendString(element, new_key, BufferData(buf));
+            }
+
+            BufferDestroy(buf);
+            JsonObjectRemoveKey(element, raw_key);
```
kacf: Could `raw_key` and `new_key` theoretically be the same? If so I assume that `JsonObjectAppendElement` would in fact not append the element, but overwrite it, with it being a map and all (I haven't actually verified this though). Then you should not remove the element afterwards.

libenv/unix_iface.h
```diff
@@ -0,0 +1,43 @@
...
+
+
+void GetInterfacesInfo(EvalContext *ctx);
+void GetNetworkingInfo(EvalContext *ctx);
+
+typedef void (*ProcPostProcessFn)(void *ctx, void *json);
+ProcPostProcessFn NetworkingPortsPostProcessInfo(ARG_UNUSED void *passed_ctx, void *json);
+JsonElement* GetProcFileInfo(EvalContext *ctx, const char* filename, const char* key, const char* extracted_key, ProcPostProcessFn post, const char* regex);
+
+#endif
```
kacf: I just removed this file, please don't bring it back! :-)

Either make the functions local to `unix_iface.c`, or put the function declarations in `sysinfo.h` instead.

libpromises/evalfunction.c
```diff
@@ -7085,6 +7086,64 @@ static FnCallResult FnCallProcessExists(ARG_UNUSED EvalContext *ctx, ARG_UNUSED
...
+    {
+        JsonObjectAppendElement(json, "udp", data);
+    }
+
+    BufferPrintf(pbuf, "%s/proc/net/udp6", procdir);
+    data = GetProcFileInfo(ctx, BufferData(pbuf), NULL, NULL, (ProcPostProcessFn) &NetworkingPortsPostProcessInfo, ports_regex);
+    if (NULL != data)
+    {
+        JsonObjectAppendElement(json, "udp6", data);
+    }
```
kacf: This stuff is still fairly Linux specific. Maybe this parsing stuff should be moved into `unix_iface.c` and this function will just call it?

tests/acceptance/00_basics/environment/proc-net-functions.cf.sub
```diff
@@ -0,0 +1,30 @@
...
+  vars:
+      "testname" string => regex_replace($(this.promise_filename), "\\.sub$", "", "");
+
+  methods:
+      "check"  usebundle => dcs_check_state(test,
+                                           "$(this.promise_filename).expected.json",
+                                           $(testname));
+      "check"  usebundle => dcs_check_state(test,
+                                           "$(this.promise_filename).expected.json",
+                                           $(this.promise_filename));
```
kacf: This is a bit confusing. I would remove the `$(this.promise_filename)` version (the last one), since the first one is the actual test name. It's not required to pass, is it?

libpromises/evalfunction.c
tzz: OK; done

libenv/unix_iface.h
tzz: OK; done

libenv/unix_iface.c
tzz: Fixed, nice catch!

tests/acceptance/00_basics/environment/proc-net-functions.cf.sub
tzz: This lets me run the subtest individually, it's very helpful while testing. I hope we can leave it.

tzz:
This is fixed, rebased, and ready. The only question I had is if we should put the network connections info under `inet -> tcp`, `inet6 -> tcp`, `inet -> udp`, `inet6 -> udp` instead of `tcp, tcp6, udp, udp6` to be consistent with the `sys.inet` and `sys.inet6` data containers. I'd leave this as is, but maybe @kacf or @bahamat have feedback?

kacf:
I think it's a good idea what you say about putting `inet -> tcp`, etc. Does it make it more difficult to use with `mergedata`? How so?

tests/acceptance/00_basics/environment/proc-net-functions.cf.sub
kacf: At least protect it with a class you have to set, maybe `debug_proc` or something. The reason is that we've had a few cases of false negatives in the past, where the test passes, but it should have failed. I really want to minimize that possibility, and this would increase the risk.

tzz:
`mergedata(foo, network_connections())` will put `inet -> tcp` etc. into `foo`. If `foo` has a pre-existing `inet` key, it will be overwritten since the merge happens only at the top level (this is not optional). So I think we're better off with the current behavior.

tzz:
@kacf I added `test_debug` to the tests as you suggested, and this is rebased and ready.

tzz:
Documentation is in https://github.com/cfengine/documentation/pull/1443

kacf:
> So I think we're better off with the current behavior.

Ok, either is fine with me.

kacf:
trigger build

kacf:
`proc-net-functions.cf` test fails on all non-Linux test machines. No messages other than the regular `FAIL` message.

I'll try to look at this, but after #2319.

tzz:
Sorry, I forgot to add the `meta` tag to skip on non-Linux. Done, this should pass now.

kacf:
Ah ok, rerunning.

kacf:
trigger build

kacf:
Alright, together with https://github.com/cfengine/enterprise/pull/307, this is green.

kacf:
Squashed and merged manually. Thanks!

tzz:
Woohoo! Epic!

PR #2496, opened 2016-02-16
Add: Test for Redmine #7947

nickanderson:
Something is wrong with this test. It's passing for me on 3.8.1, but based on
previous manual tests and tests by @pasinskim it should be failing.


kacf:
So you're saying the test doesn't work? We can't merge it as is then.

kacf:
Closing this for now. Please reopen if you fix the test.

PR #2498, opened 2016-02-17
Add: Test that agent able to setgid without root owning dir

nickanderson:


nickanderson:
trigger build

tests/acceptance/10_files/can_set_sticky_bits_without_root_owning_directory.cf
```diff
@@ -0,0 +1,80 @@
+#######################################################
+#
+# Acceptance test for issue #2745
```
kacf: This should specify that it's zendesk, not Redmine.

kacf:
trigger build

tests/acceptance/10_files/can_set_sticky_bits_without_root_owning_directory.cf
```diff
@@ -0,0 +1,80 @@
...
+
+      #######################################
+      #Clean up test directory before testing
+      #######################################
+
+  files:
+      "$(G.testdir)/test_sgid"
+        delete => tidy;
+
+}
```
kacf: Note that these section are never necessary when running under `testall`, because it always starts with a clean slate. It would be nice not to clutter every test with this.

tests/acceptance/10_files/can_set_sticky_bits_without_root_owning_directory.cf
```diff
@@ -0,0 +1,80 @@
...
+
+#######################################################
+
+bundle agent check
+{
+
+  vars:
+    test_sgid_dir_reached::
+      "expect_permoct" string => "$(test.mode)";
+      "result_permoct" string => execresult('/usr/bin/stat -c %a $(G.testdir)/test_sgid', "useshell");
```
kacf: I'm pretty sure `stat` doesn't exist on all platforms. This probably should either be changed, or we skip it on non-Linux.

kacf: It appears I was wrong, this passes everywhere.

kacf:
Merged manually.

PR #2503, opened 2016-02-25
Fixes #7956 : cf-promises -p cf doesn't produce valid cfengine code

peckpeck:
https://dev.cfengine.com/issues/7956

### DIFF
--- a/libpromises/policy.c
+++ b/libpromises/policy.c
@@ -1891,15 +1891,13 @@ static void IndentPrint(Writer *writer, int indent_level)
 static void AttributeToString(Writer *writer, Constraint *attribute, bool symbolic_reference)
 {
     WriterWriteF(writer, "%s => ", attribute->lval);
-    if (attribute->rval.type == RVAL_TYPE_SCALAR && !symbolic_reference)
+    if(symbolic_reference)
     {
-        WriterWriteChar(writer, '"');
         RvalWrite(writer, attribute->rval);
-        WriterWriteChar(writer, '"');
     }
     else
     {
-        RvalWrite(writer, attribute->rval);
+        RvalWriteQuoted(writer, attribute->rval);
     }
 }
 
@@ -1948,9 +1946,10 @@ void BodyToString(Writer *writer, Body *body)
             }
         }
 
-        WriterWriteChar(writer, '\n');
         IndentPrint(writer, 1);
         AttributeToString(writer, cp, false);
+        WriterWriteChar(writer, ';');
+        WriterWriteChar(writer, '\n');
     }
 
     WriterWrite(writer, "\n}\n");
@@ -1969,24 +1968,20 @@ void BundleToString(Writer *writer, Bundle *bundle)
 
         WriterWriteF(writer, "\n%s:\n", promise_type->name);
 
+        char *current_class = NULL;
         for (size_t ppi = 0; ppi < SeqLength(promise_type->promises); ppi++)
         {
             Promise *pp = SeqAt(promise_type->promises, ppi);
-            char *current_class = NULL;
 
             if (current_class == NULL || strcmp(pp->classes, current_class) != 0)
             {
                 current_class = pp->classes;
-
-                if (strcmp(current_class, "any") != 0)
-                {
-                    IndentPrint(writer, 1);
-                    WriterWriteF(writer, "%s::", current_class);
-                }
+                IndentPrint(writer, 1);
+                WriterWriteF(writer, "%s::\n", current_class);
             }
 
             IndentPrint(writer, 2);
-            WriterWriteF(writer, "\"%s\"", pp->promiser);
+            ScalarWrite(writer, pp->promiser, true);
 
             /* FIX: add support
              *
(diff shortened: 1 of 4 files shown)


peckpeck:
Hello, anybody on this one ?

libpromises/rlist.c
```diff
@@ -1206,35 +1204,35 @@ void RlistWrite(Writer *writer, const Rlist *list)
...
         }
-        WriterWriteChar(w, *s);
+        WriterWriteChar(writer, *s);
     }
+    if(!dontQuote)
+        WriterWriteChar(writer, '"');
 }
 
-void RvalWrite(Writer *writer, Rval rval)
+void RvalWrite(Writer *writer, Rval rval, bool dontQuote)
```
kacf: In most places we don't quote, and the extra bool is an annoying and somewhat confusing argument to include in every call. Therefore I suggest that you split this into `RvalWrite` and `RvalWriteQuoted`, both taking the same arguments as before the patch. The patch will be much smaller, since you won't have to change every place that uses it.

libpromises/rlist.c
```diff
@@ -1206,35 +1204,35 @@ void RlistWrite(Writer *writer, const Rlist *list)
...
 char *RvalToString(Rval rval)
 {
     Writer *w = StringWriter();
-    RvalWrite(w, rval);
+    RvalWrite(w, rval, true);
     return StringWriterClose(w);
 }
 
-void RvalWriteParts(Writer *writer, const void* item, RvalType type)
+void RvalWriteParts(Writer *writer, const void* item, RvalType type, bool dontQuote)
```
kacf: Same with this one, split into `RvalWriteParts` and `RvalWritePartsQuoted`.

libpromises/rlist.c
```diff
@@ -1206,35 +1204,35 @@ void RlistWrite(Writer *writer, const Rlist *list)
     WriterWriteChar(writer, '}');
 }
 
-/* Note: only single quotes are escaped, as they are used in RlistWrite to
-   delimit strings. If double quotes would be escaped, they would be mangled by
-   RlistParseShown */
-
-static void ScalarWrite(Writer *w, const char *s)
+void ScalarWrite(Writer *writer, const char *s, bool dontQuote)
```
kacf: You can leave this one with the extra bool, since it is internal to this file. However, I ask that you flip the meaning of the flag, IOW use `bool quote` instead of `bool dontQuote`, so you avoid the confusing double negative. This means you'll have to flip the true/false flags where you call it as well, of course.

libpromises/policy.c
```diff
@@ -2000,10 +1988,15 @@ void BundleToString(Writer *writer, Bundle *bundle)
             {
                 Constraint *cp = SeqAt(pp->conlist, k);
 
-                WriterWriteChar(writer, '\n');
                 IndentPrint(writer, 4);
                 AttributeToString(writer, cp, cp->references_body);
+                if(k < SeqLength(pp->conlist)-1) {
+                  WriterWriteChar(writer, ',');
+                  WriterWriteChar(writer, '\n');
+                }
```
kacf: Incorrect indentation, should be four spaces per level.

kacf:
Otherwise this seems like a good patch.

kacf:
Just let me know when you've updated it.

libpromises/rlist.c
peckpeck: But the bigger part will be the code duplication.
Moreover, i would advise to properly quote everywhere

libpromises/rlist.c
peckpeck: I removed the static part, it is not anymore internal to this file

libpromises/rlist.c
kacf: `RvalWrite` is a one liner, there is not much duplication. But see my other comment as well about `RvalWriteParts`.

libpromises/rlist.c
kacf: You're right, this function is too long to duplicate. But in fact, it's not used anywhere besides this file, so keep the name and the `quote` argument, but make it `static` instead, and remove it from `rlist.h` (possibly including the declaration at the top of this file instead).

kacf: My proposal to change `dontQuote` to `quote` still stands though.

libpromises/rlist.c
kacf: Oh you're right. But still I would change `dontQuote` to `quote`.

peckpeck:
@kacf PR updated

kacf:
Nice, thanks!

kacf:
trigger build

libpromises/rlist.c
```diff
@@ -1206,35 +1204,23 @@ void RlistWrite(Writer *writer, const Rlist *list)
...
 
-/* Note: only single quotes are escaped, as they are used in RlistWrite to
-   delimit strings. If double quotes would be escaped, they would be mangled by
-   RlistParseShown */
-
-static void ScalarWrite(Writer *w, const char *s)
+void ScalarWrite(Writer *writer, const char *s, bool quote)
 {
+    if(quote)
+        WriterWriteChar(writer, '"');
```
kacf: Code style:
* Should be space before expression after if statement.
* No if blocks without braces.

I fixed these for now, but please remember for the future.

kacf:
Thanks! Added changelog and merged manually.

PR #2504, opened 2016-02-26
libpromises/generic_agent.c: allow def.json up to 5MB instead of 4K

tzz:
Currently the maximum size for `def.json` is 4K. This increases it to 5MB, which I think is more reasonable.

See https://dev.cfengine.com/issues/7960 for more information (but it's not the primary bug report, so I didn't use it in the commit message).

### DIFF
--- a/libpromises/generic_agent.c
+++ b/libpromises/generic_agent.c
@@ -97,7 +97,8 @@ static JsonElement *ReadJsonFile(const char *filename, LogLevel log_level)
     }
 
     JsonElement *doc = NULL;
-    JsonParseError err = JsonParseFile(filename, 4096, &doc);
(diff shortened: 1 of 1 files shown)


kacf:
Thanks, I think this should be safe to include without running it through the tests. I merged manually after fixing a minor whitespace error.

For future work, check also the changelog message in https://github.com/cfengine/core/commit/f03ab8289d0775ca821b39a724c6f6f08ef8fb17. I know, it looks a bit redundant, but references to source files are irrelevant to the readers of the changelog, so I included just the message.

Cherry-picked to 3.8.x as well, and I've made a note to include it in 3.7.x if we include the overall def.json fix there.

PR #2505, opened 2016-02-29
Fix regex for matching BSD and GNU tagets

skreuzer:
Fix a typo which was resulted in the check for yp_get_default_domain being skipped on BSD and GNU targets

Standardize the regex used to check for a GNU target

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1181,7 +1181,7 @@ case "$target_os" in
    aix*)
         CPPFLAGS="$CPPFLAGS -w"
         ;;
-   linux*|*bsd*gnu)
+   linux*|*bsd*|*gnu*)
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -1196,7 +1196,7 @@ case "$target_os" in
         ;;
    openbsd*|obsd*)
         ;;
-   gnu*)
+   *gnu*)
```
kacf: I believe this can be removed. It is already included in the above case.

skreuzer:
> I believe this can be removed. It is already included in the above case.

The new pull requests removes that section completely

kacf:
Thanks!

kacf:
trigger build

kacf:
Cherry-picked to 3.6.x, 3.7.x and 3.8.x as well.

PR #2506, opened 2016-02-29
Hash collisions redmine7912

jimis:


jimis:
trigger build

jimis:
@kacf please review

jimis:
First build is green, triggering again because I rebased with a couple of cosmetic changes.

jimis:
trigger build

jimis:
All green, I'll now amend a commit with an acceptance test.

jimis:
Acceptance test added, tested failing before the patch and passing afterwards.

jimis:
trigger build

jimis:
all green

jimis:
@kacf please merge if it looks ok after you have a look

jimis:
Fixed small memory leak, squashing it to the original commit.

jimis:
trigger build

libpromises/class.c
```diff
@@ -244,11 +242,20 @@ ClassTableIterator *ClassTableIteratorNew(const ClassTable *table, const char *n
...
     {
+        Class *cls = keyvalue->value;
+
+        /* Make sure we never store "default" as namespace in the ClassTable,
+         * instead we have always ns==NULL in that case. */
+        CF_ASSERT_FIX(cls->ns == NULL ||
+                      strcmp(cls->ns, "default") != 0,
+                      (cls->ns = NULL),
+                      "Class table contained \"default\" namespace,"
+                      " should never happen!");
```
kacf: I'm confused about this part. Here you say that you don't store the namespace, but you *do* store the namespace specifically in `ClassTablePut()`, by using `fullname`. Am I missing something?

kacf:
Exactly the type of fix I was hoping for! Just two issues:

* The namespace storing issue I mentioned in the code comment.
* Changelog entry is missing.

libpromises/class.c
jimis: On Wed, Mar 2, 2016 at 9:44 AM, Kristian Amlie <notifications@github.com>
wrote:

> In libpromises/class.c
> <https://github.com/cfengine/core/pull/2506#discussion_r54693749>:
>
> >      {
> > +        Class *cls = keyvalue->value;
> > +
> > +        /* Make sure we never store "default" as namespace in the ClassTable,
> > +         * instead we have always ns==NULL in that case. */
> > +        CF_ASSERT_FIX(cls->ns == NULL ||
> > +                      strcmp(cls->ns, "default") != 0,
> > +                      (cls->ns = NULL),
> > +                      "Class table contained \"default\" namespace,"
> > +                      " should never happen!");
>
> I'm confused about this part. Here you say that you don't store the
> namespace, but you *do* store the namespace specifically in
> ClassTablePut(), by using fullname. Am I missing something?
>
>
"fullname" is the key, in this key-value store, and is indeed always
including the namespace. However this assertion ensures that the *value* of
the key-value store, i.e. struct Class, is in the correct form.

BTW this assertion is rather expensive. I'm contemplating on introducing
something like CF_DEBUG_ASSERT that completely disappears ifdef NDEBUG.

jimis:
Added changelog message and rebased.

jimis:
Thanks for reviewing @kacf. I'm merging to continue with other fixes on top of it.

PR #2507, opened 2016-02-29
Replaced a bunch of assert(false) with appropriate messages.

jimis:


jimis:
@tzz please review, this is just a blind search and replace for figuring out [redmine#7952](http://dev.cfengine.com/issues/7952).

tzz:
Seems reasonable.

jimis:
Thanks Ted. BTW I was a bit unsure whether all these should be `ProgrammingError()`, i.e. instant crash in both prod and debug versions. I can change it if there is consensus. @kacf what's your opinion?

tzz:
@jimis from my perspective, it's always better to recover gracefully and generate a problem report than to abort in any situation. It's just more work.

kacf:
Fully agree with adding more info to each assert. However, I also think it's a little risky to turn these into aborts in production builds. It's possible that we have been hitting some of those asserts in the field without knowing. As far as I have understood, `ProgrammingError` always exits and `UnexpectedError` always carries on, right? I propose adding our own `Assert()` with a message which aborts in debug mode and carries on in release mode. That way we get the best of both worlds, we always get the messages, and it will abort in our test builds, but we don't suddenly start aborting in a place where there's a problem we didn't know about.

kacf:
In looking at your other pull request, I realize that `CF_ASSERT` and `CF_ASSERT_FIX` already exist. Perhaps you can use one of those?

jimis:
CF_ASSERT is kind of unsuitable here, since there is no condition to check,
most of the errors here are unreachable parts.

It's been bugging me some time now; Why are ProgrammingErrors aborting in
non-debug builds? Wouldn't it be better if they turned into
UnexpectedErrors ifdef NDEBUG?

kacf:
> CF_ASSERT is kind of unsuitable here, since there is no condition to check, most of the errors here are unreachable parts.

That's fine, you can use false as before. The important thing is that we get a message if we *do* in fact get there.

> It's been bugging me some time now; Why are ProgrammingErrors aborting in non-debug builds? Wouldn't it be better if they turned into UnexpectedErrors ifdef NDEBUG?

Because `ProgrammingError` is meant for places where it's not possible to recover gracefully. I've had a few such cases in the past, for example in `generic_at.c` there is a section which tries to unlock a mutex (but I see now that I have used the wrong function, it's `UnexpectedError`, but should have been `ProgrammingError`). Not being able to unlock it is guaranteed to lead to a deadlock later, so the only possible action is to abort. But it should never happen of course.

jimis:
On Wed, Mar 2, 2016 at 12:03 PM, Kristian Amlie <notifications@github.com>
wrote:

> Because ProgrammingError is meant for places where it's not possible to
> recover gracefully. I've had a few such cases in the past, for example in
> generic_at.c there is a section which tries to unlock a mutex (but I see
> now that I have used the wrong function, it's UnexpectedError, but should
> have been ProgrammingError). Not being able to unlock it is guaranteed to
> lead to a deadlock later, so the only possible action is to abort. But it
> should never happen of course.
>

Then how about configuring "UnexpectedError" to abort in debug builds, and
only print a message (like it does) in non-debug builds?

kacf:
You should double check usage, but yeah, I agree.

jimis:
Great, I wouldn't like to introduce even more custom asserts. Furthermore using `CF_ASSERT(false)` is not nice IMHO.

Hmmm I'll see if I can steal some time to change `UnexpectedError()` to abort in debug builds but this pull request should be mergeable already.

jimis:
> Because ProgrammingError is meant for places where it's not possible to recover gracefully.

BTW this is not clear. `ProgrammingError()` to me means that it is the programmer's fault if this point is reached, just like an `assert()`. Recoverable or not is not obvious.

Argh I'm putting too much thought in this.

jimis:
I'm merging this as UnexpectedError can (and should) change behaviour with a separate pull request.

PR #2509, opened 2016-02-29
Don't follow symbolic links when copying extended attributes

estenberg:
We want to copy the attributes of the source link to the destination link.
Seen when using copy_from.preserve => "true".

https://cfengine.zendesk.com/agent/tickets/2792


kacf:
There are many other places where these function calls are used, including when detecting platform support. It seems prudent to use them everywhere. I resubmitted this here: https://github.com/cfengine/core/pull/2521.

PR #2511, opened 2016-03-02
Replace RBTree with Map for vars.

jimis:


jimis:
Dammit the first commit is bogus because of auto-rebasing. Fixing...

jimis:
Rebased and it's ready. Please review commit by commit, as it fixes a couple of issues separately.

jimis:
Ah, don't bother to trigger build; this is failing in two tests, where JSON output order has changed, but not in a significant way. I'll ask for opinions on this once my local test run has finished.

jimis:
OK, on test failing is `./01_vars/02_functions/getindices_returns_expected_list_from_array.cf` where the lines in `getindices_returns_expected_list_from_array.cf.expected.json` have changed order.

The second test failing is `./01_vars/02_functions/maparray_mapdata.cf` where the order is again changed in the file `maparray_mapdata.cf.expected.json`.

I could resubmit the `expected` files to have the current order, but I feel this is the wrong fix. Since the RBTree (ordered data structure) was replaced by a hash table (unordered), the order is not something we can rely on. And simply piping through `sort` will destroy the JSON syntax. Any ideas on how to handle it? @kacf @tzz what do you think?

tzz:
@jimis please resubmit with the current order, and assign a ticket to me to make these tests ignore the order by sorting. I'll take care of it later but it shouldn't block you now, and it's not a trivial fix because unfortunately we can't say `sort(maparray(...))`.

jimis:
@tzz thanks, much appreciated. I'll change the order to be the correct one and squash it to the original commit.

I'm wondering however if the order will change across platforms, because the hash function might yield different results because of endianess. Lets see...

jimis:
trigger build

jimis:
lets see how it fares in all platforms

jimis:
Wow, how unexpected; one test failed on one platform: `./10_files/02_maintain/line_truncation_at_4096_chars.cf` on HP-UX.

jimis:
Hmm it seems like a transient error on our HP-UX buildslave.

jimis:
I could use some help on this one.

It seems that perl is not discovered right in our buildslave; `G.perl` variable points to `/bin/perl` which is a symlink to `/opt/perl/bin/perl`, which is not accessible. However the test in `dcs.cf.sub` is `fileexists(/bin/perl)` which is true, despite being a symlink leading to nowhere.

@nickanderson @tzz how to change `dcs.cf.sub` to dereference potential symlinks before checking for existence? The relevant part:
```
  classes:
      # General commands.
      "$(paths_indices)_$(cmds)$(exeext)" expression => fileexists("$(paths[$(paths_indices)])$(const.dirsep)$(cmds)$(exeext)");
      "has_$(cmds)"                       expression => fileexists("$(paths[$(paths_indices)])$(const.dirsep)$(cmds)$(exeext)"),
      scope => "namespace";
```

jimis:
Wow I'm impressed by myself. @tzz @nickanderson please review the last patch that touches on `dcs.cf.sub`, commit e211309.

jimis:
trigger build

tzz:
Nice one @jimis that looks great :)

You could save some space by doing `fileexists(filestat(...))` instead of creating an intermediate variable.

jimis:
trigger build

jimis:
Thanks @tzz, I was wondering if that was the case.

jimis:
Great, now another irrelevant test fails, `./05_processes/01_matching/timed/stime.cf`.

It seems that `G.sh` is not expanded on Ubuntu. I believe it's a bug in CFEngine function `filestat()`. While running `cf-promises` on that file, the following message is logged 3 times:
```
    info: filestat bailing on link 'dash' (original '/bin/sh') because it could not be read
```
But the link is there, as `ls` shows, but it is *relative*, not *absolute* link:
```
lrwxrwxrwx 1 root root 4 Jun  4  2015 /bin/sh -> dash
```
So it seems that `filestat()` is not following it properly, and `strace` verifies this.

tzz:
Oh, that's weird. I think I've seen that error too (in Docker containers). If you can fix it, that would be terrific!

jimis:
Fixed, but not exactly so, since now the broken test passes but 4 new tests fail! ARGH!

jimis:
trigger build

jimis:
All fixed, lots of changed behaviour because of this fix, please see the commit messages, documentation must be updated as well.

jimis:
Lets see what jenkins has to say... trigger build

kacf:
I won't look through the intermediary steps, @jimis, will you let me know when I can review the final?

jimis:
@kacf SOON

tzz:
Code and docs look good to me.

jimis:
Alright this should be ready, I added one acceptance test for the `filestat()` change and squashed what needed to be squashed!

@kacf please review and merge if the build is green. :-)

jimis:
trigger build

libpromises/class.c
```diff
@@ -42,10 +47,10 @@ static void ClassDestroy(Class *cls);                /* forward declaration */
 TYPED_MAP_DECLARE(Class, char *, Class *)
 
 TYPED_MAP_DEFINE(Class, char *, Class *,
-                 (MapHashFn) &StringHash,
-                 (MapKeyEqualFn) &StringSafeEqual,
+                 StringHash_untyped,
+                 StringSafeEqual_untyped,
                  free,
-                 (MapDestroyDataFn) &ClassDestroy)
+                 ClassDestroy_untyped)
```
kacf: I like it! Admit it, you are becoming a template person! :-)

libutils/array_map.c
```diff
@@ -52,8 +52,12 @@ int ArrayMapInsert(ArrayMap *map, void *key, void *value)
     {
         if (map->equal_fn(map->values[i].key, key))
         {
-            map->destroy_key_fn(key);
+            /* Replace the key with the new one despite those two being the
+             * same, since the new key might be referenced somewhere inside
+             * the new value. */
+            map->destroy_key_fn(map->values[i].key);
             map->destroy_value_fn(map->values[i].value);
+            map->values[i].key   = key;
```
kacf: Nice. I'm amazed that this has worked so far...

libpromises/var_expressions.c
```diff
@@ -95,6 +96,13 @@ static size_t VarRefHash(const VarRef *ref)
     return (h & (INT_MAX - 1));
 }
 
+unsigned int VarRefHash_untyped(const void *ref,
+                                unsigned int seed ARG_UNUSED,
+                                unsigned int max)
+{
+    return VarRefHash(ref) & (max - 1);
+}
```
kacf: I might be completely misunderstanding what you're trying to do here, but shouldn't this be XOR (`^`)? Otherwise you risk a hash that is very biased towards `1` bits, since they are anded.

kacf: Never mind, I misunderstood. As long as `max` is a power of two, this is valid. I think that deserves a comment here though (and perhaps in some of the other places where this pops up).

tests/unit/variable_test.c
```diff
@@ -287,11 +287,13 @@ static void test_iterate_indices(void)
 
 // Below test relies on the ordering items in RB tree which is strongly
 // related to the hash function used.
+/* No more relevant, RBTree has been replaced with Map. */
+#if 0
```
kacf: Just remove all this. It's there in the history if we need it.

tests/unit/variable_test.c
```diff
@@ -361,7 +364,7 @@ int main()
         unit_test(test_get_in_default_namespace),
         unit_test(test_get_different_namespaces),
         unit_test(test_get_indices),
-        unit_test(test_iterate_indices_ordering_related),
+//        unit_test(test_iterate_indices_ordering_related),
```
kacf: And remove this too.

tests/acceptance/dcs.cf.sub
```diff
@@ -83,11 +83,18 @@ bundle common G
...
   classes:
       # General commands.
-      "$(paths_indices)_$(cmds)$(exeext)" expression => fileexists("$(paths[$(paths_indices)])$(const.dirsep)$(cmds)$(exeext)");
-      "has_$(cmds)"                       expression => fileexists("$(paths[$(paths_indices)])$(const.dirsep)$(cmds)$(exeext)"),
-      scope => "namespace";
+      "$(paths_indices)_$(cmds)$(exeext)" expression =>
+          fileexists("$(deref_paths[$(paths[$(paths_indices)])][$(cmds)])");
+      "has_$(cmds)"                       expression =>
+          fileexists("$(deref_paths[$(paths[$(paths_indices)])][$(cmds)])"),
+          scope => "namespace";
```
kacf: We may have to revisit this at some point, perhaps using some way of preserving the order. But this is good enough for now.

tests/acceptance/10_files/02_maintain/absolute_link_and_no_move_obstructions.cf
```diff
@@ -44,8 +44,8 @@ bundle agent test
 
   files:
       "$(G.testfile)"
-      classes => init_if_failed("test_ok"),
-      link_from => test_link;
+          classes => init_if_failed("test_ok"),
+          link_from => test_link;
```
kacf: The official CFEngine policy standard is two spaces per level, so that will be:

```
{
  files:
    class::
      "promiser"
        attribute => "value";
}
```

If class is missing you still behave as if it were there.

I know, this is violated a lot, but if you're fixing it anyway...

tzz: Remember you have `contrib/reindent.pl` for this :)

libpromises/class.c
jimis: NO-WAY! But I still hate dangerous casts. :-)

tests/unit/variable_test.c
jimis: > It's there in the history if we need it.

Heh, famous last words. I'm quite sure that if in 6 months or more we'll reintroduce order in the variables and we need to test it, somebody will just rewrite a test if this is removed. Plus it's a good indication of functionality change.

I'm more in favour of keeping a bit of garbage in the test file, if it reminds everyone that opens it that something important changed. For some time at least.

tests/acceptance/dcs.cf.sub
jimis: Agreed. @tzz work on `sort()` would be valuable for improving the test suite (see https://github.com/cfengine/core/pull/2514 ) .

tests/acceptance/10_files/02_maintain/absolute_link_and_no_move_obstructions.cf
jimis: Sooo I just have to edit `reindent.pl` to change the stinky indentation rules? It's perl but I think I might have to do that...

jimis:
```
      "promiser"
        attribute => "value";
```
The 2 spaces under the promises are almost always 0 spaces in our acceptance tests, which is unreadable. Furthermore, since the promisers first character is a quote, which looks like a space, the first letter of the promiser is way too close to the first letter of the attribute underneath.

Anyway, I just indented to 4 spaces based on how easily I could read that policy the moment I was editing it.

tzz: `reindent.pl` just runs Emacs to reindent the code, so you need to look at the Emacs indentation settings (see the script, they are in `contrib/cfengine-code-style.el` for C and CFE code). It should Just Work.

We used to indent attributes 0 spaces, that's why older code has that (it should all be reindented some day).

libpromises/var_expressions.c
jimis: You are right actually, I just copied this code from every other hash function we have. But now that I looked deeper into it, here are some strange issues I can't help but notice:

* the `max` argument  can indeed be only power of two for this to work right; but there is no documentation, nor assert. IMHO the `max` argument should work for any number, but for now I will amend a commit that `assert()`s power of two - it should be better than a comment.
* With a quick check, it seems our hash table **always uses 8192 size of buckets**. This is bad, I'll file a redmine ticket.
* The `max` argument in already existing in all our hash functions, the purpose is for getting a random number up to the size of the hash table (in case that size was dynamic). Still I think that `max` bounding should be done separately, not in our hash function.

So I'll amend a commit asserting max, and file a bug report for refactoring `hash_map.c`.

libutils/array_map.c
jimis: Well there was a very mysterious crash that was resolved by accident after changing this. :-p

libpromises/var_expressions.c
kacf: Sounds good. Another thing to keep in mind regarding this is performance. Since the bucket number is fixed, the hash map has the potential to be quite slow when you get variable numbers in the millions. Might be worth autogenerating a test which has millions of randomly named variables and comparing the difference in running time. We know that some customers have a ridiculous number of variables.

At the very least take a look at the running time of a host and compare the same host with the running time on master. Solaris 11 is a good candidate for that, since it's only running a single VM ATM, so is relatively disturbance free.

tzz: I had one customer with over 100,000 variables (array elements) before I converted them to data containers.

jimis: @kacf performance is an issue, as well as memory usage since hash tables are created for each EvalContext now and 8192 is possibly too big. But I don't think this should block this pull request. It will leave in master anyway, and we ought to refactor our hash table (see [redmine#7966](https://dev.cfengine.com/issues/7966)).

jimis:
Removed the commit with the indentation changes, added a commit with the new asserts.

jimis:
trigger build

jimis:
It seems `StringHash()` is already being misused with non-power-of-2 `max`, in particular in the `shuffle` function.

jimis:
Fixed some more bugs, especially bounding the hash with correct bitmasks.

jimis:
trigger build

jimis:
trigger build

jimis:
Wow, all builds are green, can it be?

jimis:
@kacf Lots of commits added and some are changed in order to fix old bugs, please do a quick re-review again and merge if it looks good!

kacf:
Code looks good, and build is green!

Just a couple of things about commits:

* Remove changelog entry from 86e2e1f44fd (Testsuite now dereferences common executables before checking their existence.), it doesn't describe a behavior change in CFEngine (there are many other testsuite changes we have not included in the changelog).
* For 55dc336 (filestat(path, "linktarget") now follows non-absolute links and returns full path of target.) I would consider changing `Changelog: Title` to `Changelog: Commit`, simply because the message is so good! :-) All the information can be useful in the changelog.

jimis:
Takk for review @kacf, messages changed in the two commits!

PR #2514, opened 2016-03-03
Allow nested container and slist function calls.

tzz:
This change allows CFEngine functions that return slists or data containers to be used as arguments to functions that expect slists or data containers. Since this work follows https://dev.cfengine.com/issues/7871 I marked some functions (but not all) as `COLLECTING` to indicate that they can accept data containers and slists this way.

The acceptance test shows the intended usage; the important part is quoted below.

All the pieces work as expected in my testing. This is a huge usability win for CFEngine users in my opinion.

```
bundle agent test
{
  vars:
      "base_list" slist => { "a", "b", "c", "aaa", "bbb", "ccc" };
      "base_data" data => '{ "x": 100, "y": "z"  }';
      "mapdata_spec" string => "<<< $(this.k)=$(this.v) >>>";

      "sort_mapdata_inline_json" slist => sort(mapdata('none', $(mapdata_spec), '[1,2,3]'), lex);
      "sort_mapdata_slist" slist => sort(mapdata('none', $(mapdata_spec), base_list), lex);
      "sort_mapdata_data" slist => sort(mapdata('none', $(mapdata_spec), base_data), lex);

      "sort_mapdata_mergedata_inline_json" slist => sort(mapdata('none', $(mapdata_spec), '[1,2,3]'), lex);
      "sort_mapdata_mergedata_slist" slist => sort(mapdata('none', $(mapdata_spec), mergedata(base_list)), lex);
      "sort_mapdata_mergedata_data" slist => sort(mapdata('none', $(mapdata_spec), mergedata(base_data)), lex);

      # we can nest functions that return slists too!
      "sort_mapdata_sort_slist" slist => sort(mapdata('none', $(mapdata_spec), sort(base_list, int)), lex);
      "sort_mapdata_mergedata_sort_slist" slist => sort(mapdata('none', $(mapdata_spec), mergedata(sort(base_list, int))), lex);
      "reverse_sort_slist" slist => reverse(sort(base_list, int));
}
```

### DIFF
--- a/examples/appgroups.cf
+++ b/examples/appgroups.cf
@@ -100,7 +100,7 @@ bundle agent appgroups(name, g)
       "$(cname)_grouplist[$(groups_$(apps))]" string => "1",
       ifvarclass => "have_app_$(apps)";
 
-      "$(cname)_grouplist" slist => getvalues("$(cname)_grouplist");
+      "$(cname)_grouplist_slist" slist => getvalues("$(cname)_grouplist");
 
       # get the keys of the array and sort them, then join with spaces
       "$(cname)_list" slist => getindices("$(cname)_grouplist");
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -7789,7 +7789,7 @@ static const FnCallArg EXPANDRANGE_ARGS[] =
 static const FnCallArg MAPARRAY_ARGS[] =
 {
     {CF_ANYSTRING, CF_DATA_TYPE_STRING, "Pattern based on $(this.k) and $(this.v) as original text"},
-    {CF_IDRANGE, CF_DATA_TYPE_STRING, "CFEngine array or data container identifier, the array variable to map"},
+    {CF_ANYSTRING, CF_DATA_TYPE_STRING, "CFEngine variable identifier or inline JSON, the array variable to map"},
     {NULL, CF_DATA_TYPE_NONE, NULL}
 };
 
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -2083,64 +2083,6 @@ static FnCallResult FnCallReadTcp(ARG_UNUSED EvalContext *ctx,
 
 /*********************************************************************/
 
-static FnCallResult FnCallRegList(EvalContext *ctx, ARG_UNUSED const Policy *policy, ARG_UNUSED const FnCall *fp, const Rlist *finalargs)
-{
-    const char *listvar = RlistScalarValue(finalargs);
-
-    if (!IsVarList(listvar))
-    {
-        Log(LOG_LEVEL_VERBOSE, "Function reglist was promised a list called '%s' but this was not found", listvar);
-        return FnFailure();
-    }
-
-    char naked[CF_MAXVARSIZE] = "";
-    GetNaked(naked, listvar);
-
-    VarRef *ref = VarRefParse(naked);
-
-    DataType value_type = CF_DATA_TYPE_NONE;
-    const Rlist *value = EvalContextVariableGet(ctx, ref, &value_type);
-    VarRefDestroy(ref);
-
-    if (!value)
-    {
-        Log(LOG_LEVEL_VERBOSE, "Function REGLIST was promised a list called '%s' but this was not found", listvar);
-        return FnFailure();
-    }
-
-    if (DataTypeToRvalType(value_type) != RVAL_TYPE_LIST)
-    {
-        Log(LOG_LEVEL_VERBOSE, "Function reglist was promised a list called '%s' but this variable is not a list",
-            listvar);
-        return FnFailure();
-    }
-
-    pcre *rx = CompileRegex(RlistScalarValue(finalargs->next));
-    if (!rx)
-    {
-        return FnFailure();
-    }
-
-    for (const Rlist *rp = value; rp != NULL; rp = rp->next)
-    {
-        if (strcmp(RlistScalarValue(rp), CF_NULL_VALUE) == 0)
-        {
-            continue;
-        }
-
-        if (StringMatchFullWithPrecompiledRegex(rx, RlistScalarValue(rp)))
-        {
-            pcre_free(rx);
-            return FnReturnContext(true);
-        }
-    }
-
-    pcre_free(rx);
-    return FnReturnContext(false);
-}
-
-/*********************************************************************/
-
 static FnCallResult FnCallRegArray(EvalContext *ctx, ARG_UNUSED const Policy *policy, ARG_UNUSED const FnCall *fp, const Rlist *finalargs)
 {
     char *arrayname = RlistScalarValue(finalargs);
@@ -2357,6 +2299,19 @@ static FnCallResult FnCallGrep(EvalContext *ctx, ARG_UNUSED const Policy *policy
                           1, // regex match = TRUE
                           0, // invert matches = FALSE
                           LONG_MAX); // max results = max int
+}
+
+/*********************************************************************/
+
+static FnCallResult FnCallRegList(EvalContext *ctx, ARG_UNUSED const Policy *policy, const FnCall *fp, const Rlist *finalargs)
+{
+    return FilterInternal(ctx,
+                          fp,
+                          RlistScalarValue(finalargs->next), // regex or string
+                          finalargs, // list identifier
+                          1,
+                          0,
+                          LONG_MAX);
 }
 
 /*********************************************************************/
@@ -3926,6 +3881,11 @@ static FnCallResult FilterInternal(EvalContext *ctx,
         ret = (match_count == 0);
     }
     else if (0 == strcmp(fp->name, "some"))
+    {
+        contextmode = 1;
+        ret = (match_count > 0);
+    }
+    else if (0 == strcmp(fp->name, "reglist"))
     {
         contextmode = 1;
         ret = (match_count > 0);
@@ -8026,7 +7986,7 @@ static const FnCallArg REGLINE_ARGS[] =
 
 static const FnCallArg REGLIST_ARGS[] =
 {
-    {CF_NAKEDLRANGE, CF_DATA_TYPE_STRING, "CFEngine list identifier"},
+    {CF_ANYSTRING, CF_DATA_TYPE_STRING, "CFEngine variable identifier or inline JSON"},
     {CF_ANYSTRING, CF_DATA_TYPE_STRING, "Regular expression"},
     {NULL, CF_DATA_TYPE_NONE, NULL}
 };
@@ -8537,7 +8497,7 @@ const FnCallType CF_FNCALL_TYPES[] =
     FnCallTypeNew("regline", CF_DATA_TYPE_CONTEXT, REGLINE_ARGS, &FnCallRegLine, "True if the regular expression in arg1 matches a line in file arg2",
                   FNCALL_OPTION_NONE, FNCALL_CATEGORY_IO, SYNTAX_STATUS_NORMAL),
     FnCallTypeNew("reglist", CF_DATA_TYPE_CONTEXT, REGLIST_ARGS, &FnCallRegList, "True if the regular expression in arg2 matches any item in the list whose id is arg1",
-                  FNCALL_OPTION_NONE, FNCALL_CATEGORY_DATA, SYNTAX_STATUS_NORMAL),
+                  FNCALL_OPTION_COLLECTING, FNCALL_CATEGORY_DATA, SYNTAX_STATUS_NORMAL),
     FnCallTypeNew("regldap", CF_DATA_TYPE_CONTEXT, REGLDAP_ARGS, &FnCallRegLDAP, "True if the regular expression in arg6 matches a value item in an ldap search",
                   FNCALL_OPTION_CACHED, FNCALL_CATEGORY_COMM, SYNTAX_STATUS_NORMAL),
     FnCallTypeNew("remotescalar", CF_DATA_TYPE_STRING, REMOTESCALAR_ARGS, &FnCallRemoteScalar, "Read a scalar value from a remote cfengine server",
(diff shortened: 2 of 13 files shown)


jimis:
Does that mean that I can compare two JSON outputs just by wrapping `mapdata()` with `sort()`? How is sorting working with nested data?

tzz:
It depends on the relevant functions. `sort()` for instance will only sort the top level scalars in whatever it gets, while `data_expand()` will descend into the container.

Specifically to answer your question, since `mapdata()` returns a data container, the order that `sort()` applies to it will depend on the contents. So `mapdata("none", "foo", '[1,2,3]')` produces `["foo", "foo", "foo"]` which `sort()` can handle. But `mapdata("json", '{ "x": 100 }', '[1,2,3]')` produces `[{ "x": 100 }, { "x": 100 }, { "x": 100 }]` which `sort()` can't handle currently. The relevant code in `FnCallSort` that gathers the values to be sorted is:

    JsonIterator iter = JsonIteratorInit(json);
    const JsonElement *e;
    while ((e = JsonIteratorNextValueByType(&iter, JSON_ELEMENT_TYPE_PRIMITIVE, true)))
    {
        RlistAppendScalar(&sorted, JsonPrimitiveGetAsString(e));
    }

Clearly there's more work to be done to improve the various functions to handle complex data, but this PR is a necessary first step.

tzz:
This will make https://dev.cfengine.com/issues/7963 much easier.

nickanderson:
Yes please +1

tzz:
@nickanderson please think up as many tests as you can. The current tests try to catch all the possible nesting combinations but more variety is always good.

libpromises/evalfunction.c
```diff
@@ -2202,7 +2212,7 @@ static FnCallResult FnCallGetIndicesClassic(EvalContext *ctx, ARG_UNUSED const P
 
 static FnCallResult FnCallGetIndices(EvalContext *ctx, ARG_UNUSED const Policy *policy, const FnCall *fp, const Rlist *finalargs)
 {
-    const char *name_str = RlistScalarValue(finalargs);
+    const char *name_str = RlistScalarValueSafe(finalargs);
```
kacf: `RlistScalarValueSafe()` seems like it should be used in cases where we will print the resulting string (`[not printable]`), but here it's used for actual variable lookup, which seems wrong to me. A more correct approach would be to not perform the variable lookup at all if we don't expect an actual variable, don't you agree?

kacf: Perhaps a shortform for `rlist->val.type == RVAL_TYPE_SCALAR` would be appropriate, like `RlistIsScalar()` or something like that.

This might be appropriate some of the other places where `RlistScalarValueSafe()` is used as well.

kacf:
I will need to continue this one later. I got about 1/3 through it.

libpromises/evalfunction.c
tzz: I've tried to use `RlistScalarValueSafe()` only where we use the `name_str` later. In this specific case, it would complicate the code because of the requirement to call the "classic" way to get indices. I think this is the only place where this happens, so we can adjust it here if you want.

My ultimate goal was to change as little as possible in the code, so if you prefer a longer code flow that explicitly branches on the rlist item type, that's fine, I'll make the change, but it will be a **lot** of code. At that point we could even look at pre-evaluating the data references around function calls, since the boilerplate is very repetitive. But I would prefer to do all that in a cleanup PR.

kacf: We will need to check for the type, yes, because calling `ResolveAndQualifyVarName()` with `name_str`, hence `[not printable]`, is definitely wrong. It may seem innocent, but it's a bug waiting to happen later.

But you're right, there are few places where this is a problem (only two that I can see, I'll mark the other one), so we only need to change it in those two places. The code change should not be that huge.

libpromises/evalfunction.c
```diff
@@ -6637,33 +6598,62 @@ static FnCallResult FnCallMakerule(EvalContext *ctx, ARG_UNUSED const Policy *po
...
         }
 
         target_time = statbuf.st_mtime;
     }
 
-    // For each file in sources, check they exist and are older than target
-
-    for (const Rlist *rp = list; rp != NULL; rp = rp->next)
+    // Check if the file name is explicit
+    if (lstat(name_str, &statbuf) != -1)
```
kacf: `name_str` may be `[not printable]` here, which could in fact be a real file. This should be changed to account for non-scalar types. I'm not even sure what this function should do if the value is not scalar. Maybe just bail out?

libpromises/evalfunction.c
```diff
@@ -2917,9 +2928,10 @@ static FnCallResult FnCallMergeData(EvalContext *ctx, ARG_UNUSED const Policy *p
 
     for (const Rlist *arg = args; arg; arg = arg->next)
     {
-        if (args->val.type != RVAL_TYPE_SCALAR)
+        if (args->val.type != RVAL_TYPE_SCALAR &&
+            args->val.type != RVAL_TYPE_CONTAINER)
         {
-            Log(LOG_LEVEL_ERR, "%s: argument '%s' is not a variable reference", fp->name, RlistScalarValue(arg));
+            Log(LOG_LEVEL_ERR, "%s: argument '%s' is not a variable reference", fp->name, RlistScalarValueSafe(arg));
```
kacf: I think the `RlistScalarValueSafe()` here is guaranteed to be `[not printable]`, due to the check above. Might as well be removed.

libpromises/evalfunction.c
kacf: > I'm not even sure what this function should do if the value is not scalar. Maybe just bail out?

Eh, no, that's the case below, obviously. So the only extra thing needed here is to check for scalar type (scalarity?) before the `lstat()`.

libpromises/fncall.c
```diff
@@ -244,10 +248,9 @@ static FnCallResult CallFunction(EvalContext *ctx, const Policy *policy, const F
         rp = rp->next;
     }
 
-    char output[CF_BUFSIZE];
     if (argnum != RlistLen(expargs) && !(fncall_type->options & FNCALL_OPTION_VARARG))
     {
-        snprintf(output, CF_BUFSIZE, "Argument template mismatch handling function %s(", fp->name);
+        fprintf(stderr, "Argument template mismatch handling function %s(", fp->name);
```
kacf: Lol! This went nowhere before? :-)

libpromises/fncall.h
```diff
@@ -58,7 +58,8 @@ typedef enum
 {
     FNCALL_OPTION_NONE = 0,
     FNCALL_OPTION_VARARG = 1 << 0,
-    FNCALL_OPTION_CACHED = 1 << 1
+    FNCALL_OPTION_CACHED = 1 << 1,
+    FNCALL_OPTION_COLLECTING = 1 << 2
```
kacf: Can you add a comment here about what the new option does? The name alone does not (and probably cannot) fully explain it.

libpromises/rlist.c
```diff
@@ -527,6 +542,17 @@ Rlist *RlistAppend(Rlist **start, const void *item, RvalType type)
...
+        {
+            JsonElement* store = JsonArrayCreate(RlistLen(item));
+            for (const Rlist *rp = item; rp; rp = rp->next)
+            {
+                JsonArrayAppendElement(store, RvalToJson(rp->val));
+            }
+
+            return RlistAppendRval(start, (Rval) { store, RVAL_TYPE_CONTAINER });
+        }
+
```
kacf: I'm a little confused by this one: If I understand correctly, the old code (`!allow_all_types`) appends element by element, whereas this appends only one element, a JSON structure. So `!allow_all_types` with two lists, `[1, 2, 3]` and `[4, 5, 6]`, would give `[1, 2, 3, 4, 5, 6]`, whereas `allow_all_types` would give `[1, 2, 3, [4, 5, 6]]`. Is this correct? If so, it's a little unexpected I think. Can you provide a bit more background for this change?

tests/acceptance/02_classes/02_functions/makerule.cf
```diff
@@ -55,13 +55,13 @@ bundle agent check
...
-      "ok5" expression => makerule("$(G.testfile)", "@(test.lold)");
+      "ok5" expression => makerule("$(G.testfile)", "test.lold");
 
       # An old file does need to be rebuilt from an newer file in a list
-      "ok6" expression => makerule("$(G.etc_passwd)", "@(test.lnew)");
+      "ok6" expression => makerule("$(G.etc_passwd)", "test.lnew");
 
       # An old file does not need to be rebuilt from sources with a missing file (fails)
-      "ok7" expression => makerule("$(G.etc_passwd)", "@(test.lnon)");
+      "ok7" expression => makerule("$(G.etc_passwd)", "test.lnon");
```
kacf: It's unfortunate that we didn't catch this inconsistency earlier, but I agree we should change it, even if it breaks backwards compatibility. This definitely deserves mentioning in the docs though, in addition to the changelog.

kacf:
Done reviewing!

libpromises/fncall.c
jimis: 1. why `fprintf()` and not `Log()`?
2. is the ending closing parenthesis in the string a typo?

libpromises/evalfunction.c
tzz: OK, fixed.

libpromises/evalfunction.c
tzz: Right, removed, thanks.

libpromises/evalfunction.c
tzz: OK, done.

libpromises/fncall.c
tzz: I followed the intent of the previous code. I suggest merging this as is and then fixing it as the developers see fit. Clearly it wasn't a big deal before, so the fix can be delayed until this PR is merged.

libpromises/fncall.h
tzz: OK, done.

libpromises/rlist.c
tzz: The old code doesn't collect Rlist elements. So you call `and(splitstring("a b"))` and it ends up as `and("a", "b")`. This expansion happens once per FnCall, not for all arguments.

With the new code, you can optionally (for `COLLECTING` functions) collect all the results of a FnCall into a single JSON array object. It requires functions to expect it, but it's the only reasonable way to preserve backwards compatibility for functions like `and()` in the example above while nesting general calls to functions that take and return compound data types.

tests/acceptance/02_classes/02_functions/makerule.cf
tzz: Yes, definitely, I'll do so once this is accepted. Fixup commit coming your way for final review.

tzz:
Rebased and ready for final review. @jimis please allow us to delay the `Log()` fix in `fncall.c` until this is merged.

libpromises/rlist.c
kacf: Ok, please make a comment that it behaves this way above the function definition.

kacf:
Thanks!

kacf:
trigger build

libpromises/rlist.c
tzz: OK, comments added to `rlist.c` and `fncall.c`.

kacf:
Great work, the build is green! All that remains is docs.

tzz:
@kacf docs in https://github.com/cfengine/documentation/pull/1342

examples/appgroups.cf
```diff
@@ -100,7 +100,7 @@ bundle agent appgroups(name, g)
       "$(cname)_grouplist[$(groups_$(apps))]" string => "1",
       ifvarclass => "have_app_$(apps)";
 
-      "$(cname)_grouplist" slist => getvalues("$(cname)_grouplist");
+      "$(cname)_grouplist_slist" slist => getvalues("$(cname)_grouplist");
```
tzz: Bugfix for pre-existing test issue.

libpromises/evalfunction.c
```diff
@@ -173,7 +173,18 @@ static FnCallResult FnFailure(void)
 
 static VarRef* ResolveAndQualifyVarName(const FnCall *fp, const char *varname)
 {
-    VarRef *ref = VarRefParse(varname);
+    VarRef *ref = NULL;
+    if (NULL != varname && IsVarList(varname))
+    {
+        char naked[CF_MAXVARSIZE] = "";
+        GetNaked(naked, varname);
```
tzz: @kacf this allows `makerule()` to be backwards compatible and allows all collecting functions to take `@(varname)`.

tzz: The documentation patch will have to be slightly updated if you agree with this change.

libpromises/evalfunction.c
```diff
@@ -238,7 +249,7 @@ static JsonElement* VarRefValueToJson(EvalContext *ctx, const FnCall *fp, const
             break;
 
         case RVAL_TYPE_CONTAINER:
-            convert = value;
+            convert = (JsonElement*) value;
```
jimis: You const-away the constness of `value`. The compiler though still thinks that `value` is const, and may optimise things the way you are not expecting. And `value` is correct to be const, because you must not change the contents of the Variables table like that.

The easy solution here is to duplicate const to a second local variable. Cost: extra allocation.
If I was writing it from scratch I would have used 2 variables. One const variable containing the unmodifiable value from the variables table, and one non-const variable containing the allocated value.

In general never cast constness off. There are very few exceptions.

tzz: All right, I'll adjust it here.

tzz:
@kacf I also converted `reglist()` to be a collecting function (docs updated too), which I think leaves only `regarray()` unconverted. Should I do that one too or leave it for later?

libpromises/evalfunction.c
tzz: Done, thanks @jimis 

I hope we find a way to optimize this later. I added a TODO note.

kacf:
> I also converted reglist() to be a collecting function (docs updated too), which I think leaves only regarray() unconverted. Should I do that one too or leave it for later?

Leave it for another PR and let's get this one through.

There was a merge conflict because Jimis cherry-picked some fixes. I resubmitted here for a final build: https://github.com/cfengine/core/pull/2524

kacf:
Thanks, merged manually in https://github.com/cfengine/core/commit/4584fd99829b6b805d2ff7b25e74d6a9bf5dfdd8.

PR #2515, opened 2016-03-07
CFE-2084: load multiple augments from "augments" string array in def.json

tzz:
See https://tracker.mender.io/browse/CFE-2084 (https://dev.cfengine.com/issues/7453)

Really simple logic currently:
- `augments` key contents (which must be an array of strings) are loaded. They can contain system variables like `sys.inputdir` (see acceptance test).
- `vars` and `classes` are merged because they are processed one by one
- `inputs` override other inputs

### DIFF
--- a/libpromises/generic_agent.c
+++ b/libpromises/generic_agent.c
@@ -83,6 +83,8 @@ static char* ReadReleaseIdFromReleaseIdFileMasterfiles(const char *maybe_dirname
 
 static bool MissingInputFile(const char *input_file);
 
+bool LoadAugmentsFiles(EvalContext *ctx, const char* filename);
+
 #if !defined(__MINGW32__)
 static void OpenLog(int facility);
 #endif
@@ -196,8 +198,10 @@ bool CheckContextOrClassmatch(EvalContext *ctx, const char* c)
     return IsDefinedClass(ctx, c);
 }
 
-void LoadAugmentsData(EvalContext *ctx, const Buffer* filename_buffer, const JsonElement* augment)
+bool LoadAugmentsData(EvalContext *ctx, const Buffer* filename_buffer, const JsonElement* augment)
 {
+    bool loaded = false;
+
     if (JsonGetElementType(augment) != JSON_ELEMENT_TYPE_CONTAINER ||
         JsonGetContainerType(augment) != JSON_CONTAINER_TYPE_OBJECT)
     {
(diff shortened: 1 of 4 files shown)


nickanderson:
@estenberg comments on this? Would be nice to get this into 3.9 so we have good soak time before the next LTS.

nickanderson:
I know this was implemented intially with a very simple merge stragety, but I thought that there was going to be a way to select the merge strategy for each augmented file (for future use). I don't see how that would be done here or in the ticket currently. Maybe I missed something.

tzz:
@nickanderson currently you specify the file as `"myfile.json"` which applies the simple overwrite merge strategy.

In future work, we'll allow the entry to also be in the form `{ file: "myfile.json", merge: ...whatever... }` but I don't think we know yet what merge strategies are needed. Maybe you can open a ticket for that spec?

nickanderson:
@tzz what happens if you dont supply any full path to the additional augmenter?
for example instead of 

```
 "augments": [ "$(sys.inputdir)/augmenter.json" ] 
```

I had:

```
 "augments": [ "augmenter.json" ] 
```

Will it error? Will it find it?

tzz:
@nickanderson the built-in CFEngine file handling functions will decide if the filename is OK, the parser doesn't check.

tzz:
This is almost 1 year old! Woo hoo!

nickanderson:
@jimis ping

nickanderson:
Happy belated birthday #2515

tzz:
Age is nothing but a number.

jimis:
A painful number, I would add.

@nickanderson @estenberg if you approve the functionality, let me know to do the code review and merge, this is not a complex pull request. But I can't comment whether having recursive loading of augments is something needed.

@tzz sorry you had to close this after more than a year of waiting. Maybe we'll have more people tending to the community pull requests soon(TM)... ;-)

basvandervlies:
@tzz it is a pity that it is closed. I cloud use the mutiple augements file in the cf_surfsaral_lib framework

nickanderson:
I still think this is a valuable feature.

tzz:
The branch is still available if anyone wants to have fun with it. I have no problem with someone else taking over :)

estenberg:
Yes, this makes sense. Useful if you have class or host-specific data that should override defaults.

nickanderson:
@jimis if this is in a good state for merging perhaps we should merge it.

We can create a new issue for implementing alternative merge strategies.

jimis:
ACK I opened #2964.

PR #2517, opened 2016-03-08
Add an option to generate full policy

peckpeck:
Currently the "cf-promises -p json" command only generate partial promises (the file being parsed), whereas the "cf-promises -p cf" command generates full promises (the file and all its dependencies).

This is an option to control that behavior.
I retained the "json" behavior by default since "cf" did not work properly before.

### DIFF
--- a/libpromises/generic_agent.h
+++ b/libpromises/generic_agent.h
@@ -33,6 +33,15 @@
 #define GENERIC_AGENT_CHECKSUM_SIZE ((2*CF_SHA1_LEN) + 1)
 #define GENERIC_AGENT_CHECKSUM_METHOD HASH_METHOD_SHA1
 
+enum generic_agent_config_common_policy_output_format
+{
+    GENERIC_AGENT_CONFIG_COMMON_POLICY_OUTPUT_FORMAT_NONE,
+    GENERIC_AGENT_CONFIG_COMMON_POLICY_OUTPUT_FORMAT_CF,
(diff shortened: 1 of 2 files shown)


libpromises/generic_agent.h
```diff
@@ -33,6 +33,15 @@
 #define GENERIC_AGENT_CHECKSUM_SIZE ((2*CF_SHA1_LEN) + 1)
 #define GENERIC_AGENT_CHECKSUM_METHOD HASH_METHOD_SHA1
 
+enum output_format
```
kacf: This will be more or less globally visible, so it should match the value names and be more specific. Although frightfully long, can you rename it to the same name as the enum values below, IOW `generic_agent_config_common_policy_output_format`?

kacf:
Looks nice! A couple of things:

* Can you change the commit message to `Add a command line option to generate full policy`, and add `Changelog: Title`?
* This really could use some tests, and I don't think they will be that hard to write, you just need to call `cf-promises` with each of the four options on a dummy policy containing at least two files, and then verify the output. Can I ask you to include these?

peckpeck:
I can't find any existing test with a call to a cf-promise binary or with a specific parameter.
How do you think it should be written ?

peckpeck:
PR updated

kacf:
Thanks, I'll take a look as soon as I can!

kacf:
Thanks, superseded by #2547.

PR #2518, opened 2016-03-08
Fix unintended UnexpectedError.

jimis:


jimis:
All green locally, and commit is platform and enterprise agnostic, merging.

PR #2520, opened 2016-03-09
Add function cache hits, misses, and size to timing instrumentation output

tzz:
@jimis this may be useful to measure function cache performance. It's a pretty trivial change.

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -1469,7 +1469,7 @@ static PromiseResult KeepAgentPromise(EvalContext *ctx, const Promise *pp, ARG_U
         result = VerifyProcessesPromise(ctx, pp);
         if (result != PROMISE_RESULT_SKIPPED)
         {
-            EndMeasurePromise(start, pp);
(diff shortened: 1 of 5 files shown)


jimis:
+1 it's a good addition. But somehow the `EndMeasurePromise()` call seems out of place for something inside `ctx`, because the counter is not zeroed in `BeginMeasure()`.

Hmmm what if you printed this information in each `EvalContextDestroy()` and `EvalContextClear()`? Is the output too much then? I would also argue that this is for DEBUG level, since the function cache is implementation details.

tzz:
@jimis this is for developers so yeah, you can move it anywhere you like and change to `DEBUG`. Can you take over this and just close the PR? It's probably more efficient at this point, since you know what you want it to do :)

jimis:
I merged it and test-run this pull request, but the output was not really useful, the hits/misses were really few which seemed particularly weird.. I'm closing this, if we need to measure I believe a bit more complex mechanism is needed, not really certain at this point but this seems to be missing things.

PR #2521, opened 2016-03-09
Don't follow symbolic links when copying extended attributes.

kacf:
We want to copy the attributes of the source link to the destination link.
Seen when using copy_from.preserve => "true".

Changelog: Title

Based on a patch by:
Eystein Måløy Stenberg <eystein.maloy.stenberg@cfengine.com>


kacf:
trigger build

kacf:
Cherry-picked to 3.7.x and 3.8.x. The patch did not apply to 3.6.x.

PR #2523, opened 2016-03-10
Examples tests not running redmine7975

jimis:


jimis:
Do not merge yet, two examples still failing:
* maplist.cf
* appgroups.cf

libpromises/evalfunction.c
```diff
@@ -238,8 +238,8 @@ static JsonElement* VarRefValueToJson(EvalContext *ctx, const FnCall *fp, const
             break;
 
         case RVAL_TYPE_CONTAINER:
-            convert = value;
-            *allocated = false;
+            convert = JsonCopy(value);
```
tzz: The title of the commit "Avoid de-constifying assignment, fixes compiler warning." doesn't tell me much, unfortunately.

jimis: Thanks for reviewing @tzz, I was ready to ping you. :-)

So this line always spewed a compiler warning:
```
evalfunction.c:241:21: warning: assignment discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
             convert = value;
                     ^
```

jimis: I tried to avoid complete refactor, thus the one extra allocation.

jimis: But this change basically reverts small part of commit da9a9fc8ff4c0d65520b07c03077136a4947c4a2 that you have committed @tzz. I am not sure why this had to be changed with that commit, still reading it.

jimis: To make clear what I mean, the `value` variable is const and can't be changed (and that's correct, `value` should be const) but it is assigned to `convert` which is not const.

jimis:
See redmine#7975, I have verified that the commit causing the failure is da9a9fc8ff4c0d65520b07c03077136a4947c4a2, before that the two examples `maplist.cf` and `appgroups.cf` were passing, after that they are failing.

@tzz @kacf any ideas?

libpromises/evalfunction.c
tzz: OK, think about it: `convert` could be a new data element we've created from inline JSON or as a copy, **or** it could be a const reference (in which case `allocated` will be false).

In https://github.com/cfengine/core/pull/2514 I actually already fixed this warning by casting `value` to `JsonElement*`. I think that's the right fix.

tzz:
`examples/maplist.cf` is fixed with https://github.com/tzz/core/commit/00f5d900400f1666e15279c476dd89ade02c8d54 (this also addresses the backwards compatibility concerns with `makerule()` because it now accepts `@(varname)` for all collecting functions so @kacf should note it).

`examples/appgroups.cf` is fixed with https://github.com/tzz/core/commit/d639a95664d85ff93304e53bc4b7685a5bc15c09 (the array and the slist were named the same).

I put both those commits in https://github.com/cfengine/core/pull/2514 because it will be merged soon, but you can cherry-pick them if you wish.

libpromises/evalfunction.c
jimis: > In #2514 I actually already fixed this warning by casting value to JsonElement*. I think that's the right fix.

Will comment there.

jimis:
Thanks Ted, I'll cherry-pick your fixes and try to put them in separately.

tzz:
It may be easier to wait for #2514 to be merged, @kacf was just waiting for the docs, and today's amends are pretty minor :)

jimis:
trigger build dammit!

jimis:
mostly green, merging, thanks TedZ!

tzz:
woo hoo!

PR #2527, opened 2016-03-15
Convert regarray() to use FnCallRegList() and clean up function docstrings

tzz:
This is the remaining cleanup for https://dev.cfengine.com/issues/7871

* convert all function docstrings that take collected lists, arrays, or data containers, or inline JSON to say so
* convert `regarray()` to use `FnCallRegList()`

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -1302,7 +1302,7 @@ static FnCallResult FnCallBundlesMatching(EvalContext *ctx, const Policy *policy
 
 static bool AddPackagesMatchingJsonLine(pcre *matcher, JsonElement *json, char *line)
 {
-    
+
     if (strlen(line) > CF_BUFSIZE - 80)
     {
         Log(LOG_LEVEL_ERR,
@@ -1310,7 +1310,7 @@ static bool AddPackagesMatchingJsonLine(pcre *matcher, JsonElement *json, char *
             strlen(line));
         return false;
     }
-    
+
 
     if (StringMatchFullWithPrecompiledRegex(matcher, line))
     {
@@ -1329,7 +1329,7 @@ static bool AddPackagesMatchingJsonLine(pcre *matcher, JsonElement *json, char *
         JsonObjectAppendString(line_obj, "version", SeqAt(list, 1));
         JsonObjectAppendString(line_obj, "arch",    SeqAt(list, 2));
         JsonObjectAppendString(line_obj, "method",  SeqAt(list, 3));
-                
+
         SeqDestroy(list);
         JsonArrayAppendObject(json, line_obj);
     }
@@ -1350,7 +1350,7 @@ static bool GetLegacyPackagesMatching(pcre *matcher, JsonElement *json, const bo
     }
 
     Log(LOG_LEVEL_DEBUG, "Reading inventory from '%s'", filename);
-    
+
     FILE *const fin = fopen(filename, "r");
     if (fin == NULL)
     {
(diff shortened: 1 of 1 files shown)


tzz:
Rebased and ready for review.

kacf:
I *love* removal of code!

libpromises/evalfunction.c
```diff
@@ -8607,7 +8573,7 @@ const FnCallType CF_FNCALL_TYPES[] =
                   FNCALL_OPTION_NONE, FNCALL_CATEGORY_COMM, SYNTAX_STATUS_NORMAL),
     FnCallTypeNew("datastate", CF_DATA_TYPE_CONTAINER, DATASTATE_ARGS, &FnCallDatastate, "Construct a container of the variable and class state",
                   FNCALL_OPTION_NONE, FNCALL_CATEGORY_UTILS, SYNTAX_STATUS_NORMAL),
-    FnCallTypeNew("difference", CF_DATA_TYPE_STRING_LIST, SETOP_ARGS, &FnCallSetop, "Returns all the unique elements of list arg1 that are not in list arg2",
+    FnCallTypeNew("difference", CF_DATA_TYPE_STRING_LIST, SETOP_ARGS, &FnCallSetop, "Returns all the unique elements of collected list or array or data container arg1 that are not in collected list or array or data container arg2",
```
kacf: I think that using the word "collected" here doesn't really help to understand what it does. In fact, now that most functions *are* collecting, I've been wondering if we should refrain from using that term at all. As a user, the "collecting" behavior is something I would expect from functions intuitively, so I don't need it specified. What do you think?

Of course, the reference to lists, arrays and containers should stay.

This goes for all the references below as well.

tzz: We need something to indicate that the behavior is special compared to 3.8 and older. I'm happy to change the term, but remember that the docs also use "collecting functions" now so they would also have to be adjusted.

kacf: Yes, but that's what I mean. This is becoming the common case now, and definitely should be mentioned in the changelog, for instance. But maybe it doesn't make sense to mention it for every function, since the default is that it *does* support collection. And yes, this would involve changing the docs as well, perhaps inverting the concept, so that only functions that don't support collection are mentioned (but are there any, after this?).

tzz: Yes, there are some functions that don't collect, like `and`, `or`, and a few others. I'll do as you suggest.

Are you asking me to also adjust the docs or only the function docstrings in this commit?

kacf: If you can, yes! :-)

tzz: OK, rebased and updated. I'll do a doc update too.

tzz:
OK, I also updated the docs in https://github.com/cfengine/documentation/pull/1454

Please take over if you want further adjustments to the docs, I made the lines pretty easy to search and replace across all the function docs in case they're still not what you want.

kacf:
Sure!

kacf:
Looks great the way it is!

kacf:
trigger build

kacf:
Alright, this is green. I modified the commit message to provide a more complete changelog, and I assembled all the function names I could see here, the previous commits, and the ticket. Could you verify that the list [here](https://github.com/cfengine/core/pull/2594/commits) looks good to you?

tzz:
@kacf yeah it looks all right to me

kacf:
Alright, thanks, it's in!

tzz:
:flags:

PR #2528, opened 2016-03-15
3.6.7 backports

kacf:


cf-serverd/server_common.c
```diff
@@ -1736,37 +1736,42 @@ void SetConnIdentity(ServerConnectionState *conn, const char *username)
...
+    if (!NovaWin_UserNameToSid(conn->username, (SID *) conn->sid,
+                               CF_MAXSIDSIZE, !is_root))
+    {
+        memset(conn->sid, 0, CF_MAXSIDSIZE);  /* is invalid sid - discarded */
+    }
 
-    if (ThreadLock(&pwnam_mtx))
+#else                                                 /* UNIX - common path */
+
+    if (conn->uid == CF_UNKNOWN_OWNER)      /* skip looking up UID for root */
```
jimis: This is a performance optimisation for hubs, maybe we should leave it out of 3.6.x

libpromises/evalfunction.c
```diff
@@ -5461,6 +5461,8 @@ static FnCallResult FnCallReadFile(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const
     }
     else
     {
+        Log(LOG_LEVEL_ERR, "Function '%s' failed to read file: %s",
+            fp->name, filename);
         return FnFailure();
```
jimis: I would also skip this since it's changing behaviour, unless somebody requested it. Hmmm after looking at redmine#7702 it seems this even broke expected behaviour in 3.7.2, so we might even revert it there.

pasinskim: I have no strong opinion, but only

>  Redmine #7460: Unable to edit the same value in multiple regions. 

looks important here. Other stuff is changing the behavior as well so I am not sure if we should add it in 3.6. @jimis what do you think about filesafe generation and interactive tty?

PR #2529, opened 2016-03-15
Allow iprange() to take an IP as a parameter

tzz:
As requested in https://dev.cfengine.com/issues/7949

Allows `iprange("1.2.3.4/32", "1.2.3.4")` to check a specific IP instead of the implied host IP. Backwards compatible.

This fixes a bug as well: even if you narrowed by interface, the `VIPADDRESS` was checked. Now it's not.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -4869,7 +4869,26 @@ static FnCallResult FnCallIPRange(EvalContext *ctx, ARG_UNUSED const Policy *pol
         return FnFailure();
     }
 
-    /* Case 1: iprange(): No second argument, just arg1: check arg1 (a range) against VIPADDRESS */
+    /* Case 1: isipinsubnet(): check arg1 (a range) against arg2, arg3, ... if they are IPs */
+    if (!iprange_mode)
+    {
+        for (Rlist *ip = finalargs->next; ip != NULL; ip = ip->next)
+        {
+            Buffer *ipb = BufferNewFrom(RlistScalarValue(ip), strlen(RlistScalarValue(ip)));
+            bool is_ip = IPAddressIsIPAddress(ipb, NULL);
+            if (is_ip && FuzzySetMatch(range, BufferData(ipb)) == 0)
+            {
+                Log(LOG_LEVEL_DEBUG, "%s: found range %s on given IP %s", fp->name, range, BufferData(ipb));
+                BufferDestroy(ipb);
+                return FnReturnContextTrue();
+            }
+            BufferDestroy(ipb);
+        }
+    }
+
+    /* Case 2: iprange(): No second argument, just arg1: check arg1 (a range)
+     * against VIPADDRESS. Below we'll check all known IPs as well, this is just
+     * for VIPADDRESS. */
     if (iprange_mode && finalargs->next == NULL)
     {
         if (FuzzySetMatch(range, VIPADDRESS) == 0)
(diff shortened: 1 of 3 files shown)


nickanderson:
+1

tzz:
Rebased and ready for review

jimis:
I got quite confused with the existing functionality of `iprange()`. Basically now we can say `iprange(address_range, interface)`. This patch will give a second meaning to the second argument if called like `iprange(address_range, address)`, right?

jimis:
May I suggest a separate function for that, for example `match_ip_subnet()` or something that makes sense to whoever writes policy... :-p

tzz:
It makes sense to me: the second parameter is an IP address either way, but the second parameter can probe by the interface's address. It's like `grep foo FILE` that works whether `FILE` is an actual file or a symlink to another file. And most of the function code is shared between the two modes. But it can certainly be broken into another function if you and others feel strongly.

jimis:
​So
* `iprange(address_range)` checks this machine's IPs,
* `iprange(address_range, interface)` checks the interface's IP,
* `iprange(address_range, address)` checks the address passed as argument

Not sure if it's clear behaviour, I'll leave it to policy experts to decide.

tzz:
@nickanderson @kacf @estenberg any opinions? Nick said he was in favor earlier but jimis' point is valid, I don't want to be the only one that understands this feature :)

kacf:
I tend to agree with @jimis. It seems too magic that `iprange` should accept an argument that can have several meanings depending on what it contains. I would make a separate function for this.

tzz:
OK; rebased and new function added `isipinsubnet()`. Internally it uses `iprange()` but won't be documented that way, so I think it's the most economical way to get this done.

jimis:
trigger build

libpromises/evalfunction.c
```diff
@@ -4867,16 +4868,39 @@ static FnCallResult FnCallIPRange(EvalContext *ctx, ARG_UNUSED const Policy *pol
...
+        if (FuzzySetMatch(range, VIPADDRESS) == 0)
+        {
+            Log(LOG_LEVEL_DEBUG, "%s: found range %s on hostip %s, no interface", fp->name, range, VIPADDRESS);
+            return FnReturnContextTrue();
+        }
+
+        return FnReturnContextFalse();
+    }
+
+    // first try to match on the given arguments as a list of IPs
```
jimis: Where is the name function name checked for being `isipinsubnet`?

libpromises/evalfunction.c
```diff
@@ -4867,16 +4868,39 @@ static FnCallResult FnCallIPRange(EvalContext *ctx, ARG_UNUSED const Policy *pol
         return FnFailure();
     }
 
-    for (const Item *ip = EvalContextGetIpAddresses(ctx); ip != NULL; ip = ip->next)
+    // we match on VIPADDRESS iff no interfaces or IPs were requested
+    if (NULL == finalargs->next)
```
jimis: style

libpromises/evalfunction.c
jimis: It would be nice if the comments were briefly but clearly indicating the 3 different modes. For example:

```
/* Case 1: iprange(arg1) */
[...]
/* Case 2: iprange(arg1, arg2) */
[...]
/* Case 3: isipinsubnet(arg1, arg2) */
[...]
```

jimis:
`isipinsubnet()` is a bit unreadable for me, don't we have other functions with underscores? What do the policy writers think of other names? I suggested `match_ip_subnet()` but I can go with one word, just asking here.

libpromises/evalfunction.c
tzz: The two functions work exactly the same way so I didn't need to check the name. But if you feel strongly about it... let me know.

tzz:
I'll rename it to whatever you want. All other "isxyz" functions in the core have underscores. Maybe we can add underscored aliases for all the others, it's not hard, just extra entries in the function table. Or just for this one. Let me know.

libpromises/evalfunction.c
tzz: Ah, this was written just before you asked me not to use this style. I'll adjust when I change the function name.

tzz: Note `evalfunction.c` has 20+ instances of `NULL == something` so maybe it makes sense to fix them all at once.

libpromises/evalfunction.c
jimis: Yes, the two functions shouldn't work by accident in vice-versa mode.

libpromises/evalfunction.c
jimis: sure

jimis:
Lets keep the ugly ;-) name `isipinsubnet()` unless somebody objects until this is merged.

libpromises/evalfunction.c
tzz: It's not an accident, it's avoiding complexity at the code level. I'll adjust it to be restrictive.

tzz:
OK, comments added and behavior restricted.

tzz:
I added a followup cleanup commit to change `NULL == foo` to `foo == NULL` just in `evalfunction.c`

tzz:
In the codebase elsewhere there are maybe 30 other cases of `NULL == foo` that you can clean up at will...

libpromises/evalfunction.c
```diff
@@ -4860,23 +4861,48 @@ static FnCallResult FnCallFormat(EvalContext *ctx, ARG_UNUSED const Policy *poli
...
+        {
+            Log(LOG_LEVEL_DEBUG, "%s: found range %s on hostip %s, no interface", fp->name, range, VIPADDRESS);
+            return FnReturnContextTrue();
+        }
+
+        return FnReturnContextFalse();
+    }
+
+    /* Case 2: isipinsubnet(): check arg1 (a range) against arg2, arg3, ... if they are IPs */
+    for (Rlist *ip = finalargs->next; !iprange_mode && ip != NULL; ip = ip->next)
```
jimis: I'm confused, why are you checking `iprange_mode` on every iteration of the loop? Same comment for the next loop.

jimis: FWIW I would change the previous `if` as

```c
if (iprange_mode)
    if (finalargs->next == NULL)    // case1: iprange(arg1)
        blah;
    else                      // case2: iprange(arg1, args...)
        blue;
else                       // case3: is ipinsubnet()
    bleh;
```

jimis:
Thanks for the cleanups, I'll take care of the rest of the codebase :-)

libpromises/evalfunction.c
tzz: I felt it was simpler my way. Please feel free to take my commit and adjust it, I'm not understanding your request well and it's faster to just make the change on your side.

libpromises/evalfunction.c
```diff
@@ -4860,23 +4861,48 @@ static FnCallResult FnCallFormat(EvalContext *ctx, ARG_UNUSED const Policy *poli
...
             return FnReturnContextTrue();
         }
-        else if (FuzzySetMatch(range, ip->name) == 0)
+        BufferDestroy(ipb);
+    }
+
+    /* Case 3: iprange(): check arg1 (a range) against arg2, arg3, ... if
+     * they are interface names, inspecting the IP for that interface. Check all
+     * IPs if no interface names are given. */
+    for (const Item *ip = EvalContextGetIpAddresses(ctx); iprange_mode && ip != NULL; ip = ip->next)
```
jimis: > /* Check all IPs if no interface names are given. */

I believe this should not be possible at this point in the code, right? Or am I missing some 4th mode for this function?

libpromises/evalfunction.c
jimis: I give up (for now?) the logic and the differents modes in this function are killing me...

libpromises/evalfunction.c
```diff
@@ -4860,25 +4861,50 @@ static FnCallResult FnCallFormat(EvalContext *ctx, ARG_UNUSED const Policy *poli
...
-        if (FuzzySetMatch(range, VIPADDRESS) == 0 && NULL == ifaces)
+        if (FuzzySetMatch(range, VIPADDRESS) == 0)
         {
-            Log(LOG_LEVEL_DEBUG, "%s: found range %s on hostip %s, no interface", fp->name, range, ip->name);
+            Log(LOG_LEVEL_DEBUG, "%s: found range %s on hostip %s, no interface", fp->name, range, VIPADDRESS);
             return FnReturnContextTrue();
         }
-        else if (FuzzySetMatch(range, ip->name) == 0)
+
+        return FnReturnContextFalse();
```
jimis: This alters old behaviour: it used to search for a match on all interfaces in case VIPADDRESS didn't match.

Ugh, this function is way too complicated, and I thought it should be easy to because of the few LOC...

tzz:
@jimis I pushed a followup commit that I think clarifies the behavior. It also fixes the case with more than 1 interface, which exited too early before (that was the 4th mode you were wondering about). I hope the comments make it all clear.

jimis:
On Fri, Jun 24, 2016 at 2:33 PM, Ted Zlatanov <notifications@github.com>
wrote:

> @jimis <https://github.com/jimis> I pushed a followup commit that I think
> clarifies the behavior. It also fixes the case with more than 1 interface,
> which exited too early before (that was the 4th mode you were wondering
> about). I hope the comments make it all clear.
>
I'm actually in the process of a rewrite myself ... :-)

jimis:
BTW I'm really convinced by now that `isipinsubnet()` needs to be a different function. :-)

tzz:
OK, it should be easy now that the bug is fixed and I split the code into two sections.

jimis:
You had challenged me to adapt it to my liking, so I couldn't stop once started! @tzz please review pull request #2632. Closing this one.

PR #2530, opened 2016-03-15
Process test spurious failure redmine7986

jimis:


jimis:
The real change is just line on the 2nd commit, the 1st commit is purely cosmetic.

jimis:
trigger build

PR #2531, opened 2016-03-15
redmine7952-36x_cherry

jimis:


jimis:
Passes `make check`, merging.

PR #2535, opened 2016-03-15
Only set VPSOPTS to 'auwx -J 0' when the FreeBSD version is > 9.3

skreuzer:
The change introduced in [9889494](https://github.com/cfengine/core/commit/98894948b26daedd76e6ff5303c6e75757231492) is causing the process promises to fails on older version of FreeBSD since -J (Display information about processes which match the specified jail IDs) @is a feature that was introduced in FreeBSD 9.3

### DIFF
--- a/libpromises/systype.c
+++ b/libpromises/systype.c
@@ -23,6 +23,9 @@
 */
 
 #include <systype.h>
+#if defined __FreeBSD__
+#include <sys/param.h>
+#endif
 
 /* Set in libenv/sysinfo.c::DetectEnvironment (called every time environment
    reload is performed).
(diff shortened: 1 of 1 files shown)


skreuzer:
Tagging @ddurieux on this PR as well so he can help me test this out

bahamat:
Is there a longer `ps -o` that could be used?

libpromises/systype.c
```diff
@@ -100,7 +103,11 @@ const char *const VPSOPTS[] =
     [PLATFORM_CONTEXT_BUSYBOX] = "",                  /* linux / busybox */
     [PLATFORM_CONTEXT_SOLARIS] = "auxww",     /* solaris >= 11 */
     [PLATFORM_CONTEXT_SUN_SOLARIS] = "auxww", /* solaris < 11 */
-    [PLATFORM_CONTEXT_FREEBSD] = "auxw -J 0",              /* freebsd */
+#if __FreeBSD_version >= 903000
+    [PLATFORM_CONTEXT_FREEBSD] = "auxw -J 0",              /* freebsd 9.3 and newer */
+#else
+    [PLATFORM_CONTEXT_FREEBSD] = "auxw",              /* freebsd 9.2 and older*/
+#endif
```
kacf: You will have to surround this with `#if defined __FreeBSD__` as well, otherwise the other platforms will complain about comparison with an unknown macro.

kacf:
Even better if this was runtime detected, but at least it solves the immediate problem.

libpromises/systype.c
kacf: Eh, never mind, it doesn't complain if it's missing apparently.

kacf:
Thanks! Cherry-picked to 3.7.x and 3.8.x as well (the option is not present in 3.6.x).

PR #2536, opened 2016-03-16
stupid

jimis:


libpromises/exec_tools.c
```diff
@@ -155,7 +155,7 @@ void ActAsDaemon()
                 GetErrorStr());
         }
 
-        if (dup2(fd, STDERR_FILENO) == -1);
+        if (dup2(fd, STDERR_FILENO) == -1)
```
jimis: @kacf @pasinskim @GregorioDiStefano how can we guard against such stupidity, by using our CI? Is there a simple tool we can run with little effort?

kacf: Man, I can't believe the compiler doesn't warn about this. Maybe there's a warning we can turn on deep down in the gcc argument list.

PR #2538, opened 2016-03-16
Fix: Test was incorrectly being skipped for *not* windows

nickanderson:
This is a *nix test, and is unsupported on *windows* platforms.


kacf:
Good catch!

kacf:
trigger build

kacf:
Oh, this test has the same problem that the other test using `stat` has, it doesn't exist on many platforms (failing on all non-Linuxes). You should perhaps rebase this on top of that other test and change this one so it also uses `has_stat` and `$(G.stat)`.

nickanderson:
superseded by #2545

PR #2539, opened 2016-03-17
Redmine #7961: Remove auto-loading of "inputs" mentioned in def.json.

kacf:
This was done in order to solve the problem that def.json is applied
to all policy files, including failsafe.cf. If there is a syntax error
in any of those files, all policy files will stop working and the host
is unable to recover.

Now one has to use the "inputs" section by referring to it as the
$(def.augments_inputs) variable, which it will populate.

Changelog: Fix bug which could render host unable to recover from a
syntax error, even if failsafe.cf was utilized. This could happen if
the file containing the syntax error was specified in the def.json
special file.

Changelog: Change: Policy files specified in the "inputs" section of
def.json will no longer be auto-loaded. One has to refer to the
$(def.augments_inputs) variable in the policy (the standard
masterfiles policies include this by default). This only affects
installations which are not based on the standard masterfiles, and
which are using the "inputs" field inside def.json.

### DIFF
--- a/libpromises/rlist.c
+++ b/libpromises/rlist.c
@@ -367,7 +367,6 @@ Rval RvalNewRewriter(const void *item, RvalType type, JsonElement *map)
                 char closing_brace = 0;
                 for (int c = 0; c < buffer_from[c]; c++)
                 {
-                    printf("In %s at %i: '%s'\n", __func__, __LINE__, buffer_from);
                     if (buffer_from[c] == '$')
                     {
                         if (buffer_from[c+1] == '(')
--- a/libpromises/eval_context.c
+++ b/libpromises/eval_context.c
@@ -1649,6 +1649,9 @@ StringSet *EvalContextStackPromisees(const EvalContext *ctx)
     return promisees;
 }
 
+/*
+ * Copies value, so you need to free your own copy afterwards.
+ */
 bool EvalContextVariablePutSpecial(EvalContext *ctx, SpecialScope scope, const char *lval, const void *value, DataType type, const char *tags)
 {
     if (strchr(lval, '['))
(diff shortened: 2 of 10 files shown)


kacf:
Do not merge yet! This is missing tests and documentation.

kacf:
@tzz: Do you mind taking a look at this? I'm still missing tests and changing the documentation, but this is a beginning.

libpromises/rlist.c
```diff
@@ -1438,6 +1437,71 @@ JsonElement *RvalToJson(Rval rval)
...
+
+    Rval rval;
+
+    switch (JsonGetElementType(json))
+    {
+    case JSON_ELEMENT_TYPE_CONTAINER:
+        switch (JsonGetContainerType(json))
+        {
+        case JSON_CONTAINER_TYPE_OBJECT:
+            rval.item = json;
```
kacf: Actually, I probably have to `JsonCopy` this.

libpromises/rlist.c
```diff
@@ -1438,6 +1437,71 @@ JsonElement *RvalToJson(Rval rval)
...
+        }
+
+        default:
+            ProgrammingError("JsonElement container with unknown type");
+            break;
+        }
+        break;
+
+    case JSON_ELEMENT_TYPE_PRIMITIVE:
+        rval.item = JsonPrimitiveGetAsString(json);
```
kacf: And `xstrdup` this.

tzz:
Looks nice.

kacf:
Test verifying that it works with masterfiles: https://github.com/cfengine/masterfiles/pull/668

kacf:
Alright, updated. I also found `ContainerToRlist` which means I could remove the whole `JsonToRval` function I made. No need for duplication.

kacf:
trigger build

kacf:
Cherry-picked to 3.8.x, 3.7.x is pending.

PR #2541, opened 2016-03-18
3.7.x backports

kacf:


kacf:
@tzz, @nickanderson: Can you please check if any important commit for the def.json parsing is missing here?

kacf:
Merging so I can start a package build, but I would still appreciate a review.

PR #2546, opened 2016-03-27
Change: Reduce noise to info level logs

nickanderson:
These messages are not useful in inform mode. I moved them to verbose
mode to match the log for indicating the size and mode but arguably
these all would be better in debug mode.

### DIFF
--- a/cf-agent/files_properties.c
+++ b/cf-agent/files_properties.c
@@ -53,11 +53,11 @@ static bool ConsiderFile(const char *nodename, const char *path, struct stat *st
     {
         if (S_ISLNK(stat->st_mode))
         {
-            Log(LOG_LEVEL_INFO, "'%s' is a symbolic link", nodename);
+            Log(LOG_LEVEL_VERBOSE, "'%s' is a symbolic link", nodename);
         }
         else if (S_ISDIR(stat->st_mode))
         {
-            Log(LOG_LEVEL_INFO, "'%s' is a directory", nodename);
+            Log(LOG_LEVEL_VERBOSE, "'%s' is a directory", nodename);
         }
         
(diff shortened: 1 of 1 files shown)


nickanderson:
try running masterfiles with -I and you see tons  of useless messages a bout things being directories.

nickanderson:
ping @jimis

jimis:
I believe it's OK to reduce the verbosity there. Which part of the policy generates the annoying messages? Is it some specific files promise?

nickanderson:
It was being generated by the log rotation policy. I tried and failed to suppress the messages by guarding the promise with ifvarclass => isplain($(file)).

jimis:
The pull request makes sense to me. The only thing is that these INFO level messages are from CFEngine 3.0, so I'm asking what kind of promise generates them to alleviate all concerns. Would these messages be expected in any case? @kacf what do you think?

kacf:
I think it makes perfect sense to reduce the verbosity, the message seems way too excessive (if it were up to me I'd even consider DEBUG log level for this).

jimis:
After some investigation with @nickanderson it seems that these message are output because of https://github.com/cfengine/core/pull/2468.

Before that pull request, these message used to come always after an error message:
```c
    Log(LOG_LEVEL_ERR, "Suspicious looking file object '%s' masquerading as hidden file in '%s'", nodename, path);

    if (S_ISLNK(stat->st_mode))
    {
        Log(LOG_LEVEL_INFO, "'%s' is a symbolic link", nodename);
    }
    else if (S_ISDIR(stat->st_mode))
    {
        Log(LOG_LEVEL_INFO, "'%s' is a directory", nodename);
    } 
```

Closing, the proper should be different.

PR #2548, opened 2016-03-31
Remove useless and annoying INFO log messages in Considerfile

amousset:
5e7c0373f918e9b089ed0eef30671ee33a1e8ea5 changed the conditions to display general stat info about the considered files. These messages are useless and should be removed.

### DIFF
--- a/cf-agent/files_properties.c
+++ b/cf-agent/files_properties.c
@@ -48,46 +48,32 @@ static const char *const SKIPFILES[] =
 static bool ConsiderFile(const char *nodename, const char *path, struct stat *stat)
 {
     int i;
-    
-    if (stat != NULL)
-    {
-        if (S_ISLNK(stat->st_mode))
-        {
-            Log(LOG_LEVEL_INFO, "'%s' is a symbolic link", nodename);
-        }
-        else if (S_ISDIR(stat->st_mode))
-        {
-            Log(LOG_LEVEL_INFO, "'%s' is a directory", nodename);
-        }
-        
-        Log(LOG_LEVEL_VERBOSE, "'%s' has size %ld and full mode %o", nodename, (unsigned long) (stat->st_size),
-              (unsigned int) (stat->st_mode));
-    }
 
-    if (strlen(nodename) < 1)
+    if (nodename[0] == '\0')
     {
-        Log(LOG_LEVEL_ERR, "Empty (null) filename detected in '%s'", path);
+        Log(LOG_LEVEL_ERR, "Empty (null) filename detected in '%s', skipping", path);
         return false;
     }
 
     if (stat != NULL && (S_ISREG(stat->st_mode) || S_ISLNK(stat->st_mode)) &&
         IsItemIn(SUSPICIOUSLIST, nodename))
     {
-        Log(LOG_LEVEL_ERR, "Suspicious file '%s' found in '%s'", nodename, path);
+        Log(LOG_LEVEL_WARNING, "Filename '%s/%s' is classified as suspiscious, skipping", path, nodename);
         return false;
     }
 
+    /* A file named '...' is likely to be a path traversal attempt, see http://cwe.mitre.org/data/definitions/32.html */
     if (strcmp(nodename, "...") == 0)
     {
-        Log(LOG_LEVEL_VERBOSE, "Possible DFS/FS cell node detected in '%s' ...", path);
-        return true;
+        Log(LOG_LEVEL_WARNING, "Possible path traversal attempt detected in '%s/%s', skipping", path, nodename);
+        return false;
     }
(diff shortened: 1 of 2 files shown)


jimis:
Thanks @amousset! Since you are going into sanitising `ConsiderFile()` I'll comment on a couple of details, feel free to amend separate commits for the fixes.

cf-agent/files_properties.c
```diff
@@ -48,32 +48,17 @@ static const char *const SKIPFILES[] =
...
-        Log(LOG_LEVEL_ERR, "Empty (null) filename detected in '%s'", path);
+        Log(LOG_LEVEL_ERR, "Empty (null) filename detected in '%s', skipping", path);
         return false;
     }
 
     if (stat != NULL && (S_ISREG(stat->st_mode) || S_ISLNK(stat->st_mode)) &&
         IsItemIn(SUSPICIOUSLIST, nodename))
     {
-        Log(LOG_LEVEL_ERR, "Suspicious file '%s' found in '%s'", nodename, path);
+        Log(LOG_LEVEL_ERR, "Filename '%s/%s' is classified as suspiscious, skipping", path, nodename);
```
jimis: Skipping a file is not an error, I would say it's not even INFO, probably VERBOSE. What do you think?

cf-agent/files_properties.c
```diff
@@ -87,7 +72,7 @@ static bool ConsiderFile(const char *nodename, const char *path, struct stat *st
     {
         if (strcmp(nodename, SKIPFILES[i]) == 0)
         {
-            Log(LOG_LEVEL_DEBUG, "Filename '%s/%s' is classified as ignorable", path, nodename);
+            Log(LOG_LEVEL_DEBUG, "Filename '%s/%s' is classified as ignorable, skipping", path, nodename);
```
jimis: Same here, probably VERBOSE, but skipping a path is not DEBUG.

cf-agent/files_properties.c
```diff
@@ -48,32 +48,17 @@ static const char *const SKIPFILES[] =
...
-        else if (S_ISDIR(stat->st_mode))
-        {
-            Log(LOG_LEVEL_INFO, "'%s' is a directory", nodename);
-        }
-        
-        Log(LOG_LEVEL_VERBOSE, "'%s' has size %ld and full mode %o", nodename, (unsigned long) (stat->st_size),
-              (unsigned int) (stat->st_mode));
-    }
 
     if (strlen(nodename) < 1)
```
jimis: it's more efficient like `if (nodename[0] == '\0')`.

jimis:
I would also suggest to convert the `"..."` check to return false, and log skipping that file. The way I understand it (feel free to comment otherwise) is that it's a security risk, and the most reasonable explanation I found is here: http://cwe.mitre.org/data/definitions/32.html, feel free to add it as a comment.

cf-agent/files_properties.c
amousset: https://github.com/cfengine/core/blob/master/libpromises/mod_common.c#L311 says:

> "suspiciousnames", "", "List of names to warn about if found during any file search"

I understand it as a WARN message and no skipping, which is not the current behavior. We can:
* Make it a WARN log message and keep the file, like it is documented
* Make it a WARN message and skip the file
* Make it a INFO/VERBOSE message and skip the file

It think we should use the WARN level, because it is a list of "suspicious" files and not simply an exclusion list, and the user probably wants to know when a file considered suspicious is found.
I don't know if we should should skip or not (i.e. if we should break the behavior or the specs), what do you think?

jimis: You are right! I agree with changing to WARN, but I'm a bit sceptical about skipping the files, since it's a change in behaviour.

I would say that skipping files would be better as a new "skipfiles" option that would populate the `SKIPFILES` array.

amousset: The suspicious file are currently skipped (and were also before #2468), so keeping them would indeed be a change in behavior.

jimis: You are right again, so go ahead and retain the same behaviour as it used to be. That would be the easiest way to get this in. Change in behaviour can be discussed afterwards.

amousset:
Updated the PR:
* Changed the documentation string for `suspiciousnames` to mention skipping
* Skip `"..."` and log it at WARN level
* Log suspicious names at WARN level

jimis:
Great, thanks @amousset.

trigger build

jimis:
All green, thank you!

amousset:
@jimis Also opened https://github.com/cfengine/documentation/issues/1438 for documentation

jimis:
Merged, thanks for completing this fully. :+1:

PR #2550, opened 2016-03-31
Change: Improve the comment regex for peers

nickanderson:


kacf:
I had to revert this, because it breaks the `04_examples/outputs/check_outputs.cf`. Be careful with examples like this.

I tried to solve this properly instead of reverting, but I couldn't get the expression to work. Is there a bug here? Try reapplying it and running the test.

PR #2553, opened 2016-03-31
Change: Improve peerleader and peerleaders examples

nickanderson:


kacf:
Had to revert this as well, same reason as #2550.

PR #2561, opened 2016-04-05
CFE-2183: Figure out version instead of hardcoding.

kacf:
This way we can build all versions we want automatically, without
having to hardcode the version number in the source, using the
EXPLICIT_VERSION variable. If the variable is not set, and git is
available, it will try to figure it out from git tags.

Note that all of this is done in the autogen stage, because git will
generally not be available at the configure stage. This is also
required in order for tarballs to come with the correct version
embedded in them, because they won't have git available to figure this
out.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -41,11 +41,27 @@ dnl AM_INIT_AUTOMAKE.
 dnl
 AC_CHECK_PROG(MAKEINFO, makeinfo, makeinfo)
 
-define([revision], m4_esyscmd_s([git rev-list -1 --abbrev-commit HEAD]))
+m4_define([revision], m4_esyscmd_s([git rev-list -1 --abbrev-commit HEAD]))
+
+m4_define([version_from_env], m4_esyscmd_s([echo $EXPLICIT_VERSION]))
+m4_ifblank(version_from_env, [
+    m4_define([version_from_detect], m4_esyscmd_s([misc/determine-version.py]))
+    m4_if(m4_sysval, 0, [], [
+        m4_fatal([Could not determine CFEngine version. Please set EXPLICIT_VERSION in the environment or make sure all git tags are up to date.])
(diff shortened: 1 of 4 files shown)


kacf:
Goes together with:

* https://github.com/cfengine/enterprise/pull/304
* https://github.com/cfengine/nova/pull/975
* https://github.com/cfengine/buildscripts/pull/115
* https://github.com/cfengine/masterfiles/pull/683

kacf:
This has already been tested on a private branch with changes applied to all repositories.

configure.ac
```diff
@@ -41,12 +41,28 @@ dnl AM_INIT_AUTOMAKE.
...
 AC_CHECK_PROG(MAKEINFO, makeinfo, makeinfo)
 
-define([revision], m4_esyscmd_s([git rev-list -1 --abbrev-commit HEAD]))
+m4_define([revision], m4_esyscmd_s([git rev-list -1 --abbrev-commit HEAD]))
+
+m4_define([version_from_env], m4_esyscmd_s([echo $EXPLICIT_VERSION]))
+m4_ifblank(version_from_env, [
+    m4_define([version_from_detect], m4_esyscmd_s([misc/determine-version.py]))
+    m4_if(m4_sysval, 0, [], [
+        m4_fatal([Could not determine CFEngine version. Please set EXPLICIT_VERSION in the environment or make sure all git tags are up to date.])
```
jimis: would it make sense to use a default version here, something like "master" or "head" etc?

jimis:
+1 looking forward to using it in practice.

configure.ac
kacf: Yes, I wasn't sure. My thinking is that we can see if it gets problematic and add a default later if we need to. Right now it would be nice to catch the cases where it fails.

PR #2562, opened 2016-04-05
Add: Test that augments file can define classes based on hard classes

nickanderson:


tzz:
@nickanderson @kacf looks like this code is not working right in `CheckContextOrClassmatch()`:

```
    ClassTableIterator *iter = EvalContextClassTableIteratorNewGlobal(ctx, NULL, true, true);
    StringSet *global_matches = ClassesMatching(ctx, iter, c, NULL, true); // returns early
```

... seems like it's returning an empty set for `c = "cfengine"`? Weird... this is pretty standard stuff.

kacf:
This is already covered by the test in #2565.

PR #2563, opened 2016-04-06
Add support for Jira tickets in test suppression meta tags.

kacf:
Strings starting with "CFE-", "ENT-", "ARCHIVE-" and "QA-" are valid
ticket identifiers.

Redmine support is retained for existing tests.


kacf:
@nickanderson: Maybe you want to take a look.

I opted for no prefix, just straight ticket names (like "CFE-1234"), but we could also require prefixing with "Jira:", like we have prefixed with "redmine" until now. Thoughts welcome.

kacf:
So one example would be:

```
"test_soft_fail" string => "any",
  meta => { "CFE-1234" };
```

kacf:
trigger build

kacf:
Build is green, merging, since we need this immediately.

kacf:

PR #2566, opened 2016-04-08
Fix: Use specifed output type with --list

nickanderson:
When using --list the output type was forced to list being an ansible
compatible seralized json string. However this made it impossible to
--list in any other output type.

This also adds an ansible_emptyhostvars output type which inserts a
_meta section defining that no host vars exist. This can speed up
working with ansible on a large scale env as it avoids making individual
executions to retrieve host data.


tzz:
YES

PR #2568, opened 2016-04-11
CFE-2161: Lots of ps parsing fixes.

kacf:
I recommend viewing commit by commit, and preferably locally with the `--histogram` option, it becomes much more readable then.

### DIFF
--- a/libpromises/processes_select.c
+++ b/libpromises/processes_select.c
@@ -1087,7 +1087,6 @@ int ZLoadProcesstable(Seq *pidlist, Seq *rootpidlist)
     int start[CF_PROCCOLS];
     int end[CF_PROCCOLS];
 
-    int index = 0;
     const char *pscmd = "/usr/bin/ps -Aleo zone,user,pid";
 
     FILE *psf = cf_popen(pscmd, "r", false);
--- a/libpromises/processes_select.c
+++ b/libpromises/processes_select.c
@@ -54,7 +54,7 @@
 #endif
 TABLE_STORAGE Item *PROCESSTABLE = NULL;
 
-#ifdef __sun
+#if defined(__sun) || defined(TEST_UNIT_TEST)
 static StringMap *UCB_PS_MAP = NULL;
 #define UCB_PS "/usr/ucb/ps"
 #define UCB_PS_ARGS "axwww"
@@ -62,15 +62,35 @@ static StringMap *UCB_PS_MAP = NULL;
 
 static int SelectProcRangeMatch(char *name1, char *name2, int min, int max, char **names, char **line);
 static bool SelectProcRegexMatch(const char *name1, const char *name2, const char *regex, bool anchored, char **colNames, char **line);
-static int SplitProcLine(const char *proc, time_t pstime, char **names, int *start, int *end, char **line);
+static bool SplitProcLine(const char *proc, time_t pstime, char **names, int *start, int *end, char **line);
 static int SelectProcTimeCounterRangeMatch(char *name1, char *name2, time_t min, time_t max, char **names, char **line);
 static int SelectProcTimeAbsRangeMatch(char *name1, char *name2, time_t min, time_t max, char **names, char **line);
 static int GetProcColumnIndex(const char *name1, const char *name2, char **names);
 static void GetProcessColumnNames(const char *proc, char **names, int *start, int *end);
 static int ExtractPid(char *psentry, char **names, int *end);
 static void ApplyPlatformExtraTable(char **names, char **columns);
+static bool ZombiesCanHaveEmptyColumns(void);
 
 /***************************************************************************/
+
+// For unit testing
+#ifndef TEST_UNIT_TEST
+
+static bool ZombiesCanHaveEmptyColumns(void)
+{
+#ifdef _AIX
+    /*
+      Ideally this should be autoconf-tested, but it's really hard to do so. ATM
+      AIX is the only platform known to produce empty columns in ps output. See
+      SplitProcLine() to see the exact effects this has.
+    */
+    return true;
+#else
+    return false;
+#endif
+}
+
+#endif // TEST_UNIT_TEST
 
 static bool SelectProcess(const char *procentry,
                           time_t pstime,
@@ -88,6 +108,8 @@ static bool SelectProcess(const char *procentry,
     assert(process_regex);
 
     StringSet *process_select_attributes = StringSetNew();
+
+    memset(column, 0, sizeof(column));
 
     if (!SplitProcLine(procentry, pstime, names, start, end, column))
     {
@@ -236,7 +258,7 @@ Item *SelectProcesses(const char *process_name, ProcessSelect a, bool attrselect
     int start[CF_PROCCOLS];
     int end[CF_PROCCOLS];
 
-    GetProcessColumnNames(processes->name, &names[0], start, end);
+    GetProcessColumnNames(processes->name, names, start, end);
 
     /* TODO: use actual time of ps-run, as time(NULL) may be later. */
     time_t pstime = time(NULL);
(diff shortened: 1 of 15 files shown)


kacf:
Being tested here: http://jenkins.usdc.cfengine.com/job/testing-kristian-privateBranch/143/

kacf:
This is now green: http://jenkins.usdc.cfengine.com/job/testing-kristian-privateBranch/145/

@jimis: If you want to look at this, now's the time! :-)

libpromises/processes_select.c
```diff
@@ -1248,6 +1272,187 @@ const char *GetProcessTableLegend(void)
...
+        Log(LOG_LEVEL_VERBOSE,
+            "%s not executable, extra process information not available",
+            UCB_PS);
+        return NULL;
+    }
+
+    FILE *cmd = cf_popen(UCB_PS " " UCB_PS_ARGS, "rt", false);
+    if (!cmd)
+    {
+        Log(LOG_LEVEL_WARNING, "Could not execute \"%s %s\", extra process "
```
jimis: The message should mention the most serious side-effect: process command line matching will be limited to 80 characters.

Additionally, maybe WARNING is too much, given that Solari are System V UNIX and are not guaranteed to have BSD tools.

Last, I've heard recently that `/usr/ucb/ps` will be deprecated on some latest solaris, but the system `ps` works correctly now so it's not needed. We could ping the Illumos experts from the community for more info, do you remember their nicknames?

tests/acceptance/05_processes/02_functions/findprocesses.cf
```diff
@@ -2,6 +2,9 @@
 #
 # Test findprocesses()
 #
+# Note: On HP-UX this depends on the
+# 00_hpux_ps_setup_serial.cf setup test having run first.
```
jimis: Why does it need to run first? Maybe edit the comment and/or the commit message?

libpromises/processes_select.c
```diff
@@ -548,6 +548,29 @@ static bool SelectProcRegexMatch(const char *name1, const char *name2,
...
+    char arrow_str[CF_BUFSIZE];
+    arrow_str[0] = '^';
+    arrow_str[1] = '\0';
+    char index_str[CF_BUFSIZE];
+    index_str[0] = '0';
+    index_str[1] = '\0';
+    for (int lineindex = 10; lineindex <= len; lineindex += 10)
+    {
+        char num[11];
+        xsnprintf(num, sizeof(num), "% 10d", lineindex);
```
jimis: typo space after '%' ?

libpromises/processes_select.c
```diff
@@ -548,6 +548,29 @@ static bool SelectProcRegexMatch(const char *name1, const char *name2,
...
+{
+    char arrow_str[CF_BUFSIZE];
+    arrow_str[0] = '^';
+    arrow_str[1] = '\0';
+    char index_str[CF_BUFSIZE];
+    index_str[0] = '0';
+    index_str[1] = '\0';
+    for (int lineindex = 10; lineindex <= len; lineindex += 10)
+    {
+        char num[11];
```
jimis: 11 is arbitrary. Better use `PRINTSIZE(lineindex)`.

libpromises/processes_select.c
```diff
@@ -548,6 +548,29 @@ static bool SelectProcRegexMatch(const char *name1, const char *name2,
...
+        xsnprintf(num, sizeof(num), "% 10d", lineindex);
+        strlcat(index_str, num, sizeof(index_str));
+        strlcat(arrow_str, "         ^", sizeof(arrow_str));
+    }
+
+    // Prefix the beginning of the indexes with the given number.
+    char fmt[100];
+    xsnprintf(fmt, sizeof(fmt), "%% %ds%%s", prefix_spaces);
```
jimis: Again, is the space after "%%" bogus?

libpromises/processes_select.c
```diff
@@ -601,7 +624,14 @@ static bool SplitProcLine(const char *line,
         return false;
     }
 
-    Log(LOG_LEVEL_DEBUG, "Trying to match ps line '%s'", line);
+    size_t linelen = strlen(line);
+
+    if (LogGetGlobalLevel() >= LOG_LEVEL_DEBUG)
+    {
+        Log(LOG_LEVEL_DEBUG, "Parsing ps line: '%s'", line);
```
jimis: I would skip the single quotes, since it's coming at the end of the line after the colon. Because these long quoted strings often contain quotes themselves, and then they are really confusing.

libpromises/processes_select.c
```diff
@@ -56,8 +56,13 @@ TABLE_STORAGE Item *PROCESSTABLE = NULL;
 
 #if defined(__sun) || defined(TEST_UNIT_TEST)
 static StringMap *UCB_PS_MAP = NULL;
-#define UCB_PS "/usr/ucb/ps"
-#define UCB_PS_ARGS "axwww"
+// These will be tried in order.
+const char *SOLARIS_UCB_STYLE_PS[] = {
+    "/usr/ucb/ps",
+    "/bin/ps",
```
jimis: I'm a bit confused now. So on Solaris 11 both commands accept BSD-style arguments? And if `/usr/ucb/ps` is not available, we'll call `/bin/ps` once with System V arguments and once with BSD-style arguments?

What if UCB commands are missing from an older system, and `/bin/ps` barfs at the BSD-style arguments? Do we detect that it returns error?

libpromises/processes_select.c
```diff
@@ -1237,29 +1242,44 @@ const char *GetProcessTableLegend(void)
...
+        char *ps_cmd;
+        xasprintf(&ps_cmd, "%s %s", SOLARIS_UCB_STYLE_PS[i],
+                  SOLARIS_UCB_STYLE_PS_ARGS);
+
+        FILE *cmd = cf_popen(ps_cmd, "rt", false);
+        if (!cmd)
+        {
+            Log(LOG_LEVEL_WARNING, "Could not execute \"%s\", extra process "
+                "information not available.", ps_cmd);
+        }
```
jimis: Is there a way for detecting that the command returned immediately with error? Or is this handled by `cf_popen`?

jimis:
Wow the main commit is monstrous, I couldn't follow! :-) Solid work anyway, can tell from the number of tests. Noted only minor remarks.
+1!

tests/acceptance/05_processes/02_functions/findprocesses.cf
kacf: I prefer not to duplicate the information. It's clearly explained why it's needed inside that test.

libpromises/processes_select.c
kacf: No, that's correct! :-) The complete expression means: "print a number, make it ten characters wide, and fill the unused space with the space character".

libpromises/processes_select.c
kacf: Right. It's extremely unlikely to be bigger, but `PRINTSIZE` is better practice. Will fix!

libpromises/processes_select.c
kacf: No, same reason.

libpromises/processes_select.c
kacf: It's perhaps better explained by demonstration. The section will produce debug output like this:

```
   debug: Parsing ps line: 'root         1     0     1  0.0  0.0  24728   0      2416    1 Apr12  2-03:59:55 00:00:01 /sbin/init'
   debug:                   ^         ^         ^         ^         ^         ^         ^         ^         ^         ^         ^
   debug:            Index: 0        10        20        30        40        50        60        70        80        90       100
```

libpromises/processes_select.c
kacf: I added it specifically because I was having trouble seeing where a line ended, since whitespace is supposed to be eaten by the parser (this is one of the tests), so I'd like to keep it.

libpromises/processes_select.c
kacf: > I'm a bit confused now. So on Solaris 11 both commands accept BSD-style arguments? And if `/usr/ucb/ps` is not available, we'll call `/bin/ps` once with System V arguments and once with BSD-style arguments?

Yes, that's right.

> What if UCB commands are missing from an older system, and `/bin/ps` barfs at the BSD-style arguments? Do we detect that it returns error?

Yes, see other comment about `cf_pclose`.

libpromises/processes_select.c
kacf: It's handled by `cf_pclose`. See that function call inside `LoadPlatformExtraTable()`.

libpromises/processes_select.c
kacf: > The message should mention the most serious side-effect: process command line matching will be limited to 80 characters.

Yes, good idea, will add that.

> Additionally, maybe WARNING is too much, given that Solari are System V UNIX and are not guaranteed to have BSD tools.

Note that this message is only if the tool exists, but the execution fails for some reason. So I think the much more common case of not having the tool is covered further up by the `stat` test.

> Last, I've heard recently that `/usr/ucb/ps` will be deprecated on some latest solaris, but the system `ps` works correctly now so it's not needed. We could ping the Illumos experts from the community for more info, do you remember their nicknames?

I lost the link now, but I read a blog post from one of the Solaris developers that `/bin/ps` would be able to do everything `/usr/ucb/ps` can. However, this is *very* recent, and only happened in February/March or so. So many systems will not have it for a while, most likely.

As long as the UCB style ps calls do not report *wrong* information, it does no harm to call it, just a small performance loss, but it will end up reporting exactly the same things.

kacf:
Thanks a lot for the review! I will fix the issues and update!

kacf:
Alright, updated with https://github.com/cfengine/core/pull/2568/commits/0835eff3d703e5823b7e7d8c5077cb501e33d538.

libpromises/processes_select.c
```diff
@@ -1268,7 +1266,8 @@ static FILE *OpenUcbPsPipe(void)
         if (!cmd)
         {
             Log(LOG_LEVEL_WARNING, "Could not execute \"%s\", extra process "
-                "information not available.", ps_cmd);
+                "information not available. Process command line length may be "
```
jimis: Can you break the literal string on the punctuation? That way grepability of messages is not affected by folding long strings.

kacf: Good point, will fix that.

PR #2570, opened 2016-04-12
Jira #2337: The lexer can go wrong with @if minimum_version macro

basvandervlies:
The lexer could not handle empty newline(s) before a ```@endif```.


kacf:
Thanks Bas! Closing in favor of #2577 to fix a couple of small issues. I took the liberty of adapting existing tests instead of using yours, it's pretty easy to do it there and it will run faster.

basvandervlies:
No problem. My first test was adjusting an existing one ;-)

PR #2571, opened 2016-04-12
Cleanup signal handling for agents and enable them earlier.

kacf:
The motivation for this is that during policy loading, certain
functions may be evaluated that open pipes, and if these pipes at any
point become disconnected, we get an EPIPE signal. This is ignored
once we get to the evaluation stage, but during loading the signal
handlers have not been enabled yet.

This was originally seen during introduction of the mapdata()
function, which opens a pipe to an external process. In the test for
this function, `cat` was used to output a prerendered response, but
`cat` was also expected to consume the data input. Depending on
timing, if the data was written to `cat` before it terminated,
everything would be fine, but if `cat` terminated before data could be
written (because it was catting a file, not from stdin), cf-promises
would get an EPIPE signal and terminate abnormally.

Upon inspection of our signal handling, it appears this can happen in
cf-agent as well, and is as such "a ticking time bomb". Therefore this
commit cleans up the signal handling a bit, consolidates it where
possible, and moves all signal handler initialization to the beginning
of each agent main().

Daemons have other signal handler requirements and were left alone for
now.

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -217,6 +217,7 @@ static const char *const HINTS[] =
 
 int main(int argc, char *argv[])
 {
+    SetupSignalsForAgent();
 #ifdef HAVE_LIBXML2
         xmlInitParser();
 #endif
@@ -661,13 +662,6 @@ static void ThisAgentInit(void)
 #ifdef HAVE_SETSID
     setsid();
 #endif
-
-    signal(SIGINT, HandleSignalsForAgent);
-    signal(SIGTERM, HandleSignalsForAgent);
-    signal(SIGHUP, SIG_IGN);
-    signal(SIGPIPE, SIG_IGN);
-    signal(SIGUSR1, HandleSignalsForAgent);
-    signal(SIGUSR2, HandleSignalsForAgent);
 
     CFA_MAXTHREADS = 30;
     EDITFILESIZE = 100000;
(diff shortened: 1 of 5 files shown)


kacf:
@jimis: I'd like your input on this one.

tzz:
Yeah, nice catch!

jimis:
Nice catch indeed! Will look into it deeper in a bit. :-)

jimis:
Makes sense. What I don't get is why `mapdata()` gets evaluated before the evaluation starts, IMO it should be clearly part of the evaluation, but this belongs to the evaluator refactoring discussion!

So +1, I can't think of any side-effects this can have.

kacf:
`mapdata` is evaluated during pre-evaluation, since it is part of a vars promise. This is expected, and you can avoid it by the known class guard workaround.

Thanks for looking at it! I will merge it together with the PR that prompted this.

jimis:
I was under the impression that *in theory* only common bundles need to be evaluated in pre-eval. So if it's a "vars" promise in a non-common bundle, it should be skipped.

kacf:
No, all variables are resolved in the pre-eval step, regardless of which bundle they are in.

jimis:
> No, all variables are resolved in the pre-eval step, regardless of which bundle they are in.

Why would that happen? The variables in common bundles being pre-evaluated, makes sense in order to resolve dynamic inputs. But what would the reason be for that to happen in all bundles, during pre-evaluation?

kacf:
I don't know what the original motivation was, it just seems to be that way now.

PR #2573, opened 2016-04-12
Add: Example illustrating promise with multiple outcomes

nickanderson:


nickanderson:
trigger build

nickanderson:
trigger build

nickanderson:
trigger build

nickanderson:
Blarg, actually having this as an active test is problematic

nickanderson:
@tzz any ideas on how to have this as an automatic test that can show multiple outcomes?

The difference in output between normal and dry-run modes messes it up.

tzz:
Take a look at `great_expectations.cf` it may do what you need.

nickanderson:
Ah thats in another repository, I was wanting to simply add an example I could include in the docs that had the output tested all the time.

Thanks @tzz

PR #2574, opened 2016-04-12
ENT-2703: Store all set classes within eval context.

maciejmrowiec:


maciejmrowiec:
@jimis

kacf:
Looks good! I took the liberty of attempting a test run with both together here: http://jenkins.usdc.cfengine.com/view/Enterprise%20private%20branches/job/testing-kristian-privateBranch/147/

maciejmrowiec:
@kacf use this to enable: https://docs.cfengine.com/docs/master/reference-components-cf-agent.html#report_class_log

kacf:
Ok, I won't test anything specific regarding the PR, I leave that to you, it's just to test that nothing *else* breaks. :-)

libpromises/eval_context.c
```diff
@@ -1443,6 +1450,8 @@ static bool EvalContextClassPut(EvalContext *ctx, const char *ns, const char *na
         return false;
     }
 
+    StringSetAdd(ctx->all_classes, SafeStringDuplicate(name));
```
jimis: I believe this pull request does not offer any functionality to Core, but this one line adds a lot of memory of overhead, because of an Enterprise-only feature. Can we make it optional?

maciejmrowiec: There is already a number of classes being set before even the triggering setting is checked for being enabled which makes it a bit more complicated to reliably switch on-off without being sure we log all classes that get set.

jimis: Hmmm I see what you mean; This is not trivial, but I'll try to think of something since the memory overhead is significant I think.

BTW, do you store the fully qualified name in the set? i.e. the namespace and the scope? Or just the class name?

kacf: We could at least limit it to Enterprise by putting this inside a `ENTERPRISE_FUNC` macro. This detection runs very early, so it should be sufficient.

maciejmrowiec: No, didn't want to introduce anything custom so this is only class name.
I don't think storing a few hundred class names is that significant in terms or ram usage (in relation to entire application), but I agree it would be nice of there was of off switch.


For masterfiles there are around 400 classes set, 300 of them are set before we even check body agent where this is configured. So even it we keep it enabled until we check config and scrap&stop when we know it's off it would not save up much.

maciejmrowiec: @kacf right, we can do that.

jimis: Another solution would be to start with a non-NULL `StringSet` inside `EvalContext` and destroy it and make it NULL when you are sure that the classes logging option is disabled.

The line in this file would just change to:
```c
if (ctx->all_classes != NULL)
     StringSetAdd(ctx->all_classes, SafeStringDuplicate(name)); 
```

maciejmrowiec: In case of masterfiles we still do 75% of the appends in this case, only thing we would optimise it to keep that 50kb map in ram for a bit shorter.

jimis: The thing is that masterfiles is only basic policy. We know (from the recent class collision issue) that customers have way more complex policies, setting hundreds of thousands of classes during the agent lifetime.

libpromises/eval_context.c
```diff
@@ -2772,3 +2781,15 @@ JsonElement* JsonExpandElement(EvalContext *ctx, const JsonElement *source)
     ProgrammingError("JsonExpandElement: unexpected container type");
     return NULL;
 }
+
+const StringSet *EvalContextAllClassesGet(const EvalContext *ctx)
```
kacf: Minor: Lose the const.

kacf: Never mind.

PR #2575, opened 2016-04-12
Add getuserinfo() function

tzz:
This is a first rough cut of the function `getuserinfo()` I proposed in #1413 

Returns:

```json
{
    "description": "root",
    "gid": 0,
    "home_dir": "/root",
    "shell": "/bin/bash",
    "uid": 0,
    "username": "root"
}
```

Still TODO: `locked` boolean, primary group name, list of secondary group names, list of secondary GIDs, and password data.

### DIFF
--- a/cf-agent/verify_users.h
+++ b/cf-agent/verify_users.h
@@ -31,5 +31,4 @@ PromiseResult VerifyUsersPromise(EvalContext *ctx, const Promise *pp);
 
 void VerifyOneUsersPromise (const char *puser, User u, PromiseResult *result, enum cfopaction action,
                             EvalContext *ctx, const Attributes *a, const Promise *pp);
-
 #endif
--- a/cf-agent/verify_users_stub.c
+++ b/cf-agent/verify_users_stub.c
@@ -33,3 +33,13 @@ void VerifyOneUsersPromise (ARG_UNUSED const char *puser, ARG_UNUSED User u,
 {
     Log(LOG_LEVEL_ERR, "Users promise type is not supported on this OS");
 }
+
+bool IsAccountLocked(const char *puser, const void *passwd_info)
+{
+    return false;
+}
+
+bool GetPasswordHash(const char *puser, const void *passwd_info, const char **result)
+{
+    return false;
+}
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -789,6 +789,73 @@ static FnCallResult FnCallIP2Host(ARG_UNUSED EvalContext *ctx,
 
 #ifdef __MINGW32__
 
+static FnCallResult FnCallGetUserInfo(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const Policy *policy, ARG_UNUSED const FnCall *fp, const Rlist *finalargs)
+{
+    // TODO: surely there's something we can do???
+    return FnFailure();
+}
+
+#else /* !__MINGW32__ */
+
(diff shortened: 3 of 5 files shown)


tzz:
* the acceptance test is broken, I will work on it later
* I can't include `verify_users.h` even though I provided the necessary stubs. I'll need help on exposing `IsAccountLocked()` to `evalfunction.c` and can then expose the other functions the same way, I just need the first one as an example.
* first target is Linux

tzz:
@skreuzer can you please look at this and give your opinion on whether `sys.init_user` with the above contents (under that name or another) would be sufficient for you to resolve #1413? If yes, I'll add that variable to this patch and we can close the original PR.

skreuzer:
I still think making this available as a variable is a bit cleaner and faster but I suppose passing sys.init_user to this function works.

tzz:
@skreuzer oh no, I don't mean `sys.init_user` will be `root`. It will be a data container, as if you had said in the `sys` context `"init_user" => getuserinfo($(this.promiser_uid));`. Does that work?

skreuzer:
I suppose that is fine but it seems more expensive to have to continually call getuserinfo() from different agents. I would still prefer to have user.foo variables because I personally think it makes the agents easier to read and at the same time would be a bit clearer in the documentation.

If someone is new to cfengine and wants to get their current home directory, seeing $(user.home) in the documentation is a bit clearer then having to call getuserinfo($(this.promiser_uid)) and parsing out the information.

I'm not quite sure why both these things cant exist within cfengine but I seem to be losing this battle so I guess this is better than nothing.

tzz:
@skreuzer I am proposing that **the user will not have to call anything**.

On startup, either in policy or in the C code, the **equivalent** of

> in the `sys` context: `"init_user" => getuserinfo($(this.promiser_uid));`

will be **automatically** run. So you will immediately find all the data under `sys.init_user` which will be a data container. For instance `$(sys.init_user[home_dir])` will give you the home directory.

The reason I don't like `user.home_dir` is that it pollutes the top-level namespace with something that's not truly global. The problem is that it's unclear if we're talking about the **current** user (`this.promiser_uid`) or the current user in the bundle, or the initial user. Since there's plans to allow changing the UID on a bundle or promise level, the top-level `user` context would be confusing. So I think `sys.init_user` is better.

tzz:
@jimis small PR with test, should be easy to merge. Very useful.

tzz:
@jimis rebased and test fixed to work cross-platform, should be OK to merge

cf-agent/verify_users_pam.c
```diff
@@ -243,13 +243,14 @@ static struct spwd *GetSpEntry(const char *puser)
 }
 #endif // HAVE_FGETSPENT
 
-static bool GetPasswordHash(const char *puser, const struct passwd *passwd_info, const char **result)
+bool GetPasswordHash(const char *puser, const void *passwd_info, const char **result)
```
jimis: I guess these are forgotten changes, judging from the comments you have commented out in `FnCallGetUserInfo()`. It's anyway wrong to call code from `cf-agent` directory, inside libpromises. Can you remove the changes in `cf-agent`?

jimis:
In general I'm a simplistic user, and @skreuzer 's syntax appeals more to me. But anyway I'm not the one coding the feature, so I'll happily merge this code from @tzz. :-)

cf-agent/verify_users_pam.c
tzz: OK; done.

nickanderson:
Yes please

nickanderson:
Docs?

jimis:
Thanks @tzz!

tzz:
docs in https://github.com/cfengine/documentation/pull/1572

@nickanderson the TODO list at the top is not done. Please put it in JIRA if you think it's useful and it may get implemented :)

PR #2577, opened 2016-04-13
Jira #2337: The lexer can go wrong with @if minimum_version macro

kacf:
Changelog: Fix the lexer which could not handle empty newline(s)
before a ```@endif```.

Modified by: Kristian Amlie <kristian.amlie@cfengine.com>
- Removed dedicated test and instead made more permutations in
  existing tests.
- Fix misquoted ParseDebug strings.


kacf:
trigger build

kacf:

PR #2578, opened 2016-04-13
CFE-2246 packagesmatching() and packageupdatesmatching() work with ne…

pasinskim:
…w package promise.

Right now both should work when new package promise is used.

Changelog: packagesmatching() and packageupdatesmatching() should work
when new package promise is used. (CFE-2246)

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -1300,6 +1300,166 @@ static FnCallResult FnCallBundlesMatching(EvalContext *ctx, const Policy *policy
 
 /*********************************************************************/
 
+static bool AddPackagesmatchingJsonLine(pcre *matcher, JsonElement *json, char *line)
+{
+    
+    if (strlen(line) > CF_BUFSIZE - 80)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Line from package inventory is too long (%zd) to be sensible",
+            strlen(line));
+        return false;
+    }
+    
+
+    if (StringMatchFullWithPrecompiledRegex(matcher, line))
+    {
+        Seq *list = SeqParseCsvString(line);
+        if (SeqLength(list) != 4)
+        {
+            Log(LOG_LEVEL_ERR,
+                "Line from package inventory '%s' did not yield correct number of elements.",
+                line);
+            SeqDestroy(list);
+            return true;
+        }
+
+        JsonElement *line_obj = JsonObjectCreate(4);
+        JsonObjectAppendString(line_obj, "name",    SeqAt(list, 0));
+        JsonObjectAppendString(line_obj, "version", SeqAt(list, 1));
+        JsonObjectAppendString(line_obj, "arch",    SeqAt(list, 2));
+        JsonObjectAppendString(line_obj, "method",  SeqAt(list, 3));
+                
+        SeqDestroy(list);
+        JsonArrayAppendObject(json, line_obj);
+    }
+
+    return true;
+}
+
+static bool GetLegacyPackagesMatching(pcre *matcher, JsonElement *json, const bool installed_mode)
+{
+    char filename[CF_MAXVARSIZE];
+    if (installed_mode)
+    {
+        GetSoftwareCacheFilename(filename);
+    }
+    else
+    {
+        GetSoftwarePatchesFilename(filename);
+    }
+
+    Log(LOG_LEVEL_DEBUG, "Reading inventory from '%s'", filename);
+    
+    FILE *const fin = fopen(filename, "r");
+    if (fin == NULL)
+    {
+        Log(LOG_LEVEL_VERBOSE,
+            "Cannot open the %s packages inventory '%s' - "
+            "This is not necessarily an error. "
+            "Either the inventory policy has not been included, "
+            "or it has not had time to have an effect yet or you are using"
+            "new package promise and check for legacy promise is made."
+            "A future call may still succeed. (fopen: %s)",
+            installed_mode ? "installed" : "available",
+            filename,
+            GetErrorStr());
+
+        return true;
+    }
+    
+    char *line;
+    while (NULL != (line = GetCsvLineNext(fin)))
+    {
+        if (!AddPackagesmatchingJsonLine(matcher, json, line))
+        {
+            free(line);
+            break;
+        }
+        free(line);
+    }
+
+    bool ret = feof(fin);
+    fclose(fin);
+    
+    return ret;
+}
+
+static bool GetPackagesMatching(pcre *matcher, JsonElement *json, const bool installed_mode, Rlist *default_inventory)
+{
+    dbid database = (installed_mode == true ? dbid_packages_installed : dbid_packages_updates);
+
+    for (const Rlist *rp = default_inventory; rp != NULL; rp = rp->next)
+    {
+        const char *pm_name =  RlistScalarValue(rp);
+        size_t pm_name_size = strlen(pm_name);
+        
+        Log(LOG_LEVEL_DEBUG, "Reading packages (%d) for package modeule [%s]", 
+                database, pm_name);
+        
+        if (StringSafeEqual(pm_name, "cf_null"))
+        {
+            continue;
+        }
+        
+        CF_DB *db_cached;
+        if (!OpenSubDB(&db_cached, database, pm_name))
+        {
+            Log(LOG_LEVEL_ERR, "Can not open database %d to get packages data.", database);
+            return false;
+        }
+
+        char *key = "<inventory>";
+        int data_size = ValueSizeDB(db_cached, key, strlen(key) + 1);
+        
+        Log(LOG_LEVEL_DEBUG, "Reading inventory from database: %d", data_size);
+        
+        /* For empty list we are storing one byte value in database. */
+        if (data_size > 1)
+        {
+            char *buff = xmalloc(data_size + 1);
+            buff[data_size] = '\0';
+            if (!ReadDB(db_cached, key, buff, data_size))
+            {
+                Log(LOG_LEVEL_WARNING, "Can not read installed packages database "
+                    "for '%s' package module.", pm_name);
+                continue;
+            }
+            
+            Seq *packages_form_module = SeqStringFromString(buff, '\n');
+            free(buff);
+            
+            if (packages_form_module)
+            {
+                // Iterate over and see where match is.
+                for (int i = 0; i < SeqLength(packages_form_module); i++)
+                {
+                    char *line = SeqAt(packages_form_module, i);
+                    size_t new_line_size = strlen(line) + pm_name_size + 2; // we need coma and terminator
+                    char new_line[new_line_size];
+                    strcpy(new_line, line);
+                    strcat(new_line, ",");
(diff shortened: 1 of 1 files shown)


pasinskim:
@kacf can you take a look at this one?

pasinskim:
test this please!

kacf:
The proper phrase is...

kacf:
trigger build

kacf:
I'll take a look at it tomorrow morning!

libpromises/evalfunction.c
```diff
@@ -1300,27 +1300,45 @@ static FnCallResult FnCallBundlesMatching(EvalContext *ctx, const Policy *policy
...
+    {
+        Seq *list = SeqParseCsvString(line);
+        if (SeqLength(list) != 4)
         {
-            return FnFailure();
+            Log(LOG_LEVEL_ERR,
+                "Line from package inventory '%s' did not yield correct number of elements.",
+                line);
+            SeqDestroy(list);
+            return true;
```
kacf: Shouldn't this return `false` if the operation failed?

libpromises/evalfunction.c
```diff
@@ -1331,79 +1349,159 @@ static FnCallResult FnCallPackagesMatching(ARG_UNUSED EvalContext *ctx, ARG_UNUS
...
+            "new package promise and check for legacy promise is made."
             "A future call may still succeed. (fopen: %s)",
-            fp->name,
             installed_mode ? "installed" : "available",
             filename,
             GetErrorStr());
 
-        pcre_free(matcher);
-        return FnFailure();
+        return true;
```
kacf: And this too, should it not return `false` here?

libpromises/evalfunction.c
```diff
@@ -1331,79 +1349,159 @@ static FnCallResult FnCallPackagesMatching(ARG_UNUSED EvalContext *ctx, ARG_UNUS
...
-                linenumber, filename, strlen(line));
             free(line);
-            break; /* or continue ? */
+            break;
         }
+        free(line);
+    }
 
-        if (StringMatchFullWithPrecompiledRegex(matcher, line))
+    bool ret = feof(fin);
```
kacf: For consistency I would say `bool ret = (feof(fin) != 0);` here. `feof()` does not return a true bool. This also makes it C++ compatible, should we ever go down that route.

libpromises/evalfunction.c
```diff
@@ -1331,79 +1349,159 @@ static FnCallResult FnCallPackagesMatching(ARG_UNUSED EvalContext *ctx, ARG_UNUS
...
+static bool GetPackagesMatching(pcre *matcher, JsonElement *json, const bool installed_mode, Rlist *default_inventory)
+{
+    dbid database = (installed_mode == true ? dbid_packages_installed : dbid_packages_updates);
+
+    for (const Rlist *rp = default_inventory; rp != NULL; rp = rp->next)
+    {
+        const char *pm_name =  RlistScalarValue(rp);
+        size_t pm_name_size = strlen(pm_name);
+        
+        Log(LOG_LEVEL_DEBUG, "Reading packages (%d) for package modeule [%s]",
```
kacf: `s/modeule/module/`

libpromises/evalfunction.c
```diff
@@ -1331,79 +1349,159 @@ static FnCallResult FnCallPackagesMatching(ARG_UNUSED EvalContext *ctx, ARG_UNUS
...
+
+    for (const Rlist *rp = default_inventory; rp != NULL; rp = rp->next)
+    {
+        const char *pm_name =  RlistScalarValue(rp);
+        size_t pm_name_size = strlen(pm_name);
+        
+        Log(LOG_LEVEL_DEBUG, "Reading packages (%d) for package modeule [%s]", 
+                database, pm_name);
+        
+        if (StringSafeEqual(pm_name, "cf_null"))
```
kacf: Horrible. But yes, that's how lists are. :-)

libpromises/evalfunction.c
```diff
@@ -1331,79 +1349,159 @@ static FnCallResult FnCallPackagesMatching(ARG_UNUSED EvalContext *ctx, ARG_UNUS
...
-                    linenumber, filename, line);
-                ++linenumber;
-                SeqDestroy(list);
-                free(line);
+                Log(LOG_LEVEL_WARNING, "Can not read installed packages database "
+                    "for '%s' package module.", pm_name);
                 continue;
             }
+            
+            Seq *packages_form_module = SeqStringFromString(buff, '\n');
```
kacf: I this actually *from* module, and not *form* module? This was a bit confusing when I read it, so should be changed, I think.

libpromises/evalfunction.c
```diff
@@ -1331,79 +1349,159 @@ static FnCallResult FnCallPackagesMatching(ARG_UNUSED EvalContext *ctx, ARG_UNUS
...
+        const char *regex_package = RlistScalarValue(finalargs);
+        const char *regex_version = RlistScalarValue(finalargs->next);
+        const char *regex_arch = RlistScalarValue(finalargs->next->next);
+        const char *regex_method = RlistScalarValue(finalargs->next->next->next);
+        char regex[CF_BUFSIZE];
 
-            JsonArrayAppendObject(json, line_obj);
+        // Here we will truncate the regex if the parameters add up to over CF_BUFSIZE
+        snprintf(regex, sizeof(regex), "^%s,%s,%s,%s$",
+                 regex_package, regex_version, regex_arch, regex_method);
```
kacf: I don't think you can do this, I think you have to match each component separately. The reason is that there are some expressions that will not work like this, for example:

* Use of any anchoring operators (`^` and `$`). This is redundant of course, since the expression is already anchored, but it's still common to use them, and they are not always guaranteed to be at the front and/or back (when combined with subgroups, for example).
* Use of `|` operator. This will divide the expression into multiple parts and the comma separation will not work anymore. This might be solvable by enclosing the subexpressions in parenthesis though.

libpromises/evalfunction.c
```diff
@@ -1331,79 +1349,159 @@ static FnCallResult FnCallPackagesMatching(ARG_UNUSED EvalContext *ctx, ARG_UNUS
...
+            {
+                // Iterate over and see where match is.
+                for (int i = 0; i < SeqLength(packages_form_module); i++)
+                {
+                    char *line = SeqAt(packages_form_module, i);
+                    size_t new_line_size = strlen(line) + pm_name_size + 2; // we need coma and terminator
+                    char new_line[new_line_size];
+                    strcpy(new_line, line);
+                    strcat(new_line, ",");
+                    strcat(new_line, pm_name);
```
kacf: Are the package strings you got from `buff` already separated by commas? I also thought they were preceded by some prefix like `<A>_` for architecture? Maybe I'm misremembering though, but an explanation of the logic here would help, a good candidate for code comments.

libpromises/evalfunction.c
```diff
@@ -1300,27 +1300,45 @@ static FnCallResult FnCallBundlesMatching(EvalContext *ctx, const Policy *policy
 
 /*********************************************************************/
 
-static FnCallResult FnCallPackagesMatching(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const Policy *policy, const FnCall *fp, const Rlist *finalargs)
+static bool AddPackagesmatchingJsonLine(pcre *matcher, JsonElement *json, char *line)
```
kacf: Minor: For consistency I would do `s/AddPackagesmatchingJsonLine/AddPackagesMatchingJsonLine/`.

libpromises/evalfunction.c
```diff
@@ -1331,79 +1349,159 @@ static FnCallResult FnCallPackagesMatching(ARG_UNUSED EvalContext *ctx, ARG_UNUS
...
-    fclose(fin);
     pcre_free(matcher);
 
-    if (fail)
+    if (ret == false)
     {
         Log(LOG_LEVEL_ERR,
-            "Unable to read (%s) package inventory from '%s'.",
-            errstr, filename);
+            "%s Unable to read package inventory.", fp->name);
```
kacf: `s/%s Unable/%s: Unable/`

kacf:
Looking nice and tidy!

In addition to the code comments, I think this needs tests. Let me know if you need help with that.

libpromises/evalfunction.c
pasinskim: Actually not. In the original implementation, this was returning ```continue``` in the for loop whereas other ```fail``` conditions here were returning ```break```. I wanted to leave original behavior unchanged.

libpromises/evalfunction.c
pasinskim: This is similar to above. In some cases you can end up here (for example when new package promise is used and package module is not expanded yet - pre-evaluation). If we return ```false``` here we will see promise error logs which is not expected.

libpromises/evalfunction.c
pasinskim: Fixing!

libpromises/evalfunction.c
pasinskim: Thanks. Fixing!

libpromises/evalfunction.c
pasinskim: This is not my invention...

libpromises/evalfunction.c
pasinskim: Yup. This is misspelling. Fixing!

libpromises/evalfunction.c
pasinskim: This is how it was in the original implementation. I am afraid that if I will change anything here we might end up witch changed behavior which is not what we want to achieve here. 
The line was simply moved from one place to another.

libpromises/evalfunction.c
pasinskim: I will add a comment, but yes. The packages are spitted by the comma. 
In the old package promise we've been using CSVs for reporting, and I left it to make sure that it will be easy to integrate this into all places when we are using package promise. This is one place, where this decisions is paying off.

libpromises/evalfunction.c
pasinskim: Fixing!

libpromises/evalfunction.c
pasinskim: Fixing!

pasinskim:
If you have some spare time, help with tests will be appreciated. Thanks for the review.

libpromises/evalfunction.c
kacf: Ok, but at least put parenthesis around each `%s`, it's an easy fix, and will enable the `|` operator to work correctly.

pasinskim: I am always scarred to do such a fixes as we don't know exactly what weird policies people are writing. Maybe actually someone is relaying on the wrong behavior and after fixing it some stuff will be broken.

kacf: Perhaps put that fix in a commit only on master then? I think this behavior is clearly broken, so this will be a bugfix, not a behavior change IMHO. After all, the old method can give incorrect matches.

pasinskim:
Acceptance tests added here: https://github.com/cfengine/masterfiles/pull/694

PR #2579, opened 2016-04-13
Add "true" as a special expression that is always EVAL_EXPRESSION_TRUE

tzz:
@jimis this is what we discussed on IRC. See the acceptance test: only `from_true` is defined, the others (`false`, `null`, and a missing value) are false.

All other acceptance tests pass.

### DIFF
--- a/libpromises/logic_expressions.c
+++ b/libpromises/logic_expressions.c
@@ -250,8 +250,19 @@ ExpressionValue EvalExpression(const Expression *expr,
         {
             return EXPRESSION_VALUE_ERROR;
         }
+        else if (0 == strcmp("true", name))
+        {
+            ret =  EXPRESSION_VALUE_TRUE;
+        }
+        else if (0 == strcmp("false", name))
+        {
+            ret =  EXPRESSION_VALUE_FALSE;
+        }
+        else
+        {
+            ret = (*nameevalfn) (name, param);
+        }
 
-        ret = (*nameevalfn) (name, param);
         free(name);
         return ret;
     }
--- a/tests/acceptance/02_classes/01_basic/from_json_booleans.cf
+++ b/tests/acceptance/02_classes/01_basic/from_json_booleans.cf
@@ -1,0 +1,38 @@
+###########################################################
+#
+# Test expansion of JSON booleans
+#
+###########################################################
+
+body common control
+{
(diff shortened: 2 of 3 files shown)


jimis:
Thanks @tzz, this makes sense. Maybe we can also hard-code the "false" string to return `EXPRESSION_VALUE_FALSE`, no matter whether a "false" class is set? This shall be only when parsing expression strings, and I think it makes sense. @kacf what do you think?

In the end, it would be nice to update the [docs](https://docs.cfengine.com/docs/3.7/reference-promise-types-classes.html#expression), as they are quite frugal regarding the "class expresions". :-(

tzz:
@jimis your suggestion was added as a second commit.

The docs will be updated when this is accepted, I'll try to make them a little clearer.

jimis:
@nickanderson @kacf @estenberg this pull request alters behaviour of our [class expressions](https://docs.cfengine.com/docs/3.7/reference-promise-types-classes.html#expression) to evaluate the string "true" to true and the string "false" to false.

It's a behavioural change but nobody should be affected unless they were using "true" and "false" class names (hmmm, in which case we might warn them).

@tzz  please add "Changelog: behaviour change: blabblah" tag in one commit message.

libpromises/logic_expressions.c
```diff
@@ -250,8 +250,20 @@ ExpressionValue EvalExpression(const Expression *expr,
         {
             return EXPRESSION_VALUE_ERROR;
         }
+        else if (0 == strcmp("true", name))
+        {
+            ret =  EXPRESSION_VALUE_TRUE;
+        }
+        else if (0 == strcmp("false", name) ||
+                 0 == strcmp("null", name))
```
jimis: Why "null"?

tzz: In case we get it as a string. We shouldn't, I'll remove it.

tzz:
OK, rebased and squashed with a better commit message.

jimis:
+1 from me

trigger build

jimis:
All green, I'll merge tomorrow if nobody comments.

tzz:
All right, I'll write the docs after merge.

jimis:
Merged, thanks.

tzz:
@jimis documented in https://github.com/cfengine/documentation/pull/1453

PR #2580, opened 2016-04-14
Reverse order of binary lookup in testall.

kacf:
It's actually the last path that ends up being used, not the first,
and it is this order of priority we want. This was motivated by trying
to run acceptance tests in a directory not named core, but core was
present next to it.


kacf:
trigger build

PR #2581, opened 2016-04-14
Improve logging when managing setuid/setgid

amousset:
Setting the setgid or setuid bits on a file owned by
root is not an error and should trigger a warning instead.

In dry-run mode, log message should clearly state that no
action is taken.


kacf:
Thanks, makes sense!

kacf:
trigger build

kacf:

PR #2584, opened 2016-04-15
Parsing an empty file should give empty list, not no list at all.

kacf:
Changelog: Behavior change: When using readintlist(), readreallist()
or readstringlist(), parsing an empty file will no longer result in a
failed function call, but instead an empty list. Failure to open the
file will still result in a failed function call.


kacf:
This was discovered during writing of an acceptance test. I'm not adding tests explicitly, because this will be indirectly tested by one of my upcoming tests.

kacf:
@nickanderson, @estenberg: I think this makes sense, and is more intuitive. Do you agree?

It is a behavior change, so it will be in a feature release, most definitely.

kacf:
trigger build

PR #2588, opened 2016-04-18
CFE-2283: Add email filtering feature for cf-execd.

kacf:
Also add some tests for cf-execd, where we mock one function which is
the connection to the SMTP server. Instead we connect to a local pipe
and feed it data.

Changelog: New feature: Emails sent by cf-execd can be filtered to get
rid of emails for unwanted log messages. The attributes
`mailfilter_include` and `mailfilter_exclude` in `body executor
control` control what to include. See documentation for `cf-execd` for
more information.


kacf:
Note, this PR contains within it #2584.

kacf:
trigger build

kacf:
trigger build

kacf:
trigger build

kacf:
Now green in Jenkins!

kacf:
@pasinskim: I would appreciate a review of this if you have the time!

kacf:
trigger build

kacf:
trigger build

cf-execd/cf-execd-runner.c
```diff
@@ -713,30 +735,33 @@ static void MailResult(const ExecConfig *config, const char *file)
...
         {
-            break;
+            continue;
         }
-
-        Log(LOG_LEVEL_DEBUG, "Mail report: %s", vbuff);
-
-        if (strlen(vbuff) > 0)
+        if (send(sd, line, read, 0) == -1 ||
+            send(sd, "\r\n", 2, 0) == -1)
```
pasinskim: I can see that previously we've been sending just '\n'.

kacf: Indeed, but this was a bug. The RFC clearly states that lines should be separated by CRLF.

cf-execd/cf-execd-runner.h
```diff
@@ -29,5 +29,7 @@
 
 void LocalExec(const ExecConfig *config);
 
+extern int ConnectToSmtpSocket(const ExecConfig *config);
```
pasinskim: Do we need ```extern``` here?

kacf: Oh you're right, we don't. Part of an earlier experiment. Will remove it!

cf-execd/exec-config.c
```diff
@@ -51,6 +51,59 @@ static char *GetIpAddresses(const EvalContext *ctx)
     return StringWriterClose(ipbuf);
 }
 
+static void RegexFree(void *ptr)
+{
+    pcre_free(ptr);
+}
+
+static void MailFilterFill(const void *input, bool rlist,
```
pasinskim: This is rather matter of taste, but I would add two functions that you can call from inside ```FillList``` and ```FillSeq```. Having this you will avoid checking ```rlist``` condition over and over again.

kacf: Yeah, this part is rather ugly. I'll see if I can restructure it a bit.

tests/acceptance/25_cf-execd/cf-execd-rpl-functions.c
```diff
@@ -0,0 +1,126 @@
...
+    { \
+        Log(LOG_LEVEL_ERR, "Failed to read SMTP response in %s:%i", __FILE__, __LINE__); \
+        return NULL; \
+    } \
+    else \
+    { \
+        fwrite(X, rcvd, 1, stdout); \
+    }
+
+    char *sendbuf;
```
pasinskim: Looks like ```sendbuf``` is ```const char *```.

tests/acceptance/25_cf-execd/cf-execd-rpl-functions.c
```diff
@@ -0,0 +1,126 @@
...
+    char recvbuf[CF_BUFSIZE];
+    int rcvd;
+    sendbuf = "220 test.com\r\n";
+    COND_SEND(sendbuf);
```
pasinskim: Why you need a ```sendbuff``` at all? Shouldn't ```COND_SEND("220 test.com\r\n")``` be enough?

tests/acceptance/25_cf-execd/cf-execd-rpl-functions.c
```diff
@@ -0,0 +1,126 @@
...
+int ConnectToSmtpSocket(ARG_UNUSED const ExecConfig *config)
+{
+    int sockets[2];
+    if (socketpair(AF_UNIX, SOCK_STREAM, 0, sockets) < 0)
+    {
+        return -1;
+    }
+
+    int *thread_socket = xmalloc(sizeof(int));
+    *thread_socket = sockets[1];
```
pasinskim: Why we can not simply pass ```sockets + 1``` or ```&sockets[1]``` as a parameter to ```pthread_create``` below?

tests/acceptance/25_cf-execd/mail_exclude_filter.cf
```diff
@@ -0,0 +1,38 @@
+# Tests that exclude mail filters, causes everything but those lines to be
+# mailed by cf-execd.
```
pasinskim: The description is hard to understand. Maybe something like "Tests that exclude mail filter causes excluded lines are not mailed by cf-execd.".

tests/acceptance/25_cf-execd/cf-execd-rpl-functions.c
kacf: Yep, it is, will fix!

tests/acceptance/25_cf-execd/mail_include_and_exclude_filters.cf
```diff
@@ -0,0 +1,46 @@
...
+                            "WrongString3",
+                          };
+      "includes" slist => {
+                            "R: CorrectString1", ".*CorrectString[2]",
+                            "R: WrongString1", ".*WrongString[2]"
+                          };
+      "excludes" slist => {
+                            "R: WrongString1", ".*WrongString[2]",
+                            # Tests anchoring.
+                            "orrectString1"
```
pasinskim: Is "orrectString1" expected here?

tests/acceptance/25_cf-execd/mail_include_filter.cf
```diff
@@ -0,0 +1,38 @@
+# Tests that include mail filters, causes only those lines to be mailed by
+# cf-execd.
```
pasinskim: Maybe " Tests that include mail filter causes only matched lines are mailed by cf-execd."?

pasinskim:
Looks nice expect some rather minor issues. The change is big though so was hard to check deeply. Definitely you've put lot of effort in adding extensive tests which is great!!!

tests/acceptance/25_cf-execd/cf-execd-rpl-functions.c
kacf: We cannot pass `sockets + 1` or `&sockets[1]` because the variable is stack allocated, and this stack frame gets destroyed after we have returned from this function. We also cannot pass the int cast as a pointer because `sizeof(int) != sizeof(int*)` on 64-bit platforms. They are 4 and 8, respectively.

tests/acceptance/25_cf-execd/mail_exclude_filter.cf
kacf: Ok, will fix!

tests/acceptance/25_cf-execd/mail_include_filter.cf
kacf: Ok, will fix!

tests/acceptance/25_cf-execd/mail_include_and_exclude_filters.cf
kacf: Yes, we are testing that a partial match does not match. I can improve the comment a bit to clarify this.

cf-execd/exec-config.c
kacf: I split this into three functions, one for each of the `Seq` and `Rlist` parts, and one common one.

kacf:
Issues fixed and branch updated. I will run this once more through Jenkins.

kacf:
Thanks for the review, @pasinskim!

kacf:
trigger build

jimis:
Dammit @kacf, the effort you've put in testing in this pull request is really scary... :-)

kacf:
Hehe, still, I think it would have taken me even longer to test this manually. It was quite tricky to get it right, and the test helped me a lot (I was plagued by a single inverted boolean for a long time during the implementation).

tests/acceptance/Makefile.am
```diff
@@ -21,6 +21,9 @@
 # (COSL) may apply to this file if you as a licensee so wish it. See
 # included file COSL.txt.
 #
+
+SUBDIRS = 25_cf-execd
```
olehermanse: @kacf Why is this needed for cf-execd but no other acceptance tests?

kacf: Because we build the `cf-execd-test` binary there (there is a `Makefile.am` added there in this PR as well). The other tests don't build anything, and hence don't need this.

olehermanse: Ah, I think I saw that. Nice.

PR #2589, opened 2016-04-19
CFE-2161: Fix missing ps fields causing segfaults on Solaris.

kacf:
The problem is that the ucb style ps can return empty fields, so we
need to account for that:
- Abstract the choice of parsing algorithm so it can be easily
  specified when running SplitProcLine().
- Make ucb ps use the "empty fields" algorithm.
- When dealing with lines that can have empty fields, never leave a
  field NULL, even if the ps line is too short to cover all the
  headers. This fixes the segfault, since in the ucb parsing case, the
  return value of SplitProcLine() would not be used, and hence we
  would try to read the string afterwards.

Also enable some tests on all platforms that should now pass
everywhere.

### DIFF
--- a/libpromises/processes_select.c
+++ b/libpromises/processes_select.c
@@ -118,7 +118,8 @@ static bool SelectProcess(const char *procentry,
 
     if (!SplitProcLine(procentry, pstime, names, start, end, column))
     {
-        return false;
+        result = false;
+        goto cleanup;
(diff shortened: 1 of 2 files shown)


kacf:
trigger build

kacf:
This is green and is blocking the build, so I will merge this now.

PR #2591, opened 2016-04-21
Fix "No such file or directory" LMDB error on heavily loaded hosts.

kacf:
There is a race condition in LMDB that will fail to open the database
environment if another process is opening it at the exact same
time. This condition is signaled by returning ENOENT, which we should
never get otherwise. This can lead to error messages on a heavily
loaded machine, so try to open it again after allowing other threads
to finish their opening process.

Ref: CFE-2300

Changelog: Title

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1039,6 +1039,9 @@ AC_REPLACE_FUNCS(pthread_attr_setstacksize)
 AC_CHECK_DECLS(pthread_sigmask, [], [], [[#include <signal.h>]])
 AC_REPLACE_FUNCS(pthread_sigmask)
 
+AC_CHECK_DECLS(sched_yield, [], [], [[#include <sched.h>]])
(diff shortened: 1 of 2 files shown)


kacf:
https://tracker.mender.io/browse/CFE-2300

jimis:
+1

kacf:
trigger build

kacf:
Hmm, I suspected that `pthread_yield` might not be available on many systems. I can make it conditional though, it should still work without it, it will just be a closer race.

jimis:
If it's not defined try using `sched_yield(2)`.

kacf:
Alright, trying `sched_yield`, seems to be available everywhere but Windows.

kacf:
trigger build

jimis:
Even a sleep/nanosleep should do.

kacf:

PR #2593, opened 2016-04-25
CFE-2339 Don't accept old protocol by default.

pasinskim:
Stop accepting clients using non TLS connections by default.

Changelog: Clients connections using non TLS protocol are rejected
by default. (CFE-2339).


pasinskim:
trigger build!

kacf:
I believe this is green in Jenkins (there is some redness, but I believe it's related to other failures in masterfiles from this morning).

kacf:
Uhm, forget that actually, I was looking at the wrong one. It doesn't look like this ran at all. I will retry.

kacf:
trigger build

PR #2595, opened 2016-04-27
Fixed unportable tarball caused by too long filenames.

kacf:
And add a test to try to prevent this in the future before we get to
the packaging stage.


kacf:
@jimis

kacf:
FYI, I pushed this directly to 3.9 since I need to test if it helps the package build.

jimis:
Thanks for adding the test. But I think the right way to test would be the by expanding the `am__tar` macro and see if there are errors creating a tarball in all platforms.

Hmmm on a second thought, I don't know if this test is a good idea after all. First of all it doesn't test the full path because the patch is not prefixed with `cfengine-3.9.0b1/` for example. Secondly it's a bit on the heavy side for a unit test - I/O, CPU etc.

Maybe the right approach is to fix `make dist` to return an error if `am__tar` fails.

kacf:
You're right, I added this test before we knew what the real problem was. Ideally this should be handled even better. But it doesn't do any harm in having it, at least it provides an early warning, even if it's not full proof.

The performance I'm not so worried about, it ran very fast on my system, after all this test doesn't gzip, it only tars, which is 100% copy operation. For comparison, the masterfiles acceptance tests install a full copy of masterfiles in several tests, which is of comparable size.

PR #2597, opened 2016-04-27
CFE-2240: enable datatype() and is_datatype() functions

tzz:
As in https://tracker.mender.io/browse/CFE-2240

Requested by @nickanderson @tzz and others

Full acceptance test included.


tests/acceptance/01_vars/02_functions/datatype.cf.expected.json
```diff
@@ -0,0 +1,123 @@
...
+  "datatype[i]": "primitive int",
+  "datatype[il]": "list ilist",
+  "datatype[init.container[array]]": "data array",
+  "datatype[init.container[false]]": "data boolean",
+  "datatype[init.container[int]]": "data number",
+  "datatype[init.container[null]]": "data null",
+  "datatype[init.container[real]]": "data number",
+  "datatype[init.container[string]]": "data string",
+  "datatype[init.container[true]]": "data boolean",
+  "datatype[init.container]": "data object",
```
nickanderson: What type would the datatype for a classic array string be?

tzz: "primitive string" or "list slist" or whatever was used originally, because arrays are just a regular variable with brackets in the name.

tzz:
@jimis small PR, very useful for libraries that need to take several kinds of inputs. Right now you don't know in advance if you have a container or not.

tzz:
@jimis rebased, tests pass, ready for review

libpromises/evalfunction.c
```diff
@@ -4429,70 +4429,123 @@ static FnCallResult FnCallFold(EvalContext *ctx,
...
+    }
+
+    if (NULL == major || NULL == minor)
+    {
+        Log(LOG_LEVEL_VERBOSE, "%s: the data type of variable '%s' could not be found", fp->name, varname);
         return FnFailure();
     }
 
-    return FnReturnNoCopy(StringWriterClose(typestring));
+    if (check_mode)
```
jimis: This is not reusing any part of the function, so why nest it in here and not create a new one?

tests/acceptance/01_vars/02_functions/datatype.cf.expected.json
jimis: I'm not really convinced about this "major minor" breakdown of types, since it's purely artificial. The only type that really has subtypes is the container one, from a C point of view.

@nickanderson do you think this "major minor" return code makes sense?

Furthermore I see this function was pre-existing and was ifdef'd out. Do you guys know if there was an old version at some point, behaving differently?

tests/acceptance/01_vars/02_functions/datatype.cf
```diff
@@ -0,0 +1,73 @@
...
+{
+  vars:
+      "s" string => "foo";
+      "i" int => "200";
+      "r" real => "300.0";
+
+      "sl" slist => { "a" };
+      "il" ilist => { "1" };
+      "rl" rlist => { "1.0" };
+
```
nickanderson: Shouldn't we have classic arrays represented here as well?

nickanderson:
@jimis I agree that these functions would be useful for framework and library maintainers. They are useful when you want to validate the data you are operating on. Currently we don't really have a way to do that within the policy language itself beyond regular expression comparison but that is limited to a data element, and can't help you to know if it was a string, list or data container.

The breakdown of major/minor does seem a bit odd, however I think it does make some sense given how classic arrays are handled.

nickanderson:
@jimis I think that these functions may have been used internally at one point. Perhaps that's why they were ifdefed.

libpromises/evalfunction.c
tzz: I thought it was cleaner.

tests/acceptance/01_vars/02_functions/datatype.cf
tzz: grrrr :)

tzz:
Comments:

* major vs. minor distinction: I think it's useful to policy writers and leaves the door open for future changes. This is a tricky one, but I think I've done it correctly to accommodate past and future data types.
* commented out: we agreed to leave it out of 3.6 IIRC because there wasn't a strong use case. Now there is (discussions on mailing list and IRC), so this is just resurrecting code from a few years ago without changing it significantly.

jimis:
OK lets move this forward then.

@tzz please take care of the stylistic issues pointed earlier on. I would also suggest to keep the "major minor" distinction only for what there is truly a distinction, currently only `data`. This will leave the subtype available for JSON data and possible future types. What I mean is that for traditional types, only one word should be output:

```
ilist
rlist
slist
int
real
string
```

etc. Only when something is of type `CF_DATA_TYPE_CONTAINER` should it display `data` as 1st word and something additional as a subtype.

This is simpler to implement, and is not lying about the hierarchy of types internally.

tzz:
@jimis my original proposal was cleaner because **everything** had a major and a minor data type. You could ask about lists without caring if they were ilists, slists, or rlists. You can find out if something is a primitive without caring about the specific data type.

I've pushed a rebased version with no minor types for primitives and lists, but I think it's less useful. Please reconsider.

tzz:
ping... happy birthday to this PR!

jimis:
Sorry @tzz, this got neglected together with the complex pull requests. But this not a complex change, so it can still make it in if you want, the code is good.

tzz:
@jimis feel free to grab the branch and bring it in, any changes you or others want are OK with me. I just don't have the cycles to work on it, unfortunately.

jimis:
Sorry no spare cycles for this one, if anyone wants to reopen it (maybe @nickanderson) I'm fine to move it forward.

PR #2598, opened 2016-04-28
Revert removed tests for both classing and TLS protocol.

pasinskim:
After removing "protocol_version" all the tests were run with
classic protocol only. This is bringing back testing using both
classic and TLS protocols.


pasinskim:
trigger build!

pasinskim:
trigger build

jimis:
build is green

PR #2599, opened 2016-04-28
Change: Adjust changelog markup to work with doc builds

nickanderson:


kacf:
Ah, so no markup in changelogs.

PR #2601, opened 2016-04-29
Improve ifelse() test

tzz:
this is partly for https://tracker.mender.io/browse/CFE-2357 but also just to clean up the test


jimis:
-1 for adding JSON file
+2 for removing the "secondpass" nightmare

trigger build

tzz:
@jimis so you're positive about it? ;)

jimis:
Indeed `1>0` :-)

jimis:
Merged, thanks.

PR #2602, opened 2016-05-02
Change: Don't load plucked.cf.sub when testing masterfiles

nickanderson:
So that tests in the masterfiles policy framework can leverege the
standard library directly.

Ref: Jira #CFE-2331


nickanderson:
trigger build

PR #2604, opened 2016-05-03
Get rid of cf_null in the serialization of empty slists to JSON

tzz:
@jimis this removes `cf_null` in the conversion of empty slists to JSON. It improves acceptance tests and cleans up a nasty corner of the code.

Only the `join.cf` test fails, and that's because of the conversion of slists outside of the JSON serialization code. So this PR will need to be merged with the work @jimis is doing.


jimis:
Thanks @tzz, I'll be needing this soon.

jimis:
I'm merging this in my own huge commit, so attribution to you will be lost @tzz, hope it's OK.

libpromises/evalfunction.c
```diff
@@ -248,10 +248,11 @@ static JsonElement* VarRefValueToJson(EvalContext *ctx, const FnCall *fp, const
             convert = JsonArrayCreate(RlistLen(value));
             for (const Rlist *rp = value; rp != NULL; rp = rp->next)
             {
-                if (rp->val.type == RVAL_TYPE_SCALAR &&
-                    strcmp(RlistScalarValue(rp), CF_NULL_VALUE) != 0)
+                char* item = RlistScalarValueSafe(rp);
```
jimis: I'm removing the _safe_ part here as we want it to crash if it turns out not being a SCALAR after all (programming error indeed).

Hmmm on a second look I'm removing the whole `if` clause, since I don't want any special handling for `cf_null`.

jimis: Hmmm what if it is an `ilist`, i.e. a list of integers? Then the value is skipped completely! Do we have examples that mix JSON and ilists?

tzz: Internally `ilist` is a list of strings also. They are **not** well supported in tests and in the language itself and generally just turn out to be strings. Here's an example that shows it:

```
bundle agent main
{
  vars:
      "ints" ilist => { 1, 2, 3 };
      "ints_to_json" data => '[ ints ]';
      "ints_to_json_string" string => format("%S", ints_to_json);

  reports:
      "converted ints to ${ints_to_json_string}";
}
```

The output:

```console
% cf-agent -KI -f ./test_ilist_to_json.cf                                                                                                      
R: converted ints to [["1","2","3"]]
```

Feel free to turn this into an acceptance test but I'm not sure what it proves.

> Hmmm on a second look I'm removing the whole `if` clause, since I don't want any special handling for `cf_null`.

I think you do want to kill `cf_null` if it's the only thing in the list, unless you have cleaned up every place in the code where an empty list gets converted to `{ cf_null }`. I didn't do that in my patch :)

tzz:
@jimis don't worry about attribution, my commit was trivial code, mostly acceptance test fixes anyhow.

jimis:
This Pull Request has been Assimilated; Closing.

jimis:
> Internally ilist is a list of strings also.

Really? I was under the impression that `CF_DATA_TYPE_INT_LIST` and `CF_DATA_TYPE_REAL_LIST` were made for exactly these cases. It would be strange if they are unused.

PR #2605, opened 2016-05-03
storage: Properly initialize the list of current mounts CFE-1803

geofft:
In commit 35523330, the sense of this check was inadvertently changed
from "If the list is empty, and LoadMountInfo fails, report an error" to
"If the list is non-empty, and LoadMountInfo fails, report an error".
This caused LoadMountInfo to never get called, so cf-agent assumed that
nothing was mounted and therefore that nothing needed to be unmounted.

### DIFF
--- a/cf-agent/verify_storage.c
+++ b/cf-agent/verify_storage.c
@@ -132,7 +132,7 @@ PromiseResult VerifyStoragePromise(EvalContext *ctx, char *path, const Promise *
     PromiseResult result = PROMISE_RESULT_NOOP;
 
 #ifndef __MINGW32__
-    if ((SeqLength(GetGlobalMountedFSList())) && (!LoadMountInfo(GetGlobalMountedFSList())))
+    if ((!SeqLength(GetGlobalMountedFSList())) && (!LoadMountInfo(GetGlobalMountedFSList())))
     {
         Log(LOG_LEVEL_ERR, "Couldn't obtain a list of mounted filesystems - aborting");
         YieldCurrentLock(thislock);


geofft:
Ping - I think this is straightforward, anything I need to do to get the fix accepted? Thanks!

tzz:
I'm not able to approve this but it seems very reasonable. If you can write a test for this, it will be even more likely to get merged.

cf-agent/verify_storage.c
```diff
@@ -132,7 +132,7 @@ PromiseResult VerifyStoragePromise(EvalContext *ctx, char *path, const Promise *
     PromiseResult result = PROMISE_RESULT_NOOP;
 
 #ifndef __MINGW32__
-    if ((SeqLength(GetGlobalMountedFSList())) && (!LoadMountInfo(GetGlobalMountedFSList())))
+    if ((!SeqLength(GetGlobalMountedFSList())) && (!LoadMountInfo(GetGlobalMountedFSList())))
```
jimis: Since SeqLength() does not return bool, please use explicit numeric comparison.

jimis:
trigger build

jimis:
Thanks @geofft, the fix looks correct, will be merged once the build succeeds and you fix the minor coding style issue.

Which versions do you know are affected? Is there an open ticket for this one?

jimis:
Merged manually with small modification in style and a "Changelog" tag in the commit message. Thanks!

geofft:
Oops, thanks for taking care of the fixup and merging it!

The commit that introduced the inverted logic was in 3.6.0 but not earlier versions, so I think it affects all versions from 3.6 to present.

PR #2606, opened 2016-05-09
ENT-2737: Fix failsafe.cf not populating modules directory.

kacf:
This has become important after the switch to the new package promise
for inventory by default, since it will immediately start looking for
the package modules.

### DIFF
--- a/libpromises/failsafe.cf
+++ b/libpromises/failsafe.cf
@@ -114,6 +114,14 @@ bundle agent failsafe_cfe_internal_update
         depth_search => failsafe_recurse("inf"),
         classes => failsafe_results("namespace", "inputdir_update");
 
+    !policy_server::
+
+      "$(sys.workdir)/modules"
+        handle => "failsafe_cfe_internal_bootstrap_update_files_sys_workdir_modules_shortcut",
+        copy_from => failsafe_scp("modules"),
+        depth_search => failsafe_recurse("inf"),
+        classes => failsafe_results("namespace", "modulesdir_update");
+
     !windows.inputdir_update_error::
 
       # When running on a *nix platform with homogeneous packages


libpromises/failsafe.cf
```diff
@@ -114,6 +118,12 @@ bundle agent failsafe_cfe_internal_update
         depth_search => failsafe_recurse("inf"),
         classes => failsafe_results("namespace", "inputdir_update");
 
+      "$(sys.workdir)/modules"
```
pasinskim: Do we need this to run on hub as well? Or we should have ```!policy_server::``` class guard?

libpromises/failsafe.cf
```diff
@@ -114,6 +118,12 @@ bundle agent failsafe_cfe_internal_update
...
+        handle => "failsafe_cfe_internal_bootstrap_update_files_sys_workdir_modules_shortcut",
```
pasinskim: Can we have some different name here? This is telling nothing. Maybe ```failsafe_cfe_internal_update_modules```?

libpromises/failsafe.cf
kacf: Oh right, good catch! On the hub the source and destination are the same, so we should not attempt a copy there.

libpromises/failsafe.cf
kacf: I copied this from the `inputdir` section. I have no opinion on what it should be. Maybe @nickanderson does?

libpromises/failsafe.cf
kacf: Fixed!

pasinskim:
Looks good. Feel free to merge it.

kacf:

PR #2607, opened 2016-05-09
updated the bundle name in compare.cf

dheerajkabra:
There was a _bundle name_ mismatch under `example/compare.cf`. Hence corrected it.


jimis:
Thanks!

jimis:
I also amended the `compare.cf` example have an `example_output` section so that it's checked during `make check`.

PR #2608, opened 2016-05-09
ENT-2738: Tolerate more than one space in dates in ps output.

kacf:
HP-UX like to produce dates space padded to two digits, like so:
"May  4".


kacf:
trigger build

kacf:

PR #2609, opened 2016-05-09
Buffer size checks

estenberg:
For some functions that write to output buffers they take as parameters, add an output buffer size parameter as well and and check it. Also add some checks to ensure we are not writing past some buffers that potentially could create instability or security risks. Some unused functions were removed.

Please backport to 3.9 once merged.

### DIFF
--- a/cf-agent/files_editline.c
+++ b/cf-agent/files_editline.c
@@ -208,9 +208,9 @@ Bundle *MakeTemporaryBundleFromTemplate(EvalContext *ctx, Policy *policy, Attrib
             // Get Action operator
             if (strncmp(buffer, "[%CFEngine", strlen("[%CFEngine")) == 0)
             {
-                char op[CF_BUFSIZE], brack[CF_SMALLBUF];
+                char op[CF_BUFSIZE], brack[4];
 
-                sscanf(buffer+strlen("[%CFEngine"), "%1024s %s", op, brack);
+                sscanf(buffer+strlen("[%CFEngine"), "%1024s %3s", op, brack);
 
                 if (strcmp(brack, "%]") != 0)
                 {
--- a/libcfnet/client_code.c
+++ b/libcfnet/client_code.c
@@ -359,8 +359,8 @@ Item *RemoteDirList(const char *dirname, bool encrypt, AgentConnection *conn)
         if (encrypt)
         {
             memcpy(in, recvbuffer, nbytes);
-            DecryptString(conn->encryption_type, in, recvbuffer,
-                          conn->session_key, nbytes);
+            DecryptString(recvbuffer, sizeof(recvbuffer), in, nbytes,
+                          conn->encryption_type, conn->session_key);
         }
 
         if (recvbuffer[0] == '\0')
--- a/libcfnet/client_code.c
+++ b/libcfnet/client_code.c
@@ -320,9 +320,17 @@ Item *RemoteDirList(const char *dirname, bool encrypt, AgentConnection *conn)
 
         snprintf(in, CF_BUFSIZE, "OPENDIR %s", dirname);
         cipherlen = EncryptString(out, sizeof(out), in, strlen(in) + 1, conn->encryption_type, conn->session_key);
+
+        tosend = cipherlen + CF_PROTO_OFFSET;
+
+        if(tosend > sizeof(sendbuffer))
+        {
+            ProgrammingError("RemoteDirList: tosend (%d) > sendbuffer (%ld)",
+                             tosend, sizeof(sendbuffer));
+        }
+
         snprintf(sendbuffer, CF_BUFSIZE - 1, "SOPENDIR %d", cipherlen);
         memcpy(sendbuffer + CF_PROTO_OFFSET, out, cipherlen);
-        tosend = cipherlen + CF_PROTO_OFFSET;
     }
     else
     {
--- a/libcfnet/client_code.c
+++ b/libcfnet/client_code.c
@@ -444,9 +444,17 @@ int CompareHashNet(const char *file1, const char *file2, bool encrypt, AgentConn
             EncryptString(out, sizeof(out), in,
                           strlen(in) + CF_SMALL_OFFSET + CF_DEFAULT_DIGEST_LEN,
                           conn->encryption_type, conn->session_key);
+
+        tosend = cipherlen + CF_PROTO_OFFSET;
+
+        if(tosend > sizeof(sendbuffer))
+        {
+            ProgrammingError("CompareHashNet: tosend (%d) > sendbuffer (%ld)",
+                             tosend, sizeof(sendbuffer));
+        }
+
         snprintf(sendbuffer, CF_BUFSIZE, "SMD5 %d", cipherlen);
         memcpy(sendbuffer + CF_PROTO_OFFSET, out, cipherlen);
-        tosend = cipherlen + CF_PROTO_OFFSET;
     }
     else
     {
--- a/libcfnet/client_code.c
+++ b/libcfnet/client_code.c
@@ -319,7 +319,7 @@ Item *RemoteDirList(const char *dirname, bool encrypt, AgentConnection *conn)
         }
 
         snprintf(in, CF_BUFSIZE, "OPENDIR %s", dirname);
-        cipherlen = EncryptString(out, in, strlen(in) + 1, conn->encryption_type, conn->session_key);
+        cipherlen = EncryptString(out, sizeof(out), in, strlen(in) + 1, conn->encryption_type, conn->session_key);
         snprintf(sendbuffer, CF_BUFSIZE - 1, "SOPENDIR %d", cipherlen);
         memcpy(sendbuffer + CF_PROTO_OFFSET, out, cipherlen);
         tosend = cipherlen + CF_PROTO_OFFSET;
@@ -441,7 +441,7 @@ int CompareHashNet(const char *file1, const char *file2, bool encrypt, AgentConn
         }
 
         cipherlen =
-            EncryptString(out, in,
+            EncryptString(out, sizeof(out), in,
                           strlen(in) + CF_SMALL_OFFSET + CF_DEFAULT_DIGEST_LEN,
                           conn->encryption_type, conn->session_key);
         snprintf(sendbuffer, CF_BUFSIZE, "SMD5 %d", cipherlen);
@@ -570,7 +570,7 @@ int EncryptCopyRegularFileNet(const char *source, const char *dest, off_t size, 
     EVP_CIPHER_CTX_init(&crypto_ctx);
 
     snprintf(in, CF_BUFSIZE - CF_PROTO_OFFSET, "GET dummykey %s", source);
-    cipherlen = EncryptString(out, in, strlen(in) + 1, conn->encryption_type, conn->session_key);
+    cipherlen = EncryptString(out, sizeof(out), in, strlen(in) + 1, conn->encryption_type, conn->session_key);
     snprintf(workbuf, CF_BUFSIZE, "SGET %4d %4d", cipherlen, blocksize);
     memcpy(workbuf + CF_PROTO_OFFSET, out, cipherlen);
     tosend = cipherlen + CF_PROTO_OFFSET;
--- a/libcfnet/client_code.c
+++ b/libcfnet/client_code.c
@@ -587,9 +587,17 @@ int EncryptCopyRegularFileNet(const char *source, const char *dest, off_t size, 
 
     snprintf(in, CF_BUFSIZE - CF_PROTO_OFFSET, "GET dummykey %s", source);
     cipherlen = EncryptString(out, sizeof(out), in, strlen(in) + 1, conn->encryption_type, conn->session_key);
+
+    tosend = cipherlen + CF_PROTO_OFFSET;
+
+    if(tosend > sizeof(workbuf))
+    {
+        ProgrammingError("EncryptCopyRegularFileNet: tosend (%d) > workbuf (%ld)",
+                         tosend, sizeof(workbuf));
+    }
+
     snprintf(workbuf, CF_BUFSIZE, "SGET %4d %4d", cipherlen, blocksize);
     memcpy(workbuf + CF_PROTO_OFFSET, out, cipherlen);
-    tosend = cipherlen + CF_PROTO_OFFSET;
 
 /* Send proposition C0 - query */
 
--- a/libcfnet/client_code.c
+++ b/libcfnet/client_code.c
@@ -319,7 +319,7 @@ Item *RemoteDirList(const char *dirname, bool encrypt, AgentConnection *conn)
         }
 
         snprintf(in, CF_BUFSIZE, "OPENDIR %s", dirname);
-        cipherlen = EncryptString(conn->encryption_type, in, out, conn->session_key, strlen(in) + 1);
+        cipherlen = EncryptString(out, in, strlen(in) + 1, conn->encryption_type, conn->session_key);
         snprintf(sendbuffer, CF_BUFSIZE - 1, "SOPENDIR %d", cipherlen);
         memcpy(sendbuffer + CF_PROTO_OFFSET, out, cipherlen);
         tosend = cipherlen + CF_PROTO_OFFSET;
@@ -441,8 +441,9 @@ int CompareHashNet(const char *file1, const char *file2, bool encrypt, AgentConn
         }
 
         cipherlen =
-            EncryptString(conn->encryption_type, in, out, conn->session_key,
-                          strlen(in) + CF_SMALL_OFFSET + CF_DEFAULT_DIGEST_LEN);
+            EncryptString(out, in,
+                          strlen(in) + CF_SMALL_OFFSET + CF_DEFAULT_DIGEST_LEN,
+                          conn->encryption_type, conn->session_key);
         snprintf(sendbuffer, CF_BUFSIZE, "SMD5 %d", cipherlen);
         memcpy(sendbuffer + CF_PROTO_OFFSET, out, cipherlen);
         tosend = cipherlen + CF_PROTO_OFFSET;
@@ -569,7 +570,7 @@ int EncryptCopyRegularFileNet(const char *source, const char *dest, off_t size, 
     EVP_CIPHER_CTX_init(&crypto_ctx);
 
     snprintf(in, CF_BUFSIZE - CF_PROTO_OFFSET, "GET dummykey %s", source);
-    cipherlen = EncryptString(conn->encryption_type, in, out, conn->session_key, strlen(in) + 1);
+    cipherlen = EncryptString(out, in, strlen(in) + 1, conn->encryption_type, conn->session_key);
     snprintf(workbuf, CF_BUFSIZE, "SGET %4d %4d", cipherlen, blocksize);
     memcpy(workbuf + CF_PROTO_OFFSET, out, cipherlen);
     tosend = cipherlen + CF_PROTO_OFFSET;
--- a/libcfnet/stat_cache.c
+++ b/libcfnet/stat_cache.c
@@ -145,8 +145,8 @@ int cf_remote_stat(AgentConnection *conn, bool encrypt, const char *file,
 
         snprintf(in, CF_BUFSIZE - 1, "SYNCH %jd STAT %s",
                  (intmax_t) tloc, file);
-        int cipherlen = EncryptString(conn->encryption_type, in, out,
-                                      conn->session_key, strlen(in) + 1);
+        int cipherlen = EncryptString(out, in, strlen(in) + 1,
+                                      conn->encryption_type, conn->session_key);
         snprintf(sendbuffer, CF_BUFSIZE - 1, "SSYNCH %d", cipherlen);
         memcpy(sendbuffer + CF_PROTO_OFFSET, out, cipherlen);
         tosend = cipherlen + CF_PROTO_OFFSET;
--- a/libcfnet/stat_cache.c
+++ b/libcfnet/stat_cache.c
@@ -145,7 +145,7 @@ int cf_remote_stat(AgentConnection *conn, bool encrypt, const char *file,
 
         snprintf(in, CF_BUFSIZE - 1, "SYNCH %jd STAT %s",
                  (intmax_t) tloc, file);
-        int cipherlen = EncryptString(out, in, strlen(in) + 1,
+        int cipherlen = EncryptString(out, sizeof(out), in, strlen(in) + 1,
                                       conn->encryption_type, conn->session_key);
         snprintf(sendbuffer, CF_BUFSIZE - 1, "SSYNCH %d", cipherlen);
         memcpy(sendbuffer + CF_PROTO_OFFSET, out, cipherlen);
--- a/libcfnet/stat_cache.c
+++ b/libcfnet/stat_cache.c
@@ -32,6 +32,7 @@
 #include <alloc.h>                            /* xmemdup */
 #include <logging.h>                          /* Log */
 #include <crypto.h>                           /* EncryptString */
+#include <misc_lib.h>                         /* ProgrammingError */
 
 static void NewStatCache(Stat *data, AgentConnection *conn)
 {
@@ -147,9 +148,17 @@ int cf_remote_stat(AgentConnection *conn, bool encrypt, const char *file,
                  (intmax_t) tloc, file);
         int cipherlen = EncryptString(out, sizeof(out), in, strlen(in) + 1,
                                       conn->encryption_type, conn->session_key);
+
+        tosend = cipherlen + CF_PROTO_OFFSET;
+
+        if(tosend > sizeof(sendbuffer))
+        {
+            ProgrammingError("cf_remote_stat: tosend (%d) > sendbuffer (%ld)",
+                             tosend, sizeof(sendbuffer));
+        }
+
         snprintf(sendbuffer, CF_BUFSIZE - 1, "SSYNCH %d", cipherlen);
         memcpy(sendbuffer + CF_PROTO_OFFSET, out, cipherlen);
-        tosend = cipherlen + CF_PROTO_OFFSET;
     }
     else
     {
--- a/libpromises/addr_lib.c
+++ b/libpromises/addr_lib.c
@@ -42,12 +42,6 @@ int FuzzySetMatch(const char *s1, const char *s2)
     short isCIDR = false, isrange = false, isv6 = false, isv4 = false;
     char address[CF_ADDRSIZE];
 
-    if (strlen(s1) >= CF_MAX_IP_LEN)
-    {
-        Log(LOG_LEVEL_ERR, "IP set string '%s' is too large to parse (>=%d)", s1, CF_MAX_IP_LEN);
-        return -1;
-    }
-
     if (strcmp(s1, s2) == 0)
     {
         return 0;
--- a/libpromises/addr_lib.c
+++ b/libpromises/addr_lib.c
@@ -157,7 +157,7 @@ int FuzzySetMatch(const char *s1, const char *s2)
         else
         {
             long i, from = -1, to = -1, cmp = -1;
-            char buffer1[CF_MAX_IP_LEN], buffer2[CF_MAX_IP_LEN];
+            char buffer1[64], buffer2[64];
 
             const char *sp1 = s1;
             const char *sp2 = s2;
@@ -165,7 +165,8 @@ int FuzzySetMatch(const char *s1, const char *s2)
             for (i = 0; i < 4; i++)
             {
                 buffer1[0] = '\0';
-                sscanf(sp1, "%[^.]", buffer1);
+                sscanf(sp1, "%63[^.]", buffer1);
+                buffer1[63] = '\0';
 
                 if (strlen(buffer1) == 0)
                 {
@@ -173,7 +174,10 @@ int FuzzySetMatch(const char *s1, const char *s2)
                 }
 
                 sp1 += strlen(buffer1) + 1;
-                sscanf(sp2, "%[^.]", buffer2);
+
+                sscanf(sp2, "%63[^.]", buffer2);
+                buffer2[63] = '\0';
+
                 sp2 += strlen(buffer2) + 1;
 
                 if (strstr(buffer1, "-"))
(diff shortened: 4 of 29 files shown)


libutils/string_lib.c
```diff
@@ -747,8 +747,14 @@ char *EscapeCharCopy(const char *str, char to_escape, char escape_with)
...
-    int in_size = strlen(str);
-    int out_size = in_size + CountChar(str, to_escape) + 1;
+    size_t in_size = strlen(str);
+
+    if(in_size > (SIZE_MAX / 2) - 1)
+    {
+        ProgrammingError("Buffer passed to EscapeCharCopy() too large (in_size=%ld)", in_size);
+    }
+
+    size_t out_size = in_size + CountChar(str, to_escape) + 1;
```
kacf: Hehe, nothing wrong with this, but extremely theoretical: You'd need more than 9 million TB of memory before hitting this!

(on 64-bit, that is)

libpromises/locks.c
```diff
@@ -889,7 +889,7 @@ void GetLockName(char *lockname, const char *locktype, const char *base, const R
     strncpy(lockname, locktype, CF_BUFSIZE / 10);
     strcat(lockname, "_");
     strncat(lockname, base, CF_BUFSIZE / 10);
-    strcat(lockname, "_");
+    strlcat(lockname, "_", CF_BUFSIZE / 10);
```
kacf: There is another one right above this one, should that be covered too?

kacf: Oh, and while you're at it, switch them all to `l` version, which is better than `n` version. I forget the exact details, but the `n` version doesn't include the null byte in the calculation or something along those lines.

libutils/buffer.c
```diff
@@ -427,38 +427,6 @@ int BufferPrintf(Buffer *buffer, const char *format, ...)
     return printed;
 }
 
-int BufferVPrintf(Buffer *buffer, const char *format, va_list ap)
```
kacf: I think we should leave this one. It's part of the API of Buffer, and this might be used later.

libpromises/crypto.c
```diff
@@ -443,6 +443,14 @@ int EncryptString(char *out, size_t out_length, const char *in, int plainlen,
...
         ProgrammingError("EncryptString: session key == NULL");
 
+    size_t max_ciphertext_length = CipherTextLengthMax(CfengineCipher(type), plainlen);
+
+    if(max_ciphertext_length > out_length)
+    {
+        ProgrammingError("EncryptString: output buffer too small: max_ciphertext_length (%ld) > out_length (%ld)",
+                          max_ciphertext_length, out_length);
+    }
+
```
kacf: This commit should be squashed together with:

* Add utiltiy function CipherTextLengthMax()...
* Pass output buffer size to EncryptString()...

libpromises/crypto.c
```diff
@@ -534,6 +534,14 @@ int DecryptString(char *out, size_t out_length, const char *in, int cipherlen,
...
         ProgrammingError("DecryptString: session key == NULL");
 
+    size_t max_plaintext_length = PlainTextLengthMax(CfengineCipher(type), cipherlen);
+
+    if(max_plaintext_length > out_length)
+    {
+        ProgrammingError("DecryptString: output buffer too small: max_plaintext_length (%ld) > out_length (%ld)",
+                          max_plaintext_length, out_length);
+    }
+
```
kacf: Same here, this commit should be squashed together with:

* Add utiltiy function PlainTextLengthMax()...
* Rearrange parameters of DecryptString()...
* Pass output buffer size to DecryptString()...

kacf:
Nice, I've gone through all three. Once you've fixed the issues I can run them all together on a private branch!

libutils/string_lib.c
estenberg: Future-proofing, I guess. :)

libpromises/locks.c
estenberg: Yes, strncat() does not guarantee \0 termination, strlcat() does. Fixed.

libutils/buffer.c
estenberg: OK, I will remove this commit, leaving the function there.

The concern is that it takes format as a parameter, which *could* be a security risk if the caller allows an attacker to inject data there. I added a comment instead.

libpromises/crypto.c
estenberg: Ack, done!

libpromises/crypto.c
estenberg: Ack, done!

estenberg:
Commits updated, please let me know if there is anything else required!

estenberg:
Oh, and please note that I do not have a great developer environment, so there might be whitespace/indentation issues (although I tried to be very careful about this).

libutils/buffer.c
kacf: Yes, but all printf-style functions suffer from this, and yet exploits are quite rare (usually `format` is a static string).

kacf:
Whitespace/indentation looked beautiful as far as I could see! :-)

I started a test build on a private branch [here](http://jenkins.usdc.cfengine.com/job/testing-kristian-privateBranch/159/), containing the changes from all three buffer size PRs.

kacf:
In nova I get the following:

```
10:08:03 report-collect-plugin.c: In function ‘Nova_QueryClientForReports__real’:
10:08:03 report-collect-plugin.c:226: error: ‘sendbuffer’ undeclared (first use in this function)
10:08:03 report-collect-plugin.c:226: error: (Each undeclared identifier is reported only once
10:08:03 report-collect-plugin.c:226: error: for each function it appears in.)
10:08:03 report-collect-plugin.c: In function ‘LoadReportSimulationCache2__real’:
10:08:03 report-collect-plugin.c:635: warning: ‘sprintf’ is deprecated (declared at /usr/include/stdio.h:363)
10:08:03 report-collect-plugin.c:648: warning: ‘sprintf’ is deprecated (declared at /usr/include/stdio.h:363)
10:08:03 make[1]: *** [report-collect-plugin.lo] Error 1
```

Also the `crypto_symmetric_test` from core seems to be failing in a lot of places. Maybe @jimis should take a look at the parts touching crypto.

libenv/sysinfo.c
```diff
@@ -144,7 +144,10 @@ static time_t GetBootTimeFromUptimeCommand(time_t now);
 
 /*****************************************************/
 
-void CalculateDomainName(const char *nodename, const char *dnsname, char *fqname, char *uqname, char *domain);
+void CalculateDomainName(const char *nodename, const char *dnsname,
+                         char *fqname, size_t fqname_length,
+                         char *uqname, size_t uqname_length,
+                         char *domain, size_t domain_length);
```
jimis: The word "length" is usually used for the length of a string, i.e. up to the '\0' terminator. We usually use "size" for buffer size, which also matches to the `sizeof` operator that is usually used to pass the correct number in these function calls.

So I would rename the occurrences here to `fqname_size`, `uqname_size`, `domain_size`, and this also applies to many other places in this pull request.

jimis:
In general the "automatic" remediation of potential buffer overflows just by replacing with `strlcpy()` is creating new security issues: string contents you never expected. The proper solution for the function to return error, and preferably abort the operation, or worst case log an error and leave the string empty or with a special value.

libpromises/crypto.c
```diff
@@ -502,10 +502,29 @@ size_t CipherTextLengthMax(const EVP_CIPHER* cipher, size_t plaintext_length)
     return plaintext_length + (CipherBlockSizeBytes(cipher) * 2) - 1;
 }
 
+size_t PlainTextLengthMax(const EVP_CIPHER* cipher, size_t ciphertext_length)
+{
+    if(ciphertext_length <= 0)
+    {
+        ProgrammingError("PlainTextLengthMax: ciphertext_length is not positive (%ld)",
```
jimis: `size_t` is unsigned, the proper way to print it is `%zu`.

libpromises/crypto.c
```diff
@@ -502,10 +502,29 @@ size_t CipherTextLengthMax(const EVP_CIPHER* cipher, size_t plaintext_length)
     return plaintext_length + (CipherBlockSizeBytes(cipher) * 2) - 1;
 }
 
+size_t PlainTextLengthMax(const EVP_CIPHER* cipher, size_t ciphertext_length)
+{
+    if(ciphertext_length <= 0)
```
jimis: Can't be negative, so I would change this to `assert(ciphertext_length == 0)`. BTW why can't it be zero?

libpromises/crypto.c
```diff
@@ -502,10 +502,29 @@ size_t CipherTextLengthMax(const EVP_CIPHER* cipher, size_t plaintext_length)
...
+    {
+        ProgrammingError("PlainTextLengthMax: ciphertext_length is not positive (%ld)",
+                         ciphertext_length);
+    }
+
+    // check for potential integer overflow, leave some buffer
+    if(ciphertext_length > INT_MAX - 1024)
```
jimis: `size_t` can't really overflow.

libpromises/crypto.c
```diff
@@ -502,10 +502,29 @@ size_t CipherTextLengthMax(const EVP_CIPHER* cipher, size_t plaintext_length)
...
+    {
+        ProgrammingError("PlainTextLengthMax: ciphertext_length is too large (%ld)",
```
jimis: Why is this an assertion failure? Can't the caller just submit a huge buffer? I believe what you want to do is return an error code (like `(size_t) -1` in this case).

jimis: BTW `ProgrammingError()` is not guaranteed to run without debug symbols. It's equivalent to an assertion.

libpromises/crypto.c
jimis: 1024 is arbitrary, I would use the exact number.

libpromises/files_names.c
```diff
@@ -515,18 +512,25 @@ const char *ReadLastNode(const char *str)
 
 /*********************************************************************/
 
-/**
- * @TODO fix the dangerous path lengths
- */
-int CompressPath(char *dest, const char *src)
+int CompressPath(char *dest, size_t dest_length, const char *src)
```
jimis: Can you change the return type to `bool` since you are at it anyway? :-)

libpromises/files_names.c
```diff
@@ -568,8 +572,7 @@ int CompressPath(char *dest, const char *src)
 
         AddSlash(dest);
 
-        /* TODO use dest_size parameter instead of CF_BUFSIZE. */
-        size_t ret = strlcat(dest, node, CF_BUFSIZE);
+        size_t ret = strlcat(dest, node, dest_length);
 
         if (ret >= CF_BUFSIZE)
         {
```
jimis: I can see that `CompressPath()` returned value is not checked in most places. The least we can do is `dest[0]='\0'` so that we avoid deleting the parent directory of an extremely long path.

libcfnet/client_code.c
```diff
@@ -587,9 +587,17 @@ int EncryptCopyRegularFileNet(const char *source, const char *dest, off_t size,
 
     snprintf(in, CF_BUFSIZE - CF_PROTO_OFFSET, "GET dummykey %s", source);
     cipherlen = EncryptString(out, sizeof(out), in, strlen(in) + 1, conn->encryption_type, conn->session_key);
+
+    tosend = cipherlen + CF_PROTO_OFFSET;
+
+    if(tosend > sizeof(workbuf))
```
jimis: Can this condition be true in any possible scenario you can imagine?

libcfnet/stat_cache.c
```diff
@@ -147,9 +148,17 @@ int cf_remote_stat(AgentConnection *conn, bool encrypt, const char *file,
                  (intmax_t) tloc, file);
         int cipherlen = EncryptString(out, sizeof(out), in, strlen(in) + 1,
                                       conn->encryption_type, conn->session_key);
+
+        tosend = cipherlen + CF_PROTO_OFFSET;
```
jimis: can cipherlen be negative?

libcfnet/stat_cache.c
```diff
@@ -147,9 +148,17 @@ int cf_remote_stat(AgentConnection *conn, bool encrypt, const char *file,
...
+
+        if(tosend > sizeof(sendbuffer))
+        {
+            ProgrammingError("cf_remote_stat: tosend (%d) > sendbuffer (%ld)",
+                             tosend, sizeof(sendbuffer));
```
jimis: Here as well, I don't think `ProgrammingError()` is appropriate. If I read this correctly, `tosend` depends on the length of `in`, and `in` depends on user input, which could be big enough to cause a crash.

libpromises/conversion.c
```diff
@@ -390,7 +390,7 @@ long IntFromString(const char *s)
 {
     long long a = CF_NOINT;
     char c = 'X';
-    char remainder[CF_BUFSIZE];
+    char remainder[4096];
```
jimis: This makes it quite hard to search for CF_BUFSIZE ugliness.

libpromises/conversion.c
```diff
@@ -409,7 +409,7 @@ long IntFromString(const char *s)
 
     remainder[0] = '\0';
 
-    sscanf(s, "%lld%c%s", &a, &c, remainder);
+    sscanf(s, "%lld%c%4095s", &a, &c, remainder);
```
jimis: I don't see what's the role of `remainder` being a string in this function, dump it completely. If you want to check if anything follows, use another `%c` and read one more character. Also remember to check the `sscanf()` return value. This is the way to kill the potential security issue here.

libpromises/evalfunction.c
```diff
@@ -7388,7 +7388,7 @@ void ModuleProtocol(EvalContext *ctx, char *command, const char *line, int print
             else if (StringMatchFullWithPrecompiledRegex(context_name_rx, content))
             {
                 Log(LOG_LEVEL_VERBOSE, "Module changed variable context from '%s' to '%s'", context, content);
-                strcpy(context, content);
+                strlcpy(context, content, context_length);
```
jimis: Is there a potential risk from truncation here?

libcfnet/misc.c
```diff
@@ -53,41 +53,6 @@ int cf_closesocket(int sd)
     return res;
 }
 
-/* Convert IP address in src (which can be struct sockaddr_storage (best
- * choice for any IP version), struct sockaddr, struct sockaddr_in or struct
- * sockaddr_in6) to string dst.
- * Better use getnameinfo(NI_NUMERICHOST) if available. */
-const char *sockaddr_ntop(const void *src, char *dst, socklen_t size)
```
jimis: Was this function flagged by the security scan? I'm honestly curious here...

libpromises/locks.c
```diff
@@ -886,10 +886,10 @@ void GetLockName(char *lockname, const char *locktype, const char *base, const R
...
     }
 
-    strncpy(lockname, locktype, CF_BUFSIZE / 10);
-    strcat(lockname, "_");
-    strncat(lockname, base, CF_BUFSIZE / 10);
-    strcat(lockname, "_");
+    strlcpy(lockname, locktype, CF_BUFSIZE / 10);
+    strlcat(lockname, "_", CF_BUFSIZE / 10);
+    strlcat(lockname, base, CF_BUFSIZE / 10);
+    strlcat(lockname, "_", CF_BUFSIZE / 10);
```
jimis: I don't get it.

libpromises/matching.c
```diff
@@ -292,7 +292,7 @@ void EscapeSpecialChars(const char *str, char *strEsc, int strEscSz, char *noEsc
                 break;
             }
 
-            strcat(strEsc, noEscSeq);
+            strlcat(strEsc, noEscSeq, strEscSz);
```
jimis: But it is checked right above that it doesn't overflow, right? Or am I reading it wrong?

libpromises/matching.c
```diff
@@ -292,7 +292,7 @@ void EscapeSpecialChars(const char *str, char *strEsc, int strEscSz, char *noEsc
                 break;
```
jimis: A message here and  emptying the string would be nice - or something better than silently truncating.

jimis:
Alright I'm pausing here. Some of the commits are clear to merge, I wish github had a "merge selectively" functionality.

libpromises/crypto.c
kacf: No, `ProgrammingError` runs regardless of debug mode. It's a pretty heavy handed mechanism though, guaranteed abortion of the program.

jimis: It is today, but it is not guaranteed for tomorrow. `ProgrammingError()` is ideally unreachable code. It is not allowed to change program behaviour. That's why I'm asking whether these fixes are for actual crashes discovered during fuzzing.

estenberg:
@kacf I attempted to fix the nova compile issues now and re-pushed. I don't have a compile environment for Nova, so sorry for these basic issues. Also, perhaps we should discuss those in the relevant PR at https://github.com/cfengine/nova/pull/979?

libpromises/crypto.c
estenberg: Thanks, fixed.

libpromises/crypto.c
estenberg: Makes sense, as size_t is unsigned this check is useless. Removed.
Also changed the overflow check.

libpromises/crypto.c
estenberg: Switched to SIZE_MAX, thanks.

estenberg: If you had other proposals on how to properly "abort" then we can switch if you want.

The thing is that if this function gets called with a buffer that's too small I can't really see how one could recover. The helper functions (e.g. PlaintextLengthMax()) should be used by the caller if uncertain about it will work or not, hence also being a valid programming error IMHO.

libpromises/files_names.c
estenberg: There is a "memset(dest, 0, dest_length);" at the top of the function, so I didn't change this. It's a bit unclear to me if you mean something else? In that case I would handle it separately (not related to this change?).

libpromises/files_names.c
estenberg: Sounds like feature creep,but ok. :)

libcfnet/stat_cache.c
estenberg: If an attacker tries to exploit, crash is better than potential overflow, right? :)
I'm not sure what else to do.

libpromises/conversion.c
estenberg: Sure, converted remainder to char.

Not sure what you mean by checking sscanf() return value being the problem, as arbitrary length written to the string reminder is now avoided?

libpromises/evalfunction.c
estenberg: I am not sure, but better not overflow at least.

libcfnet/misc.c
estenberg: Yes, but might be false positive. I did not bother to check rather remove it as it was unused.

libpromises/locks.c
estenberg: Me neither, I did not change the size limits, just made the bounds consistent.

libpromises/matching.c
estenberg: I think so, so strcat() might be safe here. I will leave strlcat() though unless you are sure and feel strongly about this.

libpromises/matching.c
estenberg: Error message added.

libpromises/conversion.c
estenberg: Switched it for a char.

libcfnet/stat_cache.c
kacf: Yes, I agree. Even if there might be better ways to handle this, it's better than it was before.

kacf:
Started a new build of all three buffer size PRs at http://jenkins.usdc.cfengine.com/job/testing-kristian-privateBranch/160/.

@jimis: I'll leave it to you to verify the issues you pointed out.

kacf:
Still lots of failures here. Windows build:

```
08:57:26 /home/jenkins/workspace/testing-kristian-privateBranch/label/PACKAGES_x86_64_linux_mingw/enterprise/libcompat-win32/uname.c:250: undefined reference to `GetMyHostInfo'
08:57:26 collect2: error: ld returned 1 exit status
```

Also the `crypto_symmetric_test` fails across the board. Doesn't `make check` fail for you, @estenberg?

libcfnet/stat_cache.c
jimis: `ProgrammingError` should be unreachable code. `abort()`ing and generating a core dump is not acceptable behaviour. Except if it's something we didn't think about, and once it happens, we *must* fix it. This is the meaning of `ProgrammingError`. 

`FatalError()` is when you want to exit with error. Still I would return an error to the caller and handle it. Is it hard in this case?

jimis:
I'll try to cherry-pick and merge the easy ones, so that this moves forward.

libpromises/addr_lib.c
```diff
@@ -42,6 +42,12 @@ int FuzzySetMatch(const char *s1, const char *s2)
     short isCIDR = false, isrange = false, isv6 = false, isv4 = false;
     char address[CF_ADDRSIZE];
 
+    if (strlen(s1) >= CF_MAX_IP_LEN)
+    {
+        Log(LOG_LEVEL_ERR, "IP set string '%s' is too large to parse (>=%d)", s1, CF_MAX_IP_LEN);
+        return -1;
```
jimis: With a deeper look, this seems like it will cause error messages printing many times on each connection, if the ACL has long hostnames included.

Where is the possible overflow? Is it later in this function? If so, just limit that one.

libpromises/matching.c
jimis: `strlcat()` with no checking is a sign of bad code - patching stuff with no real understanding. Here this is clearly not the case, checks are quite rigorous above. I woudn't write it like that, I don't care deeply though so I'm cherry-picking this.

libpromises/conversion.c
```diff
@@ -211,11 +211,11 @@ char *Rlist2String(Rlist *list, char *sep)
 
     for (rp = list; rp != NULL; rp = rp->next)
     {
-        strcat(line, RlistScalarValue(rp));
+        strlcat(line, RlistScalarValue(rp), sizeof(line));
 
         if (rp->next)
         {
-            strcat(line, sep);
+            strlcat(line, sep, sizeof(line));
```
jimis: I'm not cherry-picking this one because this is a very generic utility function and who knows what truncation can cause. Please add a `Log(ERR, "internal limit reached ...")` or similar in the case truncation happens.

libpromises/conversion.c
```diff
@@ -509,7 +509,7 @@ bool DoubleFromString(const char *s, double *value_out)
 
     remainder[0] = '\0';
 
-    sscanf(s, "%lf%c%s", &a, &c, remainder);
+    sscanf(s, "%lf%c%4095s", &a, &c, remainder);
```
jimis: Cherry-picking since it's not worse than before, I'll repeat though that there is no meaning in having 4K memory overhead. Reading "remainder" in a `char` variable would be enough.

libpromises/conversion.c
```diff
@@ -407,15 +406,14 @@ long IntFromString(const char *s)
...
 
-    remainder[0] = '\0';
 
-    sscanf(s, "%lld%c%s", &a, &c, remainder);
+    sscanf(s, "%lld%c%c", &a, &c, &remainder);
 
 // Test whether remainder is space only
 
-    if ((a == CF_NOINT) || (!IsSpace(remainder)))
+    if ((a == CF_NOINT) || (remainder == ' '))
```
jimis: Logic error, you want exactly the opposite. And don't compare with space, but do `!isspace(remainder)`. See `ctype.h` for more such functions.

libpromises/conversion.c
jimis: Err, not cherry-picking after all since you've already changed one occurence of this kind. Please do the same here.

libpromises/files_names.c
```diff
@@ -169,7 +166,7 @@ char *JoinSuffix(char *path, const char *leaf)
         return NULL;
     }
 
-    strcat(path, leaf);
+    strlcat(path, leaf, path_size);
```
jimis: There is a length check right above? What is it for?

libpromises/evalfunction.c
```diff
@@ -7388,7 +7388,7 @@ void ModuleProtocol(EvalContext *ctx, char *command, const char *line, int print
             else if (StringMatchFullWithPrecompiledRegex(context_name_rx, content))
             {
                 Log(LOG_LEVEL_VERBOSE, "Module changed variable context from '%s' to '%s'", context, content);
-                strcpy(context, content);
+                strlcpy(context, content, context_size);
```
jimis: No, this can't overflow and hides logic. Keep the `strcpy` and just use `assert` or `ProgrammingError` for checking that `context_size > 50` preferably at the beginning of the file. Or at least I think, I didn't go through the whole function.

libcfnet/stat_cache.c
jimis: Or at least `Log(ERR, "skipping file because blah blue")` and return without doing anything. Is this easy in this context?

jimis:
If you fix issues, either here or in #2617 please just append new commits, don't rebase. Since this is a big pull request, we don't want to lose track of the new changes that fix the issues.

estenberg:
Ah, GetMyHostInfo() was used on Windows. Added it as static in uname.c.

libcfnet/stat_cache.c
estenberg: FatalError() won't work as it takes EvalContext which is not available here. Prototype:

void FatalError(const EvalContext *ctx, char *s, ...) FUNC_ATTR_NORETURN FUNC_ATTR_PRINTF(2, 3);

Furthermore I don't see any benefit in returning here. Why would we do that? If it's reached it seems to be a bug and we should stop.

jimis:
The easy ones have been cherry-picked to master. Can you rebase?

libpromises/addr_lib.c
estenberg: Thanks, good catch. Addressed in two new commits.

libpromises/conversion.c
estenberg: OK, error log added.

libpromises/conversion.c
estenberg: OK, changed reminder from type string to char.

libpromises/conversion.c
estenberg: Ugh, good catch! Thanks, fixed!

libpromises/files_names.c
estenberg: The length check was broken (assuming paht is CF_BUFSIZE).
However, as we have path_size now I reverted to using that and your favourite function strcat(). :)

libpromises/evalfunction.c
estenberg: 'filename' is based on a parameter 'command', how do you know it's less than 50 chars?
Perhaps you did not see that one..?

For the second strlcpy() of context in this function it looks to be safe to make sure it's larger
than 50 chars and use strcpy(), but what if someone changes the function in the future?
Is it really worth risking?

I think we should keep strlcpy() here, why are you so insistent on strcpy()?

I added a ProgrammingError() in case < 51 char buffer is passed, though.

estenberg:
Alright, I addressed all concerns and rebased.

Can we trigger a build with enterprise (https://github.com/cfengine/enterprise/pull/312) & nova (https://github.com/cfengine/nova/pull/979) PRs and run a test round?
It passes make check for me on core & enterprise (don't have a nova build environment, though).

If there aren't any bugs with this and it improves the situation, can I also ask that we get these merged so they make it into 3.9.0?

kacf:
trigger build

kacf:
Oh right, this requires multiple branches. Started a private build here: http://jenkins.usdc.cfengine.com/job/testing-kristian-privateBranch/162/

kacf:
Seems to pass everywhere.

kacf:
Eh, wait, I was looking at the wrong build, sorry. This fails everywhere: `crypto_symmetric_test`

It also fails on Windows, with:

```
08:11:23 uname.c: In function 'GetMyHostInfo':
08:11:23 uname.c:280:13: error: implicit declaration of function 'Log' [-Werror=implicit-function-declaration]
08:11:23              Log(LOG_LEVEL_ERR, "Could not get host entry for local host. (gethostbyname: %s)", GetErrorStr());
08:11:23              ^
08:11:23 uname.c:280:17: error: 'LOG_LEVEL_ERR' undeclared (first use in this function)
08:11:23              Log(LOG_LEVEL_ERR, "Could not get host entry for local host. (gethostbyname: %s)", GetErrorStr());
08:11:23                  ^
08:11:23 uname.c:280:17: note: each undeclared identifier is reported only once for each function it appears in
08:11:23 uname.c:280:13: error: implicit declaration of function 'GetErrorStr' [-Werror=implicit-function-declaration]
08:11:23              Log(LOG_LEVEL_ERR, "Could not get host entry for local host. (gethostbyname: %s)", GetErrorStr());
08:11:23              ^
08:11:23 cc1: some warnings being treated as errors
08:11:23 Makefile:331: recipe for target 'uname.lo' failed
08:11:23 make[1]: *** [uname.lo] Error 1
```

estenberg:
Sorry about the Windows error, I can't build that in my development environment. Attempted to fix in https://github.com/cfengine/enterprise/pull/312.

estenberg:
I attempted to fix the crypto_symmetric_test failure as well. It seems to be caused by the enterprise extension not being loaded (it would always fail on core due to different implementation of CfengineCipher() between core and enterprise).

kacf:
Triggered a new build containing all three PRs here: http://jenkins.usdc.cfengine.com/job/testing-kristian-privateBranch/164/

kacf:
Nice going fixing Windows blind folded! ;-)

Still failing, but in nova only, I have posted the output there.

kacf:
All green now!

cf-agent/files_editline.c
```diff
@@ -208,9 +208,9 @@ Bundle *MakeTemporaryBundleFromTemplate(EvalContext *ctx, Policy *policy, Attrib
             // Get Action operator
             if (strncmp(buffer, "[%CFEngine", strlen("[%CFEngine")) == 0)
             {
-                char op[CF_BUFSIZE], brack[CF_SMALLBUF];
+                char op[CF_BUFSIZE], brack[4];
 
-                sscanf(buffer+strlen("[%CFEngine"), "%1024s %s", op, brack);
+                sscanf(buffer+strlen("[%CFEngine"), "%1024s %3s", op, brack);
 
                 if (strcmp(brack, "%]") != 0)
```
jimis: What happens here if `brack` is never read from `sscanf()`? Then it contains garbage and this call can overflow.

jimis:
This needs to be rebased to master in order to get tested by Travis, and some commits that touch Enterprise need to wait for Jenkins.

jimis:
Hmmm I can move this forward faster by cherry-picking again what is ready, and rebasing in another pull request what is straightforward to merge.  Sounds good @estenberg ? Mostly because the commits affecting Enterprise can't be tested now.

estenberg:
Sure, please cherry pick what is ready and I can rebase once we have the Enterprise test (there are some important ones there) and can merge quickly without rebasing again and again. Please let me know when I should rebase.

jimis:
Superseded by https://github.com/cfengine/core/pull/2740.

PR #2611, opened 2016-05-11
CFE-2371: Prevent misinterpretation of ps column header line.

kacf:
The header is always the first line, so just use that, instead of
scanning for "PID" which can occur in process names.

Changelog: Fix intermittent error message of type:
"error: Process table lacks space for last columns: <cmd>"


kacf:
trigger build

jimis:
+1

jimis:
Build is red though, failing at the `ifelapsed_and_expireafter/timed/defaults.cf` test on Ubuntu 12.04. This usually is a spurious failure of our buildslaves, right?

kacf:
Yes.

kacf:

PR #2612, opened 2016-05-11
CFE-1995: Be less verbose if a network interface doesn't have a MAC address.

kacf:
Changelog: Title


kacf:
trigger build

jimis:
Makes sense and build is green, thanks!

kacf:
Thanks! Cherry-picked to 3.7.x and 3.9.x as well.

PR #2613, opened 2016-05-16
When no file is provided when calling cf-promises with cf or json out…

amousset:
…put,

use promises.cf by default.

When parsing complete promises, libpromises uses promises.cf by default.
As the cf and json output options do not load the complete policy, we
provide the default filename in cf-promises.
Also use Cf3ParseFile to allow parsing JSON promises.

This restores the previous behavior.

### DIFF
--- a/cf-promises/cf-promises.c
+++ b/cf-promises/cf-promises.c
@@ -132,10 +132,13 @@ int main(int argc, char *argv[])
     if (format == GENERIC_AGENT_CONFIG_COMMON_POLICY_OUTPUT_FORMAT_CF ||
         format == GENERIC_AGENT_CONFIG_COMMON_POLICY_OUTPUT_FORMAT_JSON)
     {
+        // If no file was provided, use 'promises.cf' by default
+        if (config->input_file == NULL) {
+            GenericAgentConfigSetInputFile(config, GetInputDir(), "promises.cf");
+        }
+
         // Just parse and write content to output
-        Policy *output_policy = ParserParseFile(AGENT_TYPE_COMMON, config->input_file,
-                                                config->agent_specific.common.parser_warnings,
-                                                config->agent_specific.common.parser_warnings_error);
+        Policy *output_policy = Cf3ParseFile(config, config->input_file);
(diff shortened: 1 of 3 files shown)


cf-promises/cf-promises.c
```diff
@@ -132,10 +132,13 @@ int main(int argc, char *argv[])
...
+        // If no file was provided, use 'promises.cf' by default
+        if (config->input_file == NULL) {
+            GenericAgentConfigSetInputFile(config, GetInputDir(), "promises.cf");
+        }
+
         // Just parse and write content to output
-        Policy *output_policy = ParserParseFile(AGENT_TYPE_COMMON, config->input_file,
-                                                config->agent_specific.common.parser_warnings,
-                                                config->agent_specific.common.parser_warnings_error);
+        Policy *output_policy = Cf3ParseFile(config, config->input_file);
```
jimis: Why export a static function, why is `Cf3ParseFile()` better? You are mentioning that this restores the previous behaviour, how was it done before?

amousset: Since https://github.com/cfengine/core/commit/2a7c11b4789f876bc36e42d8b473936f56757162, the single file parsing is done before parsing the whole policy, using ParserParseFile directly (which does not handle empty file name and segfaults).

The previous behavior was to always call `LoadPolicy` (which checks if the file exists, and uses "promises.cf" if no file is given) before calling `ParserParseFile`. Another issue with calling directly `ParserParseFile` is that it is not able to read json policy files, contrary to the Higher-level `Cf3ParseFile` which is basically a wrapper around `ParserParseFile` that handles checks and detection of the target file.

jimis:
Thanks for explaining, I'll merge this once the build passes.

trigger build

jimis:
Merging, thanks!

jimis:
I'll merge manually because of conflicts.

jimis:
I hit "merge" and github tells me "there has been an error with this pull request" and then shows me the usual "this pull request has conflicts". The I try to merge manually on the command line and everything looks already merged.

I'm closing this, let me know if you notice anything missing.

PR #2614, opened 2016-05-23
Change: Switch processes restart_class logging to verbose

nickanderson:
This moves the logging of a class definition as the result of a
restart_class attached to a processes promise from info to verbose. No
other class definition logging happens in info.

Changelog: Title

### DIFF
--- a/cf-agent/verify_processes.c
+++ b/cf-agent/verify_processes.c
@@ -246,7 +246,7 @@ static PromiseResult VerifyProcessOp(EvalContext *ctx, Attributes a, const Promi
         else
         {
             PromiseResult status = killed ? PROMISE_RESULT_CHANGE : PROMISE_RESULT_NOOP;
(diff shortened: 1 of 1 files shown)


cf-agent/verify_processes.c
```diff
@@ -246,7 +246,7 @@ static PromiseResult VerifyProcessOp(EvalContext *ctx, Attributes a, const Promi
         else
         {
             PromiseResult status = killed ? PROMISE_RESULT_CHANGE : PROMISE_RESULT_NOOP;
-            cfPS(ctx, LOG_LEVEL_INFO, status, pp, a,
+            cfPS(ctx, LOG_LEVEL_VERBOSE, status, pp, a,
```
jimis: Verbose is better I agree. But message should contain the `C:` string like other class settings.

PR #2615, opened 2016-05-23
Change: Switch processes restart_class logging to verbose

nickanderson:
This moves the logging of a class definition as the result of a
restart_class attached to a processes promise from info to verbose. No
other class definition logging happens in info.

Changelog: Title

### DIFF
--- a/cf-agent/verify_processes.c
+++ b/cf-agent/verify_processes.c
@@ -246,8 +246,8 @@ static PromiseResult VerifyProcessOp(EvalContext *ctx, Attributes a, const Promi
         else
         {
             PromiseResult status = killed ? PROMISE_RESULT_CHANGE : PROMISE_RESULT_NOOP;
-            cfPS(ctx, LOG_LEVEL_INFO, status, pp, a,
-                 "Setting restart class '%s' for promise '%s'", a.restart_class, pp->promiser);
+            cfPS(ctx, LOG_LEVEL_VERBOSE, status, pp, a,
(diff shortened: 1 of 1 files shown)


nickanderson:
@jimis here you go. Had to close and open a new PR because github wasnt showing the updates to my branch.

ncharles:
Le 23 mai 2016 19:35:05 GMT+02:00, Nick Anderson <notifications@github.com> a écrit :
>@jimis here you go. Had to close and open a new PR because github wasnt
>showing the updates to my branch.
>
>
>---
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/cfengine/core/pull/2615#issuecomment-221039864

I'm quite happy with this log being Inform. This really helps explaining what is going on
-- 
Envoyé de mon téléphone Android avec K-9 Mail. Excusez la brièveté.

jimis:
> I'm quite happy with this log being Inform. This really helps explaining what is going on

The rule of thumb is INFO being used for system changes. Setting class is VERBOSE everywhere else. I have no strong opinion on the text being printed (it could be different to set it apart) but why would INFO be better for this specific class message?

ncharles:
Oh, sorry, i didn't see the answer, Nick just pinged me. I didn't know the rational of INFO being used for change; and it does make sense
I usually have classes defined by restart_class as a non expected behaviour, but the command afterward will be inform, so i'm ok with it moving to verbose.
Thank you for the explaination :+1:

PR #2618, opened 2016-05-25
Redmine7952 - cherry-pick for 3.7.x

jimis:


jimis:
All green with local tests, merging.

PR #2620, opened 2016-05-26
zendesk#2906: improve messages when 'mount -av' fails.

jimis:


jimis:
@nickanderson please let me know what you think

PR #2621, opened 2016-05-31
CFE-2388: resolve body parameters repeatedly

tzz:
See https://tracker.mender.io/browse/CFE-2388?focusedCommentId=73272

This fixes body inheritance with multiple levels by:
- accumulating the rewrite map across several inheritance steps
- expanding and rewriting **all** the body parameters on every inheritance step instead of just the specified ones

Acceptance test change included.

### DIFF
--- a/libpromises/promises.c
+++ b/libpromises/promises.c
@@ -128,8 +128,6 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
             break;
         }
 
-        /* First case is: we have a body template to expand lval = body(args), .. */
-
         if (bodies_and_args && SeqLength(bodies_and_args) > 0)
         {
             const Body *bp = SeqAt(bodies_and_args, 0); // guaranteed to be non-NULL
@@ -263,20 +261,24 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
                         {
                             Rval returnval = RvalNew(scp->rval.item, scp->rval.type);
 
+                            // TODO: this was only in the case of body call with no args, should it always happen?
+                            if (returnval.type == RVAL_TYPE_LIST)
+                            {
+                                Rlist *new_list = RvalRlistValue(returnval);
+                                RlistFlatten(ctx, &new_list);
+                                returnval.item = new_list;
+                            }
+
                             // First we rewrite the Rval with the rewrite map
                             Rval rewrite = RvalCopyRewriter(returnval, arg_rewrite);
                             RvalDestroy(returnval);
 
-                            // TODO: this was only in the case of body call with no args, should it always happen?
-                            if (rewrite.type == RVAL_TYPE_LIST)
-                            {
-                                Rlist *new_list = RvalRlistValue(rewrite);
-                                RlistFlatten(ctx, &new_list);
-                                rewrite.item = new_list;
-                            }
-
                             // Second we expand body vars; note it has to happen ONLY ONCE
+                            // We have to protect this.promiser because in promise iterations, it will not be always correct
+                            // For example "x" usebundle => y; followed by an iterative files promise; will set this.promiser to "x"
+                            ExpandExceptionAdd("this.promiser");
                             returnval = ExpandPrivateRval(ctx, NULL, "body", rewrite.item, rewrite.type);
+                            ExpandExceptionRemove("this.promiser");
 
                             RvalDestroy(rewrite);
 
@@ -292,14 +294,17 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
 
                             // Writer *w = StringWriter();
                             // WriterWrite(w, "'");
+                            // RvalWrite(w, scp->rval);
+                            // WriterWrite(w, "' -> expanded rval '");
+                            // WriterWrite(w, "'");
                             // RvalWrite(w, returnval);
                             // WriterWrite(w, "' -> expanded+rewritten rval '");
                             // RvalWrite(w, scp_copy->rval);
                             // WriterWrite(w, "'; rewrite map ");
                             // JsonWrite(w, arg_rewrite, 0);
 
-                            // Log(LOG_LEVEL_DEBUG, "DeRefCopyPromise: processing body %s: expanding constraint %s with rval %s",
-                            //     body_reference, scp->lval, StringWriterData(w));
+                            // Log(LOG_LEVEL_DEBUG, "DeRefCopyPromise: in promise %s, processing body %s: expanding constraint %s with rval %s",
+                            //     pcopy->promiser, body_reference, scp->lval, StringWriterData(w));
                             // WriterClose(w);
                         }
                     }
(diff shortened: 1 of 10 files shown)


libpromises/promises.c
```diff
@@ -261,18 +284,23 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
...
+                            Writer *w = StringWriter();
+                            // RvalWrite(w, returnval);
                             // WriterWrite(w, " -> copied rval ");
                             // RvalWrite(w, scp_copy->rval);
+                            // WriterWrite(w, "; rewrite map ");
+                            // JsonWrite(w, arg_rewrite, 0);
 
-                            // Log(LOG_LEVEL_DEBUG, "DeRefCopyPromise: processing body %s: at index %ld, current body %s, copying constraint %s with rval %s", body_reference, body_index, current_body->name, scp->lval, StringWriterData(w));
-                            // WriterClose(w);
+                            Log(LOG_LEVEL_DEBUG, "DeRefCopyPromise: processing body %s: expanding constraint %s with rval %s", body_reference, scp->lval, StringWriterData(w));
```
kacf: Split this (and many other) lines, they are getting very long.

libpromises/promises.c
```diff
@@ -240,6 +250,19 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
 
                         if (IsDefinedClass(ctx, scp->classes))
                         {
+                            SeqAppend(constraints_to_expand, scp);
+                        }
+                    }
+
+                    SeqReverse(constraints_to_expand);
+                    // Only do the copy expansion if we copied something while inheriting
+                    for (size_t r = 0; r < SeqLength(constraints_to_expand); r++)
```
kacf: The logic is a bit difficult to follow, is the reason for reversing that you need to resolve the constraints with the "youngest" body taking precedence?

In any case, this will have better performance if you don't reverse the sequence (especially since you reverse it back later), but rather reverse the loop, starting at `SeqLength(constraints_to_expand) - 1` and going to `0`.

libpromises/promises.c
tzz: OK; removed.

libpromises/promises.c
tzz: OK; done. I got segfaults because size_t can't be negative and forgot to investigate, thanks for catching that.

tzz:
All right, ready for another review :)

tzz:
Looking at comments on https://tracker.mender.io/browse/CFE-2388 I realized the code was split in two branches for body calls with and without arguments, for reasons unknown to me, and I had only fixed the branch with arguments. I've consolidated the code and it passes all the body tests for me, plus the case that started this work.

The only difference I saw between the two branches was that rlists were flattened in the branch without arguments. Using that seems to be OK generally so I made it the common case.

libpromises/promises.c
```diff
@@ -240,12 +249,32 @@ Promise *DeRefCopyPromise(EvalContext *ctx, const Promise *pp)
...
+
+                        if (IsDefinedClass(ctx, scp->classes))
+                        {
                             Rval returnval = RvalNew(scp->rval.item, scp->rval.type);
 
                             // First we rewrite the Rval with the rewrite map
                             Rval rewrite = RvalCopyRewriter(returnval, arg_rewrite);
                             RvalDestroy(returnval);
 
+                            // TODO: this was only in the case of body call with no args, should it always happen?
```
kacf: It is of course difficult to tell, but it sounds logical that it should. When would you ever want an unexpanded list member?

kacf:
Looks alright to me. Normally I think Jimis should look at these things since he has become quite familiar with the evaluator lately, but he is out of the office for the moment.

kacf:
trigger build

libpromises/promises.c
tzz: It's been that way since before the first Git import, different behavior between the two branches... I agree we need to commit to one way or the other, and this way is more logical.

tzz:
The latest commit:

* uses `BufferSearchAndReplace()` which was promised, instead of the hacked string replacer
* passes all the tests, including the latest by Yann in the ticket which I added to the commit

tzz:
@kacf sorry for the repeated revisions but I think you'll see the new version is significantly better because it eliminates a long complicated section of the code with a much simpler function call. Altogether this PR actually cuts 106 lines of nasty C code now, and has 2 new acceptance tests! Thanks for your help.

kacf:
Thanks, I'll look at it soon. Note that Jimis is taking over a lot of the pull request work, so I'll be a bit less active for the time being (Jimis is on vacation ATM, though).

For what it's worth, the previous revision you posted failed with these messages:

```
Error Message

FAIL (UNEXPECTED FAILURE) ./10_files/this_promiser.cf

Stacktrace

R: any_0_ok
R: template_test_4_ok
R: any_1_ok
R: any_3_ok
R: any_2_ok
R: 0: 1
R: 1: 1
R: 2: 1
R: 3: 1
R: 4: 1
R: /home/jenkins/workspace/testing-enterprise-pr_core/label/PACKAGES_x86_64_linux_SuSE_11/core/tests/acceptance/./10_files/this_promiser.cf FAIL
```

Is that likely to still be an issue?

tzz:
See https://tracker.mender.io/browse/CFE-2388?focusedCommentId=73327

* this.promiser needs to be protected from early expansion
* expansion needs to happen after the list is flattened
* constraints need to be expanded in ascending order, not descending as I had it previously
* add acceptance tests for the above, including a soft fail for the case of inheritance with the same arguments
* all the acceptance tests pass

@kacf @jimis let me know if I can help further and if you have questions

tzz:
@jimis kacf suggested that you should look this over. Perhaps there's a way to use the evaluation stack frames to simulate the layers of body inheritance but I couldn't figure out how to make it happen, and it may delay the merge of these bug fixes... so I leave it to the developers' discretion.

kacf:
trigger build

kacf:
The build is green, but I think I will still let @jimis comment on this one before merging it.

tzz:
Oh, I was hoping it would make 3.9... thanks for doing so much review and testing. Let's see what happens...

kacf:
Probably not, but seeing as how this fixes a bug it's probably eligible for 3.9.1 as well.

tzz:
@jimis this is a bigger bug fix. Let me know if you want me to rebase it now or later.

kacf:
@tzz: How is this bug fix looking from your side? I see there were some problems resolving the scope. I'm prepared to spend some time on reviewing / merging this, but I don't think I have the capacity currently to fix the issue myself.

tzz:
@kacf now that @jimis has rewritten the iteration engine, my work is useless (it was almost a year ago!). 

Maybe you can reuse the acceptance tests and see what I did, but the conflicts are too much now.

If @jimis wants to take a look, he's probably the best person to help us. This may be easily resolved with the new iteration logic anyway.

kacf:
Hehe, I see. :-) Treating this as obsolete then. If you think the acceptance tests are still useful, feel free to submit them in a separate pull request, but I will go ahead and close this.

PR #2622, opened 2016-06-01
ENT-2739: Don't skip output filtering if old output log doesn't exist.

kacf:
It was assumed we could short circuit in this case, but we can't
    because we still don't know whether the filtered output of the new log
    will produce any lines or not. So go through it even if the old log
    cannot be opened, and treat it as if the old log is a 0-byte file.
    
    Changelog: Fix bug which caused empty emails to be sent from cf-execd
    if there was no previous output log and the new log was fully filtered
    by email filters.


kacf:
trigger build

kacf:

PR #2624, opened 2016-06-09
Add readenvfile() function and read /etc/os-release

tzz:
for https://tracker.mender.io/browse/CFE-1881 as discussed with @nickanderson @kacf 
- support ENV format: `K=V` where `V` can be a quoted escaped string. Skip `#` comments and blank lines. This is really useful for working with systemd unit files or Docker env files, for instance.
- adds `readenvfile()` function with acceptance test to read ENV files
- adds `readdata()` support for `ENV` parameter and auto-detected format for `*.env` files, with acceptance test
- read `/etc/os-release` into the data container `sys.os_release` with the exported `evalfunction.c` function `ReadDataFile()`

Reference: https://www.freedesktop.org/software/systemd/man/os-release.html


nickanderson:
According to the man page we should check /etc/os-release first and then fall back to /usr/lib/os-release

tzz:
@nickanderson added, thanks

nickanderson:
Oh the readenv() function will be useful.

nickanderson:
@jimis do you see reason (other than the current conflitcs) this shouldn't be merged soon?

tzz:
@jimis this is a small PR but some code had to be moved around so the data reader function could be exported.

tzz:
@jimis no need for rebasing after the eval changes, ready for review

libpromises/evalfunction.c
```diff
@@ -6180,106 +6180,119 @@ static FnCallResult FnCallReadRealList(EvalContext *ctx, ARG_UNUSED const Policy
     return ReadList(ctx, fp, args, CF_DATA_TYPE_REAL);
 }
 
-static FnCallResult FnCallReadData(ARG_UNUSED EvalContext *ctx,
-                                   ARG_UNUSED const Policy *policy,
-                                   const FnCall *fp,
-                                   const Rlist *args)
+JsonElement *ReadDataFile(const char* log_identifier, const char* input_path, const char* requested_mode, size_t size_max)
```
jimis: Wow I just noticed how this one function does everything: JSON, YAML, CSV, ENV. The fine details of parsing should be definitely moved elsewhere. Here the leftovers should only be error checking/logging and calling whatever parsers we have ready.

jimis:
Nice feature; It would be nice to add an example showing how to extract a specific field from the `sys.os_release` variable in policy, no JSON involved.

libpromises/evalfunction.c
```diff
@@ -6180,106 +6180,119 @@ static FnCallResult FnCallReadRealList(EvalContext *ctx, ARG_UNUSED const Policy
...
-                    fp->name, linenumber, input_path, (long unsigned int)size_max);
-                free(line);
-                break;
+                ++linenumber;
+
+                byte_count += strlen(line);
+                if (byte_count > size_max)
+                {
+                    Log(LOG_LEVEL_VERBOSE, "%s: line %d from ENV file '%s' exceeded byte limit %lu, done with file",
+                        myname, linenumber, input_path, (long unsigned int)size_max);
```

libpromises/evalfunction.c
```diff
@@ -6180,106 +6180,119 @@ static FnCallResult FnCallReadRealList(EvalContext *ctx, ARG_UNUSED const Policy
...
+                {
+                    // skip comment or blank line
+                }
+                else
+                {
+                    Rlist *pieces = RlistFromSplitRegex(line, "=", 2, true);
+                    if (pieces && pieces->next)
+                    {
+                        const char* v = RlistScalarValueSafe(pieces->next);
+                        if (strlen(v) > 1 && v[0] == '"' && v[strlen(v)-1] == '"')
```
tzz: Not according to the URL I quoted.

libpromises/evalfunction.c
```diff
@@ -6180,106 +6180,119 @@ static FnCallResult FnCallReadRealList(EvalContext *ctx, ARG_UNUSED const Policy
...
+                        {
+                            char* decoded = JsonDecodeString(v);
```
tzz: It seemed like the right thing to do, based on the URL I quoted. ENV files don't support heredocs or other multiline data, as far as I know. They are not actual shell scripts, although you can source them from a shell. For instance Docker has a `--env-file FILE.env` option that doesn't know about shell variable expansion.

libpromises/evalfunction.c
```diff
@@ -6180,106 +6180,119 @@ static FnCallResult FnCallReadRealList(EvalContext *ctx, ARG_UNUSED const Policy
...
+                            // cut off the last character
+                            decoded[strlen(decoded)-1] = '\0';
```

libpromises/evalfunction.c
```diff
@@ -6180,106 +6180,119 @@ static FnCallResult FnCallReadRealList(EvalContext *ctx, ARG_UNUSED const Policy
...
+                            JsonObjectAppendString(json, RlistScalarValueSafe(pieces), decoded+1);
```
tzz: OK.

libpromises/evalfunction.c
```diff
@@ -6180,106 +6180,119 @@ static FnCallResult FnCallReadRealList(EvalContext *ctx, ARG_UNUSED const Policy
...
+                            free(decoded);
+                        }
+                        else
+                        {
+                            JsonObjectAppendString(json, RlistScalarValueSafe(pieces), v);
+                        }
+                    }
+                    else
+                    {
+                        Log(LOG_LEVEL_INFO, "%s: line %d from %s file '%s' was not in the format k=v",
```
tzz: OK.

libpromises/evalfunction.c
```diff
@@ -6180,106 +6180,119 @@ static FnCallResult FnCallReadRealList(EvalContext *ctx, ARG_UNUSED const Policy
...
+                            myname, linenumber, requested_mode, input_path);
+                    }
+
+                    if (pieces)
```

tzz:
@jimis I've rebased and added a fixup commit for all the items except the library stuff.

For me, it's not easy to move functions between libraries as you suggested to export from libutils, although I agree it's the right approach. Can you please make that change on your side as a followup commit? It would be really helpful, and probably just a matter of moving code to the right places. Similarly for the splitting on a single character, it would be great if you gave a good implementation we can use in the core. Thank you!

tzz:
I'll add a followup commit with a testable example, when I do the docs, so it can be included in the docs.

jimis:
This needs quite some work to be mergeable. Since @nickanderson is right next to me pestering me to merge this, I tried to move the functions around to reduce the complexity, but after 2 hours I'm still not done. Sorry but it will have to wait.

olehermanse:
@tzz I'm picking this up now, do you have anything you'd like to add or any uncommited work?

tzz:
@olehermanse sounds great. I appreciate your help. I hope the acceptance test makes sense.

tzz:
Rewritten elsewhere :)

PR #2626, opened 2016-06-11
Evaluator cleanup

jimis:


jimis:
Accompanied by cfengine/enterprise#314, tested successfully with testing-jimis-privateBranch build 26.

PR #2628, opened 2016-06-16
CFE-2260: allow ifelse(FALSE, $(x), "something else") to work

tzz:
see https://tracker.mender.io/browse/CFE-2260

* fixes the acceptance test (it was a bit broken before)
* special-cases the FnCall `ifelse("!any", ANYTHING, "foo")` to be allowed to evaluate even though `ANYTHING` may be an unexpanded variable. This is only done if `ifelse()` is called with exactly 3 arguments, but could be generalized fairly easily.

Use case:

```
ifelse( isvariable("init.passwd_file"), $(init.passwd_file), "/etc/passwd")
```

### DIFF
--- a/libpromises/fncall.c
+++ b/libpromises/fncall.c
@@ -345,15 +345,29 @@ FnCallResult FnCallEvaluate(EvalContext *ctx, const Policy *policy, FnCall *fp, 
 
     if (RlistIsUnresolved(expargs))
     {
-        if (LogGetGlobalLevel() >= LOG_LEVEL_DEBUG)
+        // Special case: ifelse(isvariable("x"), $(x), "default")
+        // (the first argument will come down expanded as "!any")
+        if (0 == strcmp(fp->name, "ifelse") &&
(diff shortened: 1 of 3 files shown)


jimis:
trigger build

jimis:
Special cases in the code are evil, but I can't think of something better at the moment... Will merge once the build is green.

tzz:
@jimis agreed, there's just no way to allow unresolved variables in a function call. Thanks for looking.

jimis:
Merging, thanks!

tzz:
Sure thing. I'm glad this was a fairly simple fix.

jimis:
Ugh, changelog tag was forgotten, @kacf I should amend an empty commit containing the right message, right?

kacf:
@jimis: Yep, use `git commit --allow-empty` and add it there. And remember to cherry-pick it as well if you want it somewhere besides master.

jimis:
Done, thanks @kacf.

PR #2629, opened 2016-06-18
CFE-2402: pass service policies unaltered to service bundles

yann-soubeyrand:
'enable' and 'disable' services policies were translated respectively to
'start' and 'stop' when passing them to service bundles. Also, in some cases
(when using a service_<service_name> implicit service bundle), 'reload'
service policy was translated to 'restart'.


yann-soubeyrand:
I think the code could be refactored: since all the policies are now passed through, the switch blocks seems unnecessary if we could have the string version of the service policy. Maybe we could write a function to get the string from the enum. Anyway, this commit already fixes CFE-2402.

This comment https://github.com/cfengine/core/blob/master/cf-agent/verify_services.c#L279 that a class name after the service policy could be set inside the service bundle with a scope restricted to this bundle. It would be great!

tzz:
@yann-soubeyrand I agree with your comments and patch, but have some questions:

* the new code is not backwards compatible. Is that going to be a problem?
* maybe instead of enums, the service policy should be just a string that can be anything? For systemd for instance that would be really useful. That seems most sensible and future-proof. I haven't found the current limited choices thrilling. If that's acceptable, it simplifies the code a lot.
* the bundle-local class for the policy sounds good, but there should also be one for the service name

yann-soubeyrand:
@tzz 
- I don't know if that's going to be a problem, I did it that way because it was the simplest solution and because of this comment https://tracker.mender.io/browse/CFE-2402?focusedCommentId=73531&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-73531 from @estenberg.
- I thought of this solution but my intention was to do as few modifications to the code as possible to fix the bug in the first time. The said modification is more about refactoring to me and should come in a second time. But I agree with your proposition of keeping the string instead of converting it to an enum.
- I've no opinion on the local class based on the service name.

tzz:
@yann-soubeyrand I asked about arbitrary service states in https://tracker.mender.io/browse/CFE-2402?focusedCommentId=73545

If arbitrary service states are accepted, your PR can be simplified further, removing legacy code, so I think it's better to wait for an answer to that question. I hope you're OK with that. If not, please don't consider my position blocking to your PR, it's good.

yann-soubeyrand:
@tzz I updated my PR to take your proposition into account ;-)

tzz:
@jimis @kacf this looks good to me and the code is much simpler. I'm not sure if we're waiting for approval in https://tracker.mender.io/browse/CFE-2402 or something else.

tzz:
@kacf @jimis @yann-soubeyrand based on https://tracker.mender.io/browse/CFE-2402?focusedCommentId=73899 I think this is ready for review and we'll soon be able to use arbitrary service state strings!

@yann-soubeyrand you may want to add a policy acceptance test (see `tests/acceptance`) in addition to the C unit test. I can help with that part... basically you'd make a fake service bundle and use it to trigger some observable behavior. Otherwise I think this looks great.

yann-soubeyrand:
@tzz Done! ;-)

tzz:
...and now we wait...

jimis:
Can you add a "Changelog: commit" line to the main commit message? That way it will be included in 3.10's changelog.

Other than that, this looks good! Please rebase to latest master so that it gets tested by Travis, and then I'll merge it.

yann-soubeyrand:
@jimis Is it OK or do you need me to do some changes?

jimis:
Looks good to me, thanks @yann-soubeyrand ! 

@nickanderson can you have a quick look at the acceptance tests?

tests/acceptance/09_services/custom.cf
```diff
@@ -0,0 +1,61 @@
...
+
+bundle agent test
+{
+  meta:
+      "test_suppress_fail" string => "windows",
+        meta => { "redmine4772" };
+
+  services:
+      "myservice"
+      service_policy => "custom",
```
jimis: An acceptance test for the default value would be nice to have, i.e. omit completely this line, and do not set `service_policy` at all.

cf-agent/verify_services.c
```diff
@@ -193,35 +173,13 @@ static PromiseResult VerifyServices(EvalContext *ctx, Attributes a, const Promis
...
-        RlistAppend(&args, pp->promiser, RVAL_TYPE_SCALAR);
-        RlistAppend(&args, "restart", RVAL_TYPE_SCALAR);
-        break;
-
-    case SERVICE_POLICY_STOP:
-    case SERVICE_POLICY_DISABLE:
-    default:
-        RlistAppend(&args, pp->promiser, RVAL_TYPE_SCALAR);
-        RlistAppend(&args, "stop", RVAL_TYPE_SCALAR);
-        break;
```
yann-soubeyrand: @jimis Not sure, but it seems that the default case is never hit because the service_policy is the result of a ServicePolicyFromString() call and this functions return SERVICE_POLICY_START if the string is null or if it doesn't find the string in the possible values.

libpromises/conversion.c
```diff
@@ -668,14 +668,6 @@ AclInherit AclInheritFromString(const char *string)
     return ACL_INHERIT_NOCHANGE;
 }
 
-ServicePolicy ServicePolicyFromString(const char *string)
```
yann-soubeyrand: I think the new behavior is equivalent to the behavior of this removed code.

cf-agent/verify_services.c
yann-soubeyrand: Not sure, but it seems that the default case is never hit because the service_policy is the result of a ServicePolicyFromString() call and this functions return SERVICE_POLICY_START if the string is null or if it doesn't find the string in the possible values.

libpromises/conversion.c
```diff
@@ -668,14 +668,6 @@ AclInherit AclInheritFromString(const char *string)
...
-{
-    static const char *const SERVICE_POLICY_TYPES[] =
-        { "start", "stop", "disable", "restart", "reload", NULL };
-
-    return FindTypeInArray(SERVICE_POLICY_TYPES, string, SERVICE_POLICY_START, SERVICE_POLICY_START);
-}
-
```
yann-soubeyrand: I think the new behavior is equivalent to the behavior of this removed code.

cf-agent/verify_services.c
yann-soubeyrand: Not sure, but it seems that the default case is never hit because the service_policy is the result of a ServicePolicyFromString() call and this functions returns SERVICE_POLICY_START if the string is null or if it doesn't find the string in the possible values.

PR #2631, opened 2016-06-23
CFE-2345: Rlists and thread-safety

jimis:
It all started from a segfault in rlist_test.c but reviewing the locking code in rlist.c indicated that most of it was bogus. In the end all thread-safety got removed, since it's only used in single-threaded contexts.

### DIFF
--- a/libpromises/rlist.c
+++ b/libpromises/rlist.c
@@ -930,9 +930,7 @@ void RvalDestroy(Rval rval)
     switch (rval.type)
     {
     case RVAL_TYPE_SCALAR:
-        ThreadLock(cft_lock);
         free(RvalScalarValue(rval));
-        ThreadUnlock(cft_lock);
         return;
 
     case RVAL_TYPE_LIST:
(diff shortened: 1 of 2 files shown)


jimis:
trigger build

jimis:
@kacf can you please review? This should be quick.

kacf:
Looks good to me. Could you remove `cft_lock` after this? (There is a reference in nova as well, just FYI).

jimis:
@kacf the occurrences of `cft_lock` in `locks.c` are worrying me a bit, they are guarding database operations. They shouldn't be needed with LMDB, but I'm not confident for the rest of the DBs we're using.

kacf:
@jimis: If that's the only place, can you add the lock locally for the two deprecated DBs?

kacf:
LMDB should be fine, it's thread safe. It might even improve performance. Still, I would consider adding this only to master, given that it's not critical and it could be a case of "famous last words"! :-)

jimis:
FWIW I just removed the only occurence of `cft_lock` from nova.

PR #2632, opened 2016-06-24
Iprange refactor

jimis:
Supersedes pull request #2529.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -166,16 +166,6 @@ static FnCallResult FnReturnContext(bool result)
     return FnReturn(result ? "any" : "!any");
 }
 
-static FnCallResult FnReturnContextTrue()
-{
-    return FnReturnContext(true);
-}
-
-static FnCallResult FnReturnContextFalse()
-{
-    return FnReturnContext(false);
-}
-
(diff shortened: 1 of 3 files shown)


jimis:
The change in `iprange()` is that it no longer matches on VIPADDRESS, I'm almost sure VIPADDRESS is included in the address returned from `EvalContextGetIpAddresses(ctx)`.

jimis:
trigger build

tzz:
I disagree with the removal of `FnReturnContextTrue` but the main code is good.

jimis:
On Fri, Jun 24, 2016 at 4:16 PM, Ted Zlatanov <notifications@github.com>
wrote:

> I disagree with the removal of FnReturnContextTrue
>

I did that to get rid of the warning, it was only used in the old
`iprange()` and `FnReturnContext(true)` is used all over the place. I have
no preference between them, I just wanted to use one form consistently.

tzz:
Oh, I see. Cool.

jimis:
trigger build

jimis:
trigger build

jimis:
Merging then, thanks for the initial code and the review @tzz. Would you be able to submit a documentation pull request?

PR #2636, opened 2016-06-29
Coding style

jimis:


jimis:
This is about replacing `NULL == func(whatever)` with `func(whatever) == NULL`, and similar cases.

@tzz because you asked for it! ;-)
@kacf lets try to enforce this now in our reviews.

jimis:
trigger build

tzz:
`GOOD == this_pr`

kacf:
> @kacf lets try to enforce this now in our reviews.

Ack!

PR #2637, opened 2016-06-30
Slow copy_from ENT-2769

jimis:


jimis:
trigger build

jimis:
trigger build

jimis:
trigger build

jimis:
trigger build

jimis:
trigger build

jimis:
trigger build

jimis:
trigger build

jimis:
trigger build

jimis:
trigger build

jimis:
Cherry-picked the acceptance tests from pull request #2644. Thank you @nickanderson!

jimis:
trigger build

jimis:
@nickanderson added acceptance test for remote file copying sparseness, feel free to review it.

jimis:
trigger build

nickanderson:
Cool, test looked ok. I would not have thought to use diff to compare the source and dest files.

jimis:
Apparently Ubuntu doesn't have `bc` installed by default, changed test to use `expr`.

Also fixed a spurious buildslave error, which showed up the issue of `cumulative()` for ttime_range not being able to take the range it's documented for.

trigger build.

jimis:
trigger build

@kacf this build adds six `safe_open(O_TRUNC)` tests, which should simulate @nickanderson's test case. Lets see how it goes on the exotics.

kacf:
> this build adds six safe_open(O_TRUNC) tests

Nice, very thorough! Should I review the rest, or are you just testing?

jimis:
trigger build

libutils/file_lib.c
```diff
@@ -413,6 +408,7 @@ int safe_open(const char *pathname, int flags, ...)
         return -1;
     }
 
+    size_t final_size = (size_t) -1;
```
kacf: Should you not use `ssize_t` if you want to use -1?

libutils/file_lib.c
```diff
@@ -537,19 +563,33 @@ int safe_open(const char *pathname, int flags, ...)
...
+            if (tr_ret < 0)
+            {
+                Log(LOG_LEVEL_ERR,
+                    "safe_open: unexpected failure (ftruncate: %s)",
+                    GetErrorStr());
+                close(currentfd);
+                return -1;
+            }
         }
     }
```
kacf: Looks good. You should clone the test case `test_safe_open_safe_switched_symlink`, and call it with `O_TRUNC`. This will verify that your fix works even when facing race conditions. Of course you'll have to compare against empty content instead of the content currently there.

libutils/file_lib.c
jimis: No because I don't want negative numbers, I just want a "variable is uninitialised" flag.

Furthermore I'm using unsigned since it converts directly from `off_t` which is also unsigned. Hmmm maybe using directly `off_t` would be the best.

jimis:
Added more tests, fixed another issue in `safe_open()`.

trigger build

kacf:
Can you please break off the `safe_open` fix from the rest? It really has no relation, and it's hard to review when mixed with the rest.

jimis:
@kacf the `safe_open()` fixes should be these two commits: 3797c8d, 8b48731 with tests here:18c2ff7e121dc7373a0deaddfc7b567f8f6ec0dc.

jimis:
I understand this pull request grew a lot with seemingly unrelated issues, but if I was branching off to a different branch every time I needed to fix the seemingly minor issues that occured, this would have never been finished.

kacf:
> if I was branching off to a different branch every time I needed to fix the seemingly minor issues that occured, this would have never been finished.

I understand, and for small issues like `Fix memory leak`, I would agree, I do the same. But `safe_open()` is incredibly tricky, difficult to review, and subject to many platform differences. To put it another way, if this was a pull request from an external contributor there is no way we would have accepted it without splitting it up. Let's not abandon good practice just because we have commit access. :-)

In any case, I'll review those commits now, but I think we should try to stick to isolated pull requests in general. It isn't hard to switch branches in Git.

kacf:
`safe_open` part looks good. Let me know if you need review for the rest, of if this is WIP.

jimis:
@kacf your reviews are always appreciated, so feel free to review whatever you like. Maybe the part having to do with sparse file writing - it has replaced all sparse file writing in the code, including the code that writes the LMDB backup files (commit 9009dc3e91c0516e6b9a22f8506d32f34ff845ed). FWIW I'm quite confident since this has been going back and forth in our CI for weeks (see the amount of "trigger builds" in the comments). 

Only HP-UX is a mystery still, see commit a5d61096e5f7c9592ff3fc0b35fdf9fb5534628c - you write a sparse file and "du" shows *larger* disk usage than its size; strange but probably having to do with the storage backend our buildslave is using.

jimis:
Merging, all further reviewing is welcome.

jimis:
For the record, the time line of this pull request has been well
commented in ENT-2769 JIRA ticket. But since the ticket is a
private one, here is what this is fixing and why:

* Properly count "du" usage of files; Because of wrong defines and
  autoconf configuration, some parts of the code were disk usage of
  files in the wrong way.

* Since CFEngine 3.5.0, **most files writen by CFEngine are potentially
  sparse files.** That means that parts filled with zeroes are written
  as "holes" in order to save disk space. This was done extremely
  inefficiently, and copying of big binary files consumed a lot of
  **system** CPU usage, with the total copy time being many times slower
  than CFEngine 3.4 or earlier. So this was sped up significantly.

* Certain filesystems (e.g. vboxfs) don't support `ftruncate()`ing a
  file to an offset greater than its true length. This was CFEngine's
  behaviour since 3.5.0 leading to failed transfer on rare ocasions.
  Fixed it by writing a byte and immediately taking it back, a
  workaround already existing in versions before 3.4.0.

* Same issue (`ftruncate()` failing) also existed in `safe_open()`, a
  function being utilised literally all over our codebase. Fixed and
  wrote tests, which revealed the following:

* `safe_open()` didn't properly support `O_TRUNC` flag, causing files to
  not be truncated as they should. Fix and write tests.

* Sparse file writing was re-implemented in 3 parts of the codebase, all
  subtly different. Consolidate it in one place (new functions
  `SparseFileWrite()` and `SparseFileClose()`) and use the new functions
  everywhere. This will probably help us with similar bugs in the
  future.

* Rigorously test the functions described above with unit tests and
  acceptance tests (thanks @nickanderson). Special attention has been
  given to portability for exotic platforms (Solaris ZFS, HP-UX, AIX),
  fixing issues as they were occuring in our CI.

* *Processes promise* was failing on one buildslave with huge uptime,
  causing all builds to fail. To fix this, the internal type used for
  `cumulative()` and `ttime_range` was increased to `intmax_t`. Now
  these functins are able to support their documented range of 1000
  years (BUG: previously the weren't able to, they were rolling around
  to negative values for slightly high numbers). Tests were modified.

* Fix various minor issues that occured while testing this pull request
  in our CI: preexisting compilation warnings (_FILE_OFFSET_BITS
  redefinition), makefile inconsistencies (proper cleaning), CI issues
  (yacc/lex being erroneously required on all buildslaves while it
  should be required only on the "autogen" preparatory slave).

PR #2640, opened 2016-07-01
ENT-7949: add example for isipinsubnet()

tzz:
This will be used in https://github.com/cfengine/documentation/pull/1553 instead of the current inline example.


jimis:
trigger build

jimis:
Thanks!

PR #2641, opened 2016-07-06
CFE-2397: Avoid JSON incompatible code path in getindices().

kacf:
```
If we have two indexed CFEngine variables:

vars:
  "a[x]"    string => "one";
  "a[x][y]" string => "two";

this structure is impossible to turn into JSON because if we do, we
get:

{
  "a": {
    "x": "one",
    "x": {
      "y": "two"
    }
  }
}

which contains two identical "x" keys, and is therefore invalid JSON.

In this case the old code would crash in some cases, but it's order
dependent, so it didn't always happen.

It's not necessary to use JSON for getindices(), since it returns a
list, but this was nevertheless the path taken in the code. So fix it
by avoiding this code path and taking the legacy path which handles
this case. Some extra care is needed because there are several ways in
which the input can be JSON, either by being specified inline, or by
referring to a variable that holds JSON, and in this case we need to
take the JSON path.

Changelog: Fix occasional segfault when running getindices() on a
variable that has indices of multiple depths (e.g. both "a[x]" and
"a[x][y]").
```

A couple of cleanup commits as well.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -2247,6 +2247,8 @@ static FnCallResult FnCallGetIndicesClassic(EvalContext *ctx, ARG_UNUSED const P
 static FnCallResult FnCallGetIndices(EvalContext *ctx, ARG_UNUSED const Policy *policy, const FnCall *fp, const Rlist *finalargs)
 {
     const char *name_str = RlistScalarValueSafe(finalargs);
+    bool allocated = false;
+    JsonElement *json = NULL;
 
     // Protect against collected args (their rval type will be data
     // container). This is a special case to preserve legacy behavior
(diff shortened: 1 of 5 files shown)


kacf:
I will provide an acceptance test for this as well, but submitting now if people wanna have a look.

kacf:
@jimis: I would appreciate a review of this!

kacf:
trigger build

libpromises/evalfunction.c
```diff
@@ -2256,23 +2258,57 @@ static FnCallResult FnCallGetIndices(EvalContext *ctx, ARG_UNUSED const Policy *
...
+            json = VarRefValueToJson(ctx, fp, ref, NULL, 0, true, &allocated);
         }
-    }
 
-    // Try to load directly. This case will behave correctly whether
-    // finalrgs holds a scalar or not.
-    bool allocated = false;
-    JsonElement *json = VarNameOrInlineToJson(ctx, fp, finalargs, true, &allocated);
+        VarRefDestroy(ref);
+        if (done)
```
jimis: In what cases are you not `done`? Could you add a comment?

jimis:
@kacf it's good after having a bird's eye look.

I tried hard to understand the rest of the logic, but I don't grok the JSON in the evaluator at all, sorry. Go ahead and merge though, the build looks green (besides the usual spurious fails). I'm glad you picked up the task!

libpromises/evalfunction.c
kacf: Actually, I think I will remove it. Originally I had more cleanup, which is why I wanted all the code paths to end up here, but there is only `VarRefDestroy` now, and the logic is kind of convoluted, so I'll move the cleanup before each `done = true`, and then return directly instead.

kacf: @jimis: If you take a look at the last commit now, the intention should be clearer.

kacf:
trigger build

libpromises/evalfunction.c
jimis: Indeed, thanks!

kacf:
Crap, I think I discovered another corner case.

kacf:
Alright, corner case handled. I will squash that into the original commit before merging.

kacf:
trigger build

kacf:
trigger build

kacf:
trigger build

kacf:
Cherry-picked to 3.9.x as well. The bug doesn't exist on 3.7.x.

PR #2643, opened 2016-07-06
Correct the check for a connection cache hit so that connections are reused when possible.

estenberg:


jimis:
Looks good, I'll edit the message to link to CFE-2447 and merge and cherry-pick. BTW this would have been caught by a properly trained static analysis stage in our CI.

jimis:
Oh I can't trigger builds because of Jenkins being down, so this will have to wait a bit.

jimis:
Tested and merged manually, closing.

PR #2630, opened 2016-06-22
Make sure that no_output option can be used together with module.

pasinskim:
Currently, when commands promise is used there is no option
to override default 'no_output' behavior which is true.
This will allow overwriting the default 'no_output' value.

PR #2646, opened 2016-07-13
Make sure that no_output option can be used together with module.

kacf:
Currently, when commands promise is used there is no option
to override default 'no_output' behavior which is true.
This will allow overwriting the default 'no_output' value.

Changelog: Fix 'contain' attribute 'no_output' having no effect when
the 'commands' promise is using 'module => "true"'.

Based on a patch by Marcin Pasinski <marcin.pasinski@cfengine.com>.

CFE-2412

### DIFF
--- a/libpromises/attributes.c
+++ b/libpromises/attributes.c
@@ -364,7 +364,14 @@ ExecContain GetExecContainConstraints(const EvalContext *ctx, const Promise *pp)
     e.owner = PromiseGetConstraintAsUid(ctx, "exec_owner", pp);
     e.group = PromiseGetConstraintAsGid(ctx, "exec_group", pp);
     e.preview = PromiseGetConstraintAsBoolean(ctx, "preview", pp);
-    e.nooutput = PromiseGetConstraintAsBoolean(ctx, "no_output", pp);
+    if (PromiseBundleOrBodyConstraintExists(ctx, "no_output", pp))
+    {
+        e.nooutput = PromiseGetConstraintAsBoolean(ctx, "no_output", pp);
+    }
+    else
+    {
+        e.nooutput = PromiseGetConstraintAsBoolean(ctx, "module", pp);
+    }
     e.timeout = PromiseGetConstraintAsInt(ctx, "exec_timeout", pp);
     e.chroot = PromiseGetConstraintAsRval(pp, "chroot", RVAL_TYPE_SCALAR);
     e.chdir = PromiseGetConstraintAsRval(pp, "chdir", RVAL_TYPE_SCALAR);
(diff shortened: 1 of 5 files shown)


kacf:
@jimis: Improved version of #2630. Should be easy to review.

kacf:
trigger build

libpromises/attributes.c
```diff
@@ -364,6 +364,7 @@ ExecContain GetExecContainConstraints(const EvalContext *ctx, const Promise *pp)
     e.owner = PromiseGetConstraintAsUid(ctx, "exec_owner", pp);
     e.group = PromiseGetConstraintAsGid(ctx, "exec_group", pp);
     e.preview = PromiseGetConstraintAsBoolean(ctx, "preview", pp);
+    e.have_nooutput = PromiseBundleOrBodyConstraintExists(ctx, "no_output", pp);
```
jimis: Why do we need `have_nooutput`? Can't we just set the default for `no_output` here no matter if the attribute exists, and overwrite it with the given value if is indeed given?

tests/acceptance/08_commands/01_modules/module_with_no_output.cf.sub
```diff
@@ -0,0 +1,33 @@
...
+        module => "true",
+        contain => test_contain("true");
+
+      "$(G.cat) $(G.testdir)/module.default.txt"
+        module => "true";
+
+      "$(G.cat) $(G.testdir)/no_module.output.txt"
+        contain => test_contain("false");
+
+      "$(G.cat) $(G.testdir)/no_module.no_output.txt"
```
jimis: Are the `txt` files missing?

kacf:
See also https://github.com/cfengine/documentation/pull/1559.

libpromises/attributes.c
kacf: Oh, you're right. I thought originally it wasn't possible because this is in the body, and what we would need to check would be in the surrounding promise. But it looks like both may be available here. I'll check.

tests/acceptance/08_commands/01_modules/module_with_no_output.cf.sub
kacf: They are generated in the non-`.sub` file. I didn't want to clutter the directory with all these small oneline files.

libpromises/attributes.c
kacf: Fixed!

kacf:
trigger build

kacf:
Cherry-picked to 3.7.x and 3.9.x as well. Thanks for review!

PR #2653, opened 2016-07-21
Add: Test that variable data containers can be passed

nickanderson:


nickanderson:
This test fails, but is marked with soft fail.

```
> $ cf-agent -KIf ./pass_variable_container.cf -DAUTO
R: test description: Test that variable data containers can be passed.
R: /home/nickanderson/CFEngine/core/tests/acceptance/01_vars/04_containers/./../../dcs.cf.sub SFAIL/CFE-2434
   error: List or container parameter '$(bundle).d' not found while constructing scope 'report_data' - use @(scope.variable) in calling reference
   error: List or container parameter '$(found).d' not found while constructing scope 'report_data' - use @(scope.variable) in calling reference
R: /home/nickanderson/CFEngine/core/tests/acceptance/01_vars/04_containers/./pass_variable_container.cf FAIL
```

PR #2655, opened 2016-07-22
Add: Tests for multiple candidate owners and group perms

nickanderson:


nickanderson:
rebased changes and this automatically closed itself. fun. I guess @jimis you included those tests as part of a separate commit.

PR #2658, opened 2016-07-26
Add --masterfiles option to changelog generator

nickanderson:


misc/changelog-generator/changelog-generator
```diff
@@ -35,6 +35,7 @@ LOG_TYPE = 0
 LOG_REPO = 1
 LOG_COMMUNITY = 2
 LOG_ENTERPRISE = 4
+LOG_MASTERFILES = 5
```
kacf: These are bit flags, so the next number in this sequence should be 8. Fixed that in #2670.

kacf:
Superseded by #2670.

PR #2662, opened 2016-07-30
CFE-2439: failsafe policy should not copy VCS files

yann-soubeyrand:
Failsafe policy does some file copies to update policy and modules. When doing
these file copies, it should avoid copying VCS files (like .gitignore files).

### DIFF
--- a/libpromises/failsafe.cf
+++ b/libpromises/failsafe.cf
@@ -112,6 +112,7 @@ bundle agent failsafe_cfe_internal_update
         handle => "failsafe_cfe_internal_bootstrap_update_files_sys_workdir_inputs_shortcut",
         copy_from => failsafe_scp("$(masterfiles_dir_remote)"),
         depth_search => failsafe_recurse("inf"),
+        file_select => failsafe_exclude_vcs_files,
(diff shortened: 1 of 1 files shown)


nickanderson:
Discussion in https://tracker.mender.io/browse/CFE-2439

nickanderson:
trigger build

jimis:
I'm a bit hesitant to merge this. /Maybe/ `gitignore` and `mailmap` (why exclude the latter BTW?) is needed by some. And generally I think the simplest the failsafe, the best it is.

On the other hand, `failsafe.cf` should only fetch the minimum policy, and leave the rest of the fetching and system convergence to the other policy. So if somebody needs more files he can get them with his "update.cf".

So I don't have a clear opinion in this. @nickanderson I leave this on you.

nickanderson:
I think this makes sense.

We already skip copying of .git and .svn subdirectories.

This makes me think that failsafe/bootstrap should potentially only copy *.cf files and common data files like json, yaml, csv.

neilhwatson:
.txt too please.

yann-soubeyrand:
@nickanderson Is this modification a possible candidate for a 3.7 maintenance release?

PR #2667, opened 2016-08-02
Add: Test comments on classes promsies are emitted in verbose logs

nickanderson:
Jira #CFE-2443

PR #2666, opened 2016-08-02
Add: Test that comments on vars promises are emitted in verbose output

nickanderson:
Jira #CFE-2442

PR #2668, opened 2016-08-02
Add: Test comments about reports are emited in verbose output.

nickanderson:


nickanderson:
This passes on 3.9.0.

It is related to https://github.com/cfengine/core/pull/2666 and https://github.com/cfengine/core/pull/2667

jimis:
Do we want to? Discussion on #2666.

PR #2669, opened 2016-08-02
Add: Test that expireafter works with methods

nickanderson:
Jira #CFE-2444

Note this test currently fails on 3.9.0.

cf-agent -Kf ./action_ifelapsed_respected.cf -DAUTO
R: test description: Test that methods are terminated as expected when ifelapsed is set
R: /home/nickanderson/CFEngine/core/tests/acceptance/21_methods/./../dcs.cf.sub SFAIL/CFE-2444
R: /home/nickanderson/CFEngine/core/tests/acceptance/21_methods/./action_ifelapsed_respected.cf FAIL

### DIFF
--- a/tests/acceptance/21_methods/action_ifelapsed_respected.cf
+++ b/tests/acceptance/21_methods/action_ifelapsed_respected.cf
@@ -1,0 +1,24 @@
+body common control
+{
+      inputs => { "../default.cf.sub" };
+      bundlesequence  => { default("$(this.promise_filename)") };
+      version => "1.0";
+}
+
+bundle agent test
+{
+  meta:
+    "description" string => "Test that methods are terminated as expected when ifelapsed is set";
+
+    # The subtest uses /bin/sleep so the test needs work on windows.
+    "test_soft_fail"
+      string => "any",
+      meta => { "CFE-2444" };
+}
+
+bundle agent check
+{
+  methods:
+    "" usebundle => dcs_passif_output(".*", ".*DID NOT TIMEOUT.*", "$(sys.cf_agent) -Kf $(this.promise_filename).sub", $(this.promise_filename));
+}
+
--- a/tests/acceptance/21_methods/action_ifelapsed_respected.cf.sub
+++ b/tests/acceptance/21_methods/action_ifelapsed_respected.cf.sub
@@ -1,0 +1,23 @@
+bundle agent main
+{
+  methods:
+    "test"
+      action => expireafter_ten_sec;
+}
+
+body action expireafter_ten_sec
+{
+  expireafter => "10";
+}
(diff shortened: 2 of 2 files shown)


tests/acceptance/21_methods/action_ifelapsed_respected.cf.sub
```diff
@@ -0,0 +1,23 @@
+bundle agent main
+{
+  methods:
+    "test"
+      action => expireafter_ten_sec;
+}
+
+body action expireafter_ten_sec
+{
+  expireafter => "10";
```
eysteins: This is in minutes.

tests/acceptance/21_methods/action_ifelapsed_respected.cf
```diff
@@ -0,0 +1,24 @@
...
+    # The subtest uses /bin/sleep so the test needs work on windows.
+    "test_soft_fail"
+      string => "any",
+      meta => { "CFE-2444" };
+}
+
+bundle agent check
+{
+  methods:
+    "" usebundle => dcs_passif_output(".*", ".*DID NOT TIMEOUT.*", "$(sys.cf_agent) -Kf $(this.promise_filename).sub", $(this.promise_filename));
```
eysteins: Not sure if this will work as the promise gets expired by the agent that runs afterwards, IIRC. So I think the next agent would kill the entire run of the current agent.

jimis: `-K` as argument disables all locks.

tests/acceptance/21_methods/action_ifelapsed_respected.cf.sub
jimis: @nickanderson I agree with @eysteins here, you need to set this to "1" minute and sleep something like 120 seconds.

jimis: BTW the filename and the pull request title mention "ifelapsed" test, but this is about expireafter.

jimis: Finally you might want to make this test a "timed' test, since it's going to sleep a long time. But I believe @kacf has already implemented `ifelapsed` and `expireafter` tests. Did you find them?

kacf:
Check out `tests/acceptance/00_basics/ifelapsed_and_expireafter/timed/defaults.cf` in particular. It may cover this.

jimis:
should we just close this @nickanderson ?

nickanderson:
Yes, I had a fundamental mis-understanding of how this was supposed to work. 


Docs updated and ticket rejected.

PR #2671, opened 2016-08-10
Use open instead of safe_open for hardcoded /dev/null path. 

pasinskim:
When using Solaris with some kind of NFS safe_open might cause some weird errors.


pasinskim:
trigger build

kacf:
I would add to the commit message for clarity:

"This avoids the check for insecure symlinks on `/dev`, since the path is hardcoded and we expect that path to be safe in all cases."

pasinskim:
@jimis this should be safe to merge and it is already reviewed by @kacf.

PR #2672, opened 2016-08-11
CFE-2225, CFE-2325: Fix for double logging of output_prefix, and log process name for cf-agent syslog messages

cobaugh:
This is my proposed fix for 2225 and 2325. When output_prefix is not set, cf-agent syslog messages have an empty process name. When it is set, we use that instead of the process name, and also include it elsewhere in the log message, doubling the ocurrences. By passing NULL to openlog(), we tell it to log the process name, which is desirable, and only include output_prefix once in the log message.

### DIFF
--- a/libpromises/generic_agent.c
+++ b/libpromises/generic_agent.c
@@ -852,7 +852,7 @@ bool GenericAgentArePromisesValid(const GenericAgentConfig *config)
 #if !defined(__MINGW32__)
 static void OpenLog(int facility)
 {
-    openlog(VPREFIX, LOG_PID | LOG_NOWAIT | LOG_ODELAY, facility);
+    openlog(NULL, LOG_PID | LOG_NOWAIT | LOG_ODELAY, facility);
 }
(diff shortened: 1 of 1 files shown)


cobaugh:
I also propose we apply this simple fix to at least 3.7.x, possibly others.

jimis:
I like the approach, discussion in CFE-2225.

jimis:
Looks good but before merging, please:
* rebase you branch to latest master (so it gets tested by Travis-CI)
* change your commit message to be something that can be added to the Changelog (like your pull request description) and add a separate line "Changelog: title" or "Changelog: commit" depending on your message (see HACKING.md).

jimis:
@cobaugh do you have time to do the simple changes I asked for?

jimis:
Moved your commit [here](https://github.com/cfengine/core/pull/2716) after editing your commit message to include changelog tag and rebasing to get Travis testing it. Closing this one.

PR #2673, opened 2016-08-13
CFE-2448: install binaries in bin instead of sbin

yann-soubeyrand:
https://tracker.mender.io/browse/CFE-2448

Since CFEngine can be used by normal users (for example to verify a policy using cf-promises), its binaries should be installed in bin instead of sbin when using FHS layout.

### DIFF
--- a/cf-agent/Makefile.am
+++ b/cf-agent/Makefile.am
@@ -121,7 +121,7 @@ endif
 endif
 
 if !BUILTIN_EXTENSIONS
-  sbin_PROGRAMS = cf-agent
+  bin_PROGRAMS = cf-agent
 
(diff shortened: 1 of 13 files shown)


jimis:
Can you rebase to latest master in order to be tested with Travis? It should give zero conflicts.

In general I'm a slightly against this as it's changing paths people are used to (`/usr/local/sbin`) which might break some scripts. And it is always the case that users can execute cf-promises by providing the full path.

@nickanderson @estenberg what do you think?

jimis:
I'm testing this in Jenkins and see what brakes in enterprise. Triggered as bootstrap-enterprise-pr-205.

kacf:
Thanks, I brought this up to date and merged manually.

PR #2674, opened 2016-08-13
CFE-2448: install binaries in bin instead of sbin

yann-soubeyrand:
https://tracker.mender.io/browse/CFE-2448

Since CFEngine can be used by normal users (for example to verify a policy using cf-promises), its binaries should be installed in bin instead of sbin when using FHS layout.


yann-soubeyrand:
Closing as it's not acceptable for a maintenance update.

PR #2675, opened 2016-08-13
CFEngine should put its logs in /var/log and its PID files in /var/run when using FHS layout

yann-soubeyrand:
https://tracker.mender.io/browse/CFE-2449

When using FHS layout, CFEngine logs should go to /var/log and its PID files should go to /run or /var/run.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -53,8 +53,7 @@ export TAR_OPTIONS
 
 EXTRA_DIST = ChangeLog INSTALL README.md LICENSE CFVERSION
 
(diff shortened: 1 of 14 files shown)


kacf:
This is based on the bin/sbin PR, where no consensus has bee reached yet, so I'm gonna skip that one and just cherry-pick the rest of the change, which I think looks good. I have #2823.

kacf:
Thanks @yann-soubeyrand, pushed 53b1f72f1d3acc6057af9694c8d86cf8b8d5132b manually.

PR #2676, opened 2016-08-13
CFEngine should put its logs in /var/log and its PID files in /var/run when using FHS layout

yann-soubeyrand:
https://tracker.mender.io/browse/CFE-2449

When using FHS layout, CFEngine logs should go to /var/log and its PID files should go to /run or /var/run.


yann-soubeyrand:
Closed as non acceptable for a maintenance update.

PR #2679, opened 2016-09-01
CFE-2365: Fix Allow to ignore missing sources in file copy operations

basvandervlies:
Changelog: Title

### DIFF
--- a/cf-agent/verify_files_utils.c
+++ b/cf-agent/verify_files_utils.c
@@ -2636,11 +2636,22 @@ static PromiseResult CopyFileSources(EvalContext *ctx, char *destination, Attrib
 
     if (cf_stat(BufferData(source), &ssb, attr.copy, conn) == -1)
     {
-        cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_FAIL, pp, attr,
-             "Can't stat file '%s' on '%s' in files.copy_from promise",
-             BufferData(source), conn ? conn->remoteip : "localhost");
-        BufferDestroy(source);
-        return PROMISE_RESULT_FAIL;
+        if (attr.copy.missing_ok)
+        {
+            cfPS(ctx, LOG_LEVEL_DEBUG, PROMISE_RESULT_NOOP, pp, attr,
+                 "Can't stat file '%s' on '%s' but promise is kept 'missing_ok' in files.copy_from promise",
+                 BufferData(source), conn ? conn->remoteip : "localhost");
+            BufferDestroy(source);
+            return PROMISE_RESULT_CHANGE;
+        }
+        else
+        {
+            cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_FAIL, pp, attr,
+                 "Can't stat file '%s' on '%s' in files.copy_from promise",
+                 BufferData(source), conn ? conn->remoteip : "localhost");
+            BufferDestroy(source);
+            return PROMISE_RESULT_FAIL;
+        }
     }
 
     start = BeginMeasure();
--- a/libpromises/cf3.defs.h
+++ b/libpromises/cf3.defs.h
@@ -1491,6 +1491,7 @@ typedef struct
     int purge;
     short timeout;
     ProtocolVersion protocol_version;
+    int missing_ok;
 } FileCopy;
 
 typedef struct
--- a/libpromises/attributes.c
+++ b/libpromises/attributes.c
@@ -915,6 +915,7 @@ FileCopy GetCopyConstraints(const EvalContext *ctx, const Promise *pp)
     f.encrypt = PromiseGetConstraintAsBoolean(ctx, "encrypt", pp);
     f.verify = PromiseGetConstraintAsBoolean(ctx, "verify", pp);
     f.purge = PromiseGetConstraintAsBoolean(ctx, "purge", pp);
+    f.missing_ok = PromiseGetConstraintAsBoolean(ctx, "missing_ok", pp);
     f.destination = NULL;
 
     return f;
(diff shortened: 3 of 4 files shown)


nickanderson:
Nice

basvandervlies:
there is also another fix for this problem.  See:
 * https://github.com/ensc/cfengine3/commits/missing-ok-next 

Which onne will be considered to be included in cfengine?

nickanderson:
@ensc do you plan to submit your branch as a pull request?

@jimis do you have any opinions on the difference in implementation between this pull request and [the branch](https://github.com/ensc/cfengine3/commits/missing-ok-next) @ensc has been working on?

The [feature request itself](https://tracker.mender.io/browse/CFE-2365) is a good addition IMHO.

cf-agent/verify_files_utils.c
```diff
@@ -2636,11 +2636,22 @@ static PromiseResult CopyFileSources(EvalContext *ctx, char *destination, Attrib
...
-             BufferData(source), conn ? conn->remoteip : "localhost");
-        BufferDestroy(source);
-        return PROMISE_RESULT_FAIL;
+        if (attr.copy.missing_ok)
+        {
+            cfPS(ctx, LOG_LEVEL_DEBUG, PROMISE_RESULT_NOOP, pp, attr,
+                 "Can't stat file '%s' on '%s' but promise is kept 'missing_ok' in files.copy_from promise",
+                 BufferData(source), conn ? conn->remoteip : "localhost");
+            BufferDestroy(source);
+            return PROMISE_RESULT_CHANGE;
```
jimis: Have you tested that it works like that? I believe this case should continue to the rest of the function. Also you shouldn't use `cfPS()` for just printing a DEBUG message, simple `Log()` will do.

jimis:
It's a new feature that needs acceptance tests to be merged. The branch from @ensc includes an attempt from one, but the branch itself doesn't look to be ready yet.

cf-agent/verify_files_utils.c
jimis: Hmmm on a second look this might be right, it should indeed return at this point. But isn't PROMISE_RESULT_NOOP the correct return code?

ensc:
I would consider my branch as feature complete but not style complete.  Because it changes the client - server protocol (uses 8th, currently unused octet for error information), I would like some more feedback about this before I try to make it clean.

Patchset itself is working fine; 'cf-agent' is silent when remote file is missing (I guess, with @jimis patch there is still some ugly noise and client will reconnect to server when file is missing).  'cf-servd' creates

> 192.168.12.76> Failed to canonicalise filename '...'  (realpath: No such file or directory)

log entries. Omitting these messages will require additional protocol changes.

cf-agent/verify_files_utils.c
basvandervlies: @jimis I have tested it with PROMISE_RESULT_NOOP, but that did not work. In needed the cfPS with the return code. If there is a better method to do this?

basvandervlies:
acceptance test added

jimis:
The 3.10.0 Beta is coming any time now, and I really want to merge this feature before the Beta (which is a feature freeze for 3.10 features). So I'm building'n'testing in my own branch, applying changes where due. I'll comment here what I needed to change to make this work.

libpromises/cf3.defs.h
```diff
@@ -1491,6 +1491,7 @@ typedef struct
     int purge;
     short timeout;
     ProtocolVersion protocol_version;
+    int missing_ok;
```
jimis: `bool`

tests/acceptance/16_cf-serverd/serial/copy_missing_ok.cf.sub
```diff
@@ -0,0 +1,55 @@
...
+bundle agent init
+{
+}
+
+bundle agent test
+{
+    files:
+        "$(G.testdir)/missing_ok"
+            classes => classes_generic("test"),
+            copy_from       => dcs_remote_cp_X("does_not_exists", "localhost", "true");
```
jimis: By using "localhost" here, the agent uses a shortcut and copies from the local directory without ever hitting the test server. I tested it with:
```
./testall --no-clean --extraclasses=debug_mode 16_cf-serverd/serial/copy_missing_ok.cf
```

which runs the server in debug mode and then I examined the logs; no incoming connections.

jimis:
@basvandervlies this is the debug server log from your acceptance test. As @ensc pointed out, there are some ugly messages:

```
 verbose: 127.0.0.1>      Received:    STAT /core/tests/acceptance/workdir/__16_cf-serverd_serial_copy_missing_ok_cf/tmp/TESTDIR.cfengine/dir/does_not_exists
    info: 127.0.0.1> Failed to canonicalise filename '/core/tests/acceptance/workdir/__16_cf-serverd_serial_copy_missing_ok_cf/tmp/TESTDIR.cfengine/dir/does_not_exists' (realpath: No such file or dir
   debug: 127.0.0.1> SendTransaction header: t 19
   debug: 127.0.0.1> SendTransaction data: BAD: Request denied
    info: 127.0.0.1> Closing connection due to illegal request: SYNCH 1476134577 STAT /core/tests/acceptance/workdir/__16_cf-serverd_serial_copy_missing_ok_cf/tmp/TESTDIR.cfengine/dir/does_not_exists
    info: 127.0.0.1> Closing connection, terminating thread
```

jimis:
I'm thinking this should be merged as a new feature before the beta, and we can iron out the details during the beta phase. Worst case we revert.

jimis:
No this is not ready, so I'm not merging, at least not yet.

@ensc if you think your code addresses this one better you should open a pull request in order to be considered for merge.

basvandervlies:
@jimis wat is missing? Is it due to the cf-serverd log messages?

jimis:
@basvandervlies yes, the log messages will not be trivial to deal with after the beta. 

Also needs more tests, for example what happens with a recursive copy_from, where the first file doesn't exist but the second one does? Does it recover or not?

basvandervlies:
@jimis The cf-serverd log messages is not covered in this pull request. I only focused on the client.  @ensc has gone this route. 

I think this only useful for 1 to 1 file copies or iterate over a list. With recurse you really do not care if a file exists of not. I did another test and that also succeeded

```
bundle agent main
{
    vars:
        "cp_files" slist => { "no_file", "def.json" };

    files:
            "/var/tmp/$(cp_files)"
                classes => classes_generic("$(cp_files)"),
                copy_from       => test_hash_cp( "/scratch/cfengine3/$(cp_files)", "true");

    reports:
        any::
            "FIle promise kept $(cp_files)"
                ifvarclass => "$(cp_files)_kept";
}
```

nickanderson:
ping @jimis

jimis:
Triggered jenkins job https://ci.cfengine.com/job/bootstrap-enterprise-pr/510/

PR #2681, opened 2016-09-02
Remove bootstrap_inc_uptodate_test.sh, which is unneeded now.

kacf:
The bootstrap.inc file is regenerated during the build now, and
usually at "make dist" time. This means that a host does not need to
be able to generate that file in order to build CFEngine, if it using
the tarball (and our build slaves do). However, since the test exists,
it would fail it if the tools were not available.

Since it is in fact better to test that CFEngine builds successfully
without the tools, remove the test instead, since any manner in which
it could fail will instead lead to a broken build now.


kacf:

PR #2682, opened 2016-09-05
ACL bug CFE-616

olehermanse:
Fixes CFE-616. Added acl_entries to phase 1 of VerifyFilePromise to ensure acls are applied, and changed order of applying perms and acls. ACLs are now applied after regular perms and the test passes.

### DIFF
--- a/cf-agent/verify_files.c
+++ b/cf-agent/verify_files.c
@@ -415,7 +415,7 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
 
 /* Phase 1 - */
 
-    if ((exists && ((a.haverename) || (a.haveperms) || (a.havechange) || (a.transformer))) ||
+    if ((exists && ((a.haverename) || (a.haveperms) || (a.havechange) || (a.transformer)) || (a.acl.acl_entries != NULL)) ||
         ((exists || link) && a.havedelete))
     {
         lstat(path, &oslb);     /* if doesn't exist have to stat again anyway */
--- a/cf-agent/verify_files.c
+++ b/cf-agent/verify_files.c
@@ -694,7 +694,7 @@ PromiseResult ScheduleEditOperation(EvalContext *ctx, char *filename, Attributes
         {
             strcpy(edit_bundle_name, (char *) vp);
             args = NULL;
-        }             
+        }
         else
         {
             goto exit;
@@ -737,7 +737,7 @@ PromiseResult ScheduleEditOperation(EvalContext *ctx, char *filename, Attributes
         {
             goto exit;
         }
-        
+
         Log(LOG_LEVEL_VERBOSE, "Handling file edits in edit_xml bundle '%s'", edit_bundle_name);
 
         const Bundle *bp = EvalContextResolveBundleExpression(ctx, policy, edit_bundle_name, "edit_xml");
@@ -752,7 +752,7 @@ PromiseResult ScheduleEditOperation(EvalContext *ctx, char *filename, Attributes
         }
     }
 
-    
+
     if (a.edit_template)
     {
         Log(LOG_LEVEL_VERBOSE, "Rendering '%s' using template '%s' with method '%s'",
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -109,3 +109,4 @@ cscope.*
 
 # Mac Finder
 .DS_Store
+.ftpconfig
```
jimis: Why do we need this change?

cf-agent/verify_files_utils.c
```diff
@@ -2257,7 +2257,6 @@ PromiseResult VerifyFileAttributes(EvalContext *ctx, const char *file, struct st
 
     if (attr.acl.acl_entries)
     {
-        oleprint("Will call VerifyACL!");
```
jimis: Try to remove completely your commits with your personal debugging helpers. `git rebase -i` and its `squash` subcommand are your friends there.

tests/acceptance/10_files/12_acl/acl_dir.cf
```diff
@@ -1,42 +0,0 @@
...
-#        unless => "regcmp_matched_$(N)";
-}
-
-body acl default
-{
-    acl_method => "overwrite";
-    aces =>  { "all:r:allow", "user:*:rwx:allow", "group:*:rw:allow" };
-    acl_type => "generic";
-    acl_default => "access";
-}
```
jimis: You add this file in one commit, and then remove it in another. Maybe don't add it at all then?

cf-agent/verify_files.c
```diff
@@ -416,7 +416,7 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
 /* Phase 1 - */
 
     if ((exists && ((a.haverename) || (a.haveperms) || (a.havechange) || (a.transformer))) ||
-        ((exists || link) && a.havedelete))
+        ((exists || link) && a.havedelete) ||  (a.acl.acl_entries != NULL))
```
jimis: What if the file doesn't exist? I think the check would work better inside the first set of parentheses, after the `if exists &&` clause.

cf-agent/verify_files.c
```diff
@@ -694,7 +694,7 @@ PromiseResult ScheduleEditOperation(EvalContext *ctx, char *filename, Attributes
         {
             strcpy(edit_bundle_name, (char *) vp);
             args = NULL;
-        }
```
jimis: No need to be verbose in the commit message of such a simple commit. Simply write "Fix trailing whitespace" and that's good enough.

tests/acceptance/10_files/12_acl/acl_fail.cf
```diff
@@ -0,0 +1,67 @@
+# These tests don't work
```
jimis: This file `acl_fail.cf` do you want to add it to this commit set? Then you might want to give it a different name, since it doesn't fail any more. And I you probably forgot to delete the commented policy at the top.

jimis:
@olehermanse what about the previously disabled tests that were failing? You should re-enable them, since they will be passing with your fix.

jimis:
Another generic comment is that you should add a "Changelog" tag in one of your commit messages. See file `HACKING.md` for a description and commits from us as an example.

olehermanse:
@jimis : Okay, should be good now. Still need to squash commits into just the 2 main ones(whitespace and acl changes). Also need to get changes from upstream. Looks like travis is building.

olehermanse:
Incorporated changes using rebase. Just need to wait for the build and squash commits.

jimis:
Looks good! Please squash the relevant commits (except the whitespace fix) and I'll merge. :-)

olehermanse:
@jimis Squashed and build passed.

jimis:
Congrats @olehermanse for your first bugfix in Core!

@nickanderson @kacf You want this cherry-picked to supported branches?

kacf:
> You want this cherry-picked to supported branches?

Yes, I think this should be cherry-picked.

jimis:
Alright, cherry-picking to 3.7.x and 3.9.x.

cf-agent/verify_files.c
```diff
@@ -415,7 +415,7 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
 
 /* Phase 1 - */
 
-    if ((exists && ((a.haverename) || (a.haveperms) || (a.havechange) || (a.transformer))) ||
+    if ((exists && ((a.haverename) || (a.haveperms) || (a.havechange) || (a.transformer)) || (a.acl.acl_entries != NULL)) ||
```
jimis: The compiler is warning here, and on a second look the new check is not on the right parenthesis level, there is a slight logic error here. 

I'm a bad reviewer it seems. :-) I'll adjust and commit.

PR #2686, opened 2016-09-14
tar_portability_test.sh failing because of long paths.

olehermanse:
Fixed by removing workdir in the shell script before running the test.

### DIFF
--- a/tests/unit/tar_portability_test.sh
+++ b/tests/unit/tar_portability_test.sh
@@ -8,5 +8,6 @@ fi
 
 cd "$(dirname $0)/../.."
 
-tar --format=ustar -cf /dev/null *
+tar --exclude="tests/acceptance/workdir" --format=ustar -cf /dev/null *
+
 exit $?


olehermanse:
@jimis ready for review. Could also move the rm to after the linux test(?) - Don't know which makes more sense.

jimis:
Thanks! But after looking deeper into this, I believe it's not appropriate for a test to change the user's environment. It would be more appropriate to use tar's `--exclude` flag.

olehermanse:
@jimis Okay, does this look better? On my machine the test passes with this change and fails without. (workdir has files in it).

olehermanse:
Will need to squash before merge. Waiting for travis just in case.

jimis:
Perfect! Please squash and I'll merge. :-)

jimis:
Merging, thanks! I'll also cherry-pick to our maintenance branches, 3.7.x and 3.9.x.

PR #2687, opened 2016-09-19
CFE-546 Fixed parsing of host name/IP and port number in cf-runagent

olehermanse:
This fixes CFE-546. Added some more tests, the new parser should work for most weird cases.


olehermanse:
@jimis Please review this. Will squash commits after review.

libpromises/addr_lib.c
```diff
@@ -578,23 +578,42 @@ void ParseHostPort(char *s, char **hostname, char **port)
...
-            p = (p[1] == ':') ? p+2 : NULL;
+            *p = '\0';                      // '\0' terminate host name
+            p = (p[1] == ':') ? p+2 : NULL; // move port* forward or set NULL
+        }
+    }
+    else if( first_colon == NULL )          // no colon, no port, host correct
+    {
+        // Do nothing
+    }
+    else if( first_colon < first_dot || first_dot == NULL)
```
jimis: Please check first for `first_dot == NULL` in this expression, to avoid further examining `first_dot` if that's the case. It shouldn't change anything in functionality.

olehermanse:
@jimis Made stylistic changes as requested. Plugged memory leak in unit test. Squashed redundant commits. Ready for review/merge, waiting for travis.

tests/unit/addr_lib_test.c
```diff
@@ -13,134 +13,136 @@ static void test_ParseHostPort()
 {
     char *hostname, *port;
 
+    char test_string[64];
+
     // Domain name:
-    ParseHostPort(xstrdup("www.cfengine.com"), &hostname, &port);
+    ParseHostPort(strcpy(test_string, "www.cfengine.com"), &hostname, &port);
```
jimis: Thanks for fixing the leaks! You should also remove the TODO comments then.

libpromises/addr_lib.c
```diff
@@ -576,25 +576,47 @@ bool IsInterfaceAddress(const Item *ip_addresses, const char *adr)
...
+        }
+    }
+    else if ( first_colon == NULL )        // no colon, no port, host correct
+    {
+        // Do nothing
+    }
+    else if ( first_dot == NULL || first_colon < first_dot )
+    {
+        // If only one colon: (cfengine.com:222 or localhost:)
+        if( strchr(first_colon + 1, ':') == NULL )
```
jimis: space after `if`

libpromises/addr_lib.c
```diff
@@ -576,25 +576,47 @@ bool IsInterfaceAddress(const Item *ip_addresses, const char *adr)
...
     {
-        h = s;
         p = strchr(h, ':');
         if (p != NULL)
         {
-            *p = '\0';                           /* '\0'-terminate hostname */
+            *p = '\0';                     // '\0'-terminate hostname
             p++;
         }
     }
```
jimis: Could there be a different case? If so a generic `else` statement would make sense, otherwise you must somehow `assert()` that we've covered all cases.

jimis:
Adjusted manually and committed, closing!

libpromises/addr_lib.c
jimis: I ended up changing this to an `else` and `assert()`ing that we are handling the right case. So there is no way there is an unhandled branch.

PR #2688, opened 2016-09-20
Enable unsafe tests in Travis

jimis:


jimis:
All tests green, merging.

tests/acceptance/testall
```diff
@@ -474,6 +474,9 @@ export CFENGINE_TEST_OVERRIDE_WORKDIR TEMP CFENGINE_TEST_OVERRIDE_EXTENSION_LIBR
 
         TEST_END_TIME=$(unix_seconds)
 
+        # This script does not run under GAINROOT but it needs to write logs
+        $GAINROOT chmod ugo+rwxt "$WORKDIR"
```
jimis: @kacf you don't see any side-effects of this, right?

kacf: I thought it wouldn't be necessary, since `WORKDIR` is pre-created. And it's a bit insecure, but probably we don't need to worry about that here.

kacf: In any case, I'm not opposed, the comment was just FYI.

jimis: Great I'll leave this in then! This was figured out after many trial and error attempts, it seems that only changing the permissions *after* the test run works.

PR #2689, opened 2016-09-22
CFE-2455: Fix ghost cf-serverd process remnants started in the chroot

jimis:
This execresult() is being executed 4-5 because of the different
evaluation passes. The first times, the .cf file doesn't even exist. As
a result, failsafe.cf is created and executed, which launches
cf-serverd.

The fix here is necessary after the CFE-2162 changes were introduced, to
avoid running the command before populating the variable.

### DIFF
--- a/tests/acceptance/19_security/other_writeable/group_write.cf
+++ b/tests/acceptance/19_security/other_writeable/group_write.cf
@@ -50,7 +50,8 @@ bundle agent test
 {
 
   vars:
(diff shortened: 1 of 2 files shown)


jimis:
@nickanderson @kacf we should implement some safeguard against failsafe launching randomly during our acceptance tests. Something like writing an empty `failsafe.cf`. I'm still not sure why the same issue didn't appear on more `.x.cf` tests, and non-hpux machines.

PR #2691, opened 2016-09-22
CFE-2474: Ensure that running failsafe many times in a row acts the sam

jimis:
Previously the results were changing because of ifelapsed promise locking.

### DIFF
--- a/libpromises/failsafe.cf
+++ b/libpromises/failsafe.cf
@@ -34,6 +34,9 @@ body agent control
 {
       # Bootstrapping can't continue without keys
       abortclasses => { "no_ppkeys_ABORT_kept" };
+      # Make sure that running failsafe many times in a row does not
+      # change functionality
+      ifelapsed => "0";
 }
 
 ################################################################################


jimis:
@nickanderson please review. In my manual test cases this solves the problem, but I'll try to carve a test before merging.

kacf:
Each promise should still be executed in only one pass, so this should be good!

jimis:
> I am slightly concerned about failsafe going into a loop not having locks on any of the promises.

@nickanderson or alternatively move this condition to a classes promise, instead of having it as a side-effect of a reports promise. Do you think this would be easy?

nickanderson:
Are you suggesting to move it to a classes promise because the promise locking wouldn't affect it then? It could be moved to a classes promise very easily, but i did it as a side effect of the report so that the user would get feedback about what is going on instead of the agent just bailing out as soon as the class was defined. 

```
bundle agent failsafe_cfe_internal_checkkeys
 {
   classes:
       "have_ppkeys"
         expression => fileexists("$(sys.workdir)/ppkeys/localhost.pub"),
         handle     => "failsafe_cfe_internal_bootstrap_checkkeys_classes_have_ppkeys";
 
   reports:
     !have_ppkeys::
       "No public/private key pair is loaded, please create one by running \"cf-key\""
         classes => failsafe_results("namespace", "no_ppkeys_ABORT");
}
```

To do that and still provide the user with feedback ....


   classes:
       "have_ppkeys"
         expression => fileexists("$(sys.workdir)/ppkeys/localhost.pub"),
         handle     => "failsafe_cfe_internal_bootstrap_checkkeys_classes_have_ppkeys";

       "no_ppkeys_ABORT" expression => "reported_no_keys_kept";

   reports:
     !have_ppkeys::
       "No public/private key pair is loaded, please create one by running \"cf-key\""
         classes => failsafe_results("bundle", "reported_no_keys");
}
```

Then we would also want to adjust the abort class in common control to be `no_ppkeys_ABORT` instead of `no_ppkeys_ABORT_kept`

But instead of globally setting ifelapsed to 0 we could 

```
bundle agent failsafe_cfe_internal_checkkeys
 {
   classes:
       "have_ppkeys"
         expression => fileexists("$(sys.workdir)/ppkeys/localhost.pub"),
         handle     => "failsafe_cfe_internal_bootstrap_checkkeys_classes_have_ppkeys";
 
   reports:
     !have_ppkeys::
       "No public/private key pair is loaded, please create one by running \"cf-key\""
         classes => failsafe_results("namespace", "no_ppkeys_ABORT"),
         action => immediate;
}

# Brought in from the stdlib
body action immediate
# @brief Evaluate the promise at every `cf-agent` execution.
{
      ifelapsed => "0";
}
```

jimis:
> Are you suggesting to move it to a classes promise because the promise locking wouldn't affect it then? It could be moved to a classes promise very easily, but i did it as a side effect of the report so that the user would get feedback about what is going on instead of the agent just bailing out as soon as the class was defined. 

It makes more sense to be a classes promise since it affects the logic, but your argument of printing a message is valid. I wish I could give you an answer.

nickanderson:
@jimis i wont block the global ifelapsed 0 for failsafe.

PR #2694, opened 2016-09-27
CFE-2334: apply 'perm' only on copied files

ensc:
see https://tracker.mender.io/browse/CFE-2334


nickanderson:
Please see the comments in https://tracker.mender.io/browse/CFE-2334. I believe this change goes against all other current promise implementations.

nickanderson:
Thanks @ensc we appreciate the effort, but after discussions we think that a function bringing the abilitty to find remote files will be a better way forward. If your interested please follow CFE-2530

PR #2695, opened 2016-09-27
CFE-2478: Add test for slashes in array index from module protocol

nickanderson:


jimis:
Thanks @nickanderson, will merge. First you might want to change the commit message from "backslashes" to "slashes".

PR #2696, opened 2016-09-27
CFE-2460: Add: Test bundlesequnece iteration with empty and absent lists and data containers

nickanderson:


nickanderson:
Regressed since 3.7.4
bundlesequence_iterate_over_data_container.cf

Passing as expected:
empty_data_container_does_not_trigger_undefined_bundle.cf
empty_list_does_not_trigger_undefined_bundle.cf

FAILING as expected:
absent_data_container_or_list_in_bundlesequence_triggers_error.x.cf

nickanderson:
@jimis  bundlesequence_iterate_over_data_container.cf is that regression we discussed a couple of weeks ago

PR #2699, opened 2016-10-03
CFE-953 Custom ports for cf-agent bootstrap.

olehermanse:
@jimis Would like you to review and test this. I am also doing some more tests.

Also, the issue of what to do with cf-serverd remains: https://tracker.mender.io/browse/CFE-953

### DIFF
--- a/libpromises/bootstrap.c
+++ b/libpromises/bootstrap.c
@@ -35,6 +35,8 @@
 #include <string_lib.h>
 #include <files_lib.h>
 #include <known_dirs.h>
+#include <addr_lib.h>
+#include <communication.h>
 
 #include <assert.h>
 
@@ -116,18 +118,47 @@ bool WriteAmPolicyHubFile(bool am_policy_hub)
  * policy_server.dat file has been removed. Then this function will be called
  * with NULL as new_policy_server, and cf-serverd will keep running even
  * without a policy server set. */
-void SetPolicyServer(EvalContext *ctx, const char *new_policy_server)
+void SetPolicyServer(EvalContext *ctx, const char *host, const char* port)
 {
-    if (new_policy_server)
+    if (host != NULL)
     {
-        xsnprintf(POLICY_SERVER, CF_MAX_IP_LEN, "%s", new_policy_server);
-        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, "policy_hub", new_policy_server, CF_DATA_TYPE_STRING, "source=bootstrap");
+        xsnprintf(POLICY_SERVER, CF_MAX_IP_LEN, "%s", host);
+        EvalContextVariablePutSpecial( ctx,  SPECIAL_SCOPE_SYS, "policy_hub",
+                                       host, CF_DATA_TYPE_STRING,
+                                       "source=bootstrap" );
+        if (port != NULL && port[0] != '\n')
+        {
+            EvalContextVariablePutSpecial( ctx, SPECIAL_SCOPE_SYS,
+                                           "policy_hub_port", port,
+                                           CF_DATA_TYPE_STRING,
+                                           "source=bootstrap" );
+        }
+        else
+        {
+            EvalContextVariablePutSpecial( ctx, SPECIAL_SCOPE_SYS,
+                                           "policy_hub_port", "5308",
+                                           CF_DATA_TYPE_STRING,
+                                           "source=bootstrap" );
+        }
     }
     else
     {
         strcpy(POLICY_SERVER, "");
-        EvalContextVariableRemoveSpecial(ctx, SPECIAL_SCOPE_SYS, "policy_hub");
+        EvalContextVariableRemoveSpecial( ctx, SPECIAL_SCOPE_SYS,
+                                          "policy_hub" );
+        EvalContextVariableRemoveSpecial( ctx, SPECIAL_SCOPE_SYS,
+                                          "policy_hub_port" );
     }
+}
+
+void SetPolicyServerFromFile(EvalContext *ctx, const char* workdir)
+{
+    char* host = NULL;
+    char* port = NULL;
+    ParsePolicyServerFile(workdir, &host, &port);
+    SetPolicyServer(ctx, host, port);
+    free(host);
+    free(port);
 }
 
 /* Set "sys.last_policy_update" variable. */
@@ -157,13 +188,16 @@ static char *PolicyServerFilename(const char *workdir)
     return StringFormat("%s%cpolicy_server.dat", workdir, FILE_SEPARATOR);
 }
 
+/**
+ * Reads out the content from policy_server.dat file
+ * @return: contents of file in char*, NULL if fopen failed, must be freed(!)
+ */
 char *ReadPolicyServerFile(const char *workdir)
 {
     char contents[CF_MAX_IP_LEN] = "";
 
     char *filename = PolicyServerFilename(workdir);
     FILE *fp = fopen(filename, "r");
-    free(filename);
 
     if (fp)
     {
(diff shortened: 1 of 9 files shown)


olehermanse:
Hopefully acceptance tests will pass in travis. Will do some practical tests on wednesday to see if this is working as intended.

olehermanse:
Passed on my machine. Waiting for travis while testing manually some other cases.

olehermanse:
@jimis Review and maybe test please. Think this works okay for now. Should also write tests for this.

olehermanse:
This is how I've been testing this: https://gist.github.com/olehermanse/d98b5b1e0480701db73efb75aa150dee

libpromises/bootstrap.c
```diff
@@ -177,10 +198,57 @@ char *ReadPolicyServerFile(const char *workdir)
...
+    bool file_read = ParsePolicyServerFile(workdir, &host, port);
+    if (file_read == false)
+    {
+        return false;
+    }
+    char tmp_ipaddr[CF_MAX_IP_LEN];
+    if (Hostname2IPString(tmp_ipaddr, host, sizeof(tmp_ipaddr)) == -1)
+    {
+        Log(LOG_LEVEL_ERR,
+            "LookUpPolicyServerFile: ERROR, could not resolve '%s'", host);
```
jimis: LOG_LEVEL_ERR is very user visible, try to lose the `LookUpPolicyServerFile` hint. Also the "ERROR" string is automatically printed, so it is redundant in the message. I would just print `Unable to resolve policy server host: %s`.

libpromises/bootstrap.c
```diff
@@ -116,20 +118,39 @@ bool WriteAmPolicyHubFile(bool am_policy_hub)
...
-    if (new_policy_server)
+    if (host != NULL)
     {
-        xsnprintf(POLICY_SERVER, CF_MAX_IP_LEN, "%s", new_policy_server);
-        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, "policy_hub", new_policy_server, CF_DATA_TYPE_STRING, "source=bootstrap");
+        xsnprintf(POLICY_SERVER, CF_MAX_IP_LEN, "%s", host);
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, "policy_hub",      host, CF_DATA_TYPE_STRING, "source=bootstrap");
+        if (port != NULL && port[0] != '\n')
+        {
+            EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, "policy_hub_port", port, CF_DATA_TYPE_STRING, "source=bootstrap");
```
jimis: Try to keep new code no more than 78 cols wide. We have tons of old code of course, no need to change what was written years ago.

libpromises/bootstrap.c
```diff
@@ -177,10 +198,57 @@ char *ReadPolicyServerFile(const char *workdir)
...
+    (*host) = xstrdup(*host);
+    if (*port != NULL)
+    {
+        (*port) = xstrdup(*port);
+    }
+    free(contents);
+    return true;
+}
+
+/* Combines reading the policy server file, parsing and lookup (host->ip)
```
jimis: For proper auto-recognizable doxygen style, write your comment like this:
```
/**
 * blablabab
 */
```

jimis:
> This is how I've been testing this: https://gist.github.com/olehermanse/d98b5b1e0480701db73efb75aa150dee

This is very useful, thanks! I believe from a user's experience perspective this is ready: @nickanderson can you check this link and tell me if you agree?

nickanderson:
Looking at this again, I think there is an easier way to add this support.

Until now the MPF has assumed port `5308`. If you wanted to change that, you
have to edit the MPF and update the `copy_from` body to specify a port.

If this feature introduces a feature class (similar to
`feature_def_json_preparse`) I think that we could simplify these changes so
that they only affect the `copy_from` body itself, instead of making changes to
the parameters it takes.

For example:

    body copy_from u_rcp(from,server)
     {
           source      => "$(from)";
           compare     => "digest";
           trustkey    => "false";
    
         !am_policy_hub::
    
           servers => { "$(server)" };
    
         !am_policy_hub.feature_bootstrap_specific_port::
           portnumber => "$(sys.policy_hub_port)";
    
         cfengine_internal_encrypt_transfers::
           encrypt => "true";
    
         cfengine_internal_purge_policies::
           purge => "true";
    
         cfengine_internal_preserve_permissions::
           preserve => "true";
    }

This way old clients not having the feature will behave exactly the same as
today. New clients that have the feature, will use the port that populates
`sys.policy_hub_port` (and `sys.policy_hub_port` will be `5308` if not specified
in `policy_server.dat`.

It might be more elegant to apply a constraint to the body attribute directly,
but that's not valid in the language.

    bundle agent main
    {
      reports:
          "test";
    }
    
    body copy_from test
    {
          portnumber => $(sys.policy_hub_port)
            ifvarclass => isvariable("sys.policy_hub_port");
    }

Then for future development:

1.  Populate `sys.policy_servers` with a map of addresses to ports. Initially
    this would look like this:
    
        {
            "$(sys.policy_hub)": "$(sys.policy_hub_port)"
        }

2.  Support multiple addresses in `$(sys.workdir)/policy_server.dat` in the
    following format:
    
        IP1:port
        IP2:port
    
    -   `sys.policy_hub` should contain the first entry in the file
    
    -   `sys.policy_hub_port` contain the port of the first entry in the
        file (defaulted appropriately if not specified).

olehermanse:
Rebased. No conflicts now. Changed masterfiles per Nick's request.

Now doing:
Jimis (nitpicky) review changes
Testing again with final versions of both repositories.

libpromises/bootstrap.c
```diff
@@ -116,20 +118,39 @@ bool WriteAmPolicyHubFile(bool am_policy_hub)
...
     }
     else
     {
         strcpy(POLICY_SERVER, "");
         EvalContextVariableRemoveSpecial(ctx, SPECIAL_SCOPE_SYS, "policy_hub");
+        EvalContextVariableRemoveSpecial(ctx, SPECIAL_SCOPE_SYS, "policy_hub_port");
     }
 }
 
+void SetPolicyServerFromFile(EvalContext *ctx, const char* workdir)
```
olehermanse: SetPolicyServer is used in generic_agent.c so I will not be doing this, now.

libpromises/bootstrap.c
olehermanse: Good. I've changed my page guide and updated the code.

olehermanse:
@nickanderson @jimis This is ready AFAIK. I would really like for you to test it as well as review.

Must be merged together with cfengine/masterfiles#811

I also updated the guide:
https://gist.github.com/olehermanse/d98b5b1e0480701db73efb75aa150dee

What remains for future:
1. Unit (and acceptance?) tests for this feature.
2. policy_server.dat with multiple lines
3. Storing ports with each server in some list or key-value storage.

libpromises/bootstrap.c
```diff
@@ -116,20 +118,49 @@ bool WriteAmPolicyHubFile(bool am_policy_hub)
...
+        {
+            EvalContextVariablePutSpecial( ctx, SPECIAL_SCOPE_SYS,
+                                           "policy_hub_port", port,
+                                           CF_DATA_TYPE_STRING,
+                                           "source=bootstrap" );
+        }
+        else
+        {
+            EvalContextVariablePutSpecial( ctx, SPECIAL_SCOPE_SYS,
+                                           "policy_hub_port", "5308",
```
jimis: instead of `"5308"` you should use `CFENGINE_PORT_STR`.

libpromises/generic_agent.c
```diff
@@ -1001,9 +1008,12 @@ void GenericAgentInitialize(EvalContext *ctx, GenericAgentConfig *config)
     if (config->agent_type != AGENT_TYPE_KEYGEN)
     {
         LoadSecretKeys();
-        char *bootstrapped_policy_server = ReadPolicyServerFile(workdir);
-        PolicyHubUpdateKeys(bootstrapped_policy_server);
-        free(bootstrapped_policy_server);
+        char* ipaddr = NULL;
+        char* port = NULL;
```
jimis: style: star next to identifier

libpromises/generic_agent.c
```diff
@@ -507,7 +511,10 @@ void GenericAgentDiscoverContext(EvalContext *ctx, GenericAgentConfig *config)
         {
             Log(LOG_LEVEL_VERBOSE, "This agent is bootstrapped to: %s",
                 existing_policy_server);
-            SetPolicyServer(ctx, existing_policy_server);
+            char* host = NULL;
+            char* port = NULL;
```
jimis: style

jimis:
@olehermanse small details in the review, so I went on and merged. Nevertheless please take care of the issues in a new pull request. And be a bit more careful on coding style, I had mentioned in the previous review as well. 

:+1:

PR #2701, opened 2016-10-04
Fix spurious failures caused by parallel tests on some platforms.

kacf:
Also switch to more modern way of specifying automake options, by
specifying version in AC_INIT instead and using AM_INIT_AUTOMAKE for
options. This has been the preferred way in autoconf/automake for a
long time.


kacf:
Running this in PR analyzer.

PR #2704, opened 2016-10-04
Reapplied jimis fixes

kacf:


kacf:
This was green after all. Should never have had any doubt! ;-)

PR #2713, opened 2016-10-07
Add sys.user_data container for user starting agent.

tzz:
@jimis @nickanderson @skreuzer as discussed in #2575 and #1413 

For the current execution's user info, you still have to `getuserinfo()` which implicitly looks at the current UID. Or you can use the current promise UID variable: `getuserinfo($(this.promiser_uid))`. That way, we avoid wasting time converting the passwd struct to a data container on every promise evaluation, since I think it will be needed very rarely in the context of general CFEngine execution.

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -2881,6 +2881,56 @@ static time_t GetBootTimeFromUptimeCommand(time_t now)
 
 /*****************************************************************************/
 
+JsonElement* GetUserInfo(const void *passwd)
+{
+#ifdef __MINGW32__
+    return NULL;
+#else /* !__MINGW32__ */
+
+    // we lose the const to set pw if passwd is NULL
+    struct passwd *pw = (struct passwd*) passwd;
+
+    if (pw == NULL)
+    {
+        pw = getpwuid(getuid());
+    }
+
+    if (pw == NULL)
+    {
+        return NULL;
+    }
+
+    JsonElement *result = JsonObjectCreate(10);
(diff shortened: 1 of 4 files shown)


tzz:
I'll add docs as soon as this is merged, it's pretty minimal.

tzz:
Added `getuserinfo()` example with `sys.user_data` usage.

tzz:
@nickanderson docs in https://github.com/cfengine/documentation/pull/1573 use the getuserinfo.cf example so should be merged together.

nickanderson:
@jimis please ping me when this gets merged so that I can merge the docs in https://github.com/cfengine/core/pull/2713

libenv/sysinfo.c
```diff
@@ -2881,6 +2881,56 @@ static time_t GetBootTimeFromUptimeCommand(time_t now)
 
 /*****************************************************************************/
 
+JsonElement* GetUserInfo(const void *passwd)
```
jimis: why `void*` ? Any reason not using `const struct passwd *` ?

tzz: So the function can be provided to all platforms, including Windows.

jimis:
It's good, and thanks for the nice example as well @tzz!

libenv/sysinfo.c
jimis: Windows code should use the proper types as well, so ideally both would be type safe. Small thing, I'll change myself and commit.

jimis:
@nickanderson ready for docs

nickanderson:
@jimis takk

PR #2718, opened 2016-10-10
Enable -Werror in Travis-CI.

jimis:


jimis:
@tzz I meant only commit 2309b9f about you having a look, but somehow github applied my comment to the whole pull request.

PR #2720, opened 2016-10-11
changelog-generator: Omit signed-off-by entries in changelog output.

kacf:


kacf:

PR #2727, opened 2016-10-26
Fix/improve how hostname, ip and port are handled ( CFE-953 CFE-721 CFE-385 CFE-2481 )

olehermanse:
This is a bigger pull request which addresses multiple issues with the hostname and custom port features (CFE-953 CFE-546).

Previous PR for reference: cfengine/core#2699

Must be merged together with Pull Requests in masterfiles and enterprise.

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -378,8 +378,8 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
                 OpenNetwork();
 
                 config->agent_specific.agent.bootstrap_argument = xstrdup(optarg);
-                char * host;
-                char * port;
+
+                char *host, *port;
                 ParseHostPort(optarg, &host, &port);
 
                 char ipaddr[CF_MAX_IP_LEN] = "";
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -1981,6 +1981,8 @@ static int AutomaticBootstrap(GenericAgentConfig *config)
         if (strlen(ipaddr) < sizeof(POLICY_SERVER))
         {
             config->agent_specific.agent.bootstrap_argument = xstrdup(ipaddr);
+            config->agent_specific.agent.bootstrap_ip       = xstrdup(ipaddr);
+            config->agent_specific.agent.bootstrap_host     = xstrdup(ipaddr);
             ret = 0;
         }
         else
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -73,6 +73,7 @@
 #include <syslog_client.h>
 #include <man.h>
 #include <bootstrap.h>
+#include <policy_server.h>
 #include <misc_lib.h>
 #include <buffer.h>
 #include <loading.h>
@@ -1852,7 +1853,8 @@ static PromiseResult ParallelFindAndVerifyFilesPromises(EvalContext *ctx, const 
 
 static bool VerifyBootstrap(void)
 {
-    if (NULL_OR_EMPTY(POLICY_SERVER))
+    const char *policy_server = PolicyServerGet();
+    if (NULL_OR_EMPTY(policy_server))
     {
         Log(LOG_LEVEL_ERR, "Bootstrapping failed, no policy server is specified");
         return false;
@@ -1882,7 +1884,8 @@ static bool VerifyBootstrap(void)
         return false;
     }
 
-    Log(LOG_LEVEL_NOTICE, "Bootstrap to '%s' completed successfully!", POLICY_SERVER);
+
+    Log(LOG_LEVEL_NOTICE, "Bootstrap to '%s' completed successfully!", policy_server);
     return true;
 }
 
@@ -1978,7 +1981,8 @@ static int AutomaticBootstrap(GenericAgentConfig *config)
         Log(LOG_LEVEL_NOTICE, "Autodiscovered hub installed on hostname '%s', IP address '%s'",
             hostname, ipaddr);
 
-        if (strlen(ipaddr) < sizeof(POLICY_SERVER))
+        // TODO: This is a very bad way to check for valid IP(?)
+        if (strlen(ipaddr) < CF_MAX_IP_LEN)
         {
             config->agent_specific.agent.bootstrap_argument = xstrdup(ipaddr);
             config->agent_specific.agent.bootstrap_ip       = xstrdup(ipaddr);
--- a/contrib/masterfiles/git-failsafe.cf
+++ b/contrib/masterfiles/git-failsafe.cf
@@ -198,7 +198,7 @@ bundle agent git_update
    "$(inputs_dir)/cf_promises_validated"
         comment => "Check whether a validation stamp is available for a new policy update to reduce the distributed load",
          handle => "check_valid_update",
-      copy_from => u_rcp("$(master_location)/cf_promises_validated","$(sys.policy_hub_ip)"),
+      copy_from => u_rcp("$(master_location)/cf_promises_validated","$(sys.policy_hub)"),
          action => u_immediate,
         classes => u_if_repaired("validated_updates_ready");
 
@@ -218,7 +218,7 @@ bundle agent git_update
    "$(inputs_dir)"
            comment => "Copy policy updates from master source on policy server if a new validation was acquired",
             handle => "update_files_inputs_dir",
-         copy_from => u_rcp("$(master_location)","$(sys.policy_hub_ip)"),
+         copy_from => u_rcp("$(master_location)","$(sys.policy_hub)"),
       depth_search => u_recurse("inf"),
       file_select  => u_input_files,
         depends_on => { "check_valid_update" },
--- a/contrib/masterfiles/git-failsafe.cf
+++ b/contrib/masterfiles/git-failsafe.cf
@@ -198,7 +198,7 @@ bundle agent git_update
    "$(inputs_dir)/cf_promises_validated"
         comment => "Check whether a validation stamp is available for a new policy update to reduce the distributed load",
          handle => "check_valid_update",
-      copy_from => u_rcp("$(master_location)/cf_promises_validated","$(sys.policy_hub)"),
+      copy_from => u_rcp("$(master_location)/cf_promises_validated","$(sys.policy_hub_ip)"),
          action => u_immediate,
         classes => u_if_repaired("validated_updates_ready");
 
@@ -218,7 +218,7 @@ bundle agent git_update
    "$(inputs_dir)"
            comment => "Copy policy updates from master source on policy server if a new validation was acquired",
             handle => "update_files_inputs_dir",
-         copy_from => u_rcp("$(master_location)","$(sys.policy_hub)"),
+         copy_from => u_rcp("$(master_location)","$(sys.policy_hub_ip)"),
       depth_search => u_recurse("inf"),
       file_select  => u_input_files,
         depends_on => { "check_valid_update" },
--- a/libpromises/addr_lib.c
+++ b/libpromises/addr_lib.c
@@ -575,6 +575,53 @@ bool IsInterfaceAddress(const Item *ip_addresses, const char *adr)
     return false;
 }
 
+char *TrimWhitespace(char *s)
+{
+    if ( NULL_OR_EMPTY(s) )
+    {
+        return NULL;
+    }
+
+    // Leading whitespace:
+    while (isspace(s[0]))
+    {
+        ++s;
+    }
+
+    // Empty string (only whitespace):
+    if (s[0] == '\0')
+    {
+        return NULL;
+    }
+
+    // Trailing whitespace:
+    char *end = s + strlen(s) - 1; // Last byte before '\0'
+    while ( isspace(end[0]) )
+    {
+        --end;
+    }
+    end[1] = '\0';   // Null terminate string after last non space char
+
+    return s;
+}
+
+/**
+ * Simple check to avoid writing to illegal memory addresses.
+ * NOT a proper test for valid IP.
+ */
+static AddressType AddressTypeCheckValidity(char *s, AddressType address_type)
+{
+    if(NULL_OR_EMPTY(s))
+    {
+        return ADDRESS_TYPE_OTHER;
+    }
+    if(strlen(s) >= CF_MAX_IP_LEN)
+    {
+        return ADDRESS_TYPE_OTHER;
+    }
+    return address_type;
+}
+
 /**
  * Parses "hostname:port" or "[hostname]:port", where hostname may also be
  * IPv4 or IPv6 address string.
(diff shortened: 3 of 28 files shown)


olehermanse:
Last thing missing is POLICY_SERVER global should be bootstrap arg (host:port)

olehermanse:
POLICY_SERVER is now longer and contains the same as bootstrap arg and policy_server.dat
(host:port)

olehermanse:
This PR makes both the c code and variables a lot more clear imo.
Also, bootstrapping now shows the complete host port string:
```
 notice: Bootstrap to 'olehub:5309' completed successfully!
```

olehermanse:
Do not merge yet, I am still testing.

olehermanse:
Warning: This definitely does not work as of now, because POLICY_SERVER is used internally for copy_from files promises. I am working on a fix.

cf-agent/cf-agent.c
```diff
@@ -1981,6 +1981,8 @@ static int AutomaticBootstrap(GenericAgentConfig *config)
         if (strlen(ipaddr) < sizeof(POLICY_SERVER_IP))
         {
             config->agent_specific.agent.bootstrap_argument = xstrdup(ipaddr);
+            config->agent_specific.agent.bootstrap_ip       = xstrdup(ipaddr);
+            config->agent_specific.agent.bootstrap_host     = xstrdup(ipaddr);
```
olehermanse: I don't know how to test the avahi stuff, but I populated the variables to be more similar to a regular bootstrap.

olehermanse:
Do not merge yet. I should add the new sys.policy_hub variable before we merge, to ensure old stuff doesn't break.

cf-agent/cf-agent.c
olehermanse: I don't know how to test the avahi stuff, but I populated the variables to be more similar to a regular bootstrap.

olehermanse:
https://github.com/cfengine/core/pull/2727/commits/58b073e06605dc0a2dce028455e3d6dc05f51ca4 <- Did I do this correctly with regards to Changelog?

cf-agent/cf-agent.c
olehermanse: I don't know how to test the avahi stuff, but I populated the variables to be more similar to a regular bootstrap.

olehermanse:
Here are my instructions for usage/testing:
https://gist.github.com/olehermanse/d98b5b1e0480701db73efb75aa150dee

olehermanse:
I've tested this and it seems like it's working fine. The only issue I've found is: https://tracker.mender.io/browse/CFE-2494 But I don't think that has anything to do with this patch.

olehermanse:
@kacf @jimis Please review. This is a bigger PR, but should be easy to understand commit by commit.

Must be merged with: 
https://github.com/cfengine/masterfiles/pull/824
https://github.com/cfengine/enterprise/pull/324

kacf:
Since this is network code I will let @jimis review this.

jimis:
Interesting, but I believe it will definitely cause breakage and grief. What is the rationale behind changing the "policy_hub" variable? Can we follow a more backwards-compatible way? I think it would be nice to reopen the ticket and keep track of this.

olehermanse:
(Discussed with jimis:)

sys.policy_hub and POLICY_SERVER are both exactly like before when not using the new host:port feature. Backwards compatibility should only be an issue with old policy when using the new bootstrap features.

olehermanse:
I think this PR makes the variables a lot more consistent, with policy_server.dat, POLICY_SERVER and sys.policy_hub now being exactly the same. It also improves log output, for example here:
```
  notice: Bootstrap to 'olehub:5308' completed successfully!
```

olehermanse:
@jimis: Discussed with @nickanderson and it seems to make sense to instead do this on policy side:
`policy_hub` can be either host or ip
`policy_hub_ip` is always ip
`policy_hub_port` is always port

My idea was to have policy_hub be exactly what you bootstrapped to. Having policy_hub variable include port number is not useful according to nick.

Then I should be able to fix the ACL issue for `host/16` : https://tracker.mender.io/browse/CFE-2495

jimis:
> @jimis: Discussed with @nickanderson and it seems to make sense to instead do this on policy side:
> policy_hub can be either host or ip
> policy_hub_ip is always ip
> policy_hub_port is always port

Fully agree, as we discussed extensively.

Regarding the `POLICY_SERVER` global, it's better to work either as host or ip, and keep it what we bootstrapped to. Whether it has the port or not is not clear yet, you'll figure it out. :+1:

olehermanse:
I've now changed the policy_hub variable as discussed, rebased and pushed again. 
`sys.policy_hub` is now host or ip (not port).

jimis:
Then the next step is to build in Jenkins' pull request analyzer!

contrib/masterfiles/git-failsafe.cf
```diff
@@ -198,7 +198,7 @@ bundle agent git_update
    "$(inputs_dir)/cf_promises_validated"
         comment => "Check whether a validation stamp is available for a new policy update to reduce the distributed load",
          handle => "check_valid_update",
-      copy_from => u_rcp("$(master_location)/cf_promises_validated","$(sys.policy_hub)"),
+      copy_from => u_rcp("$(master_location)/cf_promises_validated","$(sys.policy_hub_ip)"),
```
jimis: Most `$(sys.policy_hub)` in the policy would be best to remain unchanged, since the code handles both IP and hostnames (especially true after the ACLs are fixed).

Just mentioning here to make sure we're on the same ship. :-)

olehermanse:
Must be merged with: 
https://github.com/cfengine/enterprise/pull/324
https://github.com/cfengine/nova/pull/990

contrib/masterfiles/git-failsafe.cf
olehermanse: Yes I agree, however I think maybe doing it like this for now, and changing back once ACL's are fixed is better(?).

olehermanse:
I've rebase and squashed the commits into a few that make sense on their own. I've started a jenkins build and will test this and hope we can merge if it looks promising.

Remaining:
1. Build & Testing
2. Fix ACL issues with hostname
3. Move away from using sys.policy_hub_ip where not strictly needed.

@jimis Please review.

Must be merged with: 
cfengine/enterprise#324
cfengine/nova#990

Relevant jenkins build:
https://ci.cfengine.com/job/bootstrap-enterprise-pr/82/

olehermanse:
Ideas for improving policyserver.c performance:
1. When policy server is ip - Leave host as NULL and avoid alloc and lookup. 
2. Cache IP with a time limit to avoid many lookups. Cache can last **long** as a host changing ip should not happen too frequently. 

I really like that we can do these sort of changes while keeping the interface (get methods) the same.

contrib/masterfiles/git-failsafe.cf
jimis: Sounds good! Then just leave the entries that are relevant to ACLs, and change back the `copy_from` ones to `policy_hub`.

libpromises/policyserver.c
```diff
@@ -0,0 +1,300 @@
...
+    free(host);
+    return true;
+}
+
+bool WritePolicyServerFile(const char *workdir, const char *new_policy_server)
+{
+    char *filename = PolicyServerFilename(workdir);
+
+    FILE *file = fopen(filename, "w");
+    if (!file)
```
jimis: Avoid `!file` since file is no `bool`, prefer `file != NULL`

libpromises/policyserver.c
```diff
@@ -0,0 +1,300 @@
...
+    {
+        Log(LOG_LEVEL_ERR, "Unable to write policy server file '%s' (fopen: %s)", filename, GetErrorStr());
+        free(filename);
+        return false;
+    }
+
+    fprintf(file, "%s", new_policy_server);
```
jimis: Since you are at it, would you mind adding a trailing `\n` in a separate commit at the end? I don't think it will break existing code, and I've been wanting to do it a since a long time...

libpromises/policyserver.c
```diff
@@ -0,0 +1,300 @@
...
+    }
+}
+
+
+//*******************************************************************
+// POLICY SERVER GET FUNCTIONS:
+//*******************************************************************
+
+// TODO: Shouldn't be necessary? Can be inline too
+static const char* PolicyServerCheckedReturn(char* rval)
```
jimis: Don't use the word `rval` for a string, `Rval` is a very widely type used in the CFEngine source code so it creates confusion. Use `host` or `ip` or whatever this is, or even `s` is better in this context since it's simple.

libpromises/policyserver.c
```diff
@@ -0,0 +1,300 @@
...
+                                       "source=bootstrap" );
+    }
+}
+
+
+//*******************************************************************
+// POLICY SERVER GET FUNCTIONS:
+//*******************************************************************
+
+// TODO: Shouldn't be necessary? Can be inline too
```
jimis: No it's not necessary, in fact it's quite confusing since it's unclear what is checked via the parameter and what via the global.

libpromises/policyserver.h
```diff
@@ -0,0 +1,131 @@
...
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+/** @file
+ * @brief Replaces global strings for POLICY_SERVER with get and set methods
```
jimis: It's nice that you write a global purpose description in the header file. However it should mean something to someone reading it, without knowing what's in the file or the past that this file replaces. 

For example: Functions to set the policy server by using IP/hostname/port, and to query it back.

libpromises/policyserver.c
```diff
@@ -0,0 +1,300 @@
...
+             filename, GetErrorStr() );
+
+        free(filename);
+        return NULL;
+    }
+}
+
+void SetPolicyServerFromFile(EvalContext *ctx, const char* workdir)
+{
+    char* contents = ReadPolicyServerFile(workdir);
```
jimis: Star to the right, here and in other places.

libpromises/policyserver.c
```diff
@@ -0,0 +1,300 @@
...
+#include <assert.h>
+#include <logging.h>
+#include <string_lib.h>
+#include <file_lib.h>
+
+//*******************************************************************
+// POLICY SERVER VARIABLES:
+//*******************************************************************
+
+static char *POLICY_SERVER      = NULL;
```
jimis: Add a comment that `POLICY_SERVER` is the exact string to the one we bootstrapped to and can be IP or hostname and includes port. And another comment that `POLICY_SERVER_HOST` has the port removed, and can be host or IP. If I undestand this right.

libpromises/policyserver.c
```diff
@@ -0,0 +1,300 @@
...
+
+    // Set POLICY_SERVER full string:
+    POLICY_SERVER = xstrdup(new_policy_server);
+
+    // Parse new_policy_server and set POLICY_SERVER_IP + POLICY_SERVER_PORT
+    {
+        // Parse policy server in a separate buffer:
+        char *host, *port;
+        char buffer[CF_MAX_SERVER_LEN];
+        strncpy(buffer, new_policy_server, CF_MAX_SERVER_LEN);
```
jimis: `strncpy()` is seldom needed, it's not a good function. Just allocate here, since setting this is not frequent operation.

jimis:
> Ideas for improving policyserver.c performance:
> 1. When policy server is ip - Leave host as NULL and avoid alloc and lookup. 

This indeed sounds necessary!

olehermanse:
@jimis I think I have addressed all the issues you commented and we discussed.

The plan going forward is:
1. Build and test in travis + jenkins.
2. Get this merged
3. Write unit tests
4. Fix ACL Issue

olehermanse:
Done with unit tests.  Waiting for travis to see if it's still failing the one acceptance test. Onto jenkins build if travis succeeds.

olehermanse:
Started build in jenkins: 
https://ci.cfengine.com/job/bootstrap-enterprise-pr/87/
https://ci.cfengine.com/job/testing-enterprise-pr/75/

cf-agent/cf-agent.c
```diff
@@ -1852,7 +1853,8 @@ static PromiseResult ParallelFindAndVerifyFilesPromises(EvalContext *ctx, const
 
 static bool VerifyBootstrap(void)
 {
-    if (NULL_OR_EMPTY(POLICY_SERVER))
+    const char* policy_server = PolicyServerGet();
```
jimis: style

libpromises/addr_lib.c
```diff
@@ -566,6 +566,53 @@ bool IsInterfaceAddress(const Item *ip_addresses, const char *adr)
...
 
+char *TrimWhitespace(char *s)
+{
+    if ( NULL_OR_EMPTY(s) )
+    {
+        return NULL;
+    }
+
+    // Leading whitespace:
+    while ( isspace(s[0]) )
```
jimis: In general, we don't use whitespace around the expression unless needed.

It's OK when it enhances readability, for example with complex expressions or for alignment purposes.

libpromises/bootstrap.c
```diff
@@ -75,159 +73,165 @@ bool BootstrapAllowed(void)
...
     {
         Log( LOG_LEVEL_VERBOSE, "Could not open file '%s' (fopen: %s)",
              filename, GetErrorStr() );
-
-        free(filename);
         return NULL;
     }
+    if (fgets(contents, CF_MAX_SERVER_LEN, fp) == NULL)
+    {
+        fclose(fp);
```
jimis: A verbose message here would help debugging in cases this happens.

libpromises/evalfunction.c
```diff
@@ -24,6 +24,7 @@
 
 #include <evalfunction.h>
 
+#include <policy_server.h>
```
jimis: you are including it twice

libpromises/policy_server.c
```diff
@@ -0,0 +1,182 @@
...
+ * @return  Read-only string, can be hostname or NULL (bootstrap to IP)
+ */
+const char *PolicyServerGetHost()
+{
+    return POLICY_SERVER_HOST; // Don't use CheckedReturn, var should be NULL!
+}
+
+/**
+ * @brief   Gets the port part of the policy server.
+ * @return  Read-only null terminated string of port number.
```
jimis: Can  you mention in the doc comment that NULL is an acceptable value, if the port is not explicitly set? Same with the rest of the functions that this stands.

libpromises/policy_server.h
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * @brief Sets both internal C variables as well as policy sys variables.
+ *
+ * Called at bootstrap and after reading policy_server.dat.
+ * Changes sys.policy_hub, sys.policy_hub_ip and sys.policy_hub_port.
+ *
+ * @param ctx EvalContext is used to set related variables
+ * @param new_policy_server can be 'host:port', same as policy_server.dat
+ */
+void PolicyServerSet(const char *new_policy_server);
```
jimis: Don't duplicate the doc comments, having them in the implementation file is enough.

jimis:
Well, jenkins is red unfortunately: https://ci.cfengine.com/job/testing-enterprise-pr/75/

olehermanse:
You linked to the wrong build. As we discussed, it seems builds and unit tests are passing but acceptance tests are failing. Will investigate.

olehermanse:
Started new build in Jenkins: 
https://ci.cfengine.com/job/bootstrap-enterprise-pr/92/
https://ci.cfengine.com/job/testing-enterprise-pr/80/

olehermanse:
@jimis Green across the board! Please merge. ( See last comment )

jimis:
wow this merge brought in several newline changes with CRLF; please disable all CRLF endings in your editor, @olehermanse ! This might be related to the XML illegal characters we've been seeing.

jimis:
After much searching, I am not sure I found any CRLFs introduced in non-policy files. And I /think/ (as it's not easy to grep that) all the special characters you introduced are in already DOS-mode files, so nothing wrong with that.

I still get the XML errors though, now I got it on AIX...

olehermanse:
Atom uses whatever line ending is already used in the file.

About the errors, I'm not sure what's causing it.

PR #2728, opened 2016-10-28
ENT-2376 Prevent cf-serverd from being started without promises.cf on systemd hosts

nickanderson:
When the package is initially installed, the cf-serverd service unit is
enabled and started. This is before bootstrap happens, and thus before
there is a promises.cf in $(sys.inputdir). cf-serverd ends up running
from failsafe.cf and will never reload its config from promises.cf after
it appears in inputs. This change prevents cf-serverd from starting
until the host has been bootstrapped.


kacf:
Cherry-picked to 3.10.x as well.

PR #2733, opened 2016-11-11
ENT-2504: Test getvalues against a string

nickanderson:
Note, these tests currently FAIL. They are for [ENT-2504](https://tracker.mender.io/browse/ENT-2504)


jimis:
Thanks! Cherry-picked in #2762, so closing this one.

PR #2734, opened 2016-11-14
CFE-2506: Fix crash on Solaris when ps ucb variant is not available.

kacf:
The correct approach when it is not available is to make the list of
ps information empty, not to delete it. This is already what happens
if the command can't be executed, but if executed with the wrong
arguments (ucb arguments for a non-ucb ps variant), the list would be
deleted.

Changelog: Title


kacf:
Testing in Jenkins.

Enemico:
Let me know when i can merge, or merge when you are ready, so i can go back to Solaris

kacf:
Cherry-picked to 3.10.x as well. This code is new in 3.9, so not cherry-picking to 3.7.x.

PR #2738, opened 2016-11-18
FreeBSD Build Tweaks

skreuzer:
On FreeBSD, python is installed via the ports system which by default will place binaries in /usr/local/bin. Switching the shebang line to `/usr/bin/env python` in misc/determine-version.py will make this script a bit more portable across different platforms

Add 'amd64' to the list of accepted x86 platforms when checking the value of $host_cpu. On FreeBSD this was preventing XEN_CPUID_SUPPORT from being defined which compiled out the functions to detect if the machine was running as a domU under Xen. This addresses the issues documented in CFE-2203


nickanderson:
Can you please rebase this?

PR #2739, opened 2016-11-19
Add host info script

mikeweilgart:
This script allows for easy lookup of information about a given host
from the command line on the policy server.

It allows lookup by hostname, IP address or SHA.


nickanderson:
Thanks!

PR #2740, opened 2016-11-21
Estenberg buffer checks, old pull request 2609

jimis:


jimis:
Weird, Travis is failing at `./25_cf-execd/mail_no_filter.cf` test. I did run it locally though and it passes.

Merging though, since it passed in Jenkins.

PR #2741, opened 2016-11-21
CFE-1563: Fix logic to detect when running under a Xen Hypervisor

skreuzer:
On Linux the Xen_Hv_Check() function can never be called because the stat on
/proc/self/status will always return true which results in the OpenVZ_Detect()
function always being called. Move the call to Xen_Hv_Check() to happen if the
stat on /proc/xen/capabilities returns false since this is an indication that
the host is acting as dom0.

Change the Xen_Hv_Check() function to scan the CPUID range from
0x40000000 to 0x40010000 looking for Xen leaves.

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -1129,7 +1129,14 @@ static void OSClasses(EvalContext *ctx)
     {
         Xen_Domain(ctx);
     }
-
+#ifdef XEN_CPUID_SUPPORT
+    else if (Xen_Hv_Check())
+    {
+        Log(LOG_LEVEL_VERBOSE, "This appears to be a xen hv system.");
+        EvalContextClassPutHard(ctx, "xen", "inventory,attribute_name=Virtual host,source=agent");
+        EvalContextClassPutHard(ctx, "xen_domu_hv", "source=agent");
+    }
+#endif /* XEN_CPUID_SUPPORT */
     if (stat("/etc/Eos-release", &statbuf) != -1)
     {
         EOS_Version(ctx);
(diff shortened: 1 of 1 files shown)


libenv/sysinfo.c
```diff
@@ -1129,7 +1129,14 @@ static void OSClasses(EvalContext *ctx)
     {
         Xen_Domain(ctx);
     }
-
+#ifdef XEN_CPUID_SUPPORT
+    else if (Xen_Hv_Check())
```
jimis: Why `else` ? Do you want to skip this check in case `/proc/xen/capabilities` exist?

According to [this](https://www.centos.org/docs/5/html/5.2/Virtualization/sect-Virtualization-Tips_and_tricks-Identifying_guest_type_and_implementation.html) the file might exist in both DomU and Dom0.

skreuzer: If /proc/xen/capabilities exists calling the function Xen_Domain() will set the class 'xen' and also determine if its dom0 or domU so there is no need to call Xen_Hv_Check because we already set the classes that need to be set.

If that file doesn't exist (xenfs is not loaded or the guest being virtualized doesn't support it for example) you can call cpuid and check if the vendor id is 'XenVMMXenVMM' which will work on any x86 host.

However, right now this function gets compiled out unless its being built on Linux but it addresses the issue documented in CFE-1563. I plan on refactoring this code a little bit so that at the very least it can also get called if the platform is FreeBSD to address CFE-2203.

(I updated the sysutils/cfengine39 port on FreeBSD to fix this with a patch when the port but since I only need to worry about FreeBSD when that port is being built its a bit shoe-horned in. If you want to take a look at that change the code review is at https://reviews.freebsd.org/D8607)

skreuzer:
actually, disregard this pull request for now. I need to take a closer look at the Xen_Domain function because it sets the class xen_pv which might not be correct. It might be better to merge the Xen_Hv_Check() and Xen_Domain() function into a single function instead of doing an if/else

PR #2743, opened 2016-11-23
CFE-2513: sys.sysctl inventory of all sysctl variables from /proc/sys

tzz:
As proposed in https://tracker.mender.io/browse/CFE-2513

The file glob code needed to be moved from evalfunction.c to file_lib.c for reuse.

Uses 64K memory on my machine, typical Ubuntu with 1500 sysctl vars. Runs instantly for me.

Sample `cf-promises --show-vars|grep sysctl`

```
default:sys.sysctl
       {"sys.abi.vsyscall32":"1",..,"sys.vm.zone_reclaim_mode":"0"} source=agent,sysctl,/proc/sys
```

We may want to add some way to turn this off, but it's pretty harmless, doing nothing if `/proc/sys` is empty (but still creates an empty map).


tzz:
@jimis @kacf rebased and rewritten with functions, as requested in ticket.

tzz:
I added an acceptance test, so if this is acceptable, only docs will be needed.

I moved the file glob and proc rootdir functions out to the file libs.

nickanderson:
+1 from me.

libpromises/evalfunction.c
```diff
@@ -789,6 +785,122 @@ static FnCallResult FnCallIP2Host(ARG_UNUSED EvalContext *ctx,
...
+        BufferAppendChar(filenamebuf, '/');
+        BufferSearchAndReplace(key, "\\.", "/", "gT");
+        BufferAppendString(filenamebuf, BufferData(key));
+        BufferDestroy(key);
+
+        if (IsDir(BufferData(filenamebuf)))
+        {
+            BufferDestroy(procrootbuf);
+            BufferDestroy(filenamebuf);
+            return FnFailure();
```
kacf: I would add an info level log message here, like the message below.

libpromises/evalfunction.c
```diff
@@ -789,6 +785,122 @@ static FnCallResult FnCallIP2Host(ARG_UNUSED EvalContext *ctx,
...
+    // Assumes that FILE_SEPARATOR is /
+    BufferAppendString(procrootbuf, GetRelocatedProcdirRoot());
+    BufferAppendString(procrootbuf, "/proc/sys");
+
+    if (sysctlvalue_mode)
+    {
+        Buffer *key = BufferNewFrom(RlistScalarValue(finalargs),
+                                    strlen(RlistScalarValue(finalargs)));
+
+        Buffer *filenamebuf = BufferCopy(procrootbuf);
```
kacf: You are not using `procrootbuf` again inside this if statement, and all code paths lead to a return. I suggest making `filenamebuf` an alias to it, instead of a copy, or to simply use `procrootbuf` throughout. This avoids the whole copy.

libpromises/evalfunction.c
```diff
@@ -789,6 +785,122 @@ static FnCallResult FnCallIP2Host(ARG_UNUSED EvalContext *ctx,
...
+
+        char *result = StringWriterClose(w);
+        StripTrailingNewline(result, max_sysctl_data);
+        return FnReturnNoCopy(result);
+    }
+
+    JsonElement *sysctl_data = JsonObjectCreate(10);
+
+    Buffer *procrootbuf_slash = BufferCopy(procrootbuf);
+    BufferDestroy(procrootbuf);
```
kacf: Why make a copy when you're immediately destroying the source? Just use the original instead.

libpromises/evalfunction.c

libpromises/evalfunction.c
tzz: Aha, right. Thanks!

libpromises/evalfunction.c
tzz: OK; done. Also renamed `pbuf` to `filematchbuf` for clarity.

tzz:
@kacf ready for another review.

kacf:
Thanks!

But did you see [the error in Travis](https://travis-ci.org/cfengine/core/jobs/219469980)? It should be publicly accessible. I suspect you may need to add `set.c` to the source files of the `file_lib_test`.

tzz:
@kacf ugh. I had to add a bunch of new C imports. Fixed, thanks for the hint!

tzz:
@kacf the `url_get.cf` test failed, not related to this. The hostname it uses is no longer active.

kacf:
Yep, I see it. Running this once through Jenkins as well, for complete coverage.

kacf:
I don't expect any problems with this though. Would you mind providing some docs for this as well? :-)

tzz:
As soon as it's merged, I'll do a docs PR too, as usual. They'll be pretty simple.

kacf:
Alright, thanks a lot! Merged manually.

kacf:
Do you know if you'll have time to do those docs soon? The beta is coming up pretty soon.

tzz:
@kacf see https://github.com/cfengine/documentation/pull/1607

kacf:
Thanks again, Ted, I'll be closing the task then!

PR #2745, opened 2016-11-28
CFE-2479: Behaviour change: getvalues(inexistent_var) returns an empty list.

jimis:
Restores 3.7.x and earlier behaviour.

Changelog: commit


jimis:
Tested [here](https://ci.cfengine.com/job/testing-enterprise-pr/86/).

PR #2748, opened 2016-11-29
Memory corruption cfe2520

jimis:


jimis:
@kacf please review commit 9965a8b

PR #2749, opened 2016-11-30
ENT-3023: fix cf-serverd crash when reporting corrupted data.

jimis:


jimis:
The only test failure on Travis was `./01_vars/02_functions/network/url_get.cf` which is unrelated and seems spurious. Merging.

PR #2753, opened 2016-11-30
CFE-2524: Only call BundleResolvePromiseType when it is needed as it is expensive

amousset:
I am preparing a issue on Jira about this (and other performance issues), but this seems to fix the performance worsening between 3.9 and 3.10 in ncf code evaluation.


jimis:
Thanks @amousset, this makes sense with a first look. I'll put it through our Enterprise tests on Jenkins.

PR #2762, opened 2016-12-05
CFE-2504: restore old getvalues(v) behaviour;

jimis:
it should always return a list, even when v is a simple string.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -2523,13 +2523,7 @@ static FnCallResult FnCallGetValues(EvalContext *ctx, ARG_UNUSED const Policy *p
         return (FnCallResult) { FNCALL_SUCCESS, { NULL, RVAL_TYPE_LIST } };
     }
 
-    Rlist *values = NULL;
-    if (JsonGetElementType(json) != JSON_ELEMENT_TYPE_CONTAINER)
-    {
-        JsonDestroyMaybe(json, allocated);
-        return (FnCallResult) { FNCALL_SUCCESS, { values, RVAL_TYPE_LIST } };
-    }
-
+    Rlist *values = NULL;                      /* start with an empty Rlist */
     CollectContainerValues(ctx, &values, json);
 
     JsonDestroyMaybe(json, allocated);
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -2930,6 +2930,7 @@ void GetSysVars(EvalContext *ctx)
     {
         EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, "user_data", info, CF_DATA_TYPE_CONTAINER,
                                       "source=agent,user_info");
+        JsonDestroy(info);
     }
 }
 
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -2928,7 +2928,8 @@ void GetSysVars(EvalContext *ctx)
     JsonElement *info = GetUserInfo(NULL);
     if (info != NULL)
     {
-        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, "user_data", info, CF_DATA_TYPE_CONTAINER,
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, "user_data",
+                                      info, CF_DATA_TYPE_CONTAINER,
                                       "source=agent,user_info");
         JsonDestroy(info);
     }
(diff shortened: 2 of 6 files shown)


libpromises/evalfunction.c
```diff
@@ -2523,13 +2523,7 @@ static FnCallResult FnCallGetValues(EvalContext *ctx, ARG_UNUSED const Policy *p
         return (FnCallResult) { FNCALL_SUCCESS, { NULL, RVAL_TYPE_LIST } };
     }
 
-    Rlist *values = NULL;
-    if (JsonGetElementType(json) != JSON_ELEMENT_TYPE_CONTAINER)
```
tzz: Works for me, as long as the tests pass.

jimis:
Wow, that was quick. Thanks @tzz, I'll merge when Jenkins turns out green!

jimis:
Nah, Travis caught 2 failing tests already:
```
./01_vars/02_functions/211.cf
./01_vars/03_lists/001.cf
```

Investigating.

jimis:
Discussion in [CFE-2504](https://tracker.mender.io/browse/CFE-2504).

libpromises/evalfunction.c
```diff
@@ -279,38 +279,49 @@ static JsonElement* VarRefValueToJson(const EvalContext *ctx, const FnCall *fp,
...
+                                /* TODO comment WHY? */
                                 local = JsonObjectCreate(1);
                                 JsonObjectAppendObject(holder, var->ref->indices[index], local);
                             }
+                            else
+                            {
+                                JsonObjectAppendElement(holder, var->ref->indices[index],
+                                                        JsonCopy(local));
+                            }
+                            last_key = var->ref->indices[index+1];        // WUT WUT this changes
```
jimis: Hmm I'll remove this comment ... @tzz you have any idea what this loop does? `last_key` will only be the last one, so in an array like `A[i1][i2][i3]` all intermediate indices will be lost.

libpromises/evalfunction.c
```diff
@@ -279,38 +279,49 @@ static JsonElement* VarRefValueToJson(const EvalContext *ctx, const FnCall *fp,
...
                     {
+                        /* More than one indices. */
                         Log(LOG_LEVEL_DEBUG, "%s: got ref with starting depth %zu and index count %zu",
                             fp_name, ref_num_indices, var->ref->num_indices);
                         for (int index = ref_num_indices; index < var->ref->num_indices-1; index++)
                         {
                             JsonElement *local = JsonObjectGet(holder, var->ref->indices[index]);
                             if (local == NULL)
                             {
+                                /* TODO comment WHY? */
```
jimis: @tzz if you know, please tell me what to fill in here. :-)

libpromises/evalfunction.c
```diff
@@ -279,38 +279,49 @@ static JsonElement* VarRefValueToJson(const EvalContext *ctx, const FnCall *fp,
...
+                            else
+                            {
+                                JsonObjectAppendElement(holder, var->ref->indices[index],
+                                                        JsonCopy(local));
+                            }
+                            last_key = var->ref->indices[index+1];        // WUT WUT this changes
 
-                            last_key = var->ref->indices[index+1];
-                            holder = local;
+                            /* TODO fix leak if this loop runs more than once! */
```
jimis: This is the crash-avoiding change. Previously, it did `holder=local` that changed `holder` from a `JsonObject` to something possibly different, in our case `JsonArray`. This caused a crash when doing `JsonObjectAppend()` later (because for arrays you need `JsonArrayAppend()`). @tzz does this fix misses some functionality?

jimis:
@tzz another breaking change caused by `VarRefValueToJson()` refactoring in commit 652d442294767e459552babce175aeb40cc07bf0, is that `getvalues()` returned (and is documented to do so) **the contents of the 2st sub-index**. Now it returns contents of all sub-indices of the array, no matter the depth. I'm not sure how to fix so ideas are appreciated.

libpromises/evalfunction.c
tzz: If the holder has nothing under that key, we create a new object to go under that key.

libpromises/evalfunction.c
tzz: This was originally written by Sigurd, I think. It converts classic arrays to JsonElement so it's nasty. I don't recall all the details changed over the years.

libpromises/evalfunction.c
tzz: I think this is good. Nice catch.

tzz:
I would keep the existing behavior of `getvalues()` if possible, even if the code becomes messy and with lots of special cases, since your team values backwards compatibility very highly. I can't see a way to do it otherwise.

jimis:
> I would keep the existing behavior of getvalues() if possible

By existing you mean the existing after 3.9.x, or the documented one and existing in 3.7.x, i.e. printing values of 1-level deep sub-indices?

tzz:
I meant 3.7 and documented.

But practically speaking, users didn't complain about the 3.9 behavior, so either they don't use multi-level arrays (remember multi-level arrays are discouraged), or they are not using 3.9. Either way, maybe keeping the 3.9 behavior and documenting the change in the manual is the right way after all.

Sorry I'm not helpful. I don't think this question has a clear answer.

jimis:
Thanks for the feedback @tzz.

PR #2763, opened 2016-12-07
My 3.9.x cherry-picks

kacf:


kacf:
@jimis: These are my 3.9.x cherry-picks. I've gone through the rest, and none of the other commits by me should be included.

PR #2766, opened 2016-12-08
CFE-2504: Add test for getvalues on foreign array subkey

nickanderson:


jimis:
close?

nickanderson:
seems like a duplicate of #2766?

PR #2767, opened 2016-12-08
Change: Warning about depth_search on a non directory to DEBUG

nickanderson:
When promising to copy remote paths recursively if the paht is a non
directory a warning is emitted but the file is copied anyway so it does
what the use meant. The warning in this case end up being useless noise.
Perhaps it is useful for debugging, but for policy writers it only poses
a challenge to be worked around. This removes the need to implement
additional policy in order to differentate between files and directories
only for the suppression of the warning.

Ref: https://groups.google.com/forum/#!topic/help-cfengine/ekSgRU3j1W4

Changelog: Title

### DIFF
--- a/cf-agent/verify_files.c
+++ b/cf-agent/verify_files.c
@@ -390,7 +390,7 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
         {
             if (a.havedepthsearch)
             {
-                Log(LOG_LEVEL_WARNING,
+                Log(LOG_LEVEL_DEBUG,
                     "depth_search (recursion) is promised for a base object '%s' that is not a directory",
                       path);
                 goto exit;


cf-agent/verify_files.c
```diff
@@ -390,7 +390,7 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
         {
             if (a.havedepthsearch)
             {
-                Log(LOG_LEVEL_WARNING,
+                Log(LOG_LEVEL_DEBUG,
                     "depth_search (recursion) is promised for a base object '%s' that is not a directory",
                       path);
                 goto exit;
```
jimis: Are you sure that the file is copied anyway? This `goto exit` is telling otherwise, but I can't be sure, the code is complex.

jimis:
I would merge, but travis is red; I don't see a reason for that though. Can you rebase?

cf-agent/verify_files.c
nickanderson: I don't recall for certain, but I believe it is copied, that would be why I went to change a somewhat annoying message in the C.

nickanderson:
@jimis travis is no longer red.

PR #2768, opened 2016-12-09
CFE-2535: Add test for ensuring iterating over a list of dictionaries works as expected

nickanderson:


jimis:
I'm closing since Travis is failing, apparently because the bug is still unfixed. Either submit it as a soft fail, or just attach it in the ticket, to be submitted when fixed.

nickanderson:
AH in order for the soft fail stuff to work you must use the fancy bundlesequence runner .... I learned that from my own testing presentation :-)

https://github.com/nickanderson/presentation-testing-cfengine-policy/blob/11dc6fdb88d3e697c940c1469de38bbaf090e5f0/deck.org#bundle-meta-info

nickanderson:
Travis build that

PR #2775, opened 2016-12-29
Detect Amazon linux

skreuzer:
This adds support for detecting Amazon Linux which is a fork of RHEL optimized for running in EC2

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -152,6 +152,7 @@ void CalculateDomainName(const char *nodename, const char *dnsname,
 #ifdef __linux__
 static int Linux_Fedora_Version(EvalContext *ctx);
 static int Linux_Redhat_Version(EvalContext *ctx);
+static void Linux_Amazon_Version(EvalContext *ctx);
 static void Linux_Oracle_VM_Server_Version(EvalContext *ctx);
 static void Linux_Oracle_Version(EvalContext *ctx);
 static int Linux_Suse_Version(EvalContext *ctx);
@@ -1063,6 +1064,11 @@ static void OSClasses(EvalContext *ctx)
     if (stat("/etc/SuSE-release", &statbuf) != -1)
     {
         Linux_Suse_Version(ctx);
+    }
+
+    if (stat("/etc/system-release", &statbuf) != -1)
+    {
+        Linux_Amazon_Version(ctx);
     }
 
 # define SLACKWARE_ANCIENT_VERSION_FILENAME "/etc/slackware-release"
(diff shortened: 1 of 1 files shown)


libenv/sysinfo.c
```diff
@@ -2311,6 +2317,31 @@ static int Linux_Mandriva_Version_Real(EvalContext *ctx, char *filename, char *r
 
 /******************************************************************/
 
+static int Linux_Amazon_Version(EvalContext *ctx)
+{
+    char buffer[CF_BUFSIZE];
+
+    // Amazon Linux AMI release 2016.09
+
+    if (ReadLine("/etc/system-release", buffer, sizeof(buffer)))
```
nickanderson: No `/etc/os-release` or `/usr/lib/os-release`?

skreuzer: `/etc/os-release` does exist but right now it just parsing a single line from `/etc/system-release` is the easiest way to determine if the host os is Amazon Linux.

At some point it seems like a lot of the logic in sysinfo.c can be refactored to take advantage of `/etc/os-release` but even this is not that simple since different distributions take liberties with these parameters. (For instance, on a debian 8 machine, neither VERSION_ID or VERSION tell you what point release it is so you would still need to parse /etc/debian_version to get this info)

libenv/sysinfo.c
```diff
@@ -2311,6 +2317,31 @@ static int Linux_Mandriva_Version_Real(EvalContext *ctx, char *filename, char *r
...
+    {
+       if (strstr(buffer, "Amazon"))
+       {
+           char version[CF_MAXVARSIZE], class[CF_MAXVARSIZE];
+           EvalContextClassPutHard(ctx, "amazon", "inventory,attribute_name=none,source=agent");
```
skreuzer: This would be the first time derived-from= references a file (seems like its mostly for $(sys) variables). I don't have an objection to adding this but it seems like it should be added for everything. If that is the case I would be happy to open up a new pull request

tzz: @skreuzer I agree with @nickanderson but let's use `derived-from-file=...` since you're right that `derived-from` is for variables. The tag usage has to start somewhere, so rather than a big pull request, I think this is a good place.

tzz:
This looks good to me @jimis @skreuzer with some notes:

* I agree with @nickanderson that introducing a `derived-from-file` tag makes sense here
* the class name could be `amazon_linux` since `amazon` is a company name that could be ambiguous in the CFEngine context (Amazon EC2, S3, ELB, etc. could also be supported by CFEngine some day)
* as a future direction I'd really like to have BIOS reading, similar to what `dmidecode` produces, so CFEngine can tell that it's running on an AWS EC2 or on a VMWare instance. We could even run `dmidecode` on startup. Doing that BIOS check in the C code has many benefits: speed, reliability, and tighter coupling with the hardware.

libenv/sysinfo.c
```diff
@@ -2311,6 +2317,31 @@ static int Linux_Mandriva_Version_Real(EvalContext *ctx, char *filename, char *r
...
 
+static int Linux_Amazon_Version(EvalContext *ctx)
+{
+    char buffer[CF_BUFSIZE];
+
+    // Amazon Linux AMI release 2016.09
+
+    if (ReadLine("/etc/system-release", buffer, sizeof(buffer)))
+    {
+       if (strstr(buffer, "Amazon"))
```
jimis: use explicit comparison, since this is not a boolean

libenv/sysinfo.c
```diff
@@ -2311,6 +2317,31 @@ static int Linux_Mandriva_Version_Real(EvalContext *ctx, char *filename, char *r
...
+       {
+           char version[CF_MAXVARSIZE], class[CF_MAXVARSIZE];
+           EvalContextClassPutHard(ctx, "amazon", "inventory,attribute_name=none,source=agent");
+           sscanf(buffer, "%*s %*s %*s %*s %s", version);
+           CanonifyNameInPlace(version);
+           snprintf(class, CF_MAXVARSIZE, "amazon_%s", version);
```
jimis: what about `amazonlinux_...`? (since you are already setting the generic `amazonlinux` class)

skreuzer:
> Finally do you notice the generic rhel class is being set on amazon linux? I'm curious since it's set on CentOS.

It doesn't look like that class gets set. A list of the discovered classes on an Amazon Linux AMI can be found at https://gist.github.com/skreuzer/64eefc815b5da8a137a99e9f0d2a35da

jimis:
Thanks!

tzz:
@jimis @nickanderson as a followup item, please make a ticket to apply `derived-from-file` consistently across the C code. I may be able to get to it, but regardless a ticket will keep that task around so we don't forget it. Thanks!

jimis:
Done! 

https://tracker.mender.io/browse/CFE-2571

PR #2776, opened 2016-12-29
Remove reference to redmine

skreuzer:
Remove the reference to redmine since the bug tracker has been moved over to jira


jimis:
Thank you @skreuzer

PR #2777, opened 2016-12-30
CFE-2551 Add new acceptance test for mergedata

nickanderson:


nickanderson:
@jimis these tests are identical except for in how they reference the data to be merged. Note that variants 1 and 2 do not include test_soft_fail because they are already passing. Variants 3 and 4 fail differently, and it's probably caused by the `:` after the namespace.

tzz:
@nickanderson yeah the namespace is not part of the parsing logic:

```

**Comment by jimis** [comment]
Indeed too many colons:

```
     "merged_data_vars" data => mergedata('{ "vars1": default:test_merged_low.vars1 }');
```

nickanderson:
@jimis what shall I do with the two namespace tests? mark them as soft fail?

jimis:
I think we shouldn't even include them there, and document that namespaces are unsupported because JSON syntax mixup. "soft-fail" usually means that it's a known bug to be fixed. I don't see how this one can be fixed.

Or maybe add them as `.x.cf` test, which means that it is expected to fail parsing.

nickanderson:
@jimis I moved tests 3 and 4 to x.cf

jimis:
"fail to exit with error" in Travis.

nickanderson:
Removed test 3 and 4

nickanderson:
@jimis, its now passing

PR #2783, opened 2017-01-23
Update "getvalues()" entry

michaelclelland:


nickanderson:
The clarification should probably go into the master branch as well.

PR #2784, opened 2017-01-23
Update "getvalues()" entry

michaelclelland:


nickanderson:
Thank you for the suggestion @brontolinux

PR #2785, opened 2017-01-23
Test deep delayed expansion

nickanderson:
When setting a cfengine variable from another cfengine variable that
contains an un-dereferenced variable the agnet should dereference
deeply.


nickanderson:
@martingehrke

nickanderson:
oh @jimis i think the one failure was because i accidentially pushed into new branch in core repo first. Then i deleted the branch and thats probably why that one failed.

nickanderson:
hocus pocus pass travis

nickanderson:
picked to 3.10.x

PR #2786, opened 2017-01-23
CFE-1996: prevent LMDB assertion on AIX

jimis:


jimis:
@kacf what do you think of this?

jimis:
rebased and rebuilding

PR #2787, opened 2017-01-24
CFE-2575: cf-serverd: autoconfigure max open files ulimit

jimis:


cf-serverd/server_transform.c
```diff
@@ -243,6 +244,28 @@ void KeepPromises(EvalContext *ctx, const Policy *policy, GenericAgentConfig *co
...
+static bool SetMaxOpenFiles(int n)
+{
+    const struct rlimit lim = {
+        .rlim_cur = n,
+        .rlim_max = n,
+    };
+    int ret = setrlimit(RLIMIT_NOFILE, &lim);
+    if (ret == -1)
+    {
+        Log(LOG_LEVEL_INFO, "Failed setting max open files limit"
```
estenberg: LOG_LEVEL_ERROR? Info is defined to be for "changes to the system".

cf-serverd/server_transform.c
```diff
@@ -243,6 +244,28 @@ void KeepPromises(EvalContext *ctx, const Policy *policy, GenericAgentConfig *co
...
+            " (setrlimit(NOFILE, %d): %s)", n, GetErrorStr());
+        Log(LOG_LEVEL_INFO,
```
estenberg: LOG_LEVEL_ERROR? Info is defined to be for "changes to the system".

cf-serverd/server_transform.c
```diff
@@ -243,6 +244,28 @@ void KeepPromises(EvalContext *ctx, const Policy *policy, GenericAgentConfig *co
...
+            "You should tweak the 'nofile' ulimit manually to be at least 2x maxconnections");
```
estenberg: Perhaps "change" instead of "tweak" (sounds more professional).

cf-serverd/server_transform.c
jimis: > Info is defined to be for "changes to the system"

True for the agent, but cf-serverd doesn't change the system thus uses a different pattern for logging (e.g. refusals are all INFO level).

So here I was between VERBOSE or something a bit higher. Definitely not ERROR though, since it's not disrupting operation, cf-serverd will keep running OK (and this message will almost always be printed for non-root users). I can bring it back down to VERBOSE, but I'll check around the code and see how similar logging is treated.

cf-serverd/server_transform.c
jimis: OK

PR #2788, opened 2017-01-24
Allow function calls in promiser through new "with" attribute.

tzz:
Fixes https://tracker.mender.io/browse/CFE-1092

Add universal "with" attribute to allow functions calls in promiser on $(item).
Add acceptance test.

Changelog: CFE-1092: allow function calls in promiser using universal "with" attribute.

### DIFF
--- a/libpromises/eval_context.c
+++ b/libpromises/eval_context.c
@@ -1381,6 +1381,25 @@ void EvalContextStackPushPromiseFrame(EvalContext *ctx, const Promise *owner, bo
 
     EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_THIS, "bundle", PromiseGetBundle(owner)->name, CF_DATA_TYPE_STRING, "source=promise");
     EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_THIS, "namespace", PromiseGetNamespace(owner), CF_DATA_TYPE_STRING, "source=promise");
+
+    // Recompute `with`
+    for (size_t i = 0; i < SeqLength(owner->conlist); i++)
+    {
+        Constraint *cp = SeqAt(owner->conlist, i);
+        if (StringSafeEqual(cp->lval, "with"))
+        {
+            Rval final = EvaluateFinalRval(ctx, PromiseGetPolicy(owner), NULL,
+                                           "this", cp->rval, false, owner);
+            if (final.type == RVAL_TYPE_SCALAR && !IsCf3VarString(RvalScalarValue(final)))
+            {
+                EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_THIS, "with", RvalScalarValue(final), CF_DATA_TYPE_STRING, "source=promise_iteration/with");
+            }
+            else
+            {
+                RvalDestroy(final);
+            }
+        }
+    }
 }
 
 Promise *EvalContextStackPushPromiseIterationFrame(EvalContext *ctx, const PromiseIterator *iter_ctx)
--- a/libpromises/iteration.c
+++ b/libpromises/iteration.c
@@ -975,6 +975,27 @@ bool PromiseIteratorNext(PromiseIterator *iterctx, EvalContext *evalctx)
      *      hopefully overwriting the previous values, but possibly not! */
     }
 
+    // Recompute `with`
+    for (size_t i = 0; i < SeqLength(iterctx->pp->conlist); i++)
+    {
+        Constraint *cp = SeqAt(iterctx->pp->conlist, i);
+        if (StringSafeEqual(cp->lval, "with"))
+        {
+            Rval final = EvaluateFinalRval(evalctx, PromiseGetPolicy(iterctx->pp), NULL,
+                                           "this", cp->rval, false, iterctx->pp);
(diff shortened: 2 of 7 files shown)


tzz:
@jimis in its current form, this has several problems. I tried doing the iterator mapping in several ways but am currently stuck:

* resolving the `with` rval can't use functions like `maplist()`, because the `Promise*` must be NULL. If it's anything but NULL, it segfaults. This fails on G1 and G2.
* unresolved variables after rval expansion hose things. That's why the test fails on `F*` and `G*`.
* we create and then destroy a local bundle variable called `with` which will overwrite any existing `with` variable. Ideally `with` would live inside the `this` scope but that scope by design can't contain iterables?!?!?! So this is unpleasant.

jimis:
@tzz we have discussed this extensively over IRC, let me know if you have further trouble.

tzz:
@jimis your suggestion to create a temporary variable didn't work, unfortunately, because implicit iterations can "bind" a promise tightly. So for example:

```
vars:
  "iter" slist => { "1", "2", "3" };
  "test" string => "$(with)", with => { $(iter) };
```

If we create a temporary variable `with` **before** evaluating the second promise, it will iterate `1, 2, 3` and then the second promise will be evaluated `with=3`. So the insertion of `with` has to be at the iteration context level, by adding a new artificial wheel. That also avoids the mess of an extra variable. So I'm working on that, we'll see if it works...

tzz:
@jimis your example uses a reference, which would work for this one case. But it wouldn't work if a transformation was done, e.g.

```
vars:
  "iter" slist => { "1", "2", "3" }; # base
  "test" string => "$(with)", with => { canonify($(iter)) }; # test case
```

The function call is applied to each `$(iter)` so the `with` variable needs to be rewritten `maplist(canonify($(this)), iter)` in order to produce the expected results, or it needs to be evaluated at each iteration of the test case. The rewriting is too hard, I think.

jimis:
> The rewriting is too hard, I think

Indeed. To support the example you just posted, you have to first support `vars` promises like this:

```
"v" slist => { canonify($(iter)) };
```

this would make sense by itself I think.

tzz:
@jimis I modified this as we discussed. The value of `with` has to be computed twice, once outside the promise iteration frame and once inside. All the tests pass. Could you take a look?

tests/acceptance/testall
```diff
@@ -815,8 +815,9 @@ done
 
 #
 # Close stdin file descriptor to avoid tty_interactive check in cf-agent being success.
+# Do this iff GDB AND Valgrind are not in use.
 #
-if [ "$GDB" != 1 ] || [ "$USE_VALGRIND" != 1 ]
+if [ "$GDB" != 1 ] && [ "$USE_VALGRIND" != 1 ]
```
tzz: This fixes a bug in `testall` and GDB.

nickanderson: Lets see if travis agrees

nickanderson:
@jimis travis likes it

jimis:
Thanks. I'm postponing for after 3.10.1, 3.7.5.

libpromises/eval_context.c
```diff
@@ -1380,6 +1380,21 @@ void EvalContextStackPushPromiseFrame(EvalContext *ctx, const Promise *owner, bo
...
     EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_THIS, "namespace", PromiseGetNamespace(owner), CF_DATA_TYPE_STRING, "source=promise");
+
+    // Recompute `with`
+    for (size_t i = 0; i < SeqLength(owner->conlist); i++)
+    {
+        Constraint *cp = SeqAt(owner->conlist, i);
+        if (StringSafeEqual(cp->lval, "with"))
+        {
+            Rval final = EvaluateFinalRval(ctx, PromiseGetPolicy(owner), NULL,
+                                           "this", cp->rval, false, owner);
```
tzz: thanks for catching that, fixed

libpromises/iteration.c
```diff
@@ -975,6 +975,23 @@ bool PromiseIteratorNext(PromiseIterator *iterctx, EvalContext *evalctx)
...
     }
 
+    // Recompute `with`
+    for (size_t i = 0; i < SeqLength(iterctx->pp->conlist); i++)
+    {
+        Constraint *cp = SeqAt(iterctx->pp->conlist, i);
+        if (StringSafeEqual(cp->lval, "with"))
+        {
+            Rval final = EvaluateFinalRval(evalctx, PromiseGetPolicy(iterctx->pp), NULL,
+                                           "this", cp->rval, false, iterctx->pp);
```
tzz: thanks for catching that, fixed

tzz:
Added the `RvalDestroy()`s, the `testall` fix is in a separate commit, and all tests pass. Thank you.

kacf:
Running in Jenkins as well.

kacf:
This is ready to merge as soon as docs are ready.

tzz:
@kacf @jimis added example which will be used by the docs.

tzz:
Docs in https://github.com/cfengine/documentation/pull/1608 need this PR merged first for the example `with.cf`.

kacf:
Thanks for another great feature! This adds some real convenience.

PR #2794, opened 2017-02-08
unix_iface: try to look for ifconfig in /bin as well

rindeal:
ifconfig utility is installed into the BINDIR by default since commit https://sourceforge.net/p/net-tools/code/ci/36b541c9f3efe55c5871674ee926be8b20339497/, made on 2011-12-02.

### DIFF
--- a/libenv/unix_iface.c
+++ b/libenv/unix_iface.c
@@ -612,7 +612,8 @@ static void FindV6InterfacesInfo(EvalContext *ctx)
         return;
     }
 #else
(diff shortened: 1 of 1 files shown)


jimis:
Thanks, merging. @rindeal for which platform is this one? On my Ubuntu and CentOS machines it can still be found in `/sbin`.

rindeal:
@jimis Gentoo. CentOS follows Fedora, where `{,/usr}/{,s}bin` is the same dir, Ubuntu either jumped on the bandwagon or must have patched net-tools.

PR #2564, opened 2016-04-06
Move def.json parsing to happen after class discovery

tzz:
@kacf @nickanderson fixes bug in class discovery and adds to acceptance test

PR #2796, opened 2017-02-08
Fixes #2564: Splayclass is not properly spread

peckpeck:
See https://tracker.mender.io/browse/CFE-2564

@jimis  I couldn't do the double version because it is hard to be right with the rounding that happen when we are near to 1, and we don't want to hot max.

### DIFF
--- a/libutils/string_lib.c
+++ b/libutils/string_lib.c
@@ -77,7 +77,7 @@ unsigned int StringHash(const char *str, unsigned int seed, unsigned int max)
     h ^= (h >> 11);
     h += (h << 15);
 
-    return (h & (max - 1));
+    return (h % max);
 }
 
 unsigned int StringHash_untyped(const void *str, unsigned int seed, unsigned int max)


libutils/string_lib.c
```diff
@@ -77,7 +77,7 @@ unsigned int StringHash(const char *str, unsigned int seed, unsigned int max)
     h ^= (h >> 11);
     h += (h << 15);
 
-    return (h & (max - 1));
+    return (h % max);
```
jimis: 20 lines above there is this:

```
     CF_ASSERT(ISPOW2(max),
"StringHash() called with non power-of-2 max: %u", max);
```

Have you ever seen this assert triggering?

peckpeck: Never

jimis: Then this means that it is always executed with power-of-two max, so `&` should be working perfectly.

peckpeck: no i means there is a bug in the assert, here is an example call 

libpromises/evalfunction.c:        int slot = StringHash(RlistScalarValue(finalargs), 0, CF_HASHTABLESIZE) * 12 / CF_HASHTABLESIZE;

jimis: But `CF_HASHTABLESIZE` is 8192, which is a valid power of two, and `&` should work fine and the results of `StringHash()` should be distributed perfectly.

I had a quick look, I couldn't find offending code. Can you remind me how you faced the issue? Was it some specific policy?

peckpeck: Os sorry I misread the parens place.
You're right, and this means that i made the same mistake in my example code.
It seems that the algorithm is fine.

We had the problem with a setup of a few thousand of agents. We found different approaches, and the has was one of them.

jimis: Alright, glad this is resolved now, these are subtle issues! Feel free to close this issue if there is nothing to fix, or maybe submit something different.

FWIW I prefer to have ASSERT(power-of-two) plus binary AND operation, rather than one-size-fits-all modulo operation. The reason is that modulo skews results and creates uneven distribution in subtle cases. But ASSERT works as documentation, enforcing proper usage.

jimis:
@peckpeck this is resolved for you now?

peckpeck:
Yes resolved

PR #2797, opened 2017-02-08
CFE-2516: Allow opening symlinks owned by root or by the current user

amousset:
Symlinks are currently checked and only opened if the user and group
of the symlink are the same as the destination.
Some other cases are safe:
 * symlinks owned by root
 * symlinks owned by the user running the process

### DIFF
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -330,7 +330,8 @@ void switch_symlink_hook();
 
 /**
  * Opens a file safely. It will follow symlinks, but only if the symlink is trusted,
- * that is, if the owner of the symlink and the owner of the target are the same.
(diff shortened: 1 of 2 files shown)


jimis:
@amousset can you look into the failing tests?

jimis:
@amousset one unit test is failing, probably needs some adjustment.

kacf:
Thanks @amousset! Keep in mind that Github doesn't tell us when you have updated the PR, you need to leave a comment.

I have made a small amendment to the PR and resubmitted it as #2824.

amousset:
Thanks @kacf! (I did not added a comment yet because I still wanted to add some explanation about the changes in the test). Closing this PR.

PR #2799, opened 2017-02-21
CFE-2590: Add test for maparray expanding non this vars

nickanderson:


nickanderson:
Merging soft failed test.

PR #2802, opened 2017-02-24
Readfile error cfe2512

jimis:


jimis:
Thanks for the acceptance test @nickanderson !

PR #2803, opened 2017-02-27
Allow string_head and string_tail to take negative arguments

tzz:
1) Reworks the text_xfrom test to be more compact and in expected JSON format.

2) Currently the only way to extract "all but the last/first N" from a string is with a sequence of `string_reverse` calls. This change makes it easy (mentioned by @nickanderson and others):

```
string_tail("abc ANYTHING", "-4") = "ANYTHING"
string_head("ANYTHING xyz", "-4") = "ANYTHING"
```


tzz:
I'll write docs as usual if this is accepted.

nickanderson:
@jimis,This seems nice. Useful when the rare (at least for me) case comes up that I need to do this.

tzz:
Tests fixed I believe.

kacf:
Thanks, looks good to me. Running through Jenkins.

kacf:
Thanks, Ted! Merged manually.

The doc change was tiny, so [I did it myself](https://github.com/cfengine/documentation/pull/1605).

tzz:
Great, thanks. That will be useful to avoid unnecessary regexes in some cases.

kacf:
@tzz: I was about to demo this, and I discovered that if you use an inline argument, and not a variable, it doesn't accept a negative integer. This might be related to the problem I had that the documentation would only render the range as being between 0 and 9999...

It's probably related to the function argument structure, but I haven't looked at it yet. Do you want to handle this, or should I create a bug for later?

tzz:
Your call, I think just making it a string argument may be enough. I remember this issue in other places...

tzz:
@kacf simply

kacf:
Thanks, turns out `CF_INTRANGE` was the trick. Submitted #2828.

PR #2809, opened 2017-03-03
INF-422: Mingw errors

jimis:


jimis:
@Lex-2008  this did well in your private branch test, so I'm merging upstream.

PR #2811, opened 2017-03-03
3.7.5 cherry-picks

jimis:


jimis:
@olehermanse you memory leaks cherry-picked to 3.7.x!
@nickanderson have a look, only the 3 first commits are user facing.

I decided to cherry-pick `cf-key -n` because it's so useful! I'm OK to undo it if you say so.

jimis:
I removed the `cf-key -n` feature after discussing with @nickanderson. Merging now.

PR #2817, opened 2017-03-15
cf-agent: learn --show-vars and --show-classes to show final state

tzz:
@nickanderson I got sick of writing stuff like this, which doesn't even show all the classes:

```
bundle agent main
{
  methods:
      "test";

  vars:
      "test_state" data => bundlestate(test);
      "test_string" string => storejson(test_state);

  reports:
      "$(this.bundle): state of things = $(test_string)";
}

bundle agent test
{
...
}
```

So I believe there's a practical need, especially for quick demos and for debugging.

If this is merged, I'll adjust #2480 as well (it needs rebasing anyway) and write docs as usual.

The options and code changes are minimal. We're just moving the support code for `--show-vars` and `--show-classes` (plus one enum) to `generic_agent.[ch]` and write the exact same glue as cf-promises in `cf-agent/cf-agent.c` to show vars and classes before destroying the policy. The only difference is that cf-promises doesn't run policy before showing vars and classes.

This code has no side effects, performance implications, or any compatibility problems. So I hope it is acceptable.

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -189,6 +189,8 @@ static const struct option OPTIONS[] =
     {"timestamp", no_argument, 0, 'l'},
     /* Only long option for the rest */
     {"log-modules", required_argument, 0, 0},
+    {"show-evaluated-classes", optional_argument, 0, 0 },
+    {"show-evaluated-vars", optional_argument, 0, 0 },
     {NULL, 0, 0, '\0'}
 };
 
@@ -214,6 +216,8 @@ static const char *const HINTS[] =
     "Disable extension loading (used while upgrading)",
     "Log timestamps on each line of log output",
     "Enable even more detailed debug logging for specific areas of the implementation. Use together with '-d'. Use --log-modules=help for a list of available modules",
+    "Show *final* evaluated classes, including those defined in common bundles in policy. Optionally can take a regular expression.",
+    "Show *final* evaluated variables, including those defined without dependency to user-defined classes in policy. Optionally can take a regular expression.",
     NULL
 };
 
@@ -265,6 +269,18 @@ int main(int argc, char *argv[])
 
     PurgeLocks();
     BackupLockDatabase();
+
+    if (config->agent_specific.common.show_classes != NULL)
+    {
+        GenericAgentShowContextsFormatted(ctx, config->agent_specific.common.show_classes);
+        free(config->agent_specific.common.show_classes);
+    }
+
+    if (config->agent_specific.common.show_variables != NULL)
+    {
+        GenericAgentShowVariablesFormatted(ctx, config->agent_specific.common.show_variables);
+        free(config->agent_specific.common.show_variables);
+    }
 
     PolicyDestroy(policy); /* Can we safely do this earlier ? */
     int ret = 0;
@@ -546,7 +562,23 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
                     exit(EXIT_FAILURE);
                 }
             }
-            break;
+            else if (strcmp(OPTIONS[longopt_idx].name, "show-evaluated-classes") == 0)
+            {
+                if (optarg == NULL)
+                {
+                    optarg = ".*";
+                }
+                config->agent_specific.common.show_classes = xstrdup(optarg);
+            }
+            else if (strcmp(OPTIONS[longopt_idx].name, "show-evaluated-vars") == 0)
+            {
+                if (optarg == NULL)
+                {
+                    optarg = ".*";
+                }
+                config->agent_specific.common.show_variables = xstrdup(optarg);
+            }
+    break;
 
         default:
             {
--- a/libpromises/generic_agent.c
+++ b/libpromises/generic_agent.c
@@ -1968,3 +1968,137 @@ void SetupSignalsForAgent(void)
     signal(SIGUSR1, HandleSignalsForAgent);
     signal(SIGUSR2, HandleSignalsForAgent);
 }
+
+void GenericAgentShowContextsFormatted(EvalContext *ctx, const char *regexp)
+{
+    assert(regexp != NULL);
+
+    ClassTableIterator *iter = EvalContextClassTableIteratorNewGlobal(ctx, NULL, true, true);
+    Class *cls = NULL;
+
+    Seq *seq = SeqNew(1000, free);
+
+    pcre *rx = CompileRegex(regexp);
+
+    if (rx == NULL)
+    {
+        Log(LOG_LEVEL_ERR, "Sorry, we could not compile regular expression %s", regexp);
+        return;
+    }
+
+    while ((cls = ClassTableIteratorNext(iter)))
+    {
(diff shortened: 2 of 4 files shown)


nickanderson:
@tzz can you show example execution and output?

tzz:
@nickanderson it's *exactly* the same execution and output as `cf-promises --show-vars` or `cf-promises --show-classes`, but after the policy is run, so you get all the policy vars and classes.

nickanderson:
yeah, was just looking to see the additional volume :)

nickanderson:
You get bundle scoped classes too?

tzz:
Bundle scoped classes are cleared when the bundle exits so they won't show, sorry.

If the functionality is acceptable, I can add the code and acceptance tests from #2480 on top of this, which will show the expected behavior. But I need to be sure it won't be rejected :)

estenberg:
The problem statement makes sense to me, however the relation between the version in cf-promises and cf-agent becomes a bit confusing, I think.

When would I use cf-promises' version and when would I use the one in cf-agent? I should not need to make this choice - it should "just work". :) Also, I think it is a problem that you would have to make changes to the system in order to dump the inventory, variables and classes.

So all in all, this might be a good short-term fix but I think we should have a plan to have 1) one place to run this that 2) doesn't change the system and 3) shows all the vars/classes. Is that too much to ask, perhaps? ;)

In any case we need to have a plan for how to get to just one place for these options, not two places with nuanced differences that need to be understood by all users.

nickanderson:
I think that its not possible to show all the vars and classes until the end of the execution. I think part of teds desire for this is related to testing. How can I more easily prove that the state I expected was the state reached after a given agent run.

There are implementation details that seep into this.

For example in order to avoid the overhead of multiple executions (even cached functions are executed multiple times during a given because of pre-eval and because cf-promises is a separate execution from cf-agent) sometimes we resort to using commands promises for modules  so that we know it's only going to be run once (assuming its going to return zero).

cf-promises only does pre-eval, and cf-agent does pre-eval + normal order. So it might be interesting to know what classes and vars are subject to execution.

I might use cf-promises to see what vars and classes I can use to make high level decisions, but might use cf-agent to see if the commands promises and services promises defined the classes I expected them to define when restarting some service during an agent run.

estenberg:
That makes sense, however that means the use case for this feature is debugging or diagnostics.
If you run cf-agent in debug mode it should print all the classes and vars -- isn't that enough for this case?

tzz:
@estenberg it's not enough because the output is all over the logs, it's inconsistent, and doesn't include meta tags in the same place (you have to write complex parsers to extract it all). This feature's output is condensed, filtered, so it can be diff-ed against the cf-promises data. That gives users consistency.

An important use case we didn't mention is verification and auditing, which I believe is separate from debugging and diagnostics. I want to make sure the **state** of the agent (filtered by some criteria) is what I expected at the end of its run. That's very useful to prove to auditors, managers, or other applications that the agent did or did not do something critical.

Another use case is acceptance testing, both internal and external, which really helps users and paying customers.

nickanderson:
The volume of information in the DEBUG output is ... Tremendous ....

This is much more easily digestible by a human or simple scripts to find information by the common ways (variable/class name, meta tag associated with variable or class). Would have to write a script or a  nasty one-liner to easily search on either of those in the debug output.

estenberg:
I agree, these are valid cases.

I just think that having duplicate functionality that works *slightly* different will lead to usability issues. It adds a burden to the user, just selecting which one to use.

Maybe there is some other way to do it; perhaps we can dump the complete tables at the end of cf-agent debug?

tzz:
@estenberg I'd rather keep the current mechanism because it's simple and consistent. I do understand the problem, you're right that it can be confusing. We can change the option name to be `--show-final-{vars,classes}` if you want? And then `--show-{vars,classes}` can do what it does in cf-promises, run before the policy?

estenberg:
Sure, that makes sense.

Then the naming game again; final makes it sound like they could change over time (maybe this is true)?
Maybe expanded, or evaluated?

tzz:
@estenberg synonyms: last, closing, concluding, finishing, end, terminating, ultimate, eventual, irrevocable, unalterable, absolute, conclusive, irrefutable, incontrovertible, indisputable, unchallengeable, binding

Let me know, I'll add the new option and make the code changes.

estenberg:
I think my favourite is

1) --show-evaluated-classes/vars. I think evaluated describes it well, drawback is that it is a bit long and cumbersome to type.

2) --show-final-classes/vars. Short and nice, but final makes it sound like it is changing, which is not that good.

The rest I don't think are that good. Anyone else has opinions?

tzz:
ping people, ping... this is a simple change, let's settle the naming and get on with it... @estenberg @jimis @kacf @olehermanse please vote or whatever :)

jimis:
I agree with not having the same option name as in `cf-promises`. I don't have a special preference for the name to be used, @estenberg 's naming choices make sense.

jimis:
Oh, I forgot, the commit message needs a "changelog" tag since this is a user-visible change.

tzz:
OK @jimis your changes are in the PR and it has been rebased. Only docs remain I think.

Note there's a bug fix here, the `optarg` was not copied+freed in cf-promises.

libpromises/generic_agent.h
```diff
@@ -33,6 +33,14 @@
...
 #define GENERIC_AGENT_CHECKSUM_METHOD HASH_METHOD_SHA1
 
+/* Long-style only options, values must start above max ASCII value. */
+enum
+{
+    OPT_EVAL_FUNCTIONS = 256,
+    OPT_SHOW_CLASSES,
+    OPT_SHOW_VARS
+};
+
```
tzz: OK

libpromises/generic_agent.c
```diff
@@ -1968,3 +1968,100 @@ void SetupSignalsForAgent(void)
     signal(SIGUSR1, HandleSignalsForAgent);
     signal(SIGUSR2, HandleSignalsForAgent);
 }
+
+void ShowContextsFormatted(EvalContext *ctx)
```
tzz: OK; moved to generic_agent.c because I do think that feel like the right place for it. If not, tell me where, or move it after merge, it's a trivial change.

jimis: Since it only gets an `EvalContext` argument and it operates only on it, I would say it belongs to `eval_context.c`.

cf-agent/cf-agent.c
```diff
@@ -546,7 +562,23 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
...
                 }
             }
-            break;
+            else if (strcmp(OPTIONS[longopt_idx].name, "show-evaluated-classes") == 0)
+            {
+                if (optarg == NULL)
+                {
+                    optarg = ".*";
+                }
+                config->agent_specific.common.show_classes = xstrdup(optarg);
```
jimis: There is no need to `strdup()` the `optarg` as it lives for the lifetime of the process.

jimis:
Really nice, merging so that it makes it in 3.11! The review comments were minor anyway.

PR #2818, opened 2017-03-31
CFE-2524: Avoid copying variable table when expanding a promise

amousset:
The idea is to stop copying variable tables during policy expansion, which appears to be one of the most heavy operation done, and to lookup the variables in the global table.

This is related to the performance issue described in https://tracker.mender.io/browse/CFE-2524.

This is a more or less a proof of concept for now, and there are still a couple of failing tests (with `@(this.something)` notation for example), but it seems to work quite well and gives huge performance improvements with big policies.

For example, loading the ncf framework takes 2.5 seconds without this patch, 1 second after with 3.10. With big policies (like one with several 500 elements arrays), the execution time goes from 10 minutes to 16 seconds.

I did not find issues caused by this change, but I probably don't understand all the implications, and there might be a logical issue with this patch. Do you know if there is a reason for copying the variables in a local variable table, and do you see any obvious problem with this change?


libpromises/var_expressions.c
```diff
@@ -329,10 +339,32 @@ VarRef *VarRefParseFromNamespaceAndScope(const char *qualified_name,
...
     return ref;
 }
 
+/*
+ * This function will return true if the given variable is
+ * a this.something variable that is an alias to a non-special local variable.
+ */
+bool is_this_not_special(const char *scope, const char *lval) {
+    // TODO: better way to get this list?
+    const char *special_this_variables[] = {"v","k","this","service_policy","promiser","promiser_uid","promiser_gid","promiser_pid","promiser_ppid","bundle","handle","namespace","promise_filename","promise_dirname","promise_linenumber"};
```
amousset: Do you know why these variables names are not all considered as [reserved](https://github.com/cfengine/core/blob/master/libpromises/syntax.c#L815)?

amousset:
@jimis Worked more on this, it now passes the test. I also started adding explanations about the change.

amousset:
Some tests fail but I could not reproduce the errors locally for now.

amousset:
Tests are now fixed!

amousset:
@jimis @kacf Tests are now passing.

Do you know why all special "this." variables names are not all considered as [reserved](https://github.com/cfengine/core/blob/master/libpromises/syntax.c#L815)?

jimis:
> Do you know why all special "this." variables names are not all considered as reserved?

I'm not sure. maybe @nickanderson knows as this is a characteristic of the language. Does it make sense to define `this`-scoped variable names from policy?

jimis:
Triggered a jenkins build.

jimis:
3 acceptance tests in `masterfiles` are failing: 

```
 ./lib/files/manage_variable_values_ini.cf.manage_variable_values_ini.cf
 ./lib/files/manage_variable_values_ini_metachar.cf.manage_variable_values_ini_metachar.cf
 ./lib/files/set_variable_values_ini.cf.set_variable_values_ini.cf
```

Here is the diff in the output files of the first test, expected vs actual:

```
***************
*** 1,5 ****
  [section]
! keyone=1
! keytwo=valuetwo
  [section2]
  wat=WUT
--- 1,5 ----
  [section]
! keyone=valueone
! keytwo=value2
  [section2]
  wat=WUT
```

jimis:
@amousset can you run these tests manually and investigate?

amousset:
@jimis Thanks, I will investigate.

jimis:
I had a very quick look and the failure does not make *any* sense to me, like something irrelevant is broken. Could it be Jenkins misbehaving again? I'll wait for @amousset 's findings first...

amousset:
I could reproduce the issue, it is definitely related to the content of this patch. I'll look deeper to understand what's happening.

jimis:
On Wed, Apr 26, 2017 at 6:12 PM, Alexis Mousset <notifications@github.com>
wrote:

> I could reproduce the issue, it is definitely related to the content of
> this patch. I'll look deeper to understand what's happening.



Great, thanks for the update!

nickanderson:
>I'm not sure. maybe @nickanderson knows as this is a characteristic of the language. Does it make sense to define this-scoped variable names from policy?

@jimis  I don't understand the question.

AFAIK you shoudn't be able to explicitly define this scoped variables. I really only ever use this.bundle, this.namespace, this.promiser, this.promise_filename and this.promise_dirname. I didnt realize until I read the docs that every variable in a bundle can also be referenced with this.

https://docs.cfengine.com/docs/3.10/reference-special-variables-this.html

```
bundle agent main
{
  vars:
    "foo" string => "bar";

  reports:
       "You can $(foo) OR you can $(this.foo) OR you can $($(this.bundle).foo)";
}
```

```
R: You can bar OR you can bar OR you can bar
```

jimis:
> I could reproduce the issue, it is definitely related to the content of this patch. I'll look deeper to understand what's happening.

@amousset how did it go?

@nickanderson we should probably add tests from masterfiles to run in Core's pull request. Do you see any problems with that?

jimis:
Has this made its way into Rudder? Can I help in order to get it into CFEngine?

amousset:
Yes, this patch is applied since May in Rudder. I could not find the time to dig into the failing tests in masterfiles since then (and IIRC the failing tests were some corner cases that we do not have in Rudder).

nickanderson:
@jimis I don't think we need core to run the MPF tests, but adding them should not contribute to the build time significantly.

In fact if we do that, we could remove the stupid plucked.cf.sub which is just copied versions of select bodies and bundles from the stdlib in order to run some tests in core.

amousset:
I will add a test that triggers the error spotted in masterfiles test and work on a fix.

ncharles:
We workarounded the issue in masterfiles with following PR
https://github.com/Normation/ncf/pull/672

jimis:
I'll try to have a look on Monday, I really want to see this merged. :-)

tests/acceptance/01_vars/01_basic/this_variables.cf
```diff
@@ -0,0 +1,101 @@
...
+    "var"      slist => { "var in test" };
+    "var_test" slist => { "var_test in test" };
+
+  files:
+    "$(G.testfile).actual"
+    create => "true",
+    edit_line => test_edit_line("testline");
+
+  methods: 
+    "any" usebundle => inter;
```
jimis: what does this have and the bundles `inter` and `final` have to do with the edit_lines test?

jimis:
This is tricky, I'm not sure if I'll have the time to figure this out. For reference, this is the minimal policy that I'm using to debug this:

```
bundle agent main
{
  files:
    "/tmp/blah.actual"
    create => "true",
    edit_line => test_edit_line("testline");
}

bundle edit_line test_edit_line(line)
{

  vars:
    "sectionName" string => "test";

  insert_lines:
      "[$(sectionName)]
[end]"
        location => start;

      "${line}"
        select_region => INI_section(escape("$(sectionName)"));

}

body location start
# @brief Editing occurs before the matched line
{
      before_after => "before";
}

body select_region INI_section(x)
# @brief Restrict the `edit_line` promise to the lines in section `[x]`
# @param x The name of the section in an INI-like configuration file
{
      select_start => "\[$(x)\]\s*";
      select_end => "\[.*\]\s*";
      select_end_match_eof => "true";
}
```

I'm debugging this with the following command line:
```
cf-agent  -Kf ./file.cf  --debug  --log-modules=all
```

tests/acceptance/01_vars/01_basic/this_variables.cf
nickanderson: Wait a sec.
- Init (copy G.testfile.expected from .finish
- Test (Promise to create and edit G.testfile.actual with edit_line)
- check if .actual (the one we edited) is different from .expected (which is a copy of .finish)

This looks like a broken test to me. The .actual file was not seeded with any content to begin with, and none of the edit_line promises insert `testline`. That is why `testline` is missing from the end state. Perhaps there was an intention to seed the .actual file with some content  that already included `testline` similar to how .expected was seeded from .finish.

jimis:
I believe I'm getting somewhere, given the example I posted earlier. So, it seems that body-local variables are pushed into the VarTable with the scope `body` instead of the scope `this`. So when qualifying unscoped variables, we should check if we are in a STACK_FRAME_TYPE_BODY, and if so, try to qualify them using `body` scope.

I'll try this theory tomorrow.

libpromises/eval_context.c
```diff
@@ -2111,6 +2116,17 @@ static Variable *VariableResolve(const EvalContext *ctx, const VarRef *ref)
         }
     }
 
+    // Try to qualify "this." variable to "current_bundle."
+    const Bundle *last_bundle = EvalContextStackCurrentBundle(ctx);
+    if (last_bundle && strcmp(ref->scope, SpecialScopeToString(SPECIAL_SCOPE_THIS)) == 0)
+    {
+        VarRef *qref = VarRefCopy(ref);
+        VarRefQualify(qref, last_bundle->ns, last_bundle->name);
+        var = VariableResolve(ctx, qref);
```
jimis: This recursion messes me up. What is the guarantee that this will not recurse infinitely?

amousset: I would say because `last_bundle->name` cannot be `this` so the scope will necessarily be different at the next call.

libpromises/eval_context.c
```diff
@@ -2111,6 +2116,17 @@ static Variable *VariableResolve(const EvalContext *ctx, const VarRef *ref)
         }
     }
 
+    // Try to qualify "this." variable to "current_bundle."
+    const Bundle *last_bundle = EvalContextStackCurrentBundle(ctx);
+    if (last_bundle && strcmp(ref->scope, SpecialScopeToString(SPECIAL_SCOPE_THIS)) == 0)
```
jimis: Basically the fix is to modify this to also check if scope is SPECIAL_SCOPE_BODY. But I've made so many modifications until I found it, and then I also wanted to purge the recursion, that I submitted the whole thing in another PR: https://github.com/cfengine/core/pull/3024

jimis:
Closing in favour of #3024. Thanks for all the work @amousset !

PR #2824, opened 2017-04-11
Allow current user symlinks

kacf:


kacf:
Cherry-picked to 3.10.x as well. I did not cherry-pick to 3.7.x, because the section has changed substantially since then.

PR #2827, opened 2017-04-14
CFE-2611: Fix automatic service shutdown on redhat

nickanderson:


nickanderson:
picked to 3.10.x and 3.7.x

PR #2828, opened 2017-04-18
Fix incorrect value range for string_head/tail functions.

kacf:
After d08868e2d6e0d2faea494f286a84eff1bb3f94d8 they can take negative
arguments.


kacf:
Merge together with https://github.com/cfengine/documentation/pull/1606.

tzz:
Nice, thanks for working on this!

PR #2838, opened 2017-05-04
ENT-3147: fix cf-execd not exiting immediately with SIGTERM on AIX.

jimis:


jimis:
All tests are passing in Jenkins.

I also submitted a test build before the fixes were applied, and the new acceptance test for cf-execd `dies_in_time.cf` was failing on AIX and HP-UX.

So this pull request seems to do the job.

cf-execd/cf-execd.c
```diff
@@ -425,31 +428,72 @@ void StartServer(EvalContext *ctx, Policy *policy, GenericAgentConfig *config, E
...
+    int ret = pthread_create(&tid, &threads_attrs, LocalExecThread, thread_config);
+
+    Thread_AllSignalsUnblock();
+
+    if (ret != 0)
     {
         ExecConfigDestroy(thread_config);
-        Log(LOG_LEVEL_INFO, "Can't create thread. (pthread_create: %s)", GetErrorStr());
+        Log(LOG_LEVEL_ERR, "Failed to create thread (pthread_create: %s)",
+            GetErrorStr());
```
jimis: @kacf I increased the level of this message to ERR, but I'm afraid it might always trigger on windows. What do you think?

libpromises/pipes_unix.c
```diff
@@ -156,12 +156,25 @@ static pid_t GenericCreatePipeAndFork(IOPipe *pipes)
...
+
+        // Redmine #2971: reset SIGPIPE signal handler in the child to have a
+        // sane behavior of piped commands within child
         signal(SIGPIPE, SIG_DFL);
+
+        /* The child should always accept all signals after it has exec'd,
+         * else the child might be unkillable! (happened in ENT-3147). */
+        sigset_t sigmask;
+        sigemptyset(&sigmask);
+        sigprocmask(SIG_SETMASK, &sigmask, NULL);
```
jimis: Despite this intending to fix a bug in `cf-execd` (which has a block-all signal mask), the code here is in libpromises and it will affect all our binaries. I can't think of any bad side-effects though. @kacf can you?

tests/acceptance/25_cf-execd/dies_in_time.cf
```diff
@@ -0,0 +1,155 @@
+# ENT-3147, zd#3157
+
+# Ensure that cf-execd dies within 3 seconds after being signalled.
```
jimis: This new acceptance test takes around 2 minutes to complete and it is too complex to convert to a `timed` test. But I don't like it running with all our other fast tests.

What I'm thinking is to create a `slow` category (i.e. directory) of tests and put this in. This category of tests will be disabl-able (?!) from our `testall` script.

What do you think @kacf @nickanderson ?

cf-execd/cf-execd.c
kacf: I'm slightly out of context here, but I think this should work fine. `pthread_create` should be fully supported on windows (using an external pthreads library).

libpromises/pipes_unix.c
kacf: No, I think this looks like a good fix as well.

tests/acceptance/25_cf-execd/dies_in_time.cf
kacf: Yes, leaving processes running has been challenging in the timed tests. I think it's alright to leave it as a standard test.

No strong opinion on the `slow` category, but I'm wondering when this would be useful. In a parallel execution, and considering the other, near 2000 tests, I don't see it as a big deal, it will hardly be noticable. In the CI we would always need to run it regardless.

jimis:
On Mon, May 15, 2017 at 9:52 AM, Kristian Amlie <notifications@github.com>
wrote:

> No strong opinion on the slow category, but I'm wondering when this would
> be useful. In a parallel execution, and considering the other, near 2000
> tests, I don't see it as a big deal, it will hardly be noticable. In the CI
> we would always need to run it regardless.
>

In Travis, tests are split in multiple batches. It helps to have the
categorised so that splitting them equally is easy.

kacf:
> In Travis, tests are split in multiple batches. It helps to have the categorised so that splitting them equally is easy.

Oh right, in that context it makes sense.

PR #2842, opened 2017-05-11
Move ps-parsing debug logging to their own log module.

jimis:


jimis:
Cherry-picking to 3.10.x, I don't consider a DEBUG logging change as user visible.

PR #2843, opened 2017-05-11
Align incoming connection messages regardless of the peer's IP address.

jimis:


jimis:
This is a change that aligns the IPv4-address-prepended messages from cf-serverd so that they are more readable. For example:

```
    info: 127.0.0.1>       Accepting connection
 verbose: 127.0.0.1>       Setting socket timeout to 600 seconds.
    info: 127.0.0.1>       Peer sent only 6 bytes! Considering the protocol as Classic
    info: 127.0.0.1>       Connection is not using latest protocol, denying
    info: 255.255.255.255> Accepting connection
 verbose: 255.255.255.255> Setting socket timeout to 600 seconds.
    info: 255.255.255.255> Peer sent only 6 bytes! Considering the protocol as Classic
    info: 255.255.255.255> Connection is not using latest protocol, denying
```

@nickanderson I quickly coded this after being frustrated while reading long logs. However I'm not so sure if it should be in after all. What do you say? Maybe merge it for 3.11 Beta and revert if we don't like?

jimis:
pinging @nickanderson : quick feedback if I should merge it for the Beta?

jimis:
Merging. We'll see how it goes.

PR #2845, opened 2017-05-15
Rename PartialMatch to PartialRegexMatch

tzz:
Trivial.

As requested in https://github.com/cfengine/core/pull/2480#discussion_r116465231 by @jimis


tzz:
OK, function moved, using `STRING_MATCH_OVECCOUNT`, and name updated.

jimis:
Thanks! 

(Squashed and merged, and removed changelog tags as this is not a user-visible change)

PR #2849, opened 2017-05-19
Remove restriction for defining rpmvercmp (its defaulted now)

nickanderson:


jimis:
Closing, not needed as masterfiles was fixed with cfengine/masterfiles#923.

PR #2852, opened 2017-05-23
Fixed cf-net STAT to use correct argument

olehermanse:
@jimis @Lex-2008 Please cherry pick this to 3.11 branch


jimis:
We must have acceptance tests for cf-net, they would have caught that. I can help with that if you want, they should be fairly similar to the cf-serverd acceptance tests since they need a server running.

jimis:
Cherry-picked to 3.11.x, probably won't make it to the beta since we have a green build already.

PR #2853, opened 2017-05-26
adding single quoted Strings to highlighting

jonny08152:
This change should enable single quoted Strings to be highlighted as well.


nickanderson:
Thanks!

PR #2856, opened 2017-05-29
determine_version.py fixes

jimis:


misc/determine-version.py
```diff
@@ -190,7 +190,8 @@ def version_cmp(a, b):
 all_tags = sorted(all_tags, cmp=version_cmp, reverse=True)
 verbose_print("all_tags   =", all_tags)
 
-if len(all_tags) > 1:
+
+if len(all_tags) >= 1:
```
jimis: @kacf maybe it was intentional what you had before?

misc/determine-version.py
```diff
@@ -179,8 +196,11 @@ def version_cmp(a, b):
     git_rev = subprocess.Popen(["git", "rev-parse", tag + "^{}"],
                                stdout=subprocess.PIPE)
     rev = git_rev.stdout.readlines()[0].strip()
-    if rev == recent_rev:
-        # Ignore tags that point at the same commit.
+    if not in_master_branch and rev == recent_rev:
+        # Ignore tags that point at the same commit as the most recent
+        # tag, i.e. ignore the same tag (TODO COMMENT WHY?). But if we
+        # are on master branch, we need that exact tag, as it might be
+        # the most recent ever issued.
```
jimis: @kacf can you help me refine this comment and maybe the logic summary at the top?

jimis:
@kacf my fix basically consists of using `git show-ref master` to check if the SHA of `refs/remotes/origin/master` or `refs/remotes/upstream/master` matches what we currently are checking. If that's the case, we increase the minor version always.

misc/determine-version.py
kacf: You have to ignore tags that point to the same commit because otherwise the two cases "Z with 3.9, and M with 3.7" in the logic description at the top are indistinguishable, and you want to have different behavior for the two cases (increase patch version on Z, increase minor version on M).

Maybe you should update that comment btw, since you are changing the logic slightly.

misc/determine-version.py
```diff
@@ -191,17 +211,23 @@ def version_cmp(a, b):
...
 
-if len(all_tags) > 1:
-    # This is a new minor version
-    final_print("%s.%d.0a.%s" % (all_tags[0][0], int(all_tags[0][1]) + 1, abbrev_rev))
-else:
-    # This is a new patch version.
+
+if len(all_tags) == 0:
+    # No tags besides the most recent one were found, so this is a new
+    # patch version. So "increase" the patch version:
```
kacf: I honestly don't remember the exact logic here, so I can't comment on the correctness. Just make sure you run this on all series branches 3.7 and up and it behaves correctly. I know 3.8 and others aren't needed, but just for verification.

jimis:
Updated comments all over, I /think/ I understand the logic now. Testing in pull request builder.

jimis:
So the jenkins builds have failed on me quite a few times. I had to manually copy `determine-version.py` to every enterprise repository. But then it still fails because the new logic I have built detects if we are one master branch by comparing HEAD to remotes/origin/master. But apparently if we are building from a pull request, we are *not* on master! So version is misdetected.

ARGH! I'm not sure what to do next.

PR #2858, opened 2017-06-01
Fixes to readenvfile() readcsv() readfile() (CFE-2657, CFE-2656)

olehermanse:
`readfile()`, `readjson()`, `readyaml()`, `readenvfile()` and `readcsv()` now all follow the same syntax: 
 * First parameter is file path, and required
 * Second parameter is max bytes to read, and optional.

Fixed an issue where csv and env parser would output bogus error messages and destroy data when the byte limit was reached.

Need to update docs and add more tests.

https://tracker.mender.io/browse/CFE-2657
https://tracker.mender.io/browse/CFE-2656

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -8417,12 +8417,6 @@ static const FnCallArg READFILE_ARGS[] =
     {NULL, CF_DATA_TYPE_NONE, NULL}
 };
 
-static const FnCallArg READCSV_ARGS[] =
-{
-    {CF_ABSPATHRANGE, CF_DATA_TYPE_STRING, "File name"},
-    {NULL, CF_DATA_TYPE_NONE, NULL}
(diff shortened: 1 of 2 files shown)


olehermanse:
@jimis @nickanderson @kacf Please review.

jimis:
@tzz what do you think about the syntax change?

tzz:
It's lovely :)

olehermanse:
@jimis @nickanderson I think at least 2 out of the 3 commits should be cherry-picked to 3.10.x. Maybe not the argument changes, but the other 2 are bug fix / error messages. See: https://tracker.mender.io/browse/CFE-2657

jimis:
Yes, as I have noted in the ticket the bug-fix should be cherry-picked. I'll do it now and close the ticket, thanks!

PR #2872, opened 2017-06-12
CFE-1565: Acceptance test for inverting nonfalse value with mustache

nickanderson:


jimis:
ping

nickanderson:
pong

latency 2.628e+9 milliseconds

I soft failed the test.

PR #2879, opened 2017-06-14
cf-net: Fix windows issues (ENT-3216 & ENT-3225)

olehermanse:
@jimis This should be cherry-picked to 3.11

### DIFF
--- a/cf-net/cf-net.c
+++ b/cf-net/cf-net.c
@@ -39,6 +39,7 @@
 #include <writer.h>
 #include <policy_server.h>      // PolicyServerReadFile
 #include <generic_agent.h>      // GenericAgentSetDefaultDigest TODO: rm dep
+#include <cf-windows-functions.h> // TODO: move this out of libpromises.
 
 typedef struct
 {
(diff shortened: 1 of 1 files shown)


olehermanse:
Tested on windows, cf-net now works!

cf-net/cf-net.c
```diff
@@ -204,11 +206,25 @@ static void CFNetSetDefault(CFNetOptions *opts){
 
 static void CFNetInit()
 {
+    #ifdef __MINGW32__
```
jimis: indentation of preprocessor goes at level zero.

cf-net/cf-net.c
```diff
@@ -204,11 +206,25 @@ static void CFNetSetDefault(CFNetOptions *opts){
...
 
+static const char *CFNetGetWorkDir()
+{
+    //TODO: Can do this differently. I didn't want to depend on libpromises.
+    #ifdef __MINGW32__
+    return "C:\\Program Files\\Cfengine";
+    #else
+    return "/var/cfengine/bin";
+    #endif
+}
```
jimis: This will break whoever uses custom `./configure` which is everybody besides CFEngine. I think it's better to depend on libpromises here too, and add a TODO for  moving `known_dirs.h` to libutils.

PR #2881, opened 2017-07-02
libutils/man.c: allow to override build time

bmwiedemann:
in order to enable reproducible builds.

See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

### DIFF
--- a/libutils/man.c
+++ b/libutils/man.c
@@ -206,6 +206,13 @@ void ManPageWrite(Writer *out, const char *program, time_t last_modified,
                   const struct option options[],
                   const char *const option_hints[], bool accepts_file_argument)
 {
+    time_t overridetime;
+    char *source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+    if (source_date_epoch != NULL &&
+        (overridetime = (time_t)strtoll(source_date_epoch, NULL, 10)) > 0)
+    {
+        last_modified = overridetime;
+    }
     WriteCopyright(out);
     WriteHeader(out, program, last_modified);
     WriteName(out, program, short_description);


libutils/man.c
```diff
@@ -206,6 +206,11 @@ void ManPageWrite(Writer *out, const char *program, time_t last_modified,
                   const struct option options[],
                   const char *const option_hints[], bool accepts_file_argument)
 {
+    time_t overridetime;
+    char *source_date_epoch = getenv("SOURCE_DATE_EPOCH");
+    if (source_date_epoch != NULL &&
+        (overridetime = (time_t)strtoll(source_date_epoch, NULL, 10)) > 0)
+        last_modified = overridetime;
```
jimis: style: use braces even for single line statements.

bmwiedemann:
updated it. please re-review.

jimis:
Thanks! I'll also cherry-pick to 3.10.x and 3.7.x LTS branches.

jimis:
@bmwiedemann is this all it takes for CFEngine to support reproducible builds?

bmwiedemann:
> is this all it takes for CFEngine to support reproducible builds?

From the openSUSE / OBS (Open Build Service) side, yes, because we have a rather normalized build environment with constant user, path, timezone etc.
Debian guys might find more issues ;-)

jimis:
Great, thanks! 

BTW I have noticed recently that the CFEngine package in SuSE is way out of date. In general it's important to avoid non-LTS releases for stable distros, so I would suggest packaging 3.10.2.  Let me know if I can help there.

bmwiedemann:
@kkaempf might know.
In openSUSE Tumbleweed we have 3.10.0 atm, but openSUSE Leap 42.3 still comes with 3.7.1 probably because we try to avoid major version upgrades in maintenance updates (or because nobody is maintaining it properly).

kkaempf:
openSUSE Leap and SUSE Linux Enterprise will get LTS releases of cfengine. I just don't follow every minor upgrade there.

Tumbleweed always gets the 'latest & greatest' and should be at 3.11.0 (see https://build.opensuse.org/package/show/systemsmanagement/cfengine).

jimis:
@kkaempf the stable releases definitely need latest LTS, currently 3.7.6 and 3.10.2. Let me know if I can help there. You can raise any issue you find on the dev-cfengine mailing list.

Even for Tumbleweed I would suggest staying to latest LTS update, as It's quite possible that non-LTS branches don't even receive a ".1" update, because their 6-month lifetime is too short.

PR #2883, opened 2017-07-06
CFE-2613: Include and package cf-keycrypt with core

olehermanse:
WIP, do not merge yet.

TODO: 
- [x] Edit enterprise and buildscripts repos for new binaries.
- [x] Fix compiler warnings.
- [x] @jimis: Please review `cf-keycrypt.c` which is from the cf-keycrypt repository
- [x] @nickanderson: Please add an acceptance test in your own commit (you should be able to push to this branch)
- [X] @nickanderson writes documentation (https://github.com/cfengine/documentation/pull/1622)
- [x] @nickanderson add example in core
- [ ] Manually test that the built binary works for common use cases
- [ ] Reimplement crypto code to be secure for large files


olehermanse:
For future reference, these were the compiler warnings of the unmodified cf-keycrypt source:

```
Making all in cf-keycrypt
  CC     cf-keycrypt.lo
cf-keycrypt.c: In function ‘get_host_pubkey’:
cf-keycrypt.c:99:9: error: unused variable ‘error’ [-Werror=unused-variable]
cf-keycrypt.c:94:9: error: unused variable ‘ecode’ [-Werror=unused-variable]
cf-keycrypt.c:88:10: error: unused variable ‘value’ [-Werror=unused-variable]
cf-keycrypt.c:87:11: error: unused variable ‘key’ [-Werror=unused-variable]
cf-keycrypt.c: In function ‘rsa_encrypt’:
cf-keycrypt.c:166:27: error: variable ‘len’ set but not used [-Werror=unused-but-set-variable]
cc1: all warnings being treated as errors
make[2]: *** [cf-keycrypt.lo] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+
+#include <lastseen.h>
+#include <conversion.h>
+#include <files_hashes.h>
+#include <locks.h>
+#include <item_lib.h>
+#include <known_dirs.h>
+#include <dbm_api.h>
+/*
+#ifdef LMDB
```
jimis: This comment can probably be removed.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+*/
+
+#include <config.h>
+#include <cf3.defs.h>
```
jimis: Always first inclusion must be 
* either `platform.h`, which itself includes `config.h`.
* OR `cf3.defs.h` which includes itself `platform.h` and a bunch of libpromises stuff.

I think the 2nd is what you want here.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+#include <known_dirs.h>
+#include <dbm_api.h>
+/*
+#ifdef LMDB
+#include <lmdb.h>
+#endif
+*/
+
+#define STDIN     0
+#define STDOUT    1
```
jimis: Unused defines. Remove?

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+
+#define BUFSIZE 1024
+
+void usage()
+{
+    printf("Workdir is defined as %s\n", WORKDIR);
+    printf(
+    "\n"
+    "Usage: cf-keycrypt [-e public-key|-d private-key|-H hostname-or-ip] -o outfile -i infile [-h]\n"
```
jimis: I've never used this tool, but the command line usage that I would expect based on other tools is:

```
cf-keycrypt -e filename   > file.out
cf-keycrypt -o file.out -e filename
```

Those two are equivalent. Similar for `-d`. Key is found in WORKDIR unless an option like `--key=/path/to/key` is specified.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+    if (error != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to get IP from host (getaddrinfo: %s)",
+            gai_strerror(error));
+        return NULL;
+    }
+
+    for(res = result; res != NULL && !found; res = res->ai_next)
+    {
+        inet_ntop(res->ai_family, get_in_addr((struct sockaddr *)res->ai_addr), ipaddress, sizeof(ipaddress));
```
jimis: Remove `get_in_addr()` and `inet_ntop()`. The modern way to convert a struct sockaddr to textual representation of address is `getnameinfo()`. Example:

```c
{
    /* Convert IP address to string, no DNS lookup performed. */
    char txtaddr[CF_MAX_IP_LEN] = "";
    getnameinfo(ap->ai_addr, ap->ai_addrlen,
                txtaddr, sizeof(txtaddr),
                NULL, 0, NI_NUMERICHOST);
    Log(LOG_LEVEL_DEBUG, "Bound to address '%s' on '%s' = %d", txtaddr,
        CLASSTEXT[VSYSTEMHARDCLASS], VSYSTEMHARDCLASS);
}
```

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+    //char *keyname = NULL;
+    char hash[CF_HOSTKEY_STRING_SIZE];
+    //char buffer[BUFSIZE];
+    char ipaddress[BUFSIZE];
+    // int ecode;                            // TODO: this was unused(!)
+    struct addrinfo *result;
+    struct addrinfo *res;
+    bool found = false;
+
+    int error = getaddrinfo(host, NULL, NULL, &result);
```
jimis: I see `getaddrinfo()` but no `freeaddrinfo()`; leaking memory?

cf-keycrypt/cf-keycrypt.c
jimis: Now that I think about all this whole address lookup function `get_host_pubkey()`, it has many issues in the lookups, and we have `Hostname2IPString()` for that. So better use that and save some code and risk.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
+/*
+   cf-keycrypt.c
+
+   Copyright (C) 2017 cfengineers.net
```
jimis: I think we should add our copyright notice as well. Maybe also convert the language to GPLv3. Pinging @kacf and @nickanderson. 

We should definitely mention the primary author of this in the CONTRIBUTORS file, and probably ping him in this pull request, what's his handle?

cf-keycrypt/cf-keycrypt.c
jimis: Apparently the command needs more options than my previous examples. It needs a key file OR a hostname/IP. 
* `-e` or `--encrypt`
* `-d` or `--decrypt`
* `-k` or `--keyfile`
* `-H` or `--host`
* `-o` or `--output`
*  `filename` should be a mandatory argument

It's important to agree to a sensible command line usage *before* merging so that we don't break things later.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+    return NULL;
+}
+
+void *readseckey(char *secfile)
+{
+    FILE *fp = NULL;
+    RSA* PRIVKEY = NULL;
+    static char *passphrase = "Cfengine passphrase";
```
jimis: This exists all over our codebase. Move all separate definitions to `cf3.defs.h`. Consolidate this `readseckey()` function together with the existing code in crypto.c, export the one from there and import it here.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+    {
+        return (void *)PRIVKEY;
+    }
+}
+
+void *readpubkey(char *pubfile)
+{
+    FILE *fp;
+    RSA *key = NULL;
+    static char *passphrase = "Cfengine passphrase";
```
jimis: same as above for this phrase and for readpubkey function.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+    free(tmpplain);
+    free(tmpciph);
+    RSA_free(key);
+    return plsz;
+}
+
+int main(int argc, char *argv[])
+{
+    OpenSSL_add_all_algorithms();
+    ERR_load_crypto_strings();
```
jimis: `CryptoInitialize()`

jimis:
Plenty of code style issues: braces not on separate lines, implicit comparisons, unneeded casts, uneeded parentheses etc.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+        memset(tmpplain, '\0', ks);
+        len = fread(tmpciph, 1, ks, infile);
+        if (len > 0){
+            if ((size = RSA_private_decrypt(ks, tmpciph, tmpplain, key, RSA_PKCS1_PADDING)) == -1)
+            {
+                fprintf(stderr, "%s\n", ERR_error_string(ERR_get_error(), NULL));
+                return -1;
+            }
+        }
+        fwrite(tmpplain,1,strlen(tmpplain),outfile);
```
jimis: This must be checked for error return.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+        outfile = stdout;
+        //printf("Write to stdout\n");
+    }else if (!(outfile = fopen(plainfile, "w"))){
+        fprintf(stderr, "Error: Cannot create output file.\n");
+        return -1;
+    }
+
+    ks = RSA_size(key);
+    tmpciph = (unsigned char *)malloc(ks * sizeof(unsigned char));
+    tmpplain = (unsigned char *)malloc(ks * sizeof(unsigned char));
```
jimis: remove casts and sizeofs.

jimis: Also use xmalloc.

jimis: Or just use stack allocated VLAs:
```c
char tmpciph[ks], tmpplain[ks];
```

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+    while(!feof(infile))
+    {
+        memset(tmpciph, '\0', ks);
+        memset(tmpplain, '\0', ks);
+        len = fread(tmpciph, 1, ks, infile);
+        if (len > 0){
+            if ((size = RSA_private_decrypt(ks, tmpciph, tmpplain, key, RSA_PKCS1_PADDING)) == -1)
+            {
+                fprintf(stderr, "%s\n", ERR_error_string(ERR_get_error(), NULL));
+                return -1;
```
jimis: File descriptor and memory leaks in case of error return. Not sure if this is a blocker though, given the short lifetime of the program.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,388 @@
...
+        size = rsa_decrypt(key, infile, outfile);
+        if (size < 0)
+            exit(1);
+    }
+    else
+    {
+        usage();
+        exit(1);
+    }
+    exit(0);
```
jimis: Replace all zero and ones with EXIT_SUCCESS and EXIT_FAILURE. I would also suggest to just use `return retval` just at the end, if it's easy.

olehermanse:
@jonhenrik13 We are adding cf-keycrypt to core :) You should probably take a look at this PR, it's not done yet though.

cf-keycrypt/cf-keycrypt.c
kacf: A file from external contributors should have the license intact, unless we've made substantial changes. Apache can be relicensed under GPLv3, so this is no problem for the project as a whole

jimis: @kacf given that we are making significant changes to it, I suggest keeping the initial license intact, but prepend our license and copyright for 2017. Sounds right?

kacf: Yep, that's fine!

olehermanse: Prepended our license.

cf-keycrypt/cf-keycrypt.c
olehermanse: Fixed

cf-keycrypt/cf-keycrypt.c
olehermanse: Removed

cf-keycrypt/cf-keycrypt.c
olehermanse: Removed

cf-keycrypt/cf-keycrypt.c
olehermanse: I think @nickanderson is in favor of getting this merged fairly quickly, and in the long term adding it's functionality to `cf-key` and deprecating this tool before the next LTS.

cf-keycrypt/cf-keycrypt.c
olehermanse: Added.

cf-keycrypt/cf-keycrypt.c
olehermanse: @jimis This is not a straight forward replace. Consider the steps here:
1. `getaddrinfo()` returns a list of results for the given hostname
2. Loop through the list. If `127.0.0.1` or `::1` is in that list and before `Address2Hostkey` finds a hostname, return `ppkeys/localhost.pub` 
3. If `Address2Hostkey` succeeded during 2., return `ppkeys/root-<KEY>.pub` 
4. If not loop through list again. Return any `ppkeys/root-<IP-ADDRESS>.pub` where that file exists.

This is what I understand. I'm sure it's flawed, but what is the desired behavior here?

cf-keycrypt/cf-keycrypt.c
jimis: I can tell you one thing: after it gets merged, nobody will have time to actually do anything. And we have time-based releases, not feature-based, so the release *will* happen.

So what you merge is what you'll get, almost certainly.

nickanderson: Let's clean up the cli usage. I can always argue and beg for backporting, thats easier than allocating time to merge functionality into cf-key when we already have something that works and is standalone.

olehermanse:
@jimis @nickanderson Commenting here since https://tracker.mender.io/browse/CFE-2613 has the annoying Jira+Zendesk bug.

Syntax will be:
```
cf-keycrypt --encrypt --keyfile /path/to/key.pub --output FILENAME.crypt FILENAME
cf-keycrypt -e -k /path/to/key.pub FILENAME > FILENAME.cfcrypt
cf-keycrypt --encrypt --keyfile /path/to/key.pub FILENAME > FILENAME.cfcrypt
cf-keycrypt --encrypt --host hostname(from cf-key -s) FILENAME > FILENAME.cfcrypt
cf-keycrypt --decrypt --keyfile /path/to/priv -o FILENAME.crypt FILENAME
```

Only note here is that the current implementation doesn't support piping output to stdout, so I will have to add that.

jimis:
Syntax looks good to me. @nickanderson agree?

Yes if `-o` is missing, output should be always on stdout, I believe it
makes sense. BUT I see some conflicts with our `Log()` function, that
always logs to stdout.

olehermanse:
Rebased on latest master.

cf-keycrypt/cf-keycrypt.c
olehermanse: Fixed.

olehermanse:
@jimis I slightly disagree with the syntax. My proposal is:
```
$ cf-keycrypt --encrypt message.txt --key /path/to/key --output message.enc
$ cf-keycrypt --decrypt message.enc -H somehost --output message.txt
```

It is almost the same, but input file must follow `--encrypt` or `--decrypt`, since that is the file that will be encrypted/decrypted. It also allows you to structure the command like a normal sentence, for example:
 * Using key X encrypt message.txt to message.enc: `--key X --encrypt message.txt -o message.enc`
 * Decrypt message.enc to message.txt using key X: `--decrypt message.enc -o message.txt --key X`

This structure makes the most sense to me.

olehermanse:
I have implemented the new syntax, as explained in the previous comment.

The new help menu looks like this:
```
$ ./cf-keycrypt/cf-keycrypt
Usage: cf-keycrypt [OPTIONS]

Options:
  --help        , -h       - Print the help message
  --manpage     , -M       - Print the man page
  --encrypt     , -e value - Encrypt file
  --decrypt     , -d value - Decrypt file
  --key         , -k value - Use key file
  --host        , -H value - Encrypt for host (get key from lastseen database)
  --output      , -o value - Output file

Website: http://www.cfengine.com
This software is Copyright (C) 2008,2010-present CFEngine AS.
```

cf-keycrypt/cf-keycrypt.c
olehermanse: Fixed.

cf-keycrypt/cf-keycrypt.c
olehermanse: Fixed.

cf-keycrypt/cf-keycrypt.c
olehermanse: I would rather not touch `crypto` or `defs`in this PR/commit. I can do that as a separate step after, if you'd like.

cf-keycrypt/cf-keycrypt.c
olehermanse: Fixed, and a bunch of other `fread` and `fwrite`.

cf-keycrypt/cf-keycrypt.c
olehermanse: Fixed.

olehermanse:
@jimis Please review, I believe this is ready to merge. Will need to add an enterprise and buildscripts repo for it to be packaged.

Changed the syntax as we discussed. I also did a lot of refactoring and error checking. Did *not* do big changes to internals like:
 * Change `crypto.c` or `cf3.defs.h`
 * Change `cf-keycrypt` to use `getnameinfo()` or `Hostname2IPString()`

@nickanderson Maybe look over the commit messages/changelog entries before merge.

nickanderson:
This looks really good. I had to look hard to find some nit picks.

`Added sys.cf-keycrypt component variable` I think should be `Added sys.cf_keycrypt component variable` since `-` is not a valid character to have in a variable name.

`This software is Copyright (C) 2008,2010-present CFEngine AS.` I think this is `Northern.tech AS`.

Does it output to stdout? if `-o` isnt' provided? I saw in the comments that wasn't present. I wouldn`t stop this for it, just curious.

olehermanse:
@nickanderson No, I removed the stdout and stdin functionality, mainly because it doesn't play well with our Log functionality.

jimis:
> It is almost the same, but input file must follow --encrypt or --decrypt, since that is the file that will be encrypted/decrypted. It also allows you to structure the command like a normal sentence, for example:
> * Using key X encrypt message.txt to message.enc: `--key X --encrypt message.txt -o message.enc`
> * Decrypt message.enc to message.txt using key X: `--decrypt message.enc -o message.txt --key X`

I believe this exact syntax you are describing will still be valid if `[filename]` is an optional argument to `cf-keycrypt` and `--encrypt` and `--decrypt` do not accept any argument. So you can still write it like a sentence. (blergh :-p )

Additionally in that way, the future functionality of missing `[filename]` meaning encrypt stdin can be added sometime.

> This structure makes the most sense to me.

To me the syntax you described in your comment on Sep 25 is more familiar as it resembles gpg and a bunch of other really old UNIX encoding/decoding tools, like `compress`, `gzip` and the likes. I'm OK merging this one if you want though, the difference is small.

olehermanse:
@jimis Correct me if I'm wrong but `tar`, `gzip` and `compress` all require you to place options (`-` or `--`) before the filename. This is the POSIX way of using `getopt`.

In that case, this:
> I believe this exact syntax you are describing will still be valid if [filename] is an optional argument to cf-keycrypt and --encrypt and --decrypt do not accept any argument. So you can still write it like a sentence. (blergh :-p )

is wrong.

My current implementation uses the POSIX style `getopt`. To implement it like you say, I'd have to ditch that and continue parsing options after filename. It's not a lot of work either way, if you strongly prefer the other way. I tried to stay  `POSIXLY_CORRECT` as the man page calls it, and this was the best way to allow you to structure the command however you like.

tests/acceptance/00_basics/cf-keycrypt/decrypt.cf
```diff
@@ -0,0 +1,27 @@
...
+}
+
+bundle agent check
+{
+  methods:
+      "any"
+        usebundle => dcs_check_diff("$(this.promise_dirname)/plaintext",
+                                    "$(G.testfile)",
+                                    "$(this.promise_filename)");
+}
```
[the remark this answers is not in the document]
olehermanse: Yes, @nickanderson, what were you thinking? ;)

jimis:
Dear @olehermanse `tar` predates POSIX and it's very non-portable, so I'll skip that. `compress`, `gzip` and the likes are reading stdin and outputing to stdout if no filename is present, and `-d` option of theirs is accepting no filename, but filename is an optional argument of the whole utility, not a specific option, and that's because the utility is made to act on files. So that's what I had in mind.

To implement as I say, you only have to make sure to parse your arguments with POSIX `getopt()`. Then on GNU systems (and our source has `__GNU_SOURCE` defined) [the libc library makes the magic happen](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html). Indeed this will not happen on non-GNU systems if you use `getopt()` which is OK by me. But you are using `getopt_long()` which is GNU-only, so the that code will be always used to parse that.

So what I said is right: you will be able to use your beautiful sentence-like syntax. It's not POSIX correct though, and it's not the documented way to use it, and I will still use the beautiful syntax with filename at the end, that I'm so used to in all utilities. So everyone is happy. :-)

olehermanse:
@jimis Please refrain from talking smack about the almighty `tape archiver - tar`. I can change the syntax to your POSIX hostile abomination. This has been noted down in your personal file. My way was enforcing POSIX compliance (in terms of option order), and clearly far superior, even though it used the GNU function.

Sarcasm aside, I'll make the fix today :)

jimis:
Dear @olehermanse, It is my understanding that this issue has been
resolved. I will not send any further messages.

olehermanse:
@jimis The syntax is now as you wanted. Input filename can be put *anywhere*.

olehermanse:
Started build in jenkins: https://ci.cfengine.com/job/bootstrap-enterprise-pr/432/

jimis:
Nah I know it will fail, not even Travis passes (unit tests fail). But
that's not the point of the assignment.

olehermanse:
@jimis Green on all except one platform

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,504 @@
...
+        memset(tmp_ciphertext, '\0', ks);
+        len = fread(
+            tmp_plaintext, 1, ks - RSA_PKCS1_PADDING_SIZE, input_file);
+        if (len <= 0 || ferror(input_file) != 0)
+        {
+            Log(LOG_LEVEL_ERR, "Could not read file '%s'", input_path);
+            error_return = true;
+            break;
+        }
+        unsigned long size = RSA_public_encrypt(
```
jimis: `size` should be `int`, because that's what `RSA_public_encrypt` returns and because it's compared to negative numbers later on.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,504 @@
...
+            strlen(tmp_plaintext), tmp_plaintext, tmp_ciphertext,
```
jimis: shouldn't this be `len`? What happens if the file is binary and contains NULLs and you call `strlen(tmp_plaintext)` ?

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,504 @@
...
+    char tmp_ciphertext[ks], tmp_plaintext[ks];
+
+    srand(time(NULL));
+    bool error_return = false;
+    while (!feof(input_file))
+    {
+        memset(tmp_plaintext, '\0', ks);
+        memset(tmp_ciphertext, '\0', ks);
+        len = fread(
+            tmp_plaintext, 1, ks - RSA_PKCS1_PADDING_SIZE, input_file);
```
jimis: So it keeps reading from a file 245 byte chunks, and RSA-encrypts each and every one separately using PKCS #1 v1.5 padding, and appends to a file, and each block is fully independent from the other without any chaining whatsoever. This is definitely [fishy](https://stackoverflow.com/questions/13908407/openssl-rsa-public-encrypt-returned-size-always-rsa-sizersa) but I am not sure if there are specific vulnerabilities. Maybe @estenberg can comment ?

cf-keycrypt/cf-keycrypt.c
```diff
@@ -0,0 +1,504 @@
...
+    {
+        for(res = result; res != NULL; res = res->ai_next)
+        {
+            inet_ntop(
+                res->ai_family,
+                get_in_addr((struct sockaddr *)res->ai_addr),
+                ipaddress,
+                sizeof(ipaddress));
+            snprintf(
+                buffer, BUFSIZE * sizeof(char), "%s/ppkeys/root-%s.pub",
```
jimis: We don't really use this kind of line breaking elsewhere in our code, unless it is almost unavoidable. Not a blocker, just keep in mind for next time.

jimis:
The code quality looks good, and the command line too (despite the bike-shedding ;-) ).

I'll now try to review the cryptographic part.

cf-keycrypt/cf-keycrypt.c
estenberg: @jimis you are right, this is not good because 1) it leaks information as identical input chunks result in identical encrypted data 2) encrypting large volumes with a public key can leak information about the private key and 3) it is very inefficient (compared to symmetric crypto). At the very least we need to set a hard limit on the size (e.g. you can only encrypt 245 bytes or less), so it would really only work for password strings and the like.

The "right way" to do this is of course to encrypt a random symmetric key with RSA and then use that key with a block or stream cipher in a secure mote (e.g. CBC, CTR, etc.)

estenberg:
Two high-level comments:
* This is insecure if we are encrypting more than ~200 bytes (even then it is not great). Either we need to set a hard limit on this input size or we need to implement this "properly" with symmetric crypto.
* I don't think "cf-keycrypt" is a particularly catchy name. Why not just call it cf-crypt?

olehermanse:
@jimis So I don't suspect I'll have time to rewrite the encryption code in at least a month. What do you want to do? Take a stab at it yourself, or wait until november?

jimis:
Not sure how to handle this, but we can't knowingly put in weak encryption. Lets bring this one up in the next planning meeting.

P.S. to do proper public key encryption seems to be a bit complicated; found this article:  https://shanetully.com/2012/06/openssl-rsa-aes-and-c/

olehermanse:
@jimis I also found this which is a little more up to date, but doesn't show AES part. http://hayageek.com/rsa-encryption-decryption-openssl-c/

jimis:
Great @olehermanse ! I'm all for doing this right, since `cf-keycrypt` is a tool people ask for all the time and it's really almost done. But I'm not sure if the code in that link can encrypt arbitrarily long buffers. See `man RSA_public_encrypt` and you'll know what I mean.

Before coding stuff though we should make sure we are doing it the right way, maybe this is material for a meeting, or discussion on the relevant ticket.

olehermanse:
@jimis I know, hence why I said it didn't cover the AES step. (RSA to encrypt AES key, AES to encrypt data)

jimis:
Closing since CFE-2613 is not scheduled. Feel free to re-open when work resumes.

PR #2884, opened 2017-07-06
Travis fixes

olehermanse:


olehermanse:
@jimis This is important

olehermanse:
@kacf @Lex-2008 Can this be merged? As it is now, travis will not fail even though some tests fail. The `set +e` and `set -e` should fix this.

Lex-2008:
Looks good to me! Although in buildscrips we usually add `set -x` to see what command failed, too

jimis:
@olehermanse IIRC travis will show FAIL is any command in the main script returns error. Does it not honour that for some test? Which one? Can you also add this info to the commit message?

Doing `set -e` will cause early exit, skipping all subsequent tests. I'm OK with that, but lets note specifically what led to this decision.

jimis:
This has been superseded by #2911. Close?

PR #2885, opened 2017-07-11
ENT-3240: check for master branch via `git branch -r --merged`

Lex-2008:
instead of `git show-ref master`

### DIFF
--- a/misc/determine-version.py
+++ b/misc/determine-version.py
@@ -178,15 +178,16 @@ recent_version = extract_version_components(recent_tag)
 verbose_print("recent_version = ", recent_version)
 
 
-in_master_branch = False
-git_show_ref =  subprocess.Popen(["git", "show-ref", "master"],
+in_master_branch = True
+git_all_merged_branches = subprocess.Popen(["git", "branch", "-r", "--merged"],
                                  stdout=subprocess.PIPE)
-for line in git_show_ref.stdout:
-    (sha,ref,) = line.split()
-    if (sha == full_rev) and \
-       (ref == "refs/remotes/origin/master" or ref == "refs/remotes/upstream/master"):
-       in_master_branch = True
-       verbose_print("Detected that we are in master branch")
+for line in git_all_merged_branches.stdout:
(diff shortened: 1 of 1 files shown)


Lex-2008:
Note: I will need to edit ./3rdparty/core/determine-version.py in other repos, too

kacf:
This will be for @jimis to review. I deliberately stayed away from all branches in the original version, using only tags.

jimis:
> I deliberately stayed away from all branches in the original version, using only tags.

@kacf so your suggestion would be to blacklist the last bad tag IIRC?

misc/determine-version.py
```diff
@@ -178,15 +178,15 @@ def version_cmp(a, b):
...
+git_all_merged_branches = subprocess.Popen(["git", "branch", "-r", "--merged"],
                                  stdout=subprocess.PIPE)
-for line in git_show_ref.stdout:
-    (sha,ref,) = line.split()
-    if (sha == full_rev) and \
-       (ref == "refs/remotes/origin/master" or ref == "refs/remotes/upstream/master"):
-       in_master_branch = True
-       verbose_print("Detected that we are in master branch")
+for line in git_all_merged_branches.stdout:
+    match = re.search('/[0-9]\+\.[0-9]\+\.x', line)
```
jimis: Looks good, so if it works +1. Just some nitpicks to avoid catching the wrong lines:
* maybe anchor the regex to the end using `$`
* maybe also grep for `upstream` or `origin` to avoid foreign or local branches

Lex-2008:
well, `-r` should already avoid listing local branches, but let's extend the regexp, indeed

Lex-2008:
Regexp corrected!

jimis:
Great! Feel free to merge at your own judgement. ;-)

kacf:
> so your suggestion would be to blacklist the last bad tag IIRC?

No, what I mean is that I never anticipated this scenario. Using branches is probably the right approach, but I haven't written that part of the code.

PR #2886, opened 2017-07-11
Update copyright to Northern.tech and current year.

kacf:
Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>

### DIFF
--- a/BSDmakefile
+++ b/BSDmakefile
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) CFEngine AS
+#  Copyright 2017 Northern.tech AS
 #
 #  This file is part of CFEngine 3 - written and maintained by CFEngine AS.
 #
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) CFEngine AS
+#  Copyright 2017 Northern.tech AS
 #
 #  This file is part of CFEngine 3 - written and maintained by CFEngine AS.
 #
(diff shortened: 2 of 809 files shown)


BSDmakefile
```diff
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) CFEngine AS
+#  Copyright 2017 Northern.tech AS
```
jimis: Why do you delete the old copyright? Who had copyright before 2017? Does the official company name contain a dot?

I would expect something like:

```
Copyright 2017-present Northern Tech AS
Copyright 2008-2017 CFEngine AS
```

kacf: I got the string directly from @ThomasRyd. This is the one we should use.

kacf: Oh, and retaining copyright for CFEngine is definitely wrong, the company doesn't exist anymore. All its copyrights were transferred to Northern.tech.

jimis: This is the first source code that I see that deletes copyright history and mentions copyright only for the current year like it just got created.

kacf:
Passed [here](https://ci.cfengine.com/job/testing-enterprise-pr/343/).

kacf:
Cherry-picked to 3.7.x, 3.10.x and 3.11.x as well.

Makefile.am
```diff
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) CFEngine AS
+#  Copyright 2017 Northern.tech AS
 #
 #  This file is part of CFEngine 3 - written and maintained by CFEngine AS.
```
olehermanse: @kacf `written and maintained by CFEngine AS` ?

kacf: Right, good catch. I'll fix that.

PR #2887, opened 2017-07-11
ENT-3267: Fix Xen_Cpuid() corrupting the stack when built with old GCC.

jimis:


jimis:
@kacf review?

jimis:
TODO cherry-pick to all branches. TODO send patch to Xen where this code originates from? I believe this is where @skreuzer borrowed this from, according to the comment.

kacf:
You get commit-of-the-year award!

PR #2890, opened 2017-07-26
Compile against OpenSSL 1.1

FauxFaux:
These changes make cfengine build against OpenSSL 1.1.0.

This mostly involves just a few small changes in how APIs work; using accessors, or not having stack allocated objects.

The major changes for review are:

 * `tls_generic_test.c` just cannot work anymore. It is reimplementing an old version of OpenSSL, relying too heavily on the OpenSSL internals, which are no-longer exposed. I have removed it from `Makefile.am`, but not deleted the code, in this pull request. I attempted to port it, but it's pretty impossible.
 * The `session_key` changes in `libcfnet/client_protocol.c` look like they leak a `malloc`, but they don't; the `session_key` is eventually freed by normal libc-`free` already; making assumptions about old OpenSSL internals. This code is actually arguably more correct like this.
 * I have deleted the key type checks in `libcfnet/tls_generic.c`. The functions that are called immediately after this are documented to safely fail if the key is not of the right format. Checking the type directly isn't really supported anymore.

I have run the build and unit tests. I am unable to get the acceptance tests to pass without my changes (all kinds of syntax errors with `chown` and `fakeroot`), so I didn't try them with the changes.

A patch for 3.9 (instead of HEAD) is attached to the Debian bug: https://bugs.debian.org/863568

### DIFF
--- a/libcfnet/client_protocol.c
+++ b/libcfnet/client_protocol.c
@@ -179,6 +179,7 @@ static bool SetSessionKey(AgentConnection *conn)
 {
     BIGNUM *bp;
     int session_size = CfSessionKeySize(conn->encryption_type);
+    unsigned char *buf;
 
     bp = BN_new();
 
@@ -196,7 +197,9 @@ static bool SetSessionKey(AgentConnection *conn)
         return false;
     }
 
-    conn->session_key = (unsigned char *) bp->d;
+    buf = malloc(BN_num_bytes(bp));
+    BN_bn2bin(bp, buf);
+    conn->session_key = buf;
     return true;
 }
 
@@ -208,6 +211,8 @@ int AuthenticateAgent(AgentConnection *conn, bool trust_key)
     int encrypted_len, nonce_len = 0, len, session_size;
     bool need_to_implicitly_trust_server;
     char enterprise_field = 'c';
+    const BIGNUM *pubkey_n, *pubkey_e;
+    BIGNUM *newkey_n, *newkey_e;
 
     if (PRIVKEY == NULL || PUBKEY == NULL)
     {
@@ -293,13 +298,15 @@ int AuthenticateAgent(AgentConnection *conn, bool trust_key)
 /*Send the public key - we don't know if server has it */
 /* proposition C2 */
 
+    RSA_get0_key(PUBKEY, &pubkey_n, &pubkey_e, NULL);
+
     memset(sendbuffer, 0, CF_EXPANDSIZE);
-    len = BN_bn2mpi(PUBKEY->n, sendbuffer);
+    len = BN_bn2mpi(pubkey_n, sendbuffer);
     SendTransaction(conn->conn_info, sendbuffer, len, CF_DONE);        /* No need to encrypt the public key ... */
 
 /* proposition C3 */
     memset(sendbuffer, 0, CF_EXPANDSIZE);
-    len = BN_bn2mpi(PUBKEY->e, sendbuffer);
+    len = BN_bn2mpi(pubkey_e, sendbuffer);
     SendTransaction(conn->conn_info, sendbuffer, len, CF_DONE);
 
 /* check reply about public key - server can break conn_info here */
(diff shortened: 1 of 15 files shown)


jimis:
Travis fails:

```
  CC     hashes.lo
hashes.c: In function ‘FileChecksum’:
hashes.c:50:9: error: implicit declaration of function ‘EVP_MD_CTX_new’ [-Werror=implicit-function-declaration]
hashes.c:50:31: error: initialization makes pointer from integer without a cast [-Werror]
hashes.c:62:9: error: implicit declaration of function ‘EVP_MD_CTX_free’ [-Werror=implicit-function-declaration]
```

libcfnet/client_protocol.c
```diff
@@ -196,7 +197,9 @@ static bool SetSessionKey(AgentConnection *conn)
         return false;
     }
 
-    conn->session_key = (unsigned char *) bp->d;
+    buf = malloc(BN_num_bytes(bp));
+    BN_bn2bin(bp, buf);
+    conn->session_key = buf;
```
jimis: where is `conn->session_key` freed?

libcfnet/client_protocol.c
```diff
@@ -460,6 +467,8 @@ int AuthenticateAgent(AgentConnection *conn, bool trust_key)
             return false;
         }
 
+        RSA_set0_key(newkey, newkey_n, newkey_e, NULL);
```
jimis: Where are `newkey_n` and `newkey_e` freed in case of the failures above?

libcfnet/tls_generic.c
```diff
@@ -88,12 +88,6 @@ static int CompareCertToRSA(X509 *cert, RSA *rsa_key)
             TLSErrorString(ERR_get_error()));
         goto ret1;
     }
-    if (EVP_PKEY_type(cert_pkey->type) != EVP_PKEY_RSA)
```
jimis: According to [docs](https://www.openssl.org/docs/man1.1.0/crypto/EVP_PKEY_type.html) we can switch to `EVP_PKEY_base_id(pkey)`

> Previous versions of this document suggested using EVP_PKEY_type(pkey->type) to determine the type of a key. Since EVP_PKEY is now opaque this is no longer possible: the equivalent is EVP_PKEY_base_id(pkey).

libpromises/crypto.c
```diff
@@ -451,18 +459,18 @@ int EncryptString(char *out, size_t out_size, const char *in, int plainlen,
                           max_ciphertext_size, out_size);
     }
 
-    EVP_CIPHER_CTX_init(&ctx);
-    EVP_EncryptInit_ex(&ctx, CfengineCipher(type), NULL, key, iv);
+    EVP_CIPHER_CTX_init(ctx);
```
jimis: Can be skipped.

libpromises/crypto.c
```diff
@@ -534,22 +542,22 @@ int DecryptString(char *out, size_t out_size, const char *in, int cipherlen,
                           max_plaintext_size, out_size);
     }
 
-    EVP_CIPHER_CTX_init(&ctx);
-    EVP_DecryptInit_ex(&ctx, CfengineCipher(type), NULL, key, iv);
+    EVP_CIPHER_CTX_init(ctx);
```
jimis: skip

jimis:
I'll try to move forward with this, fixing the nitpicks and making the build compatible with both old and new openssl. @FauxFaux let me know if you are planning to look at this, to avoid working on the same thing.

libcfnet/client_protocol.c
jimis: It is freed in `DeleteConn()` apparently. But in this function, there is a missing `BN_clear_free()` to avoid the leak of `bp`. But I believe this was also leaking before.

jimis:
`tls_generic_test.c` is an abomination! It's impossible to fix without rewriting, so I agree that you removed it from the makefile @FauxFaux ! I'm now thinking of burning all this code with fire...

jimis:
Superseded by #2916, discussion also in [CFE-2629](https://tracker.mender.io/browse/CFE-2629). Thank you @FauxFaux for starting this!

PR #2892, opened 2017-07-27
3.7.x ENT-3147 cherry-pick

jimis:


jimis:
Build is red. Since this is a complicated and risky fix, and because it doesn't cherry-pick cleanly, I'm skipping the cherry-pick to 3.7.x.

PR #2895, opened 2017-07-29
CFE-1960: fix storage mount promise when existing mountpoint has a si…

skreuzer:
…milar path

The MatchFSInFstab will parse the each line of the fstab looking for the
existence of a mountpoint as a substring. In certain cases the call to strstr
would incorrectly match if a nfs volume with a similar mount point was already
mounted. Instead of using strstr, tokenize the line and look at each token for
an exact match.

 ChangeLog: Title

### DIFF
--- a/cf-agent/nfs.c
+++ b/cf-agent/nfs.c
@@ -667,13 +667,23 @@ PromiseResult VerifyUnmount(EvalContext *ctx, char *name, Attributes a, const Pr
 static int MatchFSInFstab(char *match)
 {
     Item *ip;
+    const char delimit[]=" \t\r\n\v\f";
+    char *fstab_line, *token;
 
     for (ip = FSTABLIST; ip != NULL; ip = ip->next)
     {
-        if (strstr(ip->name, match))
+        fstab_line = xstrdup(ip->name);
+        token = strtok(fstab_line, delimit);
+        while (token != NULL)
         {
-            return true;
+            if(strcmp(token, match) == 0)
+            {
+                free(fstab_line);
+                return true;
+            }
+            token = strtok(NULL, delimit);
         }
+        free(fstab_line);
     }
 
     return false;
--- a/cf-agent/nfs.c
+++ b/cf-agent/nfs.c
@@ -673,15 +673,18 @@ static int MatchFSInFstab(char *match)
     for (ip = FSTABLIST; ip != NULL; ip = ip->next)
     {
         fstab_line = xstrdup(ip->name);
-        token = strtok(fstab_line, delimit);
-        while (token != NULL)
+        if(strncmp(fstab_line, "#", 1) != 0)
         {
-            if(strcmp(token, match) == 0)
+            token = strtok(fstab_line, delimit);
+            while (token != NULL)
             {
-                free(fstab_line);
-                return true;
+                if(strcmp(token, match) == 0)
+                {
+                    free(fstab_line);
+                    return true;
+                }
+                token = strtok(NULL, delimit);
             }
-            token = strtok(NULL, delimit);
         }
         free(fstab_line);
     }
--- a/tests/unit/nfs_test.c
+++ b/tests/unit/nfs_test.c
@@ -1,0 +1,35 @@
+#include <test.h>
+#include <nfs.c>
+#include <nfs.h>
+#include <item_lib.h>
+
+static void test_MatchFSInFstab(void)
+{
+    AppendItem(&FSTABLIST, "fileserver1:/vol/vol10      /mnt/fileserver1/vol10     nfs     rw,intr,tcp,fg,rdirplus,noatime,_netdev", NULL);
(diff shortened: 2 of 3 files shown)


tzz:
/me is not an approver but I had some comments:

* can you add some test cases? (The ticket reference is not enough IMO.) Right now in core.git there's only one test case for `storage` promises: `tests/acceptance/06_storage/01_local/001.cf` so anything you add will be a huge improvement :)
* `strtok` is used very rarely in CFEngine source code, maybe there's a reason? When I did some parsing, I usually reached for regular expressions. Something like `libenv/unix_iface.c:GetProcFileInfo()` might do the job here.
* maybe a brief comment in the C code might be helpful to understand what it does, like a sample fstab line?

nickanderson:
Patch working as reported by user on mailing list: https://groups.google.com/d/msg/help-cfengine/VNjEiOoczwQ/xgAqf-E8AAAJ

tests/unit/nfs_test.c
```diff
@@ -0,0 +1,35 @@
...
+    AppendItem(&FSTABLIST, "fileserver2:/vol/vol11 	 /mnt/fileserver2/vol11 	 nfs 	 rw,intr,tcp,fg,noatime", NULL);
+
+    assert_true(MatchFSInFstab("/mnt/fileserver1/vol10"));
+    assert_true(MatchFSInFstab("/mnt/fileserver1/vol11"));
+    assert_false(MatchFSInFstab("/mnt/fileserver1/vol1"));
+
+    assert_true(MatchFSInFstab("/mnt/fileserver2/vol10"));
+    assert_true(MatchFSInFstab("/mnt/fileserver2/vol11"));
+    assert_false(MatchFSInFstab("/mnt/fileserver2/vol1"));
+}
```
jimis: @nickanderson can you think of more test cases that cause trouble?

nickanderson: The commented out entries were the other one I was thinking about

skreuzer:
> On a side not, and since you are messing with nfs filesystem promises, what happens if /etc/fstab contains lines commented out? Would these lines incorrectly match?

Right now this function will still match a line which has been commented out. That particular bug is being tracked in CFE-2198 and if this gets merged I plan on addressing that next.

jimis:
> Additionally could you please edit your commit to have a proper title? This way the changelog will be updated correctly.

Scratch this. I fetched your branch and it looks fine on the command line. It's the github UI that apparently cuts the title in two pieces if it's too long.

tests/unit/nfs_test.c
tzz: I'd also test

* empty FSTAB
* garbage and empty lines
* lines with just one token

cf-agent/nfs.c
```diff
@@ -667,13 +667,23 @@ PromiseResult VerifyUnmount(EvalContext *ctx, char *name, Attributes a, const Pr
 static int MatchFSInFstab(char *match)
 {
     Item *ip;
+    const char delimit[]=" \t\r\n\v\f";
+    char *fstab_line, *token;
 
     for (ip = FSTABLIST; ip != NULL; ip = ip->next)
```
tzz: I'd make the `FSTABLIST` a parameter of the function, so it's not single-purpose and so it doesn't rely on global variables.

tzz:
@skreuzer I don't understand why push code with a known issue (it will match commented-out lines)? Isn't it better to fix that issue right here before it's merged? The extra code should not be too bad.

skreuzer:
> I don't understand why push code with a known issue

It seems like it is easier to get patches accepted when they are small and concise. I just pushed a change to this branch which addresses the commented out line false positive issue. I didn't rebase since I figure these are two separate issues so they should be addressed in two separate commits

jimis:
I agree with @tzz, these functions should take parameters and not modify globals. As it is now it's hard to follow the logic - where is FSTABLIST populated (answer: in many places) and why does it contain comments in the first place. Refactorings are welcome here.

However I'm easily merging this since it contains two clear bugfixes, together with unit tests, so it's a good improvement over what we already have.

cf-agent/nfs.c
```diff
@@ -667,13 +667,26 @@ PromiseResult VerifyUnmount(EvalContext *ctx, char *name, Attributes a, const Pr
...
     {
-        if (strstr(ip->name, match))
+        fstab_line = xstrdup(ip->name);
+        if(strncmp(fstab_line, "#", 1) != 0)
```
tzz: @skreuzer I think this is not ideal either, it will miss `<SPACE>#comment`

Please consider submitting another change moving the code to regular expressions and refactoring as @jimis seconded. If you don't plan to, let us know please.

tzz:
The following tests should also be done in a followup, IMO:

* the fstab is empty
* some lines contain garbage data
* some lines are empty (length 0)
* some lines contain only spaces
* some comments have spaces before the `#`

PR #2896, opened 2017-08-02
CFE-2203: Enable Xen hypervisor detection on all x86 platforms

skreuzer:
ChangeLog: Title

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -170,14 +170,13 @@ static int EOS_Version(EvalContext *ctx);
 static int MiscOS(EvalContext *ctx);
 static void OpenVZ_Detect(EvalContext *ctx);
 
+static bool ReadLine(const char *filename, char *buf, int bufsize);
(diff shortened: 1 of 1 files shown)


jimis:
What is this doing? "Enable Xen hypervisor detection on all x86 platforms" does not tell me much, and all I see in the diff is unmodified code moved around. Can you be a bit more exact on what has changed, in your commit message?

skreuzer:
> What is this doing?

All the code to detect the Xen hypervisor gets excluded on non linux platforms because its all gated by the __linux__ preprocessor. Since Xen is capable of running i386 and x86_64 based operating systems, allow that code to be included regardless of what OS you are compiling for. (Autoconf is already configured to define XEN_CPUID_SUPPORT if its being built on a x86 platform)

The diff is a bit misleading because all it is doing is moving the Xen_Hv_Check() and Xen_Cpuid() out from under the __linux__ preprocessor. I updated the commit message to be a bit more informative.

jimis:
Thanks, looks good now. I'll merge after our CI gives the green light.

PR #2898, opened 2017-08-03
Add script to detect duplicate keys

atsaloli:
See [README](https://raw.githubusercontent.com/atsaloli/core/2287f2ace54c1b29e5386fc570597dcfbfbb56d8/contrib/enterprise/detect_duplicate_keys/README)


atsaloli:
@mikeweilgart suggested this may not catch all hosts with duplicate keys (if multiple hosts with duplicate keys come and go while the script is sleeping).  It may be more reliable/complete to enable "logallconnections" in cf_serverd.cf controls and parse that.  I'm exploring that option.

nickanderson:
Thanks for the contribution!

PR #2899, opened 2017-08-03
Update Changelog from 3.11.0b1.

jimis:


nickanderson:
Comments in https://tracker.mender.io/browse/ENT-3289

jimis:
Updated, thanks Nick.

PR #2902, opened 2017-08-04
CFE-2478: Allow '/' and '\' characters in the module protocol

nickanderson:
Changelog: Title


nickanderson:
@jimis stumbled on https://tracker.mender.io/browse/CFE-2478 and i have no idea if this is the proper fix or not, but classic arrays allow slashes between [ ] the module protocol should too. LMK what you think.

jimis:
Besides the syntax errors in the Code (backslashes need to be backslash-escaped) I'm quite sure this change will introduce breakage. Allow backslashes in identifiers? That's definitely not a good idea. :-)

nickanderson:
Well I was thinking that paths on windows use backslashes.

PR #2903, opened 2017-08-05
Set the exit value when running cf-key

skreuzer:
When running cf-key to generate new keys, set the exit value of the
program to be 0 on success and 1 on failure. This makes it easier to
catch errors during setup of a new machine.

Change the default behavior of the program to not write anything to stdout,
opting to use the Log() function which can write to stdout and will also
allow output to be sent to syslog.

Add a --inform option to set the global log level to LOG_LEVEL_INFO.

Change the permissions of the randseed file to 600 and catch the exception
if the chmod call fails.

### DIFF
--- a/cf-key/cf-key-functions.c
+++ b/cf-key/cf-key-functions.c
@@ -314,7 +314,7 @@ int RemoveKeys(const char *input, bool must_be_coherent)
     return -1;
 }
 
-void KeepKeyPromises(const char *public_key_file, const char *private_key_file)
+bool KeepKeyPromises(const char *public_key_file, const char *private_key_file)
 {
 #ifdef OPENSSL_NO_DEPRECATED
     RSA *pair = RSA_new();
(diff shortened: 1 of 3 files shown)


jimis:
I like this. But I'm not sure about the change in behaviour: cf-key becomes silent by default, printing nothing on the console. @nickanderson what do you think?

The commit message needs a big "changelog" tag btw.

nickanderson:
I don't see any issue with suppressing "Making a key pair for CFEngine, please wait, this could take a minute..." by default with cf-key.

+1 for exit code.

Not for currently released branches.

nickanderson:
@skreuzer can you ammend your commit to include the line:

```
Changelog: commit
```

skreuzer:
> @skreuzer can you ammend your commit to include the line:

i just amended the commit message with that line

jimis:
Merged, thanks! 

I still think that LOG_LEVEL_NOTICE would be better for that single message "generating keys", because it might hang for a while. Anyway merged, we can always tweak it later.

PR #2905, opened 2017-08-30
sha-deletion.sh wrapper script for record cleanup

mikeweilgart:
For bulk removals of host records this script is incomparably easier
to use than the UI or the REST API.  For single host removals it is
still slightly easier than the REST API and there is lot less to remember.

I've been using this for many months; there are really no caveats.  Input is
fully validated before use.  (It's not checked that a given SHA actually *exists*
on the hub before attempting to remove it; only that is validly formatted as a SHA.
However, this matches the behavior of the UI for hostkey removals, so not an issue.)

### DIFF
--- a/contrib/enterprise/sha-deletion.sh
+++ b/contrib/enterprise/sha-deletion.sh
@@ -1,0 +1,208 @@
+#!/bin/bash
+
+# Author: Mike Weilgart
+# Date: 19 April 2017
+
+# This script removes host records from a CFEngine Hub
+
+########################################################################
+#                      USAGE MESSAGE                                   #
+########################################################################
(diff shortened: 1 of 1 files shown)


mikeweilgart:
Of interest: I only made one bugfix to this after my initial testing and codewriting:

Add 'silent' flag to curl command

For routine purposes this doesn't do anything, but if you redirect
the output of the script (e.g. so you can put it in the background after
you get prompted for password), then you will get (without the 'silent'
flag) a lot of noise of 'downloading progress.'  This prevents that.

nickanderson:
Thank you for the contribution!

PR #2907, opened 2017-08-30
Inclusion of sysmacros.h by types.h is deprecated.

ekigwana:
Release notes for glibc-2.25 state:

The inclusion of <sys/sysmacros.h> by <sys/types.h> is deprecated.
This means that in a future release, the macros “major”, “minor”, and
“makedev” will only be available from <sys/sysmacros.h>.

These macros are not part of POSIX nor XSI, and their names frequently
collide with user code; see for instance glibc bug 19239 and Red Hat
bug 130601. <stdlib.h> includes <sys/types.h> under _GNU_SOURCE, and
C++ code presently cannot avoid being compiled under _GNU_SOURCE,
exacerbating the problem.

evalfunction.c: In function ‘FnCallFileStatDetails’:
evalfunction.c:3191:64: error: implicit declaration of function ‘minor’; did you mean ‘mknod’? [-Werror=implicit-function-declaration]
             snprintf(buffer, CF_MAXVARSIZE, "%jd", (uintmax_t) minor(statbuf.st_dev) );
                                                                ^~~~~
                                                                mknod
evalfunction.c:3199:64: error: implicit declaration of function ‘major’ [-Werror=implicit-function-declaration]
             snprintf(buffer, CF_MAXVARSIZE, "%jd", (uintmax_t) major(statbuf.st_dev) );

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -656,6 +656,7 @@ AC_CHECK_HEADERS(winsock2.h)
 AC_CHECK_HEADERS(ws2tcpip.h)
 AC_CHECK_HEADERS(zone.h)
 AC_CHECK_HEADERS(sys/uio.h)
+AC_CHECK_HEADERS_ONCE([sys/sysmacros.h]) dnl glibc deprecated inclusion in sys/type.h
 AC_CHECK_HEADERS(sys/types.h)
 AC_CHECK_HEADERS(sys/mpctl.h) dnl For HP-UX $(sys.cpus) - Mantis #1069
 AC_CHECK_HEADERS(shadow.h)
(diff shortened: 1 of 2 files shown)


jimis:
Thanks, I'm running this on all platforms on our CI. I have a feeling that this will fail, because the inclusion needs to be guarder with `#ifdef HAVE_SYS_SYSMACROS_H`, and a check `AC_CHECK_HEADERS_ONCE([sys/sysmacros.h])` needs to be added to configure.ac. We'll see.

jimis:
Indeed MinGW build failed with:

 `../../libutils/platform.h:82:27: fatal error: sys/sysmacros.h: No such file or directory`

ekigwana:
Done

jimis:
Green in all platforms, thanks!

PR #2909, opened 2017-09-04
Changelog-generator updates

kacf:
@jimis: These are the changes we have made in Mender's changelog-generator that can be ported back to CFEngine's generator. Pick and choose as you like. Even with these ports there are substantial differences, mostly due to different repository structure, but this at least brings them closer to each other.


jimis:
Takk @kacf !

PR #2910, opened 2017-09-11
Track code coverage for builds.

maciejmrowiec:
Integrate with codecov and report coverage.

Changelog: none

Signed-off-by: Maciej Mrowiec <mrowiec.maciej@gmail.com>

@jimis early xmas gift ;)

### DIFF
--- a/travis-scripts/script.sh
+++ b/travis-scripts/script.sh
@@ -20,10 +20,10 @@ then
     # On osx the default gcc is actually LLVM
     export CC=gcc-6
     NO_CONFIGURE=1 ./autogen.sh
-    ./configure --enable-debug --prefix=$INSTDIR --with-init-script --with-lmdb=/usr/local/Cellar/lmdb/  --with-openssl=/usr/local/opt/openssl
+    ./configure --enable-debug --prefix=$INSTDIR --with-init-script --with-lmdb=/usr/local/Cellar/lmdb/  --with-openssl=/usr/local/opt/openssl --enable-coverage
 else
     NO_CONFIGURE=1 ./autogen.sh
-    ./configure --enable-debug --with-tokyocabinet --prefix=$INSTDIR --with-init-script
+    ./configure --enable-debug --with-tokyocabinet --prefix=$INSTDIR --with-init-script --enable-coverage
 fi
 
 make dist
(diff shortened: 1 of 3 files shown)


maciejmrowiec:
Considering adding exclude for .*_test.c files, currently I see they are included in the coverage.

jimis:
Thanks @maciejmrowiec, awesome gift! Consider excluding everything under the `tests/` directory.

travis-scripts/script.sh
```diff
@@ -47,12 +48,14 @@ chmod -R go-w .
 if [ "$JOB_TYPE" = acceptance_tests_common ]
 then
     ./testall --tests=common
+    bash <(curl -s https://codecov.io/bash) -F acceptance_tests ;
```
jimis: `acceptance_tests_common`

travis-scripts/script.sh
```diff
@@ -47,12 +48,14 @@ chmod -R go-w .
...
     return
 fi
 
   # WARNING: the following job runs the selected tests as root!
 if [ "$JOB_TYPE" = acceptance_tests_unsafe_serial_network_etc ]
 then
     ./testall --gainroot=sudo --tests=timed,slow,errorexit,libxml2,libcurl,serial,network,unsafe
+    bash <(curl -s https://codecov.io/bash) -F acceptance_tests ;
```
jimis: `acceptance_tests_rest`

travis-scripts/script.sh
maciejmrowiec: They are called the same so it can merge coverage for distributed acceptance test executions to get same result as those whole suite was executed in single go.

maciejmrowiec:
tricky to do so because all gcov file get dumped flat in main dir after running tests, added some regexes to exclude test.c.gov files (most of unit tests follow this) and in addition -g option that says it allows you to exclude dir (not sure if have effect, setting anyways).  Added commit and lets see the how the next report will look.

jimis:
> They are called the same so it can merge coverage for distributed
> acceptance test executions to get same result as the whole suite was
> executed in single go.
>
I think it does not have the effect you want. It merges coverage reports
even with different names. In fact, I think that using the same name is
having opposite effect, and *overwrites* the coverage reports, since
cf-serverd coverage is missing.

Try with different names, and lets see how it goes.

maciejmrowiec:
It supposed to merge, we used it heavily in mender.
Exclude did not work as well as I'd like. Added commit with config to the service configuration instead tweaking coverage reports send to the service.

maciejmrowiec:
Closing this PR, want to try to open a new one with squashed commits and make it run clean instead overwriting results every time I add commits.

PR #2911, opened 2017-09-11
Fixes to travis (test failures, shebangs, mac etc.)

olehermanse:
PR will be updated with more commits

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
 # feedback ASAP. For more info read this:
 # https://docs.travis-ci.com/user/multi-os/
 matrix:
+  fast_finish: true # Build will succeed once all linux jobs succeed
   include:
     - os: linux
       sudo: required
@@ -34,19 +35,10 @@ matrix:
     - os: osx
       env: JOB_TYPE=compile_only
       sudo: false
-#    - os: osx
-#      sudo: required
-#      env: JOB_TYPE=acceptance_tests_common
-#    - os: osx
-#      sudo: required
-#      env: JOB_TYPE=acceptance_tests_unsafe_serial_network_etc
-#  allow_failures:
-#    - os: osx
-#      sudo: required
-#      env: JOB_TYPE=acceptance_tests_common
-#    - os: osx
-#      sudo: required
-#      env: JOB_TYPE=acceptance_tests_unsafe_serial_network_etc
+  allow_failures:
+    - os: osx
+      env: JOB_TYPE=compile_only
+      sudo: false
 
 before_install:
   - set -e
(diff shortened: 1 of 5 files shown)


olehermanse:
@jimis I have added 5 commits here and started travis builds for all of them. Will get back to this.

jimis:
thanks

olehermanse:
@jimis I believe this is along the lines of what you wanted. See `.travis.yml`. Now, no scripts include `set` commands. Test and install scripts must pass, deploy and after scripts can fail.

olehermanse:
Once travis is green, I believe this is ready to merge.

All changes are Travis specific, so running Jenkins is not needed.

@Lex-2008 This might be of interest to you.

olehermanse:
@jimis Please review and merge

jimis:
> Travis: Added /bin/sh shebang to all travis scripts

These are intended to be sourced with `set -e`, not executed. It would make more sense to add /bin/false shebang with a comment that says so.

> Travis: return errors if cd or echo fails

For the scripts that are meant to be sourced with set -e, it is confusing to do ` || return 1`.

jimis:
The rest of the commits are good stuff (TM), and I don't care that much for the notes above, as long as it works. Merging!

olehermanse:
> For the scripts that are meant to be sourced with set -e, it is confusing to do `|| return 1`.

But the scripts should return if cd fails, would you prefer `return 0` ?

olehermanse:
> These are intended to be sourced with set -e, not executed. It would make more sense to add /bin/false shebang with a comment that says so.

This is true, I guess I prefer `/bin/sh` because it clearly communicates (to me and my editor) that this is intendend to be an `sh` script, as opposed to a `bash` script for example.

jimis:
On Fri, Sep 22, 2017 at 5:20 PM, Ole Herman Schumacher Elgesem <
notifications@github.com> wrote:

> For the scripts that are meant to be sourced with set -e, it is confusing
> to do || return 1.
>
> But the scripts should return if cd fails, would you prefer return 0 ?
>
Of course returning with error code is the proper thing. My point is that
`set -e` assures that will happen. Doing `|| return 1` does not add
functionality, but confuses the coder to think that he needs to do that all
over the place.

olehermanse:
> Of course returning with error code is the proper thing. My point is that
`set -e` assures that will happen. Doing `|| return 1` does not add
functionality, but confuses the coder to think that he needs to do that all
over the place.

But then again, `set -e` is outside of this script.

jimis:
> But then again, set -e is outside of this script.

Yes, but this does not change the fact that this script is intended to be executed under `set -e`. Or else start adding `|| return 1` noise all over the place.

PR #2914, opened 2017-09-30
Two fixes for HashMap memory allocation

vpodzime:
One minor tweak and one fix for a potential SEGFAULT/memory corruption.


jimis:
Thanks, but `calloc()` already initialises the allocated region to zero. :-)

http://pubs.opengroup.org/onlinepubs/009695399/functions/calloc.html

vpodzime:
> Thanks, but calloc() already initialises the allocated region to zero. :-)

Oh, nice! I thought it's just the multiplication it does and always wondered if/why that was worth a separate function. :)

vpodzime:
I just removed the second patch from this PR. Not sure if the first one is worth merging on its own, but now this PR can either be merged or closed.

PR #2915, opened 2017-10-01
CFE-2301: HashMap improvements

vpodzime:
A number of commits that make the HashMap implementation more flexible. The initial size can be set plus it grows and shrinks on demand. It is also better prepared for future changes, e.g. switching the size values from powers of 2 to primes to prevent issues with poor hash functions. For now, there's at least a minor trick that can potentially deal with some kinds of poor hash functions.


vpodzime:
The test failures seem to be caused by the commit 050ec2d, probably because the expected data no longer match due the change of hash values.

jimis:
Awesome, thank you!

libutils/hash_map.c
```diff
@@ -38,8 +40,11 @@ HashMap *HashMapNew(MapHashFn hash_fn, MapKeyEqualFn equal_fn,
     map->equal_fn = equal_fn;
     map->destroy_key_fn = destroy_key_fn;
     map->destroy_value_fn = destroy_value_fn;
-    map->size = HASHMAP_BUCKETS;
+    map->size = DEFAULT_HASHMAP_BUCKETS;
     map->buckets = xcalloc(map->size, sizeof(BucketListItem *));
+    memset(map->buckets, 0, map->size * sizeof(BucketListItem *));
```
vpodzime: This is not needed, ``(x)calloc`` initializes the chunk of memory to zeroes.

libutils/hash_map.c
```diff
@@ -48,6 +53,38 @@ static unsigned int HashMapGetBucket(const HashMap *map, const void *key)
...
+    BucketListItem **old_buckets;
+
+    old_size = map->size;
+    old_buckets = map->buckets;
+
+    map->size = new_size;
+    /* map->load stays the same */
+    map->threshold = (size_t) map->size * MAX_LOAD_FACTOR;
+    map->buckets = xcalloc(map->size, sizeof(BucketListItem *));
+    memset(map->buckets, 0, map->size * sizeof(BucketListItem *));
```
vpodzime: Same here

vpodzime:
Redundant ``memset()`` calls removed, rebased to current *master* and one patch added.

vpodzime:
Rebased to current master and fixed two things:

1. Put macro values using bitshifts into brackets.
2. Do not compare load to ``MIN_HASHMAP_BUCKETS`` when deciding whether to shrink or not (it can never be bigger than ``init_size``).

I still wasn't able to run the full test suite and update the expected outputs in variables tests though.

libutils/string_lib.c
```diff
@@ -59,15 +59,13 @@ char *StringFormat(const char *fmt, ...)
     return res;
 }
 
-unsigned int StringHash(const char *str, unsigned int seed, unsigned int max)
+unsigned int StringHash(const char *str, unsigned int seed)
 {
-    CF_ASSERT(ISPOW2(max),
```
jimis: Shouldn't this assert still be somewhere, since you are still using binary and instead of division ?

libutils/hash_map.c
```diff
@@ -25,26 +25,93 @@
...
+    /* spread (XOR) higher bits into lower
+     *
+     * This is a trick used e.g. in Java HashMap implementation
+     * (http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/HashMap.java#l320).
+     * The goal is to prevent poor hash functions from causing collisions by
+     * common/related values only differing in the higher bits (e.g. when just
+     * taking the exact 32 bits of float values). Good hash functions where the
+     * input affects all 32 bits of the result don't suffer from any regressions
+     * caused by this operation.
+     */
```
jimis: > Good hash functions where the input affects all 32 bits of the result don't suffer from any regressions caused by this operation.

What regressions are caused by this?  maybe you meant something else?

libutils/string_lib.c
vpodzime: Not really. The *HashMap* now makes sure its size is a power of 2 and it's the hash map doing the limitation of the hash values returned by hash functions. See https://github.com/cfengine/core/pull/2915/files#diff-557504754a8a166d13cf3066e34c1390R49 for details.

jimis: Then it should be inside GetBucket, right before the binary and computation. I agree that it's an expensive assertion to have there though. Maybe use `assert()` which is compiled out in release builds.

vpodzime:
Jenkins job: https://ci.cfengine.com/job/testing-enterprise-pr/540/

vpodzime:
New jenkins build triggered: https://ci.cfengine.com/job/testing-enterprise-pr/578

vpodzime:
New jenkins build: https://ci.cfengine.com/job/pr-pipeline/43/

vpodzime:
yet another jenkins build, this time with @nickanderson's patch and expected to be all green: https://ci.cfengine.com/job/testing-pr/56/

olehermanse:
@vpodzime  Jenkins build all green.

@jimis Review and merge?

PR #2916, opened 2017-10-03
CFE-2629: Openssl 1.1 compatibility

jimis:


jimis:
Do not merge, it's not ready yet. Apparently needs some reviewing and squashing after all tests pass.

Superseding #2890, thanks @FauxFaux for the initial pull request. How to retain attribution if all commits are squashed into one?

libutils/libcrypto-compat.c
```diff
@@ -0,0 +1,409 @@
+/*
+ * Copyright 2016 The OpenSSL Project Authors. All Rights Reserved.
+ *
```
jimis: @kacf what do you think about this C file and its license? As I note in the commit message, this file has been taken from the [OpenSSL wiki](https://wiki.openssl.org/index.php/OpenSSL_1.1.0_Changes#Downloads).

nickanderson:
> How to retain attribution if all commits are squashed into one?

Empty commit with --author set?

libutils/libcrypto-compat.c
kacf: Should be ok. The license below ensures that we can use it as per the OpenSSL license.

vpodzime:
Great to see this! I had troubles building the code on Fedora since it no longer has the devel packages for *openssl-1.0*.

jimis:
Alright the Travis build is green, which means that at least the old-openssl build is working. I wish Travis was testing the new openssl as well.

jimis:
Jenkins is green as well on all platforms (besides known or spurious redness). Will merge soon.

jimis:
I managed to test an openssl-1.1 build of CFEngine and I got a segfault on one test:

```
   debug: Waiting at incoming select...
    info: 127.0.0.1>       Accepting connection
 verbose: 127.0.0.1>       Setting socket timeout to 600 seconds.
lt-cf-serverd: server_tls.c:189: ServerTLSPeek: Assertion `SSLSERVERCONTEXT != NULL && PRIVKEY != NULL && PUBKEY != NULL' failed.
Aborted (core dumped)
```

Happens when the cipher negotiation comes out empty, which occurs with newer openssl stricter client-side cipherlist. Will work on fixing the segfault.

jimis:
@FauxFaux when this gets merged, all commits will be squashed into one authored by myself, so that there is no tests-passing incontinuity. Are you OK with just mentioning your name and initial pull request in the commit message? I'm also thinking of using a `Co-Authored-By: ` header, as described in the [linux kernel git conventions|https://git.wiki.kernel.org/index.php/CommitMessageConventions], what do you think?

FauxFaux:
Yeah, that's fine. So long as it's in, I don't mind!

libcfnet/client_protocol.c
```diff
@@ -196,7 +199,10 @@ static bool SetSessionKey(AgentConnection *conn)
         return false;
     }
 
-    conn->session_key = (unsigned char *) bp->d;
+    conn->session_key = xmalloc(BN_num_bytes(bp));
+    BN_bn2bin(bp, conn->session_key);
```
jimis: So previously `session_key` was the internal representation directly, but now we convert it using `BN_bn2bin()` which [is not just a plain byte copy (or is it)](https://github.com/openssl/openssl/blob/OpenSSL_1_1_0f/crypto/bn/bn_lib.c#L521). So this could be a behaviour change. It needs to be checked with classic protocol connections between old CFEngine versions and new ones.

jimis:
All tests are passing, with builds of both new or old openssl!

jimis:
Thank you @FauxFaux for the initial pull request!

PR #2919, opened 2017-10-10
ENT-3395: cf-execd systemd service now only kills cf-execd itself

nickanderson:
Changelog: Title
(cherry picked from commit ff7aef4da3ee704f105685bf9a58ebbc835628cf)


nickanderson:
@Lex-2008  any idea what the test failure is about? I have rebased the changes on top of current 3.10.x I don't see how this change would break any tests.

nickanderson:
Rebased

PR #2921, opened 2017-10-11
CFE-2551: Test that merged data does not include variables from caller

nickanderson:


jimis:
./01_vars/02_functions/mergedata-CFE-2551-2.cf FAIL (The test Passed, but failure was expected) (UNEXPECTED FAILURE)

jimis:
So the test passes? There is no bug here?

olehermanse:
Ping @nickanderson

nickanderson:
I apologise for being so terrible at following up on these

nickanderson:
So, now the proper ones are passing and proper ones are soft failing. Does Jenkins agree?

nickanderson:
https://ci.cfengine.com/view/all/job/pr-pipeline/246/

nickanderson:
Yeah, but that was a different test case. Probably spurious.

FAIL (UNEXPECTED FAILURE) ./25_cf-execd/timed/mailfilter_repeated_runs.cf

Merging.

PR #2924, opened 2017-10-13
Fixing confusing log message

eli-taft-quoin:
This makes the log more clear on why promises are being skipped when viewing verbose output.  After testing locally, I've discovered the reason the promise is skipped is NOT because 'if' or 'ifvarclass' are not defined; rather, it's because the promise being evaluated is not defined in that context.

See https://tracker.mender.io/browse/CFE-2697 for details.

### DIFF
--- a/libpromises/promises.c
+++ b/libpromises/promises.c
@@ -629,7 +629,7 @@ Promise *ExpandDeRefPromise(EvalContext *ctx, const Promise *pp, bool *excluded)
         if (ifvarclass && !IsVarClassDefined(ctx, ifvarclass, pcopy))
         {
             Log(LOG_LEVEL_VERBOSE, "Skipping promise '%s'"
-                " because 'if'/'ifvarclass' is not defined", pp->promiser);
+                " because it is not defined in this context.", pp->promiser);
 
             *excluded = true;
             return pcopy;


libpromises/promises.c
```diff
@@ -629,7 +629,7 @@ Promise *ExpandDeRefPromise(EvalContext *ctx, const Promise *pp, bool *excluded)
         if (ifvarclass && !IsVarClassDefined(ctx, ifvarclass, pcopy))
         {
             Log(LOG_LEVEL_VERBOSE, "Skipping promise '%s'"
-                " because 'if'/'ifvarclass' is not defined", pp->promiser);
+                " because it is not defined in this context.", pp->promiser);
```
nickanderson: The promise is not part of the current context because the if/ifvarclass constraint was not met correct?

The if/ifvarclass constraint may not be met because it contains a variable that's not defined, or because it evaluates to non true.

eli-taft-quoin: Yes I believe this is correct.  Tracing through the code, I got to the IsDefinedClass in libpromises/eval_context.c, and that returns false because of the final line:

return r == true;

jimis:
I don't think I'm following here, the message is ugly but makes sense to me. There is some ongoing discussion in [CFE-2697](https://tracker.mender.io/browse/CFE-2697). Maybe you can post a minimal policy that reproduces the message when it does not make sense?

jimis:
Superseded by this one: https://github.com/cfengine/core/pull/2949

OK to close this @eli-taft-quoin ?

eli-taft-quoin:
Sounds good to me.

PR #2927, opened 2017-10-16
CFE-2406: cf-execd now re-parses augments on policy reload

nickanderson:
Changelog: Title


nickanderson:
:/ I don't think this is working.

Testing  a build with this patch I can see that it will pick up iniitial control values from augments but  if I modify a policy file and then wait  for the daemon to reload itself the configuration from augments doesn/t stick.

nickanderson:
This works in my testing. As I tweak settings for executor control in def.json I see them being reflectedin the behaviour.

For example splay, schedule, mailto/mailfrom.

Will squash

PR #2930, opened 2017-10-18
CFE-2704: mergedata segfaults when called on a non-container

amousset:
https://tracker.mender.io/browse/CFE-2704


amousset:
I'm not sure about the consequences of this change, particularly what is the expected behavior of mergedata on a JSON primitive.

nickanderson:
@amousset can you rebase your changes since I have merged tests from https://github.com/cfengine/core/pull/2931

If you rename them to from .x.cf to .cf then they should start passing with your change.

amousset:
Done, thank you for the tests!

nickanderson:
@jimis for your review

libpromises/evalfunction.c
```diff
@@ -3431,6 +3431,13 @@ static FnCallResult FnCallMergeData(EvalContext *ctx, ARG_UNUSED const Policy *p
             return FnFailure();
         }
 
+        // Ignore json primitives, only merge containers
+        if (JsonGetElementType(json) != JSON_ELEMENT_TYPE_CONTAINER)
+        {
+            JsonDestroy(json);
```
jimis: Shouldn't this be destroyed only `if (allocated)` ?

Regarding your question, I believe somebody else is more appropriate to answer:

> I'm not sure about the consequences of this change, particularly what is the expected behavior of mergedata on a JSON primitive.

@tzz what do you say? :-)

Finally, if this is the right way, I would like to see a VERBOSE message here indicating that mergedata skipped merging something.

tzz: I think that if some of the data is not mergeable, the function should explain in an error message and fail, to be consistent with the rest of CFE and to avoid data loss. Skipping data elements can be very hard to diagnose and debug.

amousset: Updated to fail on non mergeable arguments, it seems more appropriate indeed.

tzz:
Thank you @amousset that looks great to me.

libpromises/evalfunction.c
```diff
@@ -3431,6 +3431,18 @@ static FnCallResult FnCallMergeData(EvalContext *ctx, ARG_UNUSED const Policy *p
...
+        // Fail on json primitives, only merge containers
+        if (JsonGetElementType(json) != JSON_ELEMENT_TYPE_CONTAINER)
+        {
+            if (allocated)
+            {
+                JsonDestroy(json);
+            }
+
+            Log(LOG_LEVEL_ERR, "%s is not mergeable as it it not a container", RvalToString(arg->val));
+            return FnFailure();
```
jimis: Memory leak. You have to `SeqDestroy(containers)`.

jimis:
Thanks!

PR #2931, opened 2017-10-18
CFE-2704: Add tests for segfault from mergedata

nickanderson:


nickanderson:
Something is busted with the travis vm and trying to install its broilerplate stuff.

PR #2936, opened 2017-10-30
Fix peerleaders test

nickanderson:
The test does not work if HOSTNAME is now downcased. CFEngine always
downcases the hostname and the fqdn must be found in the list in order
to determine peers.


nickanderson:
I saw the affected test passed in Travis. Merging.

PR #2940, opened 2017-10-31
Get the hostname from hostname -f

nickanderson:


nickanderson:
I see in the test output that check_outputs passed. merging.

PR #2943, opened 2017-11-02
Fix peerleaders examples on systems where hostname -f is not working.

jimis:


jimis:
Tested [here](https://ci.cfengine.com/job/testing-enterprise-pr/413/). I will also create a pull request for 3.10.x, as that branch was the one failing.

jimis:
Wut failed:

```
Q: "...r/bin/perl /hom": +R: Unable to find "(none)" or "(none)" in /tmp/cfe_hostlist. Can"t determine peers.
```

jimis:
It becomes apparent that for unknown reasons our hosts are broken and we have to fix the provisioned VMs.

PR #2944, opened 2017-11-02
Fix peerleaders examples on systems where hostname -f is not working.

jimis:


jimis:
To be tested [here](https://ci.cfengine.com/job/testing-enterprise-pr/414/).

PR #2947, opened 2017-11-14
Do not segfault if policy_server.dat only contains whitespaces and/or line breaks

amousset:


jimis:
Thanks everyone! Cherry-picked to 3.11.x, I don't think it applies to 3.10.x. @amousset do you agree?

amousset:
We found it on 3.10.2, so I guess it also applies to 3.10.x, probably at least in libpromises/policy_server.c in PolicyServerSet()

jimis:
Alright, since you have it fixed already can you send a pull request against 3.10.x? Thanks in advance!

jimis:
@amousset I could also cherry-pick the commit from your 3.10.x branch, if you point me there.

amousset:
It is not patched in Rudder yet and I did not have time to look at the 3.10 branch, I can probably do it next week.

jimis:
Thanks, much appreciated!

amousset:
@jimis 3.10 version: https://github.com/cfengine/core/pull/2958

PR #2949, opened 2017-11-17
CFE-2697: Improve misleading verbose message

jimis:
For constraints if/ifvarclass/unless, we now print the whole rval of the constraint.

Previously the message was just "skipping variable because ifvarclass is not defined" while the variable itself was defined.

Old message example:

  verbose: Skipping promise 'mailto' because 'if'/'ifvarclass' is not defined

Changed to:

   verbose: Skipping promise 'mailto' because constraint 'ifvarclass => not(isvariable("mailto"))' is not met

Changelog: commit


nickanderson:
+1
So much easier to understand.

jimis:
Tests finished, merging.

Deployment tests failed for spurious reasons (DO failing in nested virtualisation) but the status never got updated. Maybe something to look into, @Lex-2008.

PR #2957, opened 2017-11-27
ENT-3483: Do not tag large volatile data for inventory

nickanderson:
interfaces_data is commonly larger than the maximum data size allowed to be
collected by cf-hub. Data larger than 1k is truncated. Instead of reporting
truncated data this change stops tagging the variable so that it will not be
collected to the Enterprise hub.

Changelog: Title


olehermanse:
@jimis This fix should be merged :)

nickanderson:
and picked to supported branches too i believe.

nickanderson:
Squashed and re-worded.

I suggest backporting because the data is volatile causing extra load in the db and its truncated, so it may not even contain useful information and the delta is very small. 

But if you think it should not be backported you know better than I.

nickanderson:
@jimis  ok to merge?

jimis:
Merged and cherry-picked to 3.11.x and 3.10.x (patch does not apply to 3.7.x). Thanks!

PR #2959, opened 2017-11-29
CFE-2515: Add a dumpdatatstate() function

amousset:
This function writes the current datastate into the given file. It is already used in Rudder to implement Jinja2 templating.

I can add tests if this new function makes sense to you.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -3574,6 +3574,33 @@ static FnCallResult FnCallDatastate(EvalContext *ctx,
     return  (FnCallResult) { FNCALL_SUCCESS, (Rval) { state, RVAL_TYPE_CONTAINER } };
 }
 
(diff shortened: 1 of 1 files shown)


nickanderson:
This is basically rendering the multiline json representation of a datacontainer? and returns true to set a clas on success?

similar to this?

```
files:
  "/datastate.json"
  template_method => "inline_mustache",
  edit_template => "{{%-top-}}";
```

amousset:
Indeed, thanks! I did not realize we could achieve the same on 3.9+ whith `-top-`. This patch was initially implemented for CFEngine 3.6/3.7, because doing this required manipulating variables containing the datastate (which was too big). Closing this PR (I will replace this with mustache templating in the next Rudder release, and schedule the removal of the patch on our side!)

nickanderson:
Kewl

`%-top-` and `$-top-` are really useful for rendering JSON.

PR #2960, opened 2017-11-29
3.10.x: Update travis files to master

olehermanse:
I noticed that 3.10.x was very outdated compared to master, builds were failing because of issues fixed in master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
 # feedback ASAP. For more info read this:
 # https://docs.travis-ci.com/user/multi-os/
 matrix:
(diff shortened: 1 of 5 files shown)


olehermanse:
@jimis : Some of the commits missing are here: https://github.com/cfengine/core/pull/2911

But that's not all of them. Of course cherry picking is nicer than just overwriting, but I have no good way to find all the commits missing, maybe you can work your `git magic` ;)

olehermanse:
Damn editor updates

olehermanse:
@jimis Fixed

jimis:
Thanks!

PR #2962, opened 2017-11-29
ENT-3455: Improve logging of ACL errors

craigcomstock:
Changelog: Title

### DIFF
--- a/cf-agent/acl_posix.c
+++ b/cf-agent/acl_posix.c
@@ -105,7 +105,7 @@ static int CheckPosixLinuxDefaultACEs(EvalContext *ctx, Rlist *aces, AclMethod m
         retval = true;
         break;
 
-    case ACL_DEFAULT_SPECIFY:        // default ALC is specified in promise
+    case ACL_DEFAULT_SPECIFY:        // default ACL is specified in promise
 
         retval = CheckPosixLinuxACEs(ctx, aces, method, file_path, ACL_TYPE_DEFAULT, a, pp, result);
         break;
@@ -115,7 +115,7 @@ static int CheckPosixLinuxDefaultACEs(EvalContext *ctx, Rlist *aces, AclMethod m
         retval = CheckDefaultEqualsAccessACL(ctx, file_path, a, pp, result);
         break;
 
-    case ACL_DEFAULT_CLEAR:          // default ALC should be empty
+    case ACL_DEFAULT_CLEAR:          // default ACL should be empty
 
         retval = CheckDefaultClearACL(ctx, file_path, a, pp, result);
         break;
@@ -149,6 +149,7 @@ static int CheckPosixLinuxACEs(EvalContext *ctx, Rlist *aces, AclMethod method, 
     int has_mask;
     Rlist *rp;
     char *acl_type_str;
+    char *acl_text_str;
 
     acl_new = NULL;
     acl_existing = NULL;
@@ -161,7 +162,7 @@ static int CheckPosixLinuxACEs(EvalContext *ctx, Rlist *aces, AclMethod method, 
 
     if ((acl_existing = acl_get_file(file_path, acl_type)) == NULL)
     {
-        Log(LOG_LEVEL_ERR, "No ACL for '%s' could be read. (acl_get_file: %s)", file_path, GetErrorStr());
+        Log(LOG_LEVEL_ERR, "No %s ACL for '%s' could be read. (acl_get_file: %s)", acl_type_str, file_path, GetErrorStr());
         return false;
     }
 
@@ -170,15 +171,15 @@ static int CheckPosixLinuxACEs(EvalContext *ctx, Rlist *aces, AclMethod method, 
     if ((acl_tmp = acl_init(1)) == NULL)
     {
         Log(LOG_LEVEL_ERR, "New ACL could not be allocated (acl_init: %s)", GetErrorStr());
-        acl_free((void *) acl_existing);
+        acl_free(acl_existing);
         return false;
     }
 
     if (acl_create_entry(&acl_tmp, &ace_parsed) != 0)
     {
         Log(LOG_LEVEL_ERR, "New ACL could not be allocated (acl_create_entry: %s)", GetErrorStr());
-        acl_free((void *) acl_existing);
-        acl_free((void *) acl_tmp);
+        acl_free(acl_existing);
+        acl_free(acl_tmp);
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


craigcomstock:
@jimis @olehermanse maybe? for review?

cf-agent/acl_posix.c
```diff
@@ -339,15 +341,30 @@ static int CheckPosixLinuxACEs(EvalContext *ctx, Rlist *aces, AclMethod method,
 
             if (!DONTDO)
             {
+                int last = -1;
+                acl_text_str = acl_to_any_text(acl_new, NULL, ',', 0);
+                Log(LOG_LEVEL_VERBOSE, "ACL: new acl is `%s'", acl_text_str);
```
jimis: Shouldn't this be LOG_LEVEL_DEBUG?

cf-agent/acl_posix.c
```diff
@@ -339,15 +341,30 @@ static int CheckPosixLinuxACEs(EvalContext *ctx, Rlist *aces, AclMethod method,
...
+
+                if ((retv = acl_check(acl_new, &last)) != 0)
+                {
+                    Log(LOG_LEVEL_ERR, "ACL: acl_check(%s): %s at entry %d. Refer to acl_check man page for validity rules.",
+			    acl_text_str,
+			    acl_error(retv),
+			    last);
+		    acl_free(acl_text_str);
+		    return false;
```
jimis: Free the rest of `acl_*` variables?

craigcomstock:
oops... foobarred the rebase... working on it :)

craigcomstock:
`bundle agent main
{
    files:
        "./test.tmp"
        create => "true",
        acl => template;
}

body acl template
{
    acl_method => "overwrite";
    aces => {
        "user:*:rwx",
        "user:craig:rwx",
        "group:*:rwx",
        "group:craig:rwx",
        "all:rwx"
    };
}

~
`

`craig@other:~/cfengine/core/tests/acceptance$ ../../cf-agent/cf-agent -f ./test.cf
   error: ACL: acl_check(user:craig:rwx,group:craig:rwx,mask::rwx): Missing or wrong entry at entry 0. Refer to acl_check man page for validity rules.
   error: Failed checking access ACL on ./test.tmp
   error: Promise belongs to bundle 'main' in file './test.cf' near line 4
   error: ACL: acl_``

cf-agent/acl_posix.c
```diff
@@ -339,15 +341,33 @@ static int CheckPosixLinuxACEs(EvalContext *ctx, Rlist *aces, AclMethod method,
...
+                if ((retv = acl_check(acl_new, &last)) != 0)
+                {
+                    Log(LOG_LEVEL_ERR, "ACL: acl_check(%s): %s at entry %d. Refer to acl_check man page for validity rules.",
+			            acl_text_str,
+        			    acl_error(retv),
+        			    last);
+                    acl_free((void *) acl_existing);
+                    acl_free((void *) acl_tmp);
+                    acl_free((void *) acl_new);
+         		    acl_free(acl_text_str);
```
jimis: `acl_free()` is not for strings. Just use `free()`.

cf-agent/acl_posix.c
```diff
@@ -339,15 +341,33 @@ static int CheckPosixLinuxACEs(EvalContext *ctx, Rlist *aces, AclMethod method,
 
             if (!DONTDO)
             {
+                int last = -1;
+                acl_text_str = acl_to_any_text(acl_new, NULL, ',', 0);
+                Log(LOG_LEVEL_DEBUG, "ACL: new acl is `%s'", acl_text_str);
+
+                if ((retv = acl_check(acl_new, &last)) != 0)
+                {
+                    Log(LOG_LEVEL_ERR, "ACL: acl_check(%s): %s at entry %d. Refer to acl_check man page for validity rules.",
```
jimis: Try to improve indentation and wrap better, if possible at column 78 max. And make sure that messages higher than log level debug are human-readable. For example this one:

```
                    Log(LOG_LEVEL_ERR, "Invalid ACL in '%s' at index %d (acl_check: %s)"
                        acl_text_str, last, acl_error(retv)); 
```

cf-agent/acl_posix.c
```diff
@@ -339,15 +341,33 @@ static int CheckPosixLinuxACEs(EvalContext *ctx, Rlist *aces, AclMethod method,
...
+			            acl_text_str,
+        			    acl_error(retv),
+        			    last);
+                    acl_free((void *) acl_existing);
```
jimis: No casting for `void *` parameters, in C `void *` casting happens implicitly. It's a C++ thing that is not needed in C, and even considered harmful by some. I know the old code does it, but if possible change it.

cf-agent/acl_posix.c
craigcomstock: I took the hint from acl_to_text(3) ```This function allocates any memory necessary to contain the string and returns a pointer to the string.  The caller should free
     any releasable memory, when the new string is no longer required, by calling acl_free(3) with the (void*)char returned by
     acl_to_text() as an argument.
``` Same difference I suppose but it seems more proper to use acl_free()?

jimis:
You are right, ignore this comment then!

cf-agent/acl_posix.c
craigcomstock: Do you have a reference for reading so I can get some background? Something like this? https://gustedt.wordpress.com/2014/04/02/dont-use-casts-i/

cf-agent/acl_posix.c
jimis: You are right, ignore this comment then! (copied misplaced comment below)

cf-agent/acl_posix.c
jimis: First let me say that there are valid uses of casts. But when you use a cast you are basically telling the compiler "don't warn me of anything, I know what I'm doing". So in general it's good practice to keep the casts at a minimum and fix compiler warnings at their root cause, if possible (again, sometimes casts are needed).

The link you posted is quite good. There were also some more real problems with legacy C compilers: if you casted the return of a malloc() for example and had forgotten to include the proper header file, you might get a runtime error. Nice posts [here](https://stackoverflow.com/questions/1565496/specifically-whats-dangerous-about-casting-the-result-of-malloc).

craigcomstock:
updated messages:

body acl template
{
    acl_method => "overwrite";
    aces => {
#        "user:*:rwx",
        "user:craig:rwx",
#        "group:*:rwx",
        "group:craig:rwx",
#        "all:rwx"
    };
}

gives the following because you can't have a user:craig:rwx without the user:*:rwx entry that is required.

2017-12-06T11:16:04-0600    error: Invalid ACL in 'user:craig:rwx,group:craig:rwx,mask::rwx' at index 0 (acl_check: Missing or wrong entry)
2017-12-06T11:16:04-0600    error: Failed checking access ACL on ./test.tmp
2017-12-06T11:16:04-0600    error: Promise belongs to bundle 'main' in file './craig-acl-test.cf' near line 4

Still the error doesn't quite fully describe the problem but I suppose it is better. :(

PR #2963, opened 2017-12-04
CFE-2736: Fix libxml2 tests not being skipped

jimis:
This exports now properly environment variables to testall, so other
unidentified breakage should also be fixed.


jimis:
@kacf simple change but it touches your code, can you review? Even though I don't expect implications.

jimis:
Merging, deploytests have failed for spurious reasons but status never updated. Pinging @Lex-2008 for that.

PR #2968, opened 2017-12-14
CFE-1886: parser  can handle large input buffers

basvandervlies:
Changelog: Title

### DIFF
--- a/libpromises/cf3lex.l
+++ b/libpromises/cf3lex.l
@@ -82,14 +82,14 @@ space      [ \t]+
 
 newline    ([\n]|[\xd][\xa])
 
-line       ^.*$
+eat_line  ([\n]|[\xd][\xa]).*
 
 comment    #[^\n]*
 
-macro_if_version    ^@if\ minimum_version\([0-9]{1,10}\.[0-9]{1,10}(\.[0-9]{1,10})?\)
-macro_if_feature    ^@if\ feature\([a-zA-Z0-9_]+\)
-macro_endif ^@endif
-macro_line ^[^@\n\xd\xa].*$
+macro_if_version    @if\ minimum_version\([0-9]{1,10}\.[0-9]{1,10}(\.[0-9]{1,10})?\)
+macro_if_feature    @if\ feature\([a-zA-Z0-9_]+\)
+macro_endif @endif
+macro_line [^@\n\xd\xa].*
 
 promises   bundle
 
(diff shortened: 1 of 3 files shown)


nickanderson:
@jimis  for your consideration

jimis:
@bas can you summarise what you changed here and why? The only thing I understand is that you removed the REJECT statement and that removes the errors. But can you explain with what you replaced it? Preferably in the commit message.

bas:
@basvandervlies ☝️ 😄

jimis:
Another Bas!

basvandervlies:
On github I am basvandervlies ;-).  @jimis I removed the `REJECT` statement because we can not handle large lines.  I discuss it with @nickanderson and we still want to the "^" carrot error messages:
```
./bas.cf:1:6: error: syntax error
bundl agent main
     ^
./bas.cf:1:6: error: Expected 'bundle' or 'body' keyword, wrong input 'bundl'
```

So I replaced `REJECT` with `yless(1)` and due to this change we do not get a nice error message when it is in the first line of the file.  I had to rewrite some other regular expression because the beginning of a line as no meaning anymore  and we only allow the `@if macros` at the beginning a line.  For me as policy write I do not care where I place these statements. That is in short 
what all the changes are, see discussion:
 * https://tracker.mender.io/browse/CFE-1886

Another thing is we do not do any error recovering. So if there is an syntax error we could stop Immediately instead of stopping after reaching a certain threshold.  

It is a bit hard to explain in a few words

jimis:
Thanks @basvandervlies ! I triggered a [jenkins build](https://ci.cfengine.com/job/bootstrap-enterprise-pr/541/).

jimis:
So each and every line is read, strdup'ed into a variable, and then REJECTED. And this happens in order to be able to print the full line in case of an error.

So you keep the same logic, but replace REJECT with `yyless()`.

It seems that there are [other strategies](https://www.ibm.com/developerworks/library/l-flexbison/index.html) we can follow for better error reporting. Just for your future consideration.

I have a better understanding now, thanks for explaining. I'll squash and edit your commit message to add your comment, before I merge.

basvandervlies:
@jimis I have read the IBM article.  This is a better approach I think

tests/unit/parser_test.c
```diff
@@ -243,6 +242,7 @@ int main()
         unit_test(test_bundle_body_forget_cb_bundle),
         unit_test(test_body_edit_line_common_constraints),
         unit_test(test_body_edit_xml_common_constraints),
+        unit_test(test_bundle_body_promisee_missing_arrow),
```
jimis: Can you add a short comment in the code about why this has to be last?

libpromises/cf3lex.l
```diff
@@ -135,20 +135,23 @@ varclass       (\"[^"]*\"|\'[^']*\')::
 promise_type   [a-zA-Z_]+:
 
 %%
-
-{line}                {
+{eat_line}             {
```
jimis: Can you add a comment here, that explains that `eat_line` is only used to store the line for error reporting, and all line is returned back and re-lexed ?

jimis:
Jenkins build was fully green so I went ahead and merged, after adding a bit more info in the commit message. Thanks @basvandervlies ! 

P.S. the nitpicks about missing comments in the review are optional, feel free to contribute a separate commit if you like.

PR #2970, opened 2017-12-15
CFE-2084: Improve multiple augments logging

nickanderson:
This change makes it more apparent that an additional augments file has
completed its processing rather than seeming like it was processed
multiple times.


nickanderson:
@basvandervlies  @tzz 

Any objections to this minor logging change? I think it makes it a bit more clear.

basvandervlies:
@nickanderson no problem at my side

PR #2978, opened 2018-01-02
Remove / comment on various #includes in cf-serverd

olehermanse:
@jimis

### DIFF
--- a/cf-serverd/cf-serverd-functions.h
+++ b/cf-serverd/cf-serverd-functions.h
@@ -25,28 +25,11 @@
 #ifndef CFENGINE_CF_SERVERD_FUNCTIONS_H
 #define CFENGINE_CF_SERVERD_FUNCTIONS_H
 
-
 #include <platform.h>
-
 #include <generic_agent.h>
-#include <server.h>
-
 #include <eval_context.h>
-#include <dir.h>
-#include <dbm_api.h>
-#include <lastseen.h>
-#include <crypto.h>
-#include <files_names.h>
-#include <vars.h>
-#include <promises.h>
-#include <item_lib.h>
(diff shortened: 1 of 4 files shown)


olehermanse:
@jimis I'm making smaller improvements to cf-serverd for cf-testd. Please review.

cf-serverd/cf-serverd-functions.h
```diff
@@ -25,29 +25,11 @@
 #ifndef CFENGINE_CF_SERVERD_FUNCTIONS_H
 #define CFENGINE_CF_SERVERD_FUNCTIONS_H
 
-
-#include <platform.h>
```
jimis: You need this. `platform.h` should always be first (or a header file that directly includes it first is also OK.)

olehermanse: Doesn't really matter, but first line is `generic_agent.h` which includes `cf3.defs.h`.

olehermanse:
https://ci.cfengine.com/view/all/job/bootstrap-enterprise-pr/530/

cf-serverd/cf-serverd-functions.h
jimis: It matters, because the pattern is that you find it included either at the top of your-file.c, or at the top of your-file.h. Before it used to as you say, that it includes some file that includes some file, that includes it at the top, but it was impossible to enforce and subtle breakage happened. Feel free to document this in HACKING.md. ;-)

olehermanse:
Jenkins is green except for HPUX (still running) and RHEL 4 (unrelated issue):
https://ci.cfengine.com/view/all/job/testing-enterprise-pr/539/

PR #2980, opened 2018-01-05
Added script for ordering json files during tests

olehermanse:
Example usage:
```
python3 misc/json-order.py test1.json test2.json && diff test1.json.out.json test2.json.out.json
```

### DIFF
--- a/misc/json-order.py
+++ b/misc/json-order.py
@@ -1,0 +1,24 @@
+import sys
+import json
+from collections import OrderedDict
+
+def dict_sort(old_dict):
+    if isinstance(old_dict, list):
+        return [dict_sort(x) for x in old_dict]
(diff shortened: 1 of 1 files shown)


olehermanse:
@jimis @vpodzime

vpodzime:
Looks good to me. It could be really helpful in implementing something like *dcs_if_diff_expected_json()* similar to [*dcs_if_diff_expected()*](https://github.com/cfengine/core/blob/master/tests/acceptance/dcs.cf.sub#L496] which could then be used in the tests to make them more stable/reliable.

nickanderson:
Nice, agree with @vpodzime that a new dcs_if_diff_expected() for json would be ncie.

Would be even nicer if the agent did canonical json natively. But not sure if any performance penalty is worth it.

PR #2981, opened 2018-01-05
CFE-2341: Clarify FILE argument vs --file and -f options for cf-runagent

nickanderson:


cf-runagent/cf-runagent.c
```diff
@@ -124,7 +124,7 @@ static const char *const HINTS[] =
     "Output verbose information about the behaviour of the agent",
     "All talk and no action mode - make no changes, only inform of promises not kept",
     "Output the version of the software",
-    "Specify an alternative input file than the default",
+    "Specify an alternative input file than the default. This option is overridden by FILE if supplied as argument.",
```
jimis: Isn't this the same as in cf-agent? This message was the same with the one there.

nickanderson: Yes it is. So now I have updated the string in the other component help files.

PR #2984, opened 2018-01-10
CFE-1459: Suppress output from systemctl based restart of services in bootstrap/failsafe

nickanderson:
When the agent is bootstrapped from an unprivledged user using sudo systemctl
may complain about not being able to access the a tty.

notice: Q: ".../systemctl rest": Failed to open /dev/tty: No such device or address

This is because cfengine runs commands in their own sessions.

This change simply suppresses the command output to avoid distracting operators.

Changelog: Title

### DIFF
--- a/libpromises/failsafe.cf
+++ b/libpromises/failsafe.cf
@@ -226,8 +226,9 @@ bundle agent failsafe_cfe_internal_update
       # is running, even if cf-execd isn't, for example. Here we want to be
       # sure we relaunch everything.
 
-      "/bin/systemctl restart cfengine3"
+      "/bin/systemctl restart cfengine3" -> { "CFE-1459" }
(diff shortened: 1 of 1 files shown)


nickanderson:
@jimis do you see any issues with suppressing the command output here?

For context. If you sudo cf-agent --bootstrap on a systemd manged host systemctl will error and complain about a missing tty.

jimis:
Looks good to me. I don't understand why systemctl needs /dev/tty, maybe @kacf knows?

kacf:
Can't say that I do, no. :-(

nickanderson:
systemctl supposedly wants /dev/tty for stdout and stderr, at least according to a comment in the ticket. I will go ahead and merge then. Thanks.

PR #2985, opened 2018-01-11
Travis: Overwrite OSX system gcc link

olehermanse:
Testing

PR #2986, opened 2018-01-12
Travis: Don't do set -e in .travis.yml

olehermanse:
See: 
https://github.com/travis-ci/travis-ci/issues/6307
#2985

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,17 +44,12 @@ matrix:
       sudo: false
 
 before_install:
-  - set -e
   - chmod ug+x ./travis-scripts/*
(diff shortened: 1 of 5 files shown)


olehermanse:
See also:
https://github.com/travis-ci/docs-travis-ci-com/issues/1672

BanzaiMan of Travis-CI says:
> RVM is not the only thing that can break when set errexit is set in .travis.yml. Just don't do it.

So I changed .travis.yml (for all platforms)

olehermanse:
@jimis @Lex-2008 Please review.

PR #2989, opened 2018-01-15
Don't segfault if a JSON policy file contains no bundles and bodies

vpodzime:
Somebody may accidentally run cf-agent or cf-promises on a JSON
file that is not a valid policy file in the JSON format.

### DIFF
--- a/libpromises/loading.c
+++ b/libpromises/loading.c
@@ -111,6 +111,15 @@ Policy *Cf3ParseFile(const GenericAgentConfig *config, const char *input_path)
         }
 
         policy = PolicyFromJson(json_policy);
+        if (policy == NULL)
+        {
+            Log(LOG_LEVEL_ERR,
+                "Failed to deserialize a policy from the JSON input file '%s'",
+                input_path);
+            JsonDestroy(json_policy);
+            WriterClose(contents);
+            return NULL;
+        }
 
         JsonDestroy(json_policy);
         WriterClose(contents);
--- a/libpromises/policy.c
+++ b/libpromises/policy.c
@@ -2290,17 +2290,24 @@ Policy *PolicyFromJson(JsonElement *json_policy)
 {
     Policy *policy = PolicyNew();
 
+    JsonElement *json_bundles = JsonObjectGetAsArray(json_policy, "bundles");
+    JsonElement *json_bodies = JsonObjectGetAsArray(json_policy, "bodies");
+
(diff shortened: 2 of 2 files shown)


jimis:
So what's happening now if somebody does that? It doesn't segfault but does what? Also you need a `Changelog: Title` tag in the commit message.

vpodzime:
> So what's happening now if somebody does that? It doesn't segfault but does what? Also you need a Changelog: Title tag in the commit message.

``Changelog: Title`` added. And the code now returns ``NULL`` as a policy, logs an error and later terminates. Something like this:
```
   error: Failed to deserialize a policy from the JSON input file '/tmp/test_policy.json'
   error: There are syntax errors in policy files
   error: Policy failed validation with command '"/home/vpodzime/.cfagent/bin/cf-promises" -c "/tmp/test_policy.json"'
   error: Failsafe condition triggered. Interactive session detected, skipping failsafe.cf execution.
   error: Error reading CFEngine policy. Exiting...
```

libpromises/policy.c
```diff
@@ -2290,22 +2290,29 @@ Policy *PolicyFromJson(JsonElement *json_policy)
...
-            JsonElement *json_bundle = JsonArrayGetAsObject(json_bundles, i);
-            PolicyAppendBundleJson(policy, json_bundle);
-        }
+        return NULL;
     }
 
+    if (json_bundles != NULL)
+    {
+        for (size_t i = 0; i < JsonLength(json_bundles); i++)
+            {
```
jimis: wrong brace indent :-)

vpodzime:
Comments resolved.

libpromises/loading.c
```diff
@@ -111,6 +111,14 @@ Policy *Cf3ParseFile(const GenericAgentConfig *config, const char *input_path)
         }
 
         policy = PolicyFromJson(json_policy);
+        if (policy == NULL)
+        {
+            Log(LOG_LEVEL_ERR, "Failed to deserialize a policy from the JSON "
+                "input file '%s'", input_path);
```
jimis: Try to wrap but not break string literals, because it reduces "grepability" of the codebase. I think there are some relevant guidelines in HACKING.md.

vpodzime: Good point, done.

jimis:
Looks good, I'll merge after Travis is green, this doesn't look for something needing to wait for Jenkins.

nickanderson:
Merged, travis was green.

PR #2990, opened 2018-01-15
ENT-3562: New StringToLong functions

olehermanse:
Must be merged together:
https://github.com/cfengine/core/pull/2990
https://github.com/cfengine/enterprise/pull/359
https://github.com/cfengine/nova/pull/1073

### DIFF
--- a/libutils/string_lib.c
+++ b/libutils/string_lib.c
@@ -372,13 +372,149 @@ bool EmptyString(const char *s)
 
 /*********************************************************************/
 
-long StringToLong(const char *str)
+/**
+ * @brief Converts a string of numerals in base 10 to a long integer.
+ *
+ * Result is stored in *value_out, return value should be checked.
+ * On error *value_out is unmodified and an error code is returned.
+ * Leading spaces in input string are skipped.
+ * String numeral must be terminated by NULL byte or space (isspace).
+ *
+ * @see StringToLongExitOnError()
+ * @see StringToLongDefaultOnError()
+ * @param[in] str String with numerals to convert, cannot be NULL
+ * @param[out] value_out Where to store result on success, cannot be NULL
+ * @return 0 on success, error code otherwise (see source code)
+ */
+int StringToLong(const char *str, long *value_out)
+{
+    assert(str != NULL);
+    assert(value_out != NULL);
+    assert(ERANGE != 0);
+
+    char *endptr = NULL;
+    long val;
+
+    errno = 0;
+    val = strtol(str, &endptr, 10);
+
+    if ((errno == ERANGE && (val == LONG_MAX || val == LONG_MIN)))
+    {
+        return ERANGE; // Overflow or underflow
+    }
+
+    if (endptr == str)
+    {
+        return -81; // No digits found
+    }
+
+    if (endptr == NULL)
+    {
+        return -82; // endpointer not set by strtol
+    }
+
+    if (*endptr != '\0' && !isspace(*endptr))
+    {
+        return -83; // string not properly terminated
+    }
+
+    if (errno != 0)
+    {
+        return errno; // Unknown error
+    }
+
+    *value_out = val;
+    return 0;
+}
+
+/**
+ * @brief Log StringToLong conversion error with an identifier for debugging
+ */
+void LogStringToLongError(const char *str_attempted, const char *id, int error_code)
+{
+    assert(error_code != 0);
+    const char *error_str = "Unknown";
+    switch (error_code)
+    {
+        case ERANGE:
+            error_str = "Overflow";
+            break;
+        case -81:
+            error_str = "No digits";
+            break;
+        case -82:
+            error_str = "No endpointer";
+            break;
+        case -83:
+            error_str = "Not terminated";
+            break;
+    }
+    Log(LOG_LEVEL_ERR, "Conversion error (%d - %s) on '%s' (%s)", error_code, error_str, str_attempted, id);
+}
+
+/**
+ * @brief Converts a string of numerals in base 10 to a long integer,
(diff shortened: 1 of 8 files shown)


olehermanse:
Jenkins:
https://ci.cfengine.com/view/all/job/bootstrap-enterprise-pr/549/

olehermanse:
@jimis @craigcomstock Can you please review these 3 pull requests. Might include some subtle change in behavior, good to get a second (and third) opinion.

libutils/string_lib.c
```diff
@@ -372,13 +372,140 @@ bool EmptyString(const char *s)
...
+ * Leading spaces in input string are skipped.
+ * String numeral must be terminated by NULL byte or space (isspace).
+ *
+ * @see StringToLongExitOnError()
+ * @see StringToLongDefaultOnError()
+ * @param[in] str String with numerals to convert, cannot be NULL
+ * @param[out] number Where to store result on success, cannot be NULL
+ * @return 0 on success, error code otherwise (see source code)
+ */
+int StringToLong(const char *str, long *number)
```
jimis: I would suggest to rename `number` to `value_out` in both the C and header file so that it's obvious this is an "out" parameter.

libutils/string_lib.c
```diff
@@ -372,13 +372,140 @@ bool EmptyString(const char *s)
 
 /*********************************************************************/
 
-long StringToLong(const char *str)
+/**
+ * @brief Converts a string of numerals in base 10 to a long integer.
+ *
+ * Result is stored in *number, return value should be chacked.
```
jimis: `chacked` typo

libutils/string_lib.c
```diff
@@ -372,13 +372,140 @@ bool EmptyString(const char *s)
...
+ * On error *number is unmodified and -1 is returned.
```
jimis: Not -1, but "an error code is returned".

I'm not sure if this is the best way though. You might as well return bool type and `false` on any error. Still I'm split on this, so error code is OK.

libutils/string_lib.h
```diff
@@ -59,7 +59,12 @@ char ToUpper(char ch);
 void ToUpperStrInplace(char *str);
 void ToLowerStrInplace(char *str);
 
-long StringToLong(const char *str);
+int StringToLong(const char *str, long *number) FUNC_WARN_UNUSED_RESULT;
+void LogConversionError(const char *str_attempted, const char *id);
```
jimis: Why export this function. Do you use it in nova? If not, why even have a function?

libutils/string_lib.c
```diff
@@ -372,13 +372,140 @@ bool EmptyString(const char *s)
...
+/**
+ * @brief Convert a string of numerals to a long integer (deprecated).
+ *
+ * @warning This function is deprecated, do not use it
+ * @warning This function can exit the process based on string contents
+ * @see StringToLongExitOnError()
+ * @param[in] str String with numerals to convert, cannot be NULL
+ * @return Result of conversion
+ */
+long StringToLongUnsafe(const char *str)
```
jimis: I thought you replaced this everywhere? Or where there too many occurences after all?

libutils/string_lib.h
olehermanse: Yes it is used in nova. To avoid code duplication.

libutils/string_lib.c
olehermanse: Thanks, fixed

libutils/string_lib.c
olehermanse: I like int better because it's easier to debug, also makes it easier to check for integer overflow for example.

libutils/string_lib.c
olehermanse: Thanks, done.

libutils/string_lib.c
olehermanse: If you think all of the places where this function has been replaced with new functions (with slightly differing behavior) are good, we can remove the old version.

olehermanse:
@jimis I have added your suggestions and started a new jenkins build. (Link in previous comment updated). As you know, I'm exhausting the jenkins queue daily.

libutils/string_lib.c
```diff
@@ -372,13 +372,140 @@ bool EmptyString(const char *s)
...
+    *value_out = val;
+    return 0;
+}
+
+/**
+ * @brief Log a conversion error with an identifier for debugging
+ */
+void LogConversionError(const char *str_attempted, const char *id)
+{
+    Log(LOG_LEVEL_ERR, "Conversion error on '%s' (%s)", str_attempted, id);
```
vpodzime: Would be nice if this also took the return code and printed the respective error message based on it.

olehermanse: That would make it specific to StringToLong, but I agree. I'll add this.

libutils/string_lib.c
jimis: Since you are triggering builds like crazy, why don't you add a commit that removes it and trigger another, to see if/what breaks? :-)

olehermanse: Nothing will break if I remove it, it is unused.

I am asking you to review the places where I've made changes by replacing the old StringToLong with new, different, functions.

jimis: There are very few places in this pull request and I checked them, they look OK.

olehermanse:
@jimis @vpodzime @craigcomstock I believe I have addressed all your comments. I'll start one more build in jenkins now, believing this is ready to merge.

https://ci.cfengine.com/view/all/job/bootstrap-enterprise-pr/552/

olehermanse:
Green build - merging.

PR #2991, opened 2018-01-16
CFE-2748: Issue a warning on ignored locking attributes

vpodzime:
According to the recently updated documentation on promise
locking, there are several promise types that don't support any
form of promise locking:

  * access
  * classes
  * defaults
  * meta
  * roles
  * vars

When a policy writer uses "ifelapsed" or "expireafter" attributes
(via an action body) on these promise types which don't support
them, a warning should be issued rather than the setting being
silently ignored.

Codewise, it would be great if every promise evaluation called
GetTransactionConstraints(), that way we could only have the
check and warning in one place. Unfortunately, that's not the
case, VerifyVarPromise() is not calling that particular function
and neither is handling of 'access' and 'roles' in cf-serverd.

Changelog: Title

### DIFF
--- a/cf-serverd/server_transform.c
+++ b/cf-serverd/server_transform.c
@@ -687,6 +687,28 @@ static PromiseResult KeepServerPromise(EvalContext *ctx, const Promise *pp, ARG_
         return VerifyClassPromise(ctx, pp, NULL);
     }
 
+    /* Warn if promise locking was used with a promise that doesn't support it.
+     * That applies to both of the below promise types while 'vars' and
+     * 'classes' handle this on their own. */
+    int ifelapsed = PromiseGetConstraintAsInt(ctx, "ifelapsed", pp);
+    if (ifelapsed != CF_NOINT)
+    {
+        Log(LOG_LEVEL_WARNING,
+            "ifelapsed attribute specified in action body for %s promise '%s',"
+            " but %s promises do not support promise locking",
+            pp->parent_promise_type->name, pp->promiser,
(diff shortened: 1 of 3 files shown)


cf-serverd/server_transform.c
```diff
@@ -687,6 +687,26 @@ static PromiseResult KeepServerPromise(EvalContext *ctx, const Promise *pp, ARG_
...
+    /* Warn if promise locking was used with a promise that doesn't support it.
+     * That applies to both of the below promise types while 'vars' and
+     * 'classes' handle this on their own. */
+    int ifelapsed = PromiseGetConstraintAsInt(ctx, "ifelapsed", pp);
+    if (ifelapsed != CF_NOINT)
+    {
+        Log(LOG_LEVEL_WARNING, "ifelapsed attribute specified in action "
+            "body for %s promise '%s', but %s promises do not support "
+            "promise locking.", pp->parent_promise_type->name,
+            pp->promiser, pp->parent_promise_type->name);
```
jimis: Do not break string literals, they harm grepability of the code. Try to break on punctuation or % qualifiers. I would write this one as:

```
Log(LOG_LEVEL_WARNING, 
    "ifelapsed attribute specified in action body for %s promise '%s',"
    " but %s promises do not support promise locking", 
    pp->parent_promise_type->name, pp->promiser, pp->parent_promise_type->name);
```

It's ok if string literals go a bit over 78 column. 

Also do not end log messages with punctuation.

vpodzime:
@jimis changed.

PR #2998, opened 2018-01-20
fix detection of libacl headers

crorvick:
Commit 41c0afd ("Improve readability") moved the error handling for a
missing libacl header into AC_CHECK_HEADER's action-if-not-found.  This
was a mistake, though, as this action is executed for each header not
found, which was not the intent of the original check.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -507,8 +507,11 @@ then
       [],
       [if test "x$with_libacl" != xcheck; then AC_MSG_ERROR(Cannot find libacl library); fi])
     AC_CHECK_HEADERS([acl.h sys/acl.h acl/libacl.h],
-      [libacl_header_found=yes],
-      [if test "x$with_libacl" != xcheck; then AC_MSG_ERROR(Cannot find libacl header files); fi])
+      [libacl_header_found=yes])
+    if test "x$libacl_header_found" != "xyes" && test "x$with_libacl" != xcheck;
+    then
+      AC_MSG_ERROR(Cannot find libacl library headers);
+    fi
   ])
 fi
 


jimis:
> 41c0afd ("Improve readability") moved the error handling for a missing libacl header into AC_CHECK_HEADER's action-if-not-found. This was a mistake, though, as this action is executed for each header not found. In other words, unless the first header (i.e., "libacl.h") is found the configure will fail claiming "Cannot find libacl header files".

The configure check should fail in case any of the  `acl.h sys/acl.h acl/libacl.h` is not found. Is that not the case now?

configure.ac
```diff
@@ -507,8 +507,11 @@ then
       [],
       [if test "x$with_libacl" != xcheck; then AC_MSG_ERROR(Cannot find libacl library); fi])
     AC_CHECK_HEADERS([acl.h sys/acl.h acl/libacl.h],
-      [libacl_header_found=yes],
-      [if test "x$with_libacl" != xcheck; then AC_MSG_ERROR(Cannot find libacl header files); fi])
+      [libacl_header_found=yes])
+    if test "x$libacl_header_found" != "xyes" && test "x$with_libacl" != xcheck;
```
jimis: You move the check outside the AC_CHECK_HEADERS macro, which means that the check will likely ignore the two first headers, only checking the last one, because `libacl_header_found` is set 3 times, once for each header file.

Can you explain what the issue is exactly so that we figure a better way to fix it? What error are you seeing? Which header files are present on your system?

crorvick: > You move the check outside the AC_CHECK_HEADERS macro

This is how it was done prior to 41c0afd ("Improve readability"), I'm just reverting that portion of the commit.  It seems pretty clear that commit did not intend to change the semantics of that check, so this is a regression.

> Can you explain what the issue is exactly so that we figure a better way to fix it? What error are you seeing?
```
$ ./configure --with-libacl
[SNIP]
checking for acl_init in -lacl... yes
checking acl.h usabilty... no
checking acl.h presense... no
checking for acl.h... no
configure: error: Cannot find libacl header files
```
> Which header files are present on your system?
```
ls /usr/include/acl.h /usr/include/sys/acl.h /usr/include/acl/libacl.h
/usr/include/sys/acl.h
/usr/include/acl/libacl.h
```

crorvick:
> The configure check should fail in case any of the acl.h sys/acl.h acl/libacl.h is not found. Is that not the case now?

What distro are you running and what package is providing `<acl.h>`, and at what path?  Everything works fine with only `<sys/acl.h>` and `<acl/libacl.h>` on my system (Gentoo) and AFAICT that is all that Fedora provides as well.

jimis:
Thanks, it makes more sense now, thanks! Will test again and see if reverting makes sense.

jimis:
I can reproduce it now and I understand what's wrong. I've taken your commit, and amended the code and the commit message with a small comment in #3032, will merge that one. Thanks!

PR #2999, opened 2018-01-22
Cherry-picks for 3.10.x

olehermanse:
MUST pass jenkins before merge.


olehermanse:
@nickanderson These were the commits which could be picked cleanly. Please review that we agree on the commits.

@vpodzime This PR contains one Hashmap fix from you. Should this fix be applied to old versions? (See commit https://github.com/cfengine/core/pull/2999/commits/9fcc66ed5a94e15e593a043ee4e219df3f713d5f)

vpodzime:
> @vpodzime This PR contains one Hashmap fix from you. Should this fix be applied to old versions? (See commit 9fcc66e)

This commit is just a code clarity/quality fix and has no impact on the resulting binary. It's safe but the only purpose of it on other branches would be to prevent conflicts if we ever decide to backport all the hash map changes. Which I don't think is going to happen. So NACK for this commit from me.

olehermanse:
@vpodzime Removed hashmap related commit.

jimis:
This 99328fa changes behaviour, not good for stable release.

jimis:
This 9faf9f4 changes log message, and we know that some people are parsing them in production, so it's also a change in behaviour.

jimis:
Rest look OK.

olehermanse:
@jimis: Reverted 2 commits in https://github.com/cfengine/core/pull/3010

PR #3002, opened 2018-01-24
Improvements and parallelism for cf-net

vpodzime:
The first few commits are just minor fix-ups and refactoring for *cf-net*. The last two ones add support for getting files in parallel (in threads). I tried playing with the number of threads and everything up to ~250 was working fine for me.

### DIFF
--- a/cf-net/cf-net.c
+++ b/cf-net/cf-net.c
@@ -228,7 +228,7 @@ static char *RequireHostname(char *hostnames)
         char *policy_server = PolicyServerReadFile(GetWorkDir());
         if (policy_server == NULL)
         {
-            printf("Error: no host name (and no policy_server.dat)");
+            printf("Error: no host name (and no policy_server.dat)\n");
             exit(EXIT_FAILURE);
         }
         return policy_server;
@@ -304,7 +304,7 @@ static int CFNetParse(int argc, char **argv,
             }
             default:
             {
-                printf("Default optarg = '%s', c = '%c' = %i",
+                printf("Default optarg = '%s', c = '%c' = %i\n",
                        optarg, c, (int)c);
                 break;
             }
@@ -347,7 +347,7 @@ static int CFNetCommandSwitch(CFNetOptions *opts, const char *hostname,
         default:
             break;
     }
-    printf("Bug: CFNetCommandSwitch() is unable to pick a command");
+    printf("Bug: CFNetCommandSwitch() is unable to pick a command\n");
     exit(EXIT_FAILURE);
     return -1;
 }
@@ -698,7 +698,7 @@ static int CFNetGet(ARG_UNUSED CFNetOptions *opts, const char *hostname, char **
             }
             default:
             {
-                printf("Default optarg = '%s', c = '%c' = %i",
+                printf("Default optarg = '%s', c = '%c' = %i\n",
                        optarg, c, (int)c);
                 break;
             }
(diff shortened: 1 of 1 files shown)


olehermanse:
@vpodzime I will review this when release work is done

cf-net/cf-net.c
```diff
@@ -704,6 +721,17 @@ static int CFNetGet(ARG_UNUSED CFNetOptions *opts, const char *hostname, char **
                 specified_path = true;
                 break;
             }
+            case 'j':
+            {
+                char *tail = NULL;
+                n_threads = (int) strtol(optarg, &tail, 10);
```
jimis: I believe @olehermanse has a new function for this purpose.

cf-net/cf-net.c
```diff
@@ -704,6 +721,17 @@ static int CFNetGet(ARG_UNUSED CFNetOptions *opts, const char *hostname, char **
...
+                if (tail && (*tail != '\0'))
+                {
+                    printf("Failed to parse number of threads/jobs: %m");
```
jimis: forgot the terminating LF. ;-)

cf-net/cf-net.c
```diff
@@ -740,9 +768,62 @@ static int CFNetGet(ARG_UNUSED CFNetOptions *opts, const char *hostname, char **
         return -1;
     }
 
-    int ret = CFNetGetFile(hostname, remote_file, local_file);
+    CFNetThreadData **threads = (CFNetThreadData**) xcalloc(n_threads, sizeof(CFNetThreadData*));
+    for (int i = 0; i < n_threads; i++)
+    {
+        threads[i] = (CFNetThreadData*) xcalloc(1, sizeof(CFNetThreadData));
+        threads[i]->data = (GetFileData*) xcalloc(1, sizeof(GetFileData));
```
jimis: no casting is needed nor recommended in all of the above, in C void* casting is implicit.

cf-net/cf-net.c
```diff
@@ -740,9 +768,62 @@ static int CFNetGet(ARG_UNUSED CFNetOptions *opts, const char *hostname, char **
...
+        threads[i]->data->hostname = hostname;
+        threads[i]->data->remote_file = remote_file;
+        if (n_threads > 1)
+        {
+            if (strstr(local_file, "%d") != NULL)
+            {
+                snprintf(threads[i]->data->local_file, PATH_MAX, local_file, i);
```
jimis: dangerous but I guess it's ok for a testing tool

cf-net/cf-net.c
olehermanse: @vpodzime @jimis `StringToLongExitOnError()` from `string_lib.h` would be appropriate. `StringToLong()` if you want to do custom error handling. (I plan to do some changes, like actually returning long on platforms where `long != int`, but they will not be breaking.)

cf-net/cf-net.c
vpodzime: Yeah, there's practically no chance that ``local_file`` will not be ``\0``-terminated.

jimis: it's not for the NULL termination, it's for the possibility of putting format string qualifiers in the filename. Imagine a filename "blah%s%x%d".

vpodzime: Oh, that's right. Well, we are all responsible people not doing things like this. :)

jimis:
Acceptance test or even a load test (in `tests/load`). But since this is a testing tool for now, I am lenient as to a test exists or not.

vpodzime:
> But since this is a testing tool for now, I am lenient as to a test exists or not.

IMHO the best way to test a test tool is to run tests with it.

jimis:
> IMHO the best way to test a test tool is to run tests with it.

Well that's your main task AFAIK, but feel free to include a small load test in Core, the more the better!

olehermanse:
@vpodzime Please tell me if you want this merged now.

vpodzime:
> @vpodzime Please tell me if you want this merged now.

That would be great. I plan to add some core-limited tests based on this, but that can happen in a separate PR once I get back to it. Thanks @olehermanse!

PR #3013, opened 2018-02-02
zendesk#3739: Load process table when actually needed for a processes…

vpodzime:
… promise

Process table loading is not a particularly cheap operation and
thus should only run when really needed. Instead of running it
when the 'processes' type context is entered, run it when some
'processes' promise is actually evaluated -- it may be skipped
due to class restrictions.

Note that LoadProcessTable() has a caching mechanism and unless
the cache is invalidated by ClearProcessTable(), it is a noop. By
keeping ClearProcessTable() call intact this change should not
change behavior.

Changelog: Title
(cherry picked from commit ef111df2a736de17481078ff83eef71034f66a56)


vpodzime:
jenkins job: https://ci.cfengine.com/job/pr-pipeline/159/

PR #3015, opened 2018-02-06
When there is an parse error quit after 3 errors instead of 12

basvandervlies:
Changelog: Title

### DIFF
--- a/libpromises/cf3parse.y
+++ b/libpromises/cf3parse.y
@@ -1258,7 +1258,7 @@ static void ParseErrorVColumnOffset(int column_offset, const char *s, va_list ap
 
     }
 
-    if (P.error_count > 12)
+    if (P.error_count > 3)
     {
         fprintf(stderr, "Too many errors\n");
         exit(EXIT_FAILURE);


jimis:
@basvandervlies the parser test is failing. Do

```
cd tests/unit
make parser_test
./parser_test
```

Does it work for you?

basvandervlies:
@jimis. I just saw this. It has more impact then I thought. The test program quits because due this statement in `cf3parse.y`
```
    if (P.error_count > 3)
    {   
        fprintf(stderr, "Too many errors\n");
        exit(EXIT_FAILURE);
    }
```

The test program can not handle this correcly. That is why we have the value 12 which is insane.

jimis:
Can you fix the parser test?

basvandervlies:
@jimis the `parser_test` will exit due the above `exit...` statement in `cf3parse.y` can I catch this exit code or must I make a lot different parser test files? I am not that familiar with the `assert_false` stuff.

jimis:
I see: the tested function `exit()`s instead of returning false, so the `assert_false()` never gets to verify the return value.

It's tricky and I can't think of any straightforward fix. Maybe have a global variable `CF_PARSER_MAX_NUMBER_OF_ERRORS = 3`. At the beginning of the test this could be changed to 12? Dirty but should do the trick...

olehermanse:
@basvandervlies Are you able to update the parser test to reflect this change?

basvandervlies:
@olehermanse I did not look at the parser test code and the coming 2 weeks I am on vacation

olehermanse:
Closing since it's been more than a year. Feel free to open a new PR if you get back to this.

PR #3016, opened 2018-02-07
CFE-2310: Add IPv6 hard classes with the "ipv6_" prefix

vpodzime:
Hard classes for the IPv4 addresses have the "ipv4_" prefix so
the IPv6 ones should use the same pattern. However, we shouldn't
just prefix the existing hard classes as users may rely on them
being available without the prefix. Thus add new ones.

Changelog: Title

### DIFF
--- a/libenv/unix_iface.c
+++ b/libenv/unix_iface.c
@@ -55,6 +55,8 @@
 #define CF_IFREQ 2048           /* Reportedly the largest size that does not segfault 32/64 bit */
 #define CF_IGNORE_INTERFACES "ignore_interfaces.rx"
 
+#define IPV6_PREFIX "ipv6_"
+
 #ifndef __MINGW32__
 
 # if defined(HAVE_STRUCT_SOCKADDR_SA_LEN) && !defined(__NetBSD__)
@@ -657,9 +659,13 @@ static void FindV6InterfacesInfo(EvalContext *ctx)
 
                 if ((IsIPV6Address(ip->name)) && ((strcmp(ip->name, "::1") != 0)))
                 {
+                    char prefixed_ip[CF_MAX_IP_LEN + sizeof(IPV6_PREFIX)] = {0};
                     Log(LOG_LEVEL_VERBOSE, "Found IPv6 address %s", ip->name);
                     EvalContextAddIpAddress(ctx, ip->name, NULL); // interface unknown
                     EvalContextClassPutHard(ctx, ip->name, "inventory,attribute_name=none,source=agent");
+
+                    xsnprintf(prefixed_ip, sizeof(prefixed_ip), IPV6_PREFIX"%s", ip->name);
+                    EvalContextClassPutHard(ctx, prefixed_ip, "inventory,attribute_name=none,source=agent");
                 }
             }
 
--- a/tests/acceptance/02_classes/01_basic/ipaddr_classes.cf
+++ b/tests/acceptance/02_classes/01_basic/ipaddr_classes.cf
@@ -1,0 +1,34 @@
+body common control
+{
+      inputs => { "../../default.cf.sub" };
+      bundlesequence  => { default("$(this.promise_filename)") };
+      version => "1.0";
+}
+
+bundle agent check
+{
+  vars:
+      "ipv6_classes" slist => classesmatching("ipv6_.*");
+
+  classes:
+      "has_ipv6" expression => returnszero("ifconfig -a | grep 'inet6 [a-z0-9].*'", "useshell");
+
+    has_ipv6::
+      "ok" expression => some(".*", "ipv6_classes");
+
+    !has_ipv6::
+      "ok" expression => "any";
+
+  reports:
+    ok::
+      "$(this.promise_filename) Pass";
+
+    !ok::
+      "$(this.promise_filename) FAIL";
+
+    DEBUG::
+      "ipv6_classes: $(ipv6_classes)";
+
+    DEBUG.has_ipv6::
+      "Has IPv6";
+}


vpodzime:
Any ideas how to test this?

jimis:
`cf-promises --show-classes` shows all classes set from environment detection. This could be good enough for manual testing.

jimis:
Also `cf-agent --verbose` will show when the classes get set.

libenv/unix_iface.c
```diff
@@ -54,6 +54,8 @@
 
 #define CF_IFREQ 2048           /* Reportedly the largest size that does not segfault 32/64 bit */
 #define CF_IGNORE_INTERFACES "ignore_interfaces.rx"
+#define IPV6_PREFIX "ipv6_"
+#define PREFIXED_IPV6_LEN (32 + 7 + sizeof(IPV6_PREFIX))
```
jimis: There is a `CF_MAX_IP_LENGTH` somewhere I believe.

libenv/unix_iface.c
```diff
@@ -657,9 +659,23 @@ static void FindV6InterfacesInfo(EvalContext *ctx)
...
+
+                    strcpy(prefixed_ip, "ipv6_");
+                    ret = strlcat(prefixed_ip + (sizeof(IPV6_PREFIX) - 1), ip->name, PREFIXED_IPV6_LEN);
+                    if (ret <= PREFIXED_IPV6_LEN)
+                    {
+                        EvalContextClassPutHard(ctx, prefixed_ip, "inventory,attribute_name=none,source=agent");
+                    }
+                    else
+                    {
+                        Log(LOG_LEVEL_WARNING, "Failed to prefix IPv6 address  %s", ip->name);
```
jimis: Use `:` delimiter before the final `%s`. And do you expect this ever happening? I believe the buffer should never overflow, in which case a simple `assert()` after `strlcat()` should be enough.

BTW you might want to have a look at `StrCat()`. It's supposed to be a faster alternative of `strlcat()` for when you know all lengths. Not sure if it makes sense to use it here.

vpodzime:
Yeah, I used ``cf-promises --show-classes`` for manual testing.

nickanderson:
niiice

nickanderson:
probably @neilhwatson  will enjoy this

libenv/unix_iface.c
vpodzime: Good point. It's a bit larger than needed, but it probably doesn't matter much:
```
#define CF_MAX_IP_LEN 64
```

neilhwatson:
🍰

libenv/unix_iface.c
vpodzime: Yeah, ``assert()`` should be enough here. I replaced the multiple steps with a single ``snprintf()`` call, it's just easier to read and the efficiency should be the same.

vpodzime:
I'll add some tests tomorrow.

libenv/unix_iface.c
```diff
@@ -657,9 +661,15 @@ static void FindV6InterfacesInfo(EvalContext *ctx)
 
                 if ((IsIPV6Address(ip->name)) && ((strcmp(ip->name, "::1") != 0)))
                 {
+                    char prefixed_ip[PREFIXED_IPV6_LEN] = {0};
+                    size_t ret;
                     Log(LOG_LEVEL_VERBOSE, "Found IPv6 address %s", ip->name);
                     EvalContextAddIpAddress(ctx, ip->name, NULL); // interface unknown
                     EvalContextClassPutHard(ctx, ip->name, "inventory,attribute_name=none,source=agent");
+
+                    ret = snprintf(prefixed_ip, PREFIXED_IPV6_LEN, "ipv6_%s", ip->name);
```
jimis: `xsnprintf()` is for you! It combines `snprintf()` and `assert()`!

vpodzime: Ha, nice! I thought we only have these variants for functions that allocate memory. :)

vpodzime: Fixed.

libenv/unix_iface.c
```diff
@@ -55,6 +55,10 @@
 #define CF_IFREQ 2048           /* Reportedly the largest size that does not segfault 32/64 bit */
 #define CF_IGNORE_INTERFACES "ignore_interfaces.rx"
 
+#define IPV6_PREFIX "ipv6_"
+/* 8 groups of 4 characters + 7 colons as separators + the prefix (incl. the NUL byte)*/
+#define PREFIXED_IPV6_LEN (32 + 7 + sizeof(IPV6_PREFIX))
```
jimis: Why didn't you go with `CF_MAX_IP_LEN` after all? There are some pathological cases of IPv6 addresses that are larger than 32+7 (like IPv4 mapped IPv6 address), and it's better to keep such things in one place.

vpodzime: It was in an attempt to catch those pathological cases. But it's probably better to just avoid them. So replaced with ``CF_MAX_IP_LEN``.

vpodzime:
Jenkins build for this PR: https://ci.cfengine.com/job/pr-pipeline/77/

libenv/unix_iface.c
```diff
@@ -657,9 +657,13 @@ static void FindV6InterfacesInfo(EvalContext *ctx)
 
                 if ((IsIPV6Address(ip->name)) && ((strcmp(ip->name, "::1") != 0)))
                 {
+                    char prefixed_ip[CF_MAX_IP_LEN] = {0};
```
jimis: @vpodzime I think the proper one would be `CF_MAX_IP_LEN + sizeof("ipv6_")`

jimis:
test looks good!

libenv/unix_iface.c
vpodzime: It doesn't really matter because ``CF_MAX_IP_LEN`` is 64 for some unspecified reason, but yeah, this code shouldn't blindly rely on the current definition. Fixed.

vpodzime:
@jimis ready for merging, now, I believe. Well, if travis is green, of course.

libenv/unix_iface.c
```diff
@@ -657,9 +659,13 @@ static void FindV6InterfacesInfo(EvalContext *ctx)
 
                 if ((IsIPV6Address(ip->name)) && ((strcmp(ip->name, "::1") != 0)))
                 {
+                    char prefixed_ip[CF_MAX_IP_LEN + sizeof(IPV6_PREFIX)] = {0};
                     Log(LOG_LEVEL_VERBOSE, "Found IPv6 address %s", ip->name);
                     EvalContextAddIpAddress(ctx, ip->name, NULL); // interface unknown
                     EvalContextClassPutHard(ctx, ip->name, "inventory,attribute_name=none,source=agent");
+
+                    xsnprintf(prefixed_ip, CF_MAX_IP_LEN, IPV6_PREFIX"%s", ip->name);
```
jimis: You forgot to update this. Better use `sizeof(prefixed_ip)`. I tend to always use `sizeof()` instead of the length directly, in order to avoid this kind of mishaps.

vpodzime: Oh, that's a nice trick. it didn't occur to me that the compiler knows the size even when it's just a pointer in the end.

vpodzime:
@jimis ready for merging, now, I yet again believe.

jimis:
Great, thanks @vpodzime !

PR #3017, opened 2018-02-08
Added note about error codes in return values to HACKING.md

olehermanse:
Also fixed some markdown syntax.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -13,8 +13,8 @@ mailing lists. You'll find us chatting on Freenode's IRC channels
 Normally, bug fixes have a higher chance of getting accepted than new
 features, but we certainly welcome feature contributions. If you have an idea
 for a new feature, it might be a good idea to open up a feature ticket in
-our [bug-tracker](https://tracker.mender.io/projects/CFE/issues/) and send a 
-message to dev-cfengine mailing list, before actually contributing the code, 
+our [bug-tracker](https://tracker.mender.io/projects/CFE/issues/) and send a
+message to dev-cfengine mailing list, before actually contributing the code,
 in order to get discussion going.
 
 Merged features and larger changes will be released in the first minor release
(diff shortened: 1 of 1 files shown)


olehermanse:
@vpodzime @jimis

olehermanse:
Relevant addition:

* In functions which can fail, error code (int) or success/failure (bool) should be returned.
    * `true`(bool) and `0`(int) should always signify success.
    * Only return an error code (int) when there are multiple different return values for different errors. If a function can only return `0` (success) or `-1` (error) use `bool` instead.
    * Compiler can enforce checking of return value, output of function can be in an output parameter (pointer).

HACKING.md
```diff
@@ -203,25 +203,29 @@ Coding Style
...
     ```c
     do
     {
-        /* ... */
+        // ...
 
     } while (condition);
     ```
+* In functions which can fail, error code (int) or success/failure (bool) should be returned.
+    * `true`(bool) and `0`(int) should always signify success.
```
jimis: zero for success in int functions might be a bit dangerous, because zero is `false`. In the end it's just a convention so I'm OK, I'm not sure if I would use it personally.

HACKING.md
```diff
@@ -203,25 +203,29 @@ Coding Style
...
+    * Only return an error code (int) when there are multiple different return values for different errors. If a function can only return `0` (success) or `-1` (error) use `bool` instead.
```
jimis: Many times it's useful to combine error code and success return value in the same return number. I know I have coded several functions in the past that return `size_t` since it's about memory sizes, but in case of error they return `(size_t) -1`. Just keep that in mind while changing stuff.

olehermanse: This is true, I might add a point about this. Return value like `recv()` where positive integers are valid and -1 means error is totally acceptable.

HACKING.md
olehermanse: I agree, but `success = 0` seems the most standard. It is already used many places in our codebase, and fits in nicely with exit codes, where success is also 0.

PR #3019, opened 2018-02-08
Add an option to specify RSA key size to cf-key

amousset:
First draft of a new option to set the size of the key generated by cf-key. Do you think "-b" is ok for this?

### DIFF
--- a/cf-key/cf-key.c
+++ b/cf-key/cf-key.c
@@ -49,6 +49,7 @@ const char *remove_keys_host = NULL;                            /* GLOBAL_A */
 static char *print_digest_arg = NULL;                           /* GLOBAL_A */
 static char *trust_key_arg = NULL;                              /* GLOBAL_A */
 static char *KEY_PATH = NULL;                                   /* GLOBAL_A */
+static int KEY_SIZE = 2048;                                     /* GLOBAL_A */
 bool LOOKUP_HOSTS = true;                                       /* GLOBAL_A */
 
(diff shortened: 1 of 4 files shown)


jimis:
Thanks!

How about `-T` or `--key-type`? For now it will only be a number signifying the RSA key size. In the future it could be more complicated like `ecdsa:256` etc. @nickanderson what do you think?

nickanderson:
--key-type seems OK to me. I don't really care for single letter options.

amousset:
Changed option to `key-type`.

cf-key/cf-key.c
```diff
@@ -249,6 +252,10 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
             KEY_PATH = optarg;
             break;
 
+        case 'T':
+            KEY_SIZE = atoi(optarg);
```
jimis: You need some error checking here. @olehermanse what is the function you suggest here?

olehermanse: `KEY_SIZE = StringToLongExitOnError(optarg);`

amousset:
Also changed `atoi` to `StringToLongExitOnError` in cf-runagent (from where it was initially copied).

olehermanse:
Thanks @amousset. I will run this through our build/test infrastructure now.

nickanderson:
@olehermanse the pr pipeline you referenced notes 33,689 tests 0 failed.

PR #3020, opened 2018-02-08
Fix segfault when no show-evaluated-vars/classes is specified

amousset:
When running cf-agent (compiled on Fedora 27), I get a segfault in pcre because of no initial value for the `show_classes` and `show_variables` regexes.

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -272,16 +272,16 @@ int main(int argc, char *argv[])
     PurgeLocks();
     BackupLockDatabase();
 
-    if (config->agent_specific.common.show_classes != NULL)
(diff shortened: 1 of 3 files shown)


amousset:
Would you prefer duplicate variables for AGENT and COMMON types, or a non agent-specific variable?

jimis:
Separate variables would be best, set only in the respective sections of
the "switch" statement.

amousset:
Updated with separate variables.

amousset:
Updated.

jimis:
Thanks @amousset ! Do you know if it needs to be cherry-picked to 3.10 ?

amousset:
No, it only appeared in 3.11.0.

jimis:
Alright, thanks for the fix!

PR #3022, opened 2018-02-09
Create cf-diag.sh

mikeweilgart:
This is a standalone diagnostic shell script that should run properly on ANY system that can run CFEngine.
It is intended for operator use in quickly diagnosing a system.  However, it uses "test anything protocol" so can
also be used from any "harness" designed for that.  (https://testanything.org/)

One test ("cf-agent ran in last 15 minutes") may falsely fail on systems without GNU coreutils (I depend on "stat"
for that test), but otherwise I would expect this script to work on a wide variety of OSes.


olehermanse:
@nickanderson Please review

contrib/cf-diag.sh
```diff
@@ -0,0 +1,240 @@
...
+
+desc="Expected Enterprise version"
+if printf '%s\n' "$enterprise_version" | grep -qFx $ok_versions; then
+  succeed "$desc"
+else
+  fail "$desc"
+fi
+
+desc="Core version matches Enterprise version"
+# In pathological cases these may not match.
```
nickanderson: +1

contrib/cf-diag.sh
```diff
@@ -0,0 +1,240 @@
...
+    fail "$desc"
+  fi
+done
+
+# Test for recent run:
+
+# 15 minutes is three times the out-of-the-box default run interval (5 min)
+# However, as noted by Neil Watson
+# (in this ticket: https://tracker.mender.io/browse/CFE-769),
+# most large CFEngine installations use longer (less frequent) run intervals.
```
nickanderson: s/most/some/

I am actually quite surprised at the number of large installs that do run a 5-minute interval. They use other methodologies to do parts of the policies at different intervals.

contrib/cf-diag.sh
```diff
@@ -0,0 +1,240 @@
...
+if [ -f /var/cfengine/policy_server.dat ]; then
+  succeed "$desc"
+  hub_ip="$(cat /var/cfengine/policy_server.dat)"
+  printf '# Hub IP is %s\n' "$hub_ip"
+else
+  fail "$desc"
+  bail 'Not connected to any hub.'
+fi
+
+desc="Hub is connecting to host"
```
nickanderson: This one is most useful in the context where you have an enterprise agent capable of serving reports. Also, it's not useful if the host is doing call collect/client initiated reporting, instead of the default hub initiated collection.

In the MPF there is augments support for setting the `client_initiated_reporting_enabled` class and the `control_server_call_collect_interval` variable. It might be useful to check/report these in the same place.
`cf-promises -c --show-classes` and `cf-promises -c --show-vars` (note in 3.12 these can take regex values)

contrib/cf-diag.sh
```diff
@@ -0,0 +1,240 @@
...
+if grep -q cf-diag_comment /var/cfengine/inputs/promises.cf; then
+  fail "$desc"
+else
+  succeed "$desc"
+fi
+
+############################### CUSTOM TESTS ##################################
+
+# Add custom tests here (if you really need them).
+# Use the general form:
```
nickanderson: Humm, seems like a good place to be looking for a directory full of tests matching some pattern.

Splitting the tests out into directories full of a kind of test might be intersting as well.
Then you could add notes about adding a `.gitlab-ci.yml` spec which runs some of the scripts as tests on your policy repo.

nickanderson:
I like it, but can you please move it to cf-diag/cf-diag.sh.

Would be nice to be accompanied by some kind of README.org or README.md

contrib/cf-diag.sh
```diff
@@ -0,0 +1,240 @@
...
+#
+#   desc="Name of test (descriptive string)"
+#   if some-command-that-should-succeed; then
+#     succeed "$desc"
+#   else
+#     fail "$desc"
+#   fi
```
Lex-2008: Nice stuff! But what about replacing `if (condition); then succeed "$desc" else fail "$desc"; fi` with something like 
```
(condition)
check $? "$desc"
```
where
```
check() {
  if [ $1 = 0 ]; then
    succeed "$2"
  else
    fail "$2"
  fi
}
```
You can also add third optional argument to bail if the test fails, and/or add `return $1` at the end to make it chain-able like this:
```
check $? "$desc" || bail "Fatal error"
```

mikeweilgart: I did consider that approach.  However, not all tests actually follow that pattern.  Some tests are conditional on the success or failure of a different test, as you can see.

If there's one thing I try to avoid like the plague in shell scripts, it's building up frameworks that don't actually match the requirements.  Shell scripts are simple enough that they don't *need* wrappers.

Lex-2008: Yes, like nested `if`s...

But maybe for simple cases?

contrib/cf-diag.sh
mikeweilgart: I did consider that.  But really, the purpose of this script is emphatically NOT to be a general-purpose testing framework.  Customized tests are not the point.  A correctly sequenced battery of standard universally applicable (or close to that) tests, is the point.

If I wanted to write a whole shell scripting testing framework, I would have just reinvented BATS.  But that's not the point of this script.  This script is intended to be ultra-lightweight, simple, standalone, and to work even on very old servers (e.g. RHEL 4).  And is intended to give a quick overview of the CFEngine health factors on the system.  And really, nothing beyond that.

contrib/cf-diag.sh
mikeweilgart: Interesting.  But, any check for those variables would ignore all pathological cases: host has no policy at all, host has invalid policy, host has a different policy from the hub (since the hub would have to be told not to initiate the reporting checks, something that's only visible from the host if you assume they're running the same policy).

Note that in the sequence written, this check comes long before the check for the *existence* of promises.cf - not even mentioning that promises.cf would have to be runnable for cf-promises to work on it, and also not mentioning that such a test would have to assume the standard MPF is more or less in use.

If call collect is in use, the operators in that environment would presumably be aware of that and would simply ignore the failing test or add a SKIP exception for that test.

PR #3023, opened 2018-02-12
Fix clang compiler warnings

olehermanse:


olehermanse:
@jimis Please review. This is the first round of warnings, from just building with clang, no scan-builder.

cf-serverd/server_classic.c
```diff
@@ -1083,7 +1083,7 @@ int BusyWithClassicConnection(EvalContext *ctx, ServerConnectionState *conn)
     case PROTOCOL_COMMAND_AUTH_PLAIN:
         SetConnectionData(conn, (char *) (recvbuffer + strlen("CAUTH ")));
 
-        if (conn->username == NULL || !IsUserNameValid(conn->username))
```
jimis: This is not bad code, just overcautious code. Now the compiler will optimise this away. But if in the future `username` becomes a pointer from an array, this check will be needed.

jimis:
Good job, can you introduce a clang parallel build with clang?

olehermanse:
Added 3 more fixes as separate commits. These were reported by clang scan-builder.

A new clang job has been added to Travis. I also checked that it fails if I introduce a warning on purpose: https://travis-ci.org/cfengine/core/builds/340506930

Jenkins build: https://ci.cfengine.com/view/all/job/pr-pipeline/76/

@jimis Please review and merge when green.

olehermanse:
@jimis It's green.

libutils/unix_dir.c
```diff
@@ -194,8 +194,10 @@ static size_t GetDirentBufferSize(size_t name_len)
 
 struct dirent *AllocateDirentForFilename(const char *filename)
 {
-    struct dirent *entry = xcalloc(1, GetDirentBufferSize(strlen(filename)));
+    int length = strlen(filename);
+    struct dirent *entry = xcalloc(1, GetDirentBufferSize(length));
 
-    strcpy(entry->d_name, filename);
+    // d_name is fixed length, but we have allocated extra space using xcalloc:
+    strncpy((char *)entry->d_name, filename, length + 1);
```
jimis: `strncpy` is generally considered bad practice for string copying, slow and easy to get it wrong. Here all you need is `memcpy` or `strcpy`, or if you want to be extra pedantic go with `strlcpy`.

jimis: The "fix" here is basically the cast, so add to your comment that this cast is used to silence compiler warnings.

travis-scripts/script.sh
```diff
@@ -39,7 +39,7 @@ then
     make CFLAGS=-Werror
 elif [ "$JOB_TYPE" = compile_and_unit_test ]
 then
-    make CFLAGS=-Werror  &&
+    make CFLAGS="-Werror -Wno-pointer-sign -Wno-duplicate-decl-specifier"  &&
```
jimis: Can you add an explanation to the commit message why it's OK to skip these two? Or maybe some TODO comment in the travis file to fix that in the future?

libutils/unix_dir.c
olehermanse: Thanks, used memcpy instead.

travis-scripts/script.sh
olehermanse: Thanks, added.

olehermanse:
https://ci.cfengine.com/view/all/job/pr-pipeline/80/

olehermanse:
Builds look green, merging.

jimis:
+1

PR #3024, opened 2018-02-15
CFE-2524: Variable table copying amousset cfe2524

jimis:
Continues from where #2818 was.


jimis:
If all tests go fine, I'll squash and merge with your commit @amousset if that's OK with you.

jimis:
Simplified your testcase in the last commit, @amousset .

jimis:
Wow all ENT tests are green! Please have a look before I squash as much as possible.

ncharles:
<3 thank you @jimis @amousset

amousset:
ncf tests are green too!

libpromises/expand.c
```diff
@@ -256,7 +256,8 @@ PromiseResult ExpandPromise(EvalContext *ctx, const Promise *pp,
      *    (including body inheritance). */
     Promise *pcopy = DeRefCopyPromise(ctx, pp);
 
-    EvalContextStackPushPromiseFrame(ctx, pcopy, true);
+    // TODO: Remove last parameter?
```
amousset: we can remove this TODO

amousset: we can remove this TODO

jimis: removing it, thanks

PR #3025, opened 2018-02-15
I got buffer overflow crashes due to the pipe being -1 and borking FD…

craigcomstock:
…_SET()

@olehermanse  any thoughts?

### DIFF
--- a/cf-testd/cf-testd.c
+++ b/cf-testd/cf-testd.c
@@ -414,6 +414,9 @@ int CFTestD_StartServer()
     }
 
     int sd = InitServer(10);
+
+    MakeSignalPipe();
+
     int selected = 0;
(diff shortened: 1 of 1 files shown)


olehermanse:
Weird. Does this commit fix it? How did you compile and run?

craigcomstock:
Yeah, this commit fixes it. I did autogen.sh; make; on core, enterprise (not needed right?) and nova. I did this on ubuntu 14.04.5. Then I ran as root in nova/tests/reporting "bash run.sh" with those adjustments in my other PR for core.

olehermanse:
enterprise is definitely needed. The reporting code which is the whole purpose of cf-testd is in the enterprise plugin.

PR #3027, opened 2018-02-16
CFE-2310: Add IPv6 hard classes with the "ipv6_" prefix

vpodzime:
Hard classes for the IPv4 addresses have the "ipv4_" prefix so
the IPv6 ones should use the same pattern. However, we shouldn't
just prefix the existing hard classes as users may rely on them
being available without the prefix. Thus add new ones.

Changelog: Title
(cherry picked from commit 307dd25e2379765724c334e8b4278f5c56e20ce1)

### DIFF
--- a/libenv/unix_iface.c
+++ b/libenv/unix_iface.c
@@ -53,6 +53,8 @@
 #define CF_IFREQ 2048           /* Reportedly the largest size that does not segfault 32/64 bit */
 #define CF_IGNORE_INTERFACES "ignore_interfaces.rx"
 
+#define IPV6_PREFIX "ipv6_"
+
 #ifndef __MINGW32__
 
 # if defined(HAVE_STRUCT_SOCKADDR_SA_LEN) && !defined(__NetBSD__)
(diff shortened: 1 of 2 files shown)


vpodzime:
https://ci.cfengine.com/job/pr-pipeline/104/

vpodzime:
All green on the supported configurations.

libenv/unix_iface.c
```diff
@@ -648,9 +650,13 @@ static void FindV6InterfacesInfo(EvalContext *ctx)
 
                 if ((IsIPV6Address(ip->name)) && ((strcmp(ip->name, "::1") != 0)))
                 {
+                    char prefixed_ip[CF_MAX_IP_LEN + sizeof(IPV6_PREFIX)] = {0};
                     Log(LOG_LEVEL_VERBOSE, "Found IPv6 address %s", ip->name);
                     EvalContextAddIpAddress(ctx, ip->name);
                     EvalContextClassPutHard(ctx, ip->name, "inventory,attribute_name=none,source=agent");
+
+                    xsnprintf(prefixed_ip, sizeof(prefixed_ip), IPV6_PREFIX"%s", ip->name);
+                    EvalContextClassPutHard(ctx, prefixed_ip, "inventory,attribute_name=none,source=agent");
```
nickanderson: My first instinct was that we should inventory the ipv6 addresses like we do the ipv4 ones. But I believe we make that choice in policy. We copy the variable and tag the copy for inventory.

PR #3030, opened 2018-02-20
No expected unittest failures on AIX anymore

vpodzime:
With the new linker flags introduced in cfengine/buildscripts#328
no unittests should be failing on AIX.


jimis:
Two tests fixed, unexpected bonus!

PR #3031, opened 2018-02-20
coverage fixes

jimis:


jimis:
Coverage was reduced because I removed a bunch of non-deterministic tests on purpose. Merging.

PR #3033, opened 2018-02-21
Hashes in file paths in filesexist()

vpodzime:
Adding a test in the first commit because it turns out things just work as [expected/documented](https://docs.cfengine.com/docs/3.10/reference-functions-filesexist.html). The other commit is a related minor fix-up.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -6942,48 +6942,44 @@ static FnCallResult FnCallStringSplit(ARG_UNUSED EvalContext *ctx,
 
 static FnCallResult FnCallFileSexist(EvalContext *ctx, ARG_UNUSED const Policy *policy, ARG_UNUSED const FnCall *fp, const Rlist *finalargs)
 {
-    char naked[CF_MAXVARSIZE];
-    char *listvar = RlistScalarValue(finalargs);
+    bool allocated = false;
+    JsonElement *json = VarNameOrInlineToJson(ctx, fp, finalargs, false, &allocated);
 
-    if (IsVarList(listvar))
-    {
-        GetNaked(naked, listvar);
-    }
-    else
-    {
-        Log(LOG_LEVEL_VERBOSE, "Function filesexist was promised a list called '%s' but this was not found", listvar);
-        return FnFailure();
-    }
-
-    VarRef *ref = VarRefParse(naked);
-    DataType input_list_type;
-    const Rlist *input_list = EvalContextVariableGet(ctx, ref, &input_list_type);
-    VarRefDestroy(ref);
-    if (input_list_type == CF_DATA_TYPE_NONE)
+    // we failed to produce a valid JsonElement, so give up
+    if (json == NULL)
     {
         Log(LOG_LEVEL_VERBOSE,
-            "Function filesexist was promised a list called"
-            " '%s' but this was not found", listvar);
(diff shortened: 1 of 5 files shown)


vpodzime:
Jenkins job: https://ci.cfengine.com/job/pr-pipeline/130/

vpodzime:
Related modification for the documentation: https://github.com/cfengine/documentation/pull/1735

vpodzime:
jenkins job: https://ci.cfengine.com/job/pr-pipeline/133/

tests/acceptance/02_classes/02_functions/empty_list_or_data_doesnt_return_true_for_filesexist.cf
```diff
@@ -10,10 +10,6 @@ bundle agent test
       "description"
```
nickanderson: Since we are yanking the soft fail which documented the issue number I think we should add a promisee to the description.

```
"description" -> { "CFE-2631" }
```

vpodzime: > Since we are yanking the soft fail which documented the issue number I think we should add a promisee to the description.

Good point. Added.

vpodzime:
@jimis and/or @olehermanse please review the C part so that this can be merged. [The jenkins job](https://ci.cfengine.com/job/testing-pr/132/) is all green except for some spurious package upload failure on Debian 7.

libpromises/fncall.c
```diff
@@ -271,19 +271,19 @@ static FnCallResult CallFunction(EvalContext *ctx, const Policy *policy, const F
 
     if (argnum != RlistLen(expargs) && !(fncall_type->options & FNCALL_OPTION_VARARG))
     {
-        fprintf(stderr, "Argument template mismatch handling function %s(", fp->name);
+        Log(LOG_LEVEL_ERR, "Argument template mismatch handling function %s(", fp->name);
```
jimis: This won't work, Log() always prints whole lines. Try compositing the string and printing it in the end.

libpromises/evalfunction.c
```diff
@@ -6942,48 +6942,45 @@ static FnCallResult FnCallStringSplit(ARG_UNUSED EvalContext *ctx,
...
+        JsonDestroyMaybe(json, allocated);
         return FnFailure();
     }
 
-    for (const Rlist *rp = input_list; rp != NULL; rp = rp->next)
+    JsonIterator iter = JsonIteratorInit(json);
+    const JsonElement *el = JsonIteratorNextValueByType(&iter, JSON_ELEMENT_TYPE_PRIMITIVE, true);
+
+    /* no elements mean 'false' should be returned, otherwise let's see if the files exist */
+    ret = el != NULL;
```
jimis: Declare `bool ret` on this line. We use C99 where it makes sense. EDIT: no need to check ret here it seems. Just initiatlise it to true, and name it accordingly (like `file_found`).

libpromises/fncall.c
vpodzime: Good catch, fixed.

libpromises/evalfunction.c
vpodzime: Done. We need to check if ``el != NULL`` because ``filesexist({})`` needs to be ``false``.

vpodzime:
@jimis ready for a re-review.

PR #3034, opened 2018-02-24
Test that executing user has valid ppkey

nickanderson:


nickanderson:
@mikeweilgart  yay or nay?

contrib/cf-diag/cf-diag.sh
```diff
@@ -53,9 +53,22 @@ trap 'printf "1..%d\n" "$test_number"' EXIT
 # I recommend custom checks be added only at the end.
 # Skip ahead to CUSTOM TESTS.
 
-# Show host key (digest)
-printf '# '
-cf-key -p /var/cfengine/ppkeys/localhost.pub 2>/dev/null || echo SHA not available
+desc="Executing user has valid ppkey"
+if ! [ $(id -u) = 0 ]; then
```
vpodzime: I think using ``!=`` would be better readable here.

mikeweilgart:
I would say nay.  Many/most of the later checks will fail given a non-root user; I don't see any real advantage to adding a lot of complicated logic to account for non-root user, anyway.  If it were to be done, it should be done on *all* tests—and that's definitely going in the direction of scope creep and, what's far worse, complexity creep.

What might be useful is to add a test right at the beginning that will fail and bail if the script is being run by a non-root user.  But even then, I don't really see a case for it.  Some of the tests *will* succeed even with a non-root user, and some will fail, and each will be reported accordingly...so, I don't see any problem.  Maybe the tests that do succeed will still be informative (better than nothing).

Okay, what would be a good idea, rather than a fail and bail, would be a warning (printed in a comment in the output, as allowed by Test Anything Protocol) if the script is run as a non-root user.  So something like:

```
# Note: this script is running as a non-root user; some tests will therefore fail.
```

contrib/cf-diag/cf-diag.sh
mikeweilgart: Or just use `=` or better yet, `-eq`.  It's not as though it matters which path goes in "then" and which goes in "else."

PR #3041, opened 2018-03-13
Added SeqStringLength function from ent

olehermanse:
Renamed this function from ENT.

### DIFF
--- a/libutils/sequence.c
+++ b/libutils/sequence.c
@@ -25,6 +25,7 @@
 
 #include <sequence.h>
 #include <alloc.h>
+#include <string_lib.h>
 
 
 static const size_t EXPAND_FACTOR = 2;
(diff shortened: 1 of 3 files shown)


libutils/sequence.c
```diff
@@ -392,3 +392,17 @@ Seq *SeqStringFromString(const char *str, char delimiter)
...
     return seq;
 }
+
+int SeqStringLength(Seq *seq)
+{
+    assert (seq);
+
+    int total_length = 0;
+
+    for (size_t i = 0; i < SeqLength(seq); i++)
```
vpodzime: We should avoid calling the ``SeqLength`` function over and over in every iteration of the cycle.

olehermanse: Yep

olehermanse:
https://ci.cfengine.com/view/all/job/pr-pipeline/174/

PR #3042, opened 2018-03-14
ENT-3526: extend reporting tests

craigcomstock:
- added ability to specify report contents via option
- flushed stdout more often to make logging work better for tests

merge with https://github.com/cfengine/nova/pull/1113


cf-testd/cf-testd.c
```diff
@@ -230,7 +234,41 @@ bool CFTestD_GetServerQuery(
         return false;
     }
 
-    return CFTestD_ReturnQueryData(conn, query);
+    if (config->report_len == 0) {
```
olehermanse: Style: Open brace on separate line. Other places as well. I recommend clang-format ;)

olehermanse:
@craigcomstock I left some general comments and questions on both PRs. I'll test this locally tomorrow!

craigcomstock:
@olehermanse fwiw, I can't remove the WIP tag. Not sure why.

craigcomstock:
@olehermanse I ran clang-format on cf-testd.c. Lots of change there. Not sure if I like formatting changes along-side functional changes so maybe I can roll this back to two commits? Thoughts?

olehermanse:
@craigcomstock For running first time on a file there will naturally be quite a few changes. Yes I definitely prefer formatting changes in a separate commit, just like any other whitespace changes. For this PR, I can review it either way, so it's up to you whether you want to split it or not. 

Also note that you don't have to use all the clang-format changes, you can just skip the ones you don't like. (`git add -p` is your friend). You can also disable formatting for parts of a file:

```
void main(void)
{
	int a;
	// clang-format off
	// No formatting here :)
	char a,     b;
	char alpha, beta;
	// clang-format on
}
```

craigcomstock:
@olehermanse yeah, for this one I'd say just review it as-is. I have some more changes coming for nova but consider this core PR stable as-is so let's go with it. Read up on git add -p and will give it a try next time.

craigcomstock:
I couldn't stand leaving this like it was. So I am submitting a new commit with first the whitespace/formatting changes, then the functional changes.

cf-testd/cf-testd.c
```diff
@@ -1,7 +1,8 @@
 /*
    Copyright 2018 Northern.tech AS
 
-   This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+   This file is part of CFEngine 3 - written and maintained by Northern.tech
+  AS.
```
vpodzime: Weird indentation here.

cf-testd/cf-testd.c
```diff
@@ -24,62 +25,60 @@
...
-#include <signals.h>                    // ReloadConfigRequested
-#include <openssl/err.h>                // ERR_get_error
+#include <server_tls.h>       // ServerTLSInitialize
+#include <server_transform.h> // Summarize
+#include <signals.h>          // ReloadConfigRequested
+#include <sysinfo.h>          // DetectEnvironment
+#include <systype.h>          // CLASSTEXT
+#include <time_classes.h>     // UpdateTimeClasses
+#include <timeout.h>          // SetReferenceTime
+#include <tls_generic.h>      // TLSLogError
```
vpodzime: I liked this better before.

cf-testd/cf-testd.c
```diff
@@ -89,12 +88,12 @@ static const char *const HINTS[] =
...
     "Output the version of the software",
-    NULL
-};
+    NULL};
 
 CFTestD_Config *CFTestD_ConfigInit()
 {
-    CFTestD_Config* r = (CFTestD_Config *)(xcalloc(1, sizeof(CFTestD_Config)));
+    CFTestD_Config *r =
+        (CFTestD_Config *)(xcalloc(1, sizeof(CFTestD_Config)));
```
vpodzime: I think this is ugly, but YMMV.

cf-testd/cf-testd.c
```diff
@@ -118,52 +117,35 @@ CFTestD_Config *CFTestD_CheckOpts(int argc, char **argv)
...
-            CFTestD_Help();
-            exit(EXIT_SUCCESS);
-        case 'I':
-            LogSetGlobalLevel(LOG_LEVEL_INFO);
-            break;
+        case 'a': SetBindInterface(optarg); break;
+        case 'd': LogSetGlobalLevel(LOG_LEVEL_DEBUG); break;
+        case 'f': config->file = xstrdup(optarg); break;
+        case 'h': CFTestD_Help(); exit(EXIT_SUCCESS);
+        case 'I': LogSetGlobalLevel(LOG_LEVEL_INFO); break;
```
vpodzime: I like the original multi-line version better.

cf-testd/cf-testd.c
```diff
@@ -202,23 +188,22 @@ bool CFTestD_TLSSessionEstablish(ServerConnectionState *conn)
     /* Send/Receive "CFE_v%d" version string, agree on version, receive
        identity (username) of peer. */
     char username[sizeof(conn->username)] = "";
-    bool id_success = ServerIdentificationDialog(conn->conn_info,
-                                                 username, sizeof(username));
+    bool id_success                       = ServerIdentificationDialog(
+        conn->conn_info, username, sizeof(username));
```
vpodzime: WTH?

cf-testd/cf-testd.c
```diff
@@ -239,17 +224,18 @@ static bool CFTestD_ProtocolError(
...
 static bool CFTestD_BusyLoop(ServerConnectionState *conn)
 {
-    char recvbuffer[CF_BUFSIZE + CF_BUFEXT] = "";
+    char recvbuffer[CF_BUFSIZE + CF_BUFEXT]        = "";
     char sendbuffer[CF_BUFSIZE - CF_INBAND_OFFSET] = "";
 
-    const int received = ReceiveTransaction(conn->conn_info,
-                                            recvbuffer, NULL);
+    const int received =
+        ReceiveTransaction(conn->conn_info, recvbuffer, NULL);
```
vpodzime: Again, I liked the original better.

cf-testd/cf-testd.c
```diff
@@ -312,7 +299,9 @@ static ServerConnectionState *CFTestD_NewConn(ConnectionInfo *info)
     struct sockaddr_storage addr;
     socklen_t size = sizeof(addr);
 
-    if (getsockname(ConnectionInfoSocket(info), (struct sockaddr *)&addr, &size) == -1)
+    if (getsockname(
+            ConnectionInfoSocket(info), (struct sockaddr *)&addr, &size) ==
+        -1)
```
vpodzime: This is just gross.

cf-testd/cf-testd.c
```diff
@@ -215,7 +220,54 @@ bool CFTestD_GetServerQuery(ServerConnectionState *conn, char *recvbuffer)
...
+            "report file must have a CFR timestamp header line and at least two lines of items\n");
+        return false;
+    }
+
+    StringRef ts_ref =
+        StringGetToken(config->report, config->report_len, 0, "\n");
+    char *ts = xstrndup(ts_ref.data, ts_ref.len);
+    header   = StringFormat("CFR: 0 %s %d\n", ts, config->report_len);
+    SendTransaction(
+        conn->conn_info, header, SafeStringLength(header), CF_MORE);
```
vpodzime: This block looks to me like it got scattered... (I know it's "clang-format-great")

cf-testd/cf-testd.c
```diff
@@ -215,7 +220,54 @@ bool CFTestD_GetServerQuery(ServerConnectionState *conn, char *recvbuffer)
...
+        Log(LOG_LEVEL_INFO,
+            "No report file argument so returning canned data from enterprise plugin server\n");
+        return CFTestD_ReturnQueryData(conn, query);
+    }
+
+    Log(LOG_LEVEL_INFO,
+        "Report file argument specified. Returning report of length %d",
+        config->report_len);
+
+    char *header;
```
vpodzime: Please declare the variable right next to the block where it is used.

cf-testd/cf-testd.c
```diff
@@ -458,8 +512,27 @@ int CFTestD_StartServer()
...
+    case SIGTERM:
+    case SIGINT:
+        // flush all logging before process ends.
+        fflush(stdout);
+        exit(EXIT_FAILURE);
+        break;
+    default: break;
+    }
+
+    signal(signum, HandleSignal);
```
vpodzime: Why this call?

cf-testd/cf-testd.c
```diff
@@ -469,12 +542,49 @@ int main(int argc, char *argv[])
...
+        if ((fp = fopen(config->file, "r")) == NULL)
+        {
+            Log(LOG_LEVEL_INFO,
+                "Could not read report file at %s (fopen: %s)",
+                config->file,
+                GetErrorStr());
+            exit(EXIT_FAILURE);
+        }
+
+        while ((c = getc(fp)) != EOF)
```
vpodzime: Sorry for the nitpicking, but is there any reason why the file is read character after character?

cf-testd/cf-testd.c
olehermanse: I agree, I can change the config for this, I think almost all of our codebase has these on separate lines.

cf-testd/cf-testd.c
olehermanse: This looks like there was another assignment here at some point, and it was trying to align the equal signs, but @craigcomstock moved that assignment and didn't run clang-format again.

cf-testd/cf-testd.c
olehermanse: Do you mean the spaces before comments or the ordering?

It seems googles default style orders includes alphabetically. I can turn that off. If you like more spaces before the comments that can be adjusted, but I like just 1.

cf-testd/cf-testd.c
olehermanse: I think the original is gross as well ;)

cf-testd/cf-testd.c
olehermanse: In this specific case, I somewhat agree, but I think they're very close. I'm not going to force others to use clang-format, manual formatting should still be an option. However, I really like that clang-format, with the current config, follows a few very simple and predictable rules.

cf-testd/cf-testd.c
olehermanse: I think this is fine.

cf-testd/cf-testd.c
vpodzime: The order doesn't really matter to me. But I like when the comments are all aligned to the column not colliding with any non-comment characters on other lines. i.e, just like it was before.

cf-testd/cf-testd.c
vpodzime: Hard to argue with that.

cf-testd/cf-testd.c
vpodzime: Yeah, just please limit its effect to the new and changed code.

cf-testd/cf-testd.c
olehermanse: Nitpicking is good :)

cf-testd/cf-testd.c
olehermanse: What do you mean scattered? Let's have a discussion here: https://github.com/cfengine/core/pull/3045

cf-testd/cf-testd.c
craigcomstock: Agreed. I don't mind nitpicking at all. No particular reason, what would you suggest?

cf-testd/cf-testd.c
craigcomstock: there is an assignment right above this. The line is too long without some breaking.

I would maybe do this: 
```    
    char username[sizeof(conn->username)] = "";
    bool id_success = ServerIdentificationDialog(
        conn->conn_info, 
        username, 
        sizeof(username)
    );
```

cf-testd/cf-testd.c
craigcomstock: similar to the other long line maybe something like this:
```
    if (getsockname(
            ConnectionInfoSocket(info),
            (struct sockaddr *)&addr, 
            &size) == -1)
    {
```

cf-testd/cf-testd.c
craigcomstock: that one is only two chars over 78, I'd say put it back to a single line.

```
    SendTransaction(conn->conn_info, header, SafeStringLength(header), CF_MORE);
```

But like you said @olehermanse discuss over in #3045 and maybe I should just revert all these clang-format related changes and stick with plain and simple, minimal changes to conform to things like curly brackets and such.

cf-testd/cf-testd.c
olehermanse: Move ConnectionInfoSocket() and getsockname outside the if :)

cf-testd/cf-testd.c
craigcomstock: I don't have a ton of experience with signal handling. I thought I needed to reset the signal handler. Of course it is a no-op with the current code since I only register SIGINT and SIGTERM and both of those cause an exit. I will remove.

cf-testd/cf-testd.c
craigcomstock: @vpodzime probably just use read() then eh? I'll rewrite it that way.

craigcomstock: Looking at https://wiki.sei.cmu.edu/confluence/display/c/FIO19-C.+Do+not+use+fseek%28%29+and+ftell%28%29+to+compute+the+size+of+a+regular+file and other bits online, maybe I should use fstat and dynamically allocate enough space for a reasonably sized file... say 1MB? I looked at a sample previous_state dir and got 260K. Certainly much bigger than the previous CF_BUFSIZE of 4K which certainly would have failed us in more strenuous/realistic testing.

craigcomstock: Oh but wait... it only works with regular files to do any sort of fstat or fseek to determine the size and I want/need to support device files like /dev/stdin. So... @vpodzime let me know what you would suggest.

cf-testd/cf-testd.c
```diff
@@ -230,7 +220,54 @@ bool CFTestD_GetServerQuery(
...
+        char *item = xstrndup(item_ref.data, item_ref.len);
+
+        SendTransaction(
+            conn->conn_info, item, SafeStringLength(item), CF_MORE);
+    }
+
+    const char end_reply[] = "QUERY complete";
+    SendTransaction(
+        conn->conn_info, end_reply, SafeStringLength(end_reply), CF_DONE);
+
```
olehermanse: This function has quite a few repeated derefs. I doubt it makes a performance difference, but I think it makes the code more readable if you deref once. I.e. put something like this further up in the function:
```

File: cf-testd/cf-testd.c (hunk 12 above)
Comment (craigcomstock) [reply cf-testd/cf-testd.c]: maybe use our existing helper functions:
```
    if (config->file != NULL)
    {
        Log(LOG_LEVEL_NOTICE, "Got file argument: '%s'", config->file);
        if (!FileCanOpen(config->file, "r"))
        {
            Log(LOG_LEVEL_ERR, "Can't open file '%s' for reading", config->file);
            exit(EXIT_FAILURE);
        }

        Writer *contents = FileRead(config->file, SIZE_MAX, NULL);
        if (!contents)
        {
            Log(LOG_LEVEL_ERR, "Error reading report file '%s'", config->file);
            exit(EXIT_FAILURE);
        }
        config->report = StringWriterData(contents);
        config->report_len = SafeStringLength(config->report);

        Log(LOG_LEVEL_NOTICE, "Read %d bytes for report contents", config->report_len);

        if (config->report_len == 0)
        {
            Log(LOG_LEVEL_ERR, "Report file contained no bytes");
            exit(EXIT_FAILURE);
        }

        Log(LOG_LEVEL_DEBUG, "Got report file contents: %s", config->report);
    }
```

File: cf-testd/cf-testd.c [hunk 14]
```diff
@@ -471,12 +542,49 @@ int main(int argc, char *argv[])
     if (config->file != NULL)
     {
         Log(LOG_LEVEL_NOTICE, "Got file argument: '%s'", config->file);
-        Log(LOG_LEVEL_ERR, "File input not yet supported");
-        // TODO: implement file input
-        exit(EXIT_FAILURE);
+        FILE *fp = NULL;
```
olehermanse: fopen here instead.

cf-testd/cf-testd.c
```diff
@@ -471,12 +542,49 @@ int main(int argc, char *argv[])
...
+        char buffer[CF_BUFSIZE];
+        int c, i = 0;
```
olehermanse: Move these down to where they are used.

cf-testd/cf-testd.c
```diff
@@ -471,12 +542,49 @@ int main(int argc, char *argv[])
...
+
+        if ((fp = fopen(config->file, "r")) == NULL)
```
olehermanse: Don't fopen inside condition. Move it up like my comment above suggests, and just do:
```C

File: cf-testd/cf-testd.c (hunk 14 above)
Comment (olehermanse) [reply cf-testd/cf-testd.c]: Then it can also be const:
```C
FILE *const fp = fopen(config->file, "r");
```

olehermanse: (Make sure this line is directly above the if test, so `GetErrorStr()` is correct)

cf-testd/cf-testd.c
```diff
@@ -471,12 +542,49 @@ int main(int argc, char *argv[])
...
+            {
+                Log(LOG_LEVEL_ERR,
+                    "Report file was larger than max buffer size %d",
+                    CF_BUFSIZE);
+                exit(EXIT_FAILURE);
+            }
+            buffer[i++] = c;
+        }
+        Log(LOG_LEVEL_NOTICE, "Read %d bytes for report contents", i);
+        buffer[i] = 0;
```
olehermanse: Buffer overflow! You can write 2 bytes outside of buffer here;

- Last byte in loop is written at buffer[CF_BUFSIZE] (because you are not checking for `>=`)

cf-testd/cf-testd.c
```diff
@@ -118,52 +121,35 @@ CFTestD_Config *CFTestD_CheckOpts(int argc, char **argv)
...
-        case 'f':
-            config->file = xstrdup(optarg);
-            break;
-        case 'h':
-            CFTestD_Help();
-            exit(EXIT_SUCCESS);
-        case 'I':
-            LogSetGlobalLevel(LOG_LEVEL_INFO);
-            break;
+        case 'a': SetBindInterface(optarg); break;
```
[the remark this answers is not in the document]
craigcomstock: cool.

.clang-format
```diff
@@ -0,0 +1,35 @@
...
+BreakBeforeBraces: Allman
+BraceWrapping:
+  AfterControlStatement: true
+  AfterFunction: true
+  AfterStruct: true
+  AfterUnion: true
+  AfterEnum: true
+  BeforeElse: true
+---
+
```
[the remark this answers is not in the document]
craigcomstock: ack

cf-testd/cf-testd.c
```diff
@@ -471,12 +542,49 @@ int main(int argc, char *argv[])
...
+        buffer[i] = 0;
+
+        if (i == 0)
+        {
+            Log(LOG_LEVEL_ERR, "Report file contained no bytes");
+            exit(EXIT_FAILURE);
+        }
+
+        config->report     = buffer;
+        config->report_len = i - 1;
```
olehermanse: Just to be 100% clear, these derefs cannot be removed, since these are assignments. My comment above is about the usage of `config->report` and `config->report_len` in arguments etc.

cf-testd/cf-testd.c
vpodzime: Yes, that looks like the best approach. :)

cf-testd/cf-testd.c
vpodzime: > @craigcomstock Put an empty line between the 2 assignments and it won't align them.

That's exactly stuff I would like to avoid doing.

cf-testd/cf-testd.c
vpodzime: I don't know, I just like the original line the best.

PR #3043, opened 2018-03-16
ENT-2713: Make the PipeIsReadWriteReady OS-specific

vpodzime:
We need to do totally different tricks on windows.


vpodzime:
Needed for https://github.com/cfengine/enterprise/pull/394

PR #3045, opened 2018-03-20
Added clang-format config

olehermanse:


olehermanse:
@vpodzime @craigcomstock Please comment on the clang-format related stuff here.

I don't wanna force a strict style on everyone or force everyone to use clang-format, but I'd like to have a working config that produces a style that we can all find acceptable.

I added an example C file for github commenting only, will remove before merge.

example.c
```diff
@@ -0,0 +1,79 @@
...
+}
+
+int main(int argc, char **argv)
+{
+    // Most of the time function calls and assignments should fit on one line:
+    int a_number = SimpleFunction(1, 2, 3);
+
+    // Assignments are aligned:
+    int a   = 0;
+    int abc = 10;
```
vpodzime: I don't like this. And with multiple assignments, in particular if some are also long ("multi-line function calls"), this results in a code that looks like scattered words to me and is very hard to read for me.

.clang-format
```diff
@@ -0,0 +1,35 @@
+---
+BasedOnStyle: Google
+---
+Language: Cpp
+ColumnLimit: 78
```
vpodzime: I have always hated this limitation. I think 98 is a much better value. But of course, YMMV.

example.c
olehermanse: You can get around this by putting a newline in front of the long function calls ones.

.clang-format
olehermanse: I think 98 is acceptable too. Tbh I'm not too annoyed with the 78 limit, but the reason why I prefer it is mainly because it's what the rest of the codebase  uses (for the most part), and also seems most "standard".

example.c
vpodzime: Yeah, I just don't want to be forced to do that. But then again, I'm probably not going to use this tool so I'm fine with this. I will just object to the way it mangles things if you forget to add that extra line.

olehermanse: This one's not a big deal for me either way. Let's hear what @craigcomstock thinks about it.

olehermanse: @vpodzime I'll turn this off since I agree. I wish it was possible to set some limit for how far it would align something, but this feature doesn't exist (yet).

vpodzime: Thanks

example.c
```diff
@@ -0,0 +1,79 @@
...
+        if (a_number == 10)
+        {
+            return 1;
+        }
+
+        // Switch cases must be on separate lines:
+        switch (a_number)
+        {
+        case 9:
+            return 1;
```
craigcomstock: The case lines not being indented looks strange to me.

olehermanse: This is the way it is through most of our code-base. Picked up from [Linux Kernel coding style](https://www.kernel.org/doc/html/latest/process/coding-style.html). It also avoids the "double indent" of case contents, so I think I'll stick with this one.

PR #3046, opened 2018-03-20
CFE-1915: Disallow modifications of foreign variables

vpodzime:
As described in [[CFE-1915]](https://tracker.mender.io/browse/CFE-1915).

### DIFF
--- a/libpromises/verify_vars.c
+++ b/libpromises/verify_vars.c
@@ -60,6 +60,7 @@ static bool IsValidVariableName(const char *var_name)
     static pcre *rx = NULL;
     if (!rx)
     {
+        /* \200-\377 is there for multibyte unicode characters */
         rx = CompileRegex("[a-zA-Z0-9_\200-\377.]+(\\[.+\\])*"); /* Known leak, see TODO. */
     }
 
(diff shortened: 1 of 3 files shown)


vpodzime:
jenkins job: https://ci.cfengine.com/job/pr-pipeline/191/

ncharles:
Hi,
This PR would break a lot of things within Rudder - we use this behaviour for setting audit mode and going back to enforce mode.
What is worse is that this PR doesn't offer any workaround for this legitimate use case - without this, we can't use the CFEngine agent anymore

vpodzime:
> What is worse is that this PR doesn't offer any workaround for this legitimate use case - without this, we can't use the CFEngine agent anymore

Please see the [JIRA ticket](https://tracker.mender.io/browse/CFE-1915) for the discussion and suggestions for workarounds.

vpodzime:
> jenkins job: https://ci.cfengine.com/job/pr-pipeline/191/

All green except for some non-related issues with nested VMs.

fanf:
Wow. Even with a big "warning" of the impact of that modification, you didn't care? This is extremelly disturbing. I don't know what we, Rudder dev (and the other people who gave voice to that) should understand about that. But the general message is extremelly bad.

olehermanse:
@fanf We do care. Our decision is to fix this bug, which will break compatibility for those "using" it, however we want to help those affected by the change (including Normation / Rudder). It is my understanding that someone from the CFEngine team will contact you about this shortly.

PR #3052, opened 2018-03-26
CFE-741 Do not error when file does not exist with isexecutable()

nickanderson:
Changelog: Title

Since isexecutable() is only checking to see if a file is executable we do not
need to have an error message that is always emitted if the file does not exist.
The function will return non-true and the class will not be set. If the file
does not exist, it's not executable.

### DIFF
--- a/libpromises/unix.c
+++ b/libpromises/unix.c
@@ -120,7 +120,7 @@ int IsExecutable(const char *file)
 
     if (stat(file, &sb) == -1)
     {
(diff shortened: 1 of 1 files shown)


nickanderson:
Thanks, squashed. will merge after tests pass.

Opinion on backport?

vpodzime:
> Opinion on backport?

It's just a cosmetic change in the end, isn't it? In that case I don't think it needs to be backported.

PR #3065, opened 2018-04-04
CFE-2793 Identify bundle name when triggering bundle abortion

nickanderson:
Changelog: Title

This change is intended to make it easier to identify what bundle was aborted when `abortbundleclasses` are triggered.

Now when a bundle is aborted the log message says Bundle x aborted on defined class y. Previously when a bundle was aborted the log message only indicated the class that triggered the abortion.


nickanderson:
In testing a package with this change I found no difference in the log messages emitted.

Looking at the code, I found an identical log entry that I had not changed. Took a wild guess, not really hopeful that my WAG will work.

nickanderson:
Closing these PRs since my patch completely fails.

olehermanse:
See #3154

PR #3068, opened 2018-04-05
ENT-3525: jquery URL timing out, use traviscistatus.com instead

craigcomstock:
WAG


olehermanse:
Fantastic

tests/acceptance/01_vars/02_functions/network/url_get.cf
```diff
@@ -27,8 +37,19 @@ bundle agent init
       "options_clean" data => '{}';
       "res1" data => url_get("http://acme.com", options_clean);
       "res2" data => url_get("http://nosuchcfenginehost.com", options_clean);
-      # static content, we hope won't change much
-      "res3" data => url_get("http://ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js", options_limit);
+      "res3" data => url_get("http://localhost:8000/index.html", options_limit);
+
+  commands:
+    "/usr/bin/python3 -m http.server 8000"
```
olehermanse: You should be able to omit `/usr/bin/` since you're using useshell.

tests/acceptance/01_vars/02_functions/network/url_get.cf
```diff
@@ -27,8 +37,19 @@ bundle agent init
...
+      action => background,
+      contain => useshell;
```
olehermanse: You could add another commands promise to write `<html>Hello url_get</html>` to `index.html`, making the expected output more readable.

craigcomstock: Yes. I thought of that too but wasn't sure where CWD would be when running. I suppose if I write it out in a promise that will be predictable. Cool.

tests/acceptance/01_vars/02_functions/network/url_get.cf
craigcomstock: :+1:

tests/acceptance/01_vars/02_functions/network/url_get.cf
olehermanse: @craigcomstock I don't know what it will be either, but it should be the same as when running the python command. As long as it has write access, (I think it does, it's probably some test specific work dir) it should be good, and `http.server` will look for index.html in the same cwd.

olehermanse: That would be the advantage of doing it as a commands promise (echo) and not a files promise.

vpodzime:
@craigcomstock do we still need to care about this? (doing an "open PRs cleanup").

PR #3070, opened 2018-04-09
CFE-2536: Resolve subkey conflicts when converting to JSON

vpodzime:
Whenever there is a conflict of array variable definitions prefer
the container subkeys over simple values when converting to JSON.

Changelog: Commit
(cherry picked from commit baaad412021cfd7d9baf9c5863df23fae19e99b2)

Conflicts:
   libpromises/evalfunction.c


olehermanse:
👍 I would run it through jenkins pipeline just to be safe.

vpodzime:
> +1 I would run it through jenkins pipeline just to be safe.

Yeap. And it turns out there's a problem.

vpodzime:
There is an issue with the backport probably due to some other earlier change(s) missing on the *3.7.x* branch.

PR #3074, opened 2018-04-12
CFE-2066 Do not warn about meta promises in server bundles

nickanderson:
meta type promises are valid in all bundle types.


nickanderson:
Thanks for the review @vpodzime . Title changed in commit and PR. will merge after green again.

PR #3082, opened 2018-04-13
CFE-2125: Add a test for a workaround for mustache section iterations

vpodzime:
Our implementation of the mustache templating iterates over not
only JSON arrays but also over JSON objects (their
properties). That can easily result in duplicates when rendering
the output (see CFE-2125 for more details). However, the
iterations are done on purpose to make our extensions to the
referential implementation work. So instead of disabling them
let's just make sure an easy workaround works.

(cherry picked from commit 0555a8e3a1bd0b0a7e72f0fa1b5c0fd353fe85d6)


vpodzime:
jenkins job: https://ci.cfengine.com/job/pr-pipeline/302/

PR #3085, opened 2018-04-14
Add in description of the policy and output

nickanderson:


nickanderson:
I don't understand why adding an example decreases code cov. This example has an example_output section, so it's part of the check_outputs test.

PR #3095, opened 2018-04-25
On the way to serving multiple reports in parallel from a single cf-testd process

vpodzime:
The only externally visible changes here are the ``--output-file`` docs fix and the new ``--key-file`` option. However, the rest of the commits is where the actual fun is happening.

With these changes, it should be possible to start multiple threads in *cf-testd*, each serving a different report, each bound to a different address and each using a different key pair. Thus effectively acting as multiple different hosts submitting reports.


vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/426/)

cf-serverd/server_tls.c
```diff
@@ -90,87 +115,91 @@ bool ServerTLSInitialize()
...
     {
         Log(LOG_LEVEL_ERR,
             "Failed to generate in-memory certificate from private key");
-        goto err2;
+        goto err;
     }
 
-    SSL_CTX_use_certificate(SSLSERVERCONTEXT, SSLSERVERCERT);
+    SSL_CTX_use_certificate(*ssl_ctx, cert);
+    X509_free(cert);
```
craigcomstock: Did you find that the cert is not needed after SSL_CTX_use_certificate? Previously the code waited until later to free the cert. I looked into the openssl code a bit and didn't see something that clearly said you could free the cert.

cf-testd/cf-testd.c
```diff
@@ -136,6 +143,12 @@ CFTestD_Config *CFTestD_CheckOpts(int argc, char **argv)
         case 'I':
             LogSetGlobalLevel(LOG_LEVEL_INFO);
             break;
+        case 'l':
+            LoggingEnableTimestamps(true);
+            break;
+        case 'k':
+            config->key_file = xstrdup(optarg);
+            break;
         case 'p':
```
craigcomstock: swap 'l' and 'k' to put them in alpha order?

libpromises/crypto.c
```diff
@@ -159,30 +159,48 @@ static void RandomSeed(void)
 static const char priv_passphrase[] = "Cfengine passphrase";
 
 /**
+ * @param[in] priv_key_path path to the private key to use (%NULL to use the default)
+ * @param[in] pub_key_path path to the private key to use (%NULL to use the default)
+ * @param[out] priv_key a place to store the loaded private key (or %NULL to
```
craigcomstock: what is the significance of % in %NULL? - I'm not too familiar with doxygen.

cf-serverd/server_tls.c
vpodzime: I did the same with the same result. So I tried it. And it seems to be working just fine. There's no way to get the cert back from the SSL_CTX so I think it must somehow load the cert into it's own internal state.

cf-testd/cf-testd.c
vpodzime: yeap, good catch

libpromises/crypto.c
vpodzime: I believe that's how you mark literal values. It definitely works in gtk-doc, but I'm not sure about doxygen. Anyway, my Emacs highlights it so I like it. :)

craigcomstock: :+1:

PR #3099, opened 2018-04-26
Remove un-necessary markup

nickanderson:


nickanderson:
I didn't think of them until after I had the example and docs working.

nickanderson:
This is master, the other one is a backport to 3.10

PR #3103, opened 2018-04-27
Parallel cf-testd

vpodzime:
Tested this with 200 threads and everything seems to be working fine.

### DIFF
--- a/cf-testd/cf-testd.c
+++ b/cf-testd/cf-testd.c
@@ -53,6 +53,7 @@
 #include <time_classes.h>       // UpdateTimeClasses
 #include <timeout.h>            // SetReferenceTime
 #include <tls_generic.h>        // TLSLogError
+#include <logging.h>            // thread-specific log prefix
 
 #ifndef __MINGW32__
 #include <sys/socket.h>
@@ -574,9 +575,26 @@ int CFTestD_StartServer(CFTestD_Config *config)
     return ret;
 }
 
+static char *LogAddPrefix(LoggingPrivContext *log_ctx,
+                          ARG_UNUSED LogLevel level,
+                          const char *raw)
+{
+    const char *ip_addr = log_ctx->param;
+    return ip_addr ? StringConcatenate(4, "[", ip_addr, "] ", raw) : xstrdup(raw);
+}
+
 static void *CFTestD_ServeReport(void *config_arg)
 {
     CFTestD_Config *config = (CFTestD_Config *) config_arg;
+
+    /* Set prefix for all Log()ging: */
+    LoggingPrivContext *prior = LoggingPrivGetContext();
+    LoggingPrivContext log_ctx = {
+        .log_hook = LogAddPrefix,
+        .param = config->address
+    };
+    LoggingPrivSetContext(&log_ctx);
+
     char *priv_key_path = NULL;
     char *pub_key_path = NULL;
     if (config->key_file != NULL)
@@ -634,6 +652,8 @@ static void *CFTestD_ServeReport(void *config_arg)
      * terminate, but it's a good way the cleanup actually works and doesn't
      * cause a segfault or something */
     ServerTLSDeInitialize(&(config->priv_key), &(config->pub_key), &(config->ssl_ctx));
+
+    LoggingPrivSetContext(prior);
 
     return NULL;
 }
--- a/cf-testd/cf-testd.c
+++ b/cf-testd/cf-testd.c
@@ -54,6 +54,10 @@
 #include <timeout.h>            // SetReferenceTime
 #include <tls_generic.h>        // TLSLogError
 
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+
 #define CFTESTD_QUEUE_SIZE 10
 
 // ============================= CFTestD_Config ==============================
@@ -67,6 +71,8 @@ typedef struct
     RSA *pub_key;
     SSL_CTX *ssl_ctx;
     char *address;
+    pthread_t t_id;
+    int ret;
 } CFTestD_Config;
 
 /*******************************************************************/
@@ -78,6 +84,7 @@ static const struct option OPTIONS[] = {
     {"debug", no_argument, 0, 'd'},
     {"help", no_argument, 0, 'h'},
     {"inform", no_argument, 0, 'I'},
+    {"jobs", required_argument, 0, 'j'},
     {"key-file", required_argument, 0, 'k'},
     {"timestamp", no_argument, 0, 'l'},
     {"port", required_argument, 0, 'p'},
@@ -91,6 +98,9 @@ static const char *const HINTS[] = {
     "Enable debugging output",
     "Print the help message",
     "Print basic information about what cf-testd does",
+    "Number of jobs (threads) to run in parallel. Use '%d' in the report path"
+    " and key file path (will be replaced by the thread number, starting with"
+    " 0). Threads will bind to different addresses incrementally.",
     "Specify a path to the key (private) to use for communication",
     "Log timestamps on each line of log output",
     "Set the port cf-testd will listen on",
@@ -122,14 +132,14 @@ void CFTestD_Help()
     FileWriterDetach(w);
 }
 
-CFTestD_Config *CFTestD_CheckOpts(int argc, char **argv)
+CFTestD_Config *CFTestD_CheckOpts(int argc, char **argv, long *n_threads)
 {
     extern char *optarg;
     int c;
     CFTestD_Config *config = CFTestD_ConfigInit();
     assert(config != NULL);
 
-    while ((c = getopt_long(argc, argv, "a:df:hIk:lp:vV", OPTIONS, NULL)) != -1)
+    while ((c = getopt_long(argc, argv, "a:df:hIj:k:lp:vV", OPTIONS, NULL)) != -1)
     {
         switch (c)
         {
@@ -145,6 +155,16 @@ CFTestD_Config *CFTestD_CheckOpts(int argc, char **argv)
         case 'I':
             LogSetGlobalLevel(LOG_LEVEL_INFO);
             break;
+        case 'j':
+        {
+            int ret = StringToLong(optarg, n_threads);
+            if (ret != 0)
+            {
+                Log(LOG_LEVEL_ERR, "Failed to parse number of threads/jobs from '%s'\n", optarg);
+                *n_threads = 1;
+            }
+            break;
+        }
         case 'k':
             config->key_file = xstrdup(optarg);
             break;
@@ -509,8 +529,6 @@ int CFTestD_StartServer(CFTestD_Config *config)
 
     int sd = InitServer(CFTESTD_QUEUE_SIZE, config->address);
 
-    MakeSignalPipe();
-
     int selected = 0;
     while (selected != -1)
     {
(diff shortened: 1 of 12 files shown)


cf-testd/cf-testd.c
```diff
@@ -90,6 +104,9 @@ static const char *const HINTS[] = {
     "Enable debugging output",
     "Print the help message",
     "Print basic information about what cf-testd does",
+    "Number of jobs (threads) to run in parallel. Use '%d' in the report and"
+    " key file paths (will be replaced by the thread number, starting with 0)."
```
craigcomstock: maybe s/report/report filename provided by the --report option/ or something that's a bit more clear that it references the --report option value.

cf-testd/cf-testd.c
```diff
@@ -607,16 +620,153 @@ int main(int argc, char *argv[])
...
+    if (ret == 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to parse address: '%s'", ip_str);
+        return NULL;
+    }
+
+    int step = 1;
+    char *last_dot = strrchr(ip_str, '.');
+    if (strcmp(last_dot + 1, "255") == 0)
+    {
```
craigcomstock: I guess inet_aton() should confirm that the IP has a dot in it except in the case that ip_str is just a number. Looks pretty good to me. I was thinking maybe a check for last_dot != NULL would be good but then thought about inet_aton() call above there.

From inet_aton(3) one of the forms the string can take:
```
       a         The  value  a  is  interpreted as a 32-bit value that is stored directly into the
                 binary address without any byte rearrangement.
```

cf-testd/cf-testd.c
```diff
@@ -607,16 +620,153 @@ int main(int argc, char *argv[])
...
+            thread_configs[i]->address = xstrdup(config->address);
+        }
+        else
+        {
+            thread_configs[i]->address = IncrementIPaddress(thread_configs[i-1]->address);
+        }
+    }
+
+    /* CFTestD_ConfigDestroy(config); */
+
```
craigcomstock: go ahead and delete this comment right?

cf-testd/cf-testd.c
vpodzime: Changing to "*report path and key file path*", hope that makes it clearer.

cf-testd/cf-testd.c
vpodzime: I had to replace this with ``inet_addr`` due to windows (in)compatibility. And that one seems to be stricter. However, this is an internal function and it's documentation says it has to be given an IP address with "*4 numbers separated by dots*". So I don't think this is a problem

cf-testd/cf-testd.c
vpodzime: Actually, it should only be uncommented. :)

vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/458/)

libcfnet/server_code.c
```diff
@@ -65,11 +68,16 @@ static int OpenReceiverChannel(void)
...
+
     /* Listen to INADDR(6)_ANY if BINDINTERFACE unset. */
     char *ptr = NULL;
-    if (BINDINTERFACE[0] != '\0')
+    if (bind_address[0] != '\0')
     {
-        ptr = BINDINTERFACE;
+        ptr = bind_address;
         query.ai_flags |= AI_NUMERICHOST;
     }
```
olehermanse: STUPID github won't let me comment on code below this line. There's one more instance of you should update, right?

```

Comment (vpodzime) [reply libcfnet/server_code.c]: Yes, good catch! Fixing.

**Comment by vpodzime** [comment]
[Build Status](https://ci.cfengine.com/job/pr-pipeline/475/)

File: cf-testd/cf-testd.c [hunk 5]
```diff
@@ -78,6 +84,7 @@ static const struct option OPTIONS[] = {
     {"debug", no_argument, 0, 'd'},
     {"help", no_argument, 0, 'h'},
     {"inform", no_argument, 0, 'I'},
+    {"jobs", required_argument, 0, 'j'},
```
olehermanse: I appreciate that you respect the alphabetic ordering :)

cf-testd/cf-testd.c
```diff
@@ -145,6 +155,16 @@ CFTestD_Config *CFTestD_CheckOpts(int argc, char **argv)
         case 'I':
             LogSetGlobalLevel(LOG_LEVEL_INFO);
             break;
+        case 'j':
+        {
+            int ret = StringToLong(optarg, n_threads);
+            if (ret != 0)
+            {
+                printf("Failed to parse number of threads/jobs from '%s'\n", optarg);
```
olehermanse: `Log(LOG_LEVEL_ERR, ...`

cf-testd/cf-testd.c
```diff
@@ -536,31 +554,9 @@ int CFTestD_StartServer(CFTestD_Config *config)
...
-        fflush(stdout);
-        exit(EXIT_FAILURE);
-        break;
-    default:
-        break;
-    }
-}
-
-int main(int argc, char *argv[])
+static void * CFTestD_ServeReport(void *config_arg)
```
olehermanse: Asterisk with function name (style).

cf-testd/cf-testd.c
```diff
@@ -609,16 +605,151 @@ int main(int argc, char *argv[])
...
+        exit(EXIT_FAILURE);
+        break;
+    default:
+        break;
+    }
+}
+
+/**
+ * @param ip_str string representation of an IPv4 address (the usual one, with
+ *               4 numbers separated by dots)
```
olehermanse: I believe the *numbers* are commonly called octets.

cf-testd/cf-testd.c
```diff
@@ -609,16 +605,151 @@ int main(int argc, char *argv[])
...
+ */
+static char *IncrementIPaddress(const char *ip_str)
```
olehermanse: I would add a note about freeing the new string this function allocates.

cf-testd/cf-testd.c
```diff
@@ -609,16 +605,151 @@ int main(int argc, char *argv[])
...
+    int ret = inet_aton(ip_str, &ip);
+    if (ret == 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to parse address: '%s'", ip_str);
+        return NULL;
+    }
+
+    int step = 1;
+    char *last_dot = strrchr(ip_str, '.');
+    if (strcmp(last_dot + 1, "255") == 0)
```
olehermanse: `StringSafeEqual()`

cf-testd/cf-testd.c
```diff
@@ -609,16 +605,151 @@ int main(int argc, char *argv[])
...
+    struct in_addr ip;
+    int ret = inet_aton(ip_str, &ip);
+    if (ret == 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to parse address: '%s'", ip_str);
+        return NULL;
+    }
+
+    int step = 1;
+    char *last_dot = strrchr(ip_str, '.');
```
olehermanse: check for `NULL`.

cf-testd/cf-testd.c
```diff
@@ -609,16 +605,151 @@ int main(int argc, char *argv[])
...
+    if (strcmp(last_dot + 1, "255") == 0)
+    {
+        /* avoid the network address (ending with 0) */
+        step = 2;
+    }
+    else if (strcmp(last_dot + 1, "254") == 0)
```
olehermanse: `StringSafeEqual()`

cf-testd/cf-testd.c
```diff
@@ -609,16 +605,151 @@ int main(int argc, char *argv[])
...
+}
+
+int main(int argc, char *argv[])
+{
+    signal(SIGINT, HandleSignal);
+    signal(SIGTERM, HandleSignal);
+
+    Log(LOG_LEVEL_VERBOSE, "Starting cf-testd");
+    CryptoInitialize();         /* TODO: remove */
+    cfnet_init(NULL, NULL);
```
olehermanse: `CryptoInitialize()` is done in `cfnet_init()` and should be safe to remove. (I know you didn't introduce this)

cf-testd/cf-testd.c
```diff
@@ -609,16 +605,151 @@ int main(int argc, char *argv[])
...
+    }
+
+    CFTestD_Config **thread_configs = (CFTestD_Config**) xcalloc(n_threads, sizeof(CFTestD_Config*));
+    for (int i = 0; i < n_threads; i++)
+    {
+        thread_configs[i] = (CFTestD_Config*) xmalloc(sizeof(CFTestD_Config));
+
+        if (config->report_file != NULL && strstr(config->report_file, "%d") != NULL)
+        {
+            asprintf(&(thread_configs[i]->report_file), config->report_file, i);
```
olehermanse: This replaces the `%d` in file name with thread number right? Would comment.

olehermanse: Same below for key file.

cf-testd/cf-testd.c
```diff
@@ -609,16 +605,151 @@ int main(int argc, char *argv[])
...
+            thread_configs[i]->key_file = SafeStringDuplicate(config->key_file);
+        }
+
+        if (i == 0)
+        {
+            thread_configs[i]->address = xstrdup(config->address);
+        }
+        else
+        {
+            thread_configs[i]->address = IncrementIPaddress(thread_configs[i-1]->address);
```
olehermanse: Nice

cf-testd/cf-testd.c
```diff
@@ -509,8 +529,6 @@ int CFTestD_StartServer(CFTestD_Config *config)
 
     int sd = InitServer(CFTESTD_QUEUE_SIZE, config->address);
 
-    MakeSignalPipe();
-
     int selected = 0;
     while (selected != -1)
```
olehermanse: `WaitForIncoming()` returns `-2` for interrupt, so maybe check for this also?

cf-testd/cf-testd.c
```diff
@@ -609,16 +605,151 @@ int main(int argc, char *argv[])
...
+    CFTestD_ConfigDestroy(config);
+
+    bool failure = false;
+    for (int i = 0; !failure && (i < n_threads); i++)
+    {
+        int ret = pthread_create(&(thread_configs[i]->t_id), NULL, CFTestD_ServeReport, thread_configs[i]);
+        if (ret != 0)
+        {
+            fprintf(stderr, "Failed to create a new thread nr. %d: %s\n",
+                    i, strerror(ret));
```
olehermanse: I guess it is intentional to use `fprintf` instead of `Log(LOG_LEVEL_ERR, ...`? Might make sense for a testing only tool.

cf-testd/cf-testd.c
```diff
@@ -61,6 +61,11 @@
 #define CFTESTD_QUEUE_SIZE 10
 #define WAIT_CHECK_TIMEOUT 10
 
+/* Strictly speaking/programming, this should use a lock, but it's not needed
+ * for this testing tool. The worse that can happen is that some threads would
```
olehermanse: The **worst**

cf-testd/cf-testd.c
```diff
@@ -676,11 +680,14 @@ static char *IncrementIPaddress(const char *ip_str)
...
+    uint32_t ip_num = ntohl(ip);
     ip_num += step;
-    ip.s_addr = htonl(ip_num);
+    ip = htonl(ip_num);
+
+    struct in_addr ip_struct;
+    ip_struct.s_addr = ip;
 
-    return xstrdup(inet_ntoa(ip));
+    return xstrdup(inet_ntoa(ip_struct));
```
olehermanse: It's interesting how `inet_ntoa()` returns a non-const `char *` to a statically allocated buffer.

cf-testd/cf-testd.c
vpodzime: I think we just want to retry in such cases.

cf-testd/cf-testd.c
vpodzime: Yes, it was intentional. But we don't do this in our testing tools in general so I'll change it to ``Log()``.

cf-testd/cf-testd.c
vpodzime: Yeah, but it doesn't say you cannot modify it. Just that a next call will override it. I'm more concerned by the fact that this function cannot be thread-safe then.

vpodzime:
All comments addressed.

vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/477/)

vpodzime:
Green build. Merging.

cf-testd/cf-testd.c
```diff
@@ -607,16 +643,158 @@ int main(int argc, char *argv[])
...
+static void HandleSignal(int signum)
+{
+    switch (signum)
+    {
+    case SIGTERM:
+    case SIGINT:
+        // flush all logging before process ends.
+        fflush(stdout);
+        fprintf(stderr, "Terminating...\n");
+        TERMINATE = true;
```
olehermanse: @vpodzime Looks like this broke our reporting tests, because it can now take up to ~10 seconds to terminate cf-testd. Do you have a better way to do this, at least for the single threaded mode?

vpodzime: See https://github.com/cfengine/core/pull/3109 and https://github.com/cfengine/nova/pull/1180

PR #3104, opened 2018-04-27
ENT-3638: Add commit title to changelog by default, if it has ticket …

Lex-2008:
…number

From now on, you don't need to write "Changelog: title" if you prepend
title with JIRA ticket number

Signed-off-by: Aleksei Shpakovskii <aleksei.shpakovskii@cfengine.com>

### DIFF
--- a/misc/changelog-generator/changelog-generator
+++ b/misc/changelog-generator/changelog-generator
@@ -20,13 +20,16 @@ LINKED_SHAS = {}
 # messages.
 SHA_TO_TRACKER = {}
 
+# more relaxed regexp to find JIRA issues anywhere in commit message
 JIRA_REGEX = r"(?:Jira:? *)?(?:https?://tracker.mender.io/browse/)?((?:CFE|ENT|INF|ARCHIVE|MEN|QA)-[0-9]+)"
+# more strict regexp to find JIRA issues only in the beginning of title
+JIRA_TITLE_REGEX = r"^(?:CFE|ENT|INF|ARCHIVE|MEN|QA)-[0-9]+"
 TRACKER_REGEX = r"\(?(?:Ref:? *)?%s\)?:? *" % (JIRA_REGEX)
 
 POSSIBLE_MISSED_TICKETS = {}
 
 # Only for testing.
-SORT_CHANGELOG = False
+SORT_CHANGELOG = True
 
 # Type of log to generate, this is bitwise.
 LOG_TYPE = 0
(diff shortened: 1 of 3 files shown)


Lex-2008:
HACKING.md has this line at the end:

> In fact if anywhere in the commit message the string CFE-1234 is found, it will be automatically added to the ChangeLog.

which looks silly now that I added line about CFE-1234 in the title

Lex-2008:
oh, Kristian, you're right: `Changelog: None` will suppress nothing this way. I should either rewrite this script completely, or read it more.

Lex-2008:
Ok, 2 months and 1 week later I looked at this code once again and now I realised now to fix it without completely rewriting everything. Also, two new tests added and all tests are passing. Thanks for the hint, @kacf !

misc/changelog-generator/changelog-generator
```diff
@@ -20,13 +20,16 @@ LINKED_SHAS = {}
 # messages.
 SHA_TO_TRACKER = {}
 
+# more relaxed regexp to find JIRA issues anywhere in commit message
 JIRA_REGEX = r"(?:Jira:? *)?(?:https?://tracker.mender.io/browse/)?((?:CFE|ENT|INF|ARCHIVE|MEN|QA)-[0-9]+)"
+# more strict regexp to find JIRA issues only in the beginning of title
+JIRA_TITLE_REGEX = r"(?:CFE|ENT|INF|ARCHIVE|MEN|QA)-[0-9]+"
```
vpodzime: Shouldn't this regex start with `^` then?

Lex-2008: not necessary. Later it's used in `re.match` function, which does an "ancored match" - i.e. implies `^` at the beginning of regexp. For non-anchored match, Python provides a `re.search` function:
```
lex@flower:~$ python3
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import re
>>> re.match('a','abba')
<_sre.SRE_Match object; span=(0, 1), match='a'>
>>> re.match('b','abba')
>>> re.match('c','abba')
>>> re.search('b','abba')
<_sre.SRE_Match object; span=(1, 2), match='b'>
>>> re.search('c','abba')
>>> 
lex@flower:~$ 
```

but maybe you're right and it's better to include it for readability

vpodzime: Yeah, the code is quite cryptic. I think it's better to be explicit.

PR #3111, opened 2018-05-09
os-release file is now used for hard classes and sys.flavor on all li…

olehermanse:
…nuxes

This will improve platform detection on newer operating systems where
/etc/os-release (or /usr/lib/os-release) is present.

A hard class will be set for the value of the ID key (canonified with
underscores), if it exists. If both ID and VERSION_ID exist, multiple
hard classes will be set for all parts of the version number. The
special variable sys.flavor will also be set by determining major
version from VERSION_ID.

Example os-release file:
	ID=coreos
	VERSION_ID=1185.3.0

For the example above, `sys.flavor` will be `coreos_1185` and 4 hard
classes will be set; `coreos_1185_3_0`, `coreos_1185_3`, `coreos_1185`,
and `coreos`.

For backwards compatibility, older distribution specific logic is still
executed and may overwrite sys.flavor and define hard classes as before.

Changelog: Commit
Signed-off-by: Ole Herman Schumacher Elgesem <ole.elgesem@northern.tech>
(cherry picked from commit 3537fc83581983df0e49b6efcc21095ac347b1ba)

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -1058,6 +1058,58 @@ static void SetFlavour(EvalContext *ctx, const char *flavour)
 }
 
 #ifdef __linux__
+
+/**
+ * @brief Combines OS and version string to define flavor variable and class
+ *
+ * @note Input strings should be canonified before calling
+ */
+static void SetFlavour2(
+    EvalContext *const ctx,
+    const char *const id,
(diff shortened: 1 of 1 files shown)


nickanderson:
+1

basvandervlies:
+1

PR #3112, opened 2018-05-09
WIP ENT-3756: try to flush each line of logging while in debug level

craigcomstock:


libutils/logging.c
```diff
@@ -175,6 +175,12 @@ static void LogToConsole(const char *msg, LogLevel level, bool color)
 
     fprintf(stdout, "%8s: %s\n", LogLevelToString(level), msg);
 
+    if (level == LOG_LEVEL_DEBUG)
+    {
+        fflush(stdout);
+        // TODO also possibly call sync()? Not sure if that's too "expensive" or not.
```
vpodzime: I think `fsync()` should be no problem, it's a debug mode after all.

craigcomstock:
I included just the fflush(stdout) in #3180 so this smaller one can be closed.

PR #3119, opened 2018-05-16
ENT-3756: Debug db mutex locks

craigcomstock:


olehermanse:
@cf-bottom Could you please start a jenkins build for this PR?

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com/job/pr-pipeline/706/)

https://ci.cfengine.com/job/pr-pipeline/706/

craigcomstock:
@vpodzime @olehermanse I guess you can only remove the WIP tag. This is ready for code review. I added a bunch of TODOs in cf-serverd related code thinking we might want to do the same thing there. So that's something for you to consider and whether or not any other binaries need this treatment for windows such as cf-monitord or cf-execd.

craigcomstock:
I left the debug log level forcing a fflush(stdout) on each call to Log() as a separate commit thinking it might be nice as a separate PR.

cf-serverd/cf-serverd-functions.c
```diff
@@ -222,6 +222,7 @@ GenericAgentConfig *CheckOpts(int argc, char **argv)
                 GenericAgentWriteVersion(w);
                 FileWriterDetach(w);
             }
+// TODO re-use SafeExit() here for windows
```
vpodzime: So why don't we do this here? Is there any reason why we shouldn't?

libpromises/mutex.c
```diff
@@ -51,11 +51,9 @@ int __ThreadLock(pthread_mutex_t *mutex,
 
     if (result != 0)
     {
-        /* Log() is blocking on mutexes itself inside malloc(), so maybe not
-         * the best idea here. */
-        Log(LOG_LEVEL_ERR,
-            "Locking failure at %s:%d function %s! (pthread_mutex_lock: %s)",
-            filename, lineno, funcname, GetErrorStrFromCode(result));
+        // printf instead of Log() to avoid problems with mutexes in Log()
+        printf("thread: %d, error: Locking failure at %s:%d function %s! (pthread_mutex_lock: %s)",
```
vpodzime: I think this should be ``fprintf(stderr,...)``. And probably followed by ``fflush(); fsync()``.

libpromises/mutex.c
```diff
@@ -69,11 +67,10 @@ int __ThreadUnlock(pthread_mutex_t *mutex,
...
     if (result != 0)
     {
-        /* Log() is blocking on mutexes itself inside malloc(), so maybe not
-         * the best idea here. */
-        Log(LOG_LEVEL_ERR,
-            "Locking failure at %s:%d function %s! (pthread_mutex_unlock: %s)",
-            filename, lineno, funcname, GetErrorStrFromCode(result));
+        // printf instead of Log() to avoid problems with mutexes in Log()
+        printf("thread: %d, error: Locking failure at %s:%d function %s! (pthread_mutex_unlock: %s)",
+               GetCurrentThreadId(), filename, lineno, funcname, GetErrorStrFromCode(result));
```
vpodzime: Same here.

libutils/logging.c
```diff
@@ -178,7 +178,6 @@ static void LogToConsole(const char *msg, LogLevel level, bool color)
     if (level == LOG_LEVEL_DEBUG)
     {
         fflush(stdout);
-        // TODO also possibly call sync()? Not sure if that's too "expensive" or not.
```
vpodzime: This should not be part of this commit.

libutils/atexit.c
```diff
@@ -71,7 +71,11 @@ void RegisterAtExitFunction(AtExitFn fn)
...
     pthread_mutex_unlock(&atexit_functions_mutex);
-#endif
-
+/*
+    Don't register atexit() functions on windows due to race conditions 
+    around lock cleanup.
+*/
+#else 
     atexit(fn);
+#endif
```
vpodzime: Since this change is in *libutils* I think we need to make sure all the tools use the ``SafeExit()`` mechanism, right? Otherwise the functions registered here are never run, right?

cf-serverd/cf-serverd-functions.c
craigcomstock: I think we should. Just wasn't my focus during the bug fixing phase.

libpromises/mutex.c
craigcomstock: Will change. Agreed.

libutils/logging.c
craigcomstock: Right! Will fix.

libutils/atexit.c
craigcomstock: Yes. I will look for all use of RegisterAtExitFunction() and ensure that they also use SafeExit(). Probably move SafeExit to atexit.h/c for sharing purposes right!?

craigcomstock:
@vpodzime after looking through the code more it seems there are many places that exit() is used which can affect our binaries. I'm wondering if the best strategy is to refactor all those calls to something like SafeExit() or to rather redefine exit() on Windows with something like:

```
#if defined(__MINGW32__)
  #define exit(x) SafeExit(x)
#endif
```

So that we don't have to refactor so much code and the change is fairly minimal and "just works" as expected.

vpodzime:
> I'm wondering if the best strategy is to refactor all those calls to something like SafeExit() or to rather redefine exit() on Windows

I am definitely for using ``SafeExit()`` in all those places because that makes it explicit (for both the developers and the development/debugging tools). Redefining ``exit()`` could significantly complicate debugging and other contributors may find it really surprising.

craigcomstock:
Good. I liked the clarity of SafeExit() but then wondered about catching use of exit() in new code. I suppose we could add static analysis rules maybe to guard against that if we wanted.

craigcomstock:
after some testing might close this PR in favor of a new one with more thorough refactoring: https://github.com/cfengine/core/pull/3159

vpodzime:
> after some testing might close this PR in favor of a new one with more thorough refactoring: #3159

OK, closing. Feel free to reopen this if it's still needed.

PR #3126, opened 2018-05-18
Added acceptance tests for __main__ bundles

olehermanse:


olehermanse:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/569/)

olehermanse:
> What if there are two bundle agent main in the same file?

See `main_duplicate.x.cf`

nickanderson:
I see it has main and __main__. But what about __main__ and __main__ in the same file? I see the tests cover having __main__ in one file and then including another file that also contains it.

olehermanse:
> I see it has `main` and `__main__`. But what about `__main__` and `__main__` in the same file? I see the tests cover having `__main__` in one file and then including another file that also contains it.

You cannot define two bundles with the same name. `__main__` and `__main__` is the same as `main` and `main`.

I didn't exhaust all options (where do you stop), but I could add your example too.

tests/acceptance/00_basics/04_bundles/main_entry_point.cf
```diff
@@ -0,0 +1,12 @@
...
+{
+      inputs => { "main_entry_point.cf.sub" };
+}
+
+bundle agent __main__
+{
+    methods:
+        "imported" usebundle => imported();
+    reports:
+        "$(this.promise_filename) Pass";
```
vpodzime: Shouldn't we print ``Pass`` from in ``reports:`` in the bundle *imported* so that we know it really was processed?

tests/acceptance/00_basics/04_bundles/main_entry_point.cf.sub
```diff
@@ -0,0 +1,11 @@
+bundle agent __main__
+{
+    reports:
+        "$(this.promise_filename) FAIL"; # This shouldn't run
+}
+
+bundle agent imported
+{
+    reports:
+        "This file can be imported and the __main__ above will be ignored";
```
vpodzime: Shouldn't this be guarded with ``DEBUG::``?

libpromises/files_names.c
```diff
@@ -674,11 +674,29 @@ char *GetRealPath(const char *const path)
...
-    char *const real_path = realpath(abs_path, NULL);
+#else // Pre POSIX 2008 - realpath arg cannot be NULL
+    char *const path_buf = xcalloc(1, PATH_MAX);
+    char *real_path = realpath(abs_path, path_buf);
+    if (NULL_OR_EMPTY(real_path))
+    {
+        free(path_buf);
+        real_path = NULL;
+    }
+#endif
```
vpodzime: Why don't we always pass a statically allocated buffer of size ``PATH_MAX``? To me that looks better than this ifdef-else-endif machinery.

olehermanse: From realpath man page:

olehermanse: > BUGS

> The POSIX.1-2001 standard version of this function is broken by design, since it is impossible to determine a suitable size for the output buffer, resolved_path.
> According to POSIX.1-2001 a buffer of size PATH_MAX suffices, but PATH_MAX need not be a defined constant, and may have to be obtained using pathconf(3).
> And asking pathconf(3) does not really help, since, on the one hand POSIX warns that the result of pathconf(3) may be huge and unsuitable for mallocing memory, and on the other hand pathconf(3) may return -1 to signify that PATH_MAX is not bounded.
> The resolved_path == NULL feature, not standardized in POSIX.1-2001, but standardized in POSIX.1-2008, allows this design problem to be avoided.

vpodzime: Yes, but then you cannot rely on ``PATH_MAX`` at all.

olehermanse: Well, I don't have a good alternative. I prefer the default (linux) to use the NULL argument which is more correct, and safer. We do set `PATH_MAX` to 4096 by default, so this way, it should still work using PATH_MAX on exotics. (Our code already relies on `PATH_MAX`, in many other places.)

olehermanse:
@vpodzime I have addressed your comments. Merging and cherry-picking.

vpodzime:
> @vpodzime I have addressed your comments. Merging and cherry-picking.

Thanks!

PR #3127, opened 2018-05-18
Don't require codecov to pass in travis

olehermanse:


codecov.yaml
```diff
@@ -1,6 +1,6 @@
 codecov:
   notify:
-    require_ci_to_pass: yes
+    require_ci_to_pass: no
```
vpodzime: I think this does a different thing. At least judging by the name of the config item and e.g. this sample from elsewhere:
```
Codecov:
  # yes: will delay sending notifications until all ci is finished
  notify:
    require_ci_to_pass: yes
```

PR #3128, opened 2018-05-18
Examples for bundle __main__ documentation

nickanderson:


examples/bundlesequence.cf
```diff
@@ -0,0 +1,20 @@
+# Example showing how to override the default bundlesequence
+body common control
+{
+      bundlesequence  => { "main" };
```
olehermanse: Should be hello, I guess.

nickanderson: oh your awake still

olehermanse: Of course. I just stopped working.

PR #3135, opened 2018-05-22
INF-402 pattern-matching

Lex-2008:
try fixing building on new AIX


Lex-2008:
works without it - issue was that something in out buildsystem doesn't tolerate /home/jenkins being symlinked to other partition

PR #3144, opened 2018-05-30
CFE-2194: Do not render templates when passed invalid data

vpodzime:
Passing no template_data should result in using datastate()
instead as our documentation says. But passing invalid data is a
different case that should result in an error and the template
not being rendered.


vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/620/)

PR #3145, opened 2018-05-30
CFE-2125: Do not iterate over JSON objects' properties in mustache

vpodzime:
Only lists are supposed to be iterated over. Otherwise duplicates
appear when using JSON with objects that have multiple
properties. See CFE-2125 for details.

However, our '{{@}}' extension to the mustache specification
requires JSON objects' properties to be iterated over because
it's the names of these properties (keys) it should be
substituted with.

Changelog: title
(cherry picked from commit 1d3c7313b85c2d27dcfdf29f51e5fe038f12f9e9)

### DIFF
--- a/libutils/mustache.c
+++ b/libutils/mustache.c
@@ -525,6 +525,124 @@ static bool SetDelimiters(const char *content, size_t content_len,
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


vpodzime:
Unrelated spurious failure on Solaris, everything else is green.

PR #3151, opened 2018-06-06
Support real-life reports in cf-testd

vpodzime:
Real-life reports have newlines in the CSV data rows so a more
clever approach has to be taken to store and load such
data. Length-prefixed data is a simple but powerful approach
which can easily be supported.

### DIFF
--- a/cf-testd/cf-testd.c
+++ b/cf-testd/cf-testd.c
@@ -76,7 +76,8 @@ static bool TERMINATE = false;
 typedef struct
 {
     char *report_file;
-    char *report;
+    char *report_data;
+    Seq *report;
     int report_len;
     char *key_file;
     RSA *priv_key;
@@ -133,7 +134,7 @@ CFTestD_Config *CFTestD_ConfigInit()
 void CFTestD_ConfigDestroy(CFTestD_Config *config)
 {
     free(config->report_file);
-    free(config->report);
+    SeqDestroy(config->report);
     free(config->key_file);
     free(config->address);
     free(config);
@@ -278,8 +279,8 @@ bool CFTestD_GetServerQuery(
     ServerConnectionState *conn, char *recvbuffer, CFTestD_Config *config)
 {
     char query[CF_BUFSIZE];
-    const int report_len            = config->report_len;
-    const char *const report        = config->report;
+    Seq *report          = config->report;
+    const int report_len = config->report_len;
     ConnectionInfo *const conn_info = conn->conn_info;
 
     query[0] = '\0';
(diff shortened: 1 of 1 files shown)


vpodzime:
This is doomed to fail in Travis because it requires changes done in https://github.com/cfengine/nova/pull/1193

cf-testd/cf-testd.c
```diff
@@ -632,8 +633,32 @@ static void *CFTestD_ServeReport(void *config_arg)
...
+        SeqAppend(report, ts);
+
+        /* start right after the newline after the timestamp header */
+        char *position = ts + ts_ref.len + 1;
+        char *row;
+        size_t row_len;
+        while (CFTestD_GetRow(position, &row, &row_len))
+        {
+            *(row + row_len) = '\0';
+            SeqAppend(report, row);
```
craigcomstock: How is the memory for `char *row` allocated and freed?

cf-testd/cf-testd.c
```diff
@@ -632,8 +633,32 @@ static void *CFTestD_ServeReport(void *config_arg)
             Log(LOG_LEVEL_ERR, "Error reading report file '%s'", report_file);
             exit(EXIT_FAILURE);
         }
-        config->report     = StringWriterClose(contents);
-        config->report_len = SafeStringLength(config->report);
+
+        size_t report_data_len = StringWriterLength(contents);
+        char *report_data = StringWriterClose(contents);
+
```
craigcomstock: Also wondering how StringWriterClose() works... seems like we are returned `char *data = writer->string.data` and that function does a `free(writer)` so wouldn't the memory pointed to by `return data` be freed?


```
char *StringWriterClose(Writer *writer)
// NOTE: transfer of ownership for allocated return value
{
    if (writer->type != WT_STRING)
    {
        ProgrammingError("Wrong writer type");
    }
    char *data = writer->string.data;

    free(writer);
    return data;
}
```

Do we need to free `char *report_data`?

olehermanse: @craigcomstock yes, StringWriterClose gives you "ownership" of the string, so you are correct in saying that report_data needs to be freed.

cf-testd/cf-testd.c
```diff
@@ -632,8 +633,32 @@ static void *CFTestD_ServeReport(void *config_arg)
...
+        Seq *report = SeqNew(64, NULL);
```
olehermanse: @craigcomstock `NULL` here means that there is no destroy function, the elements are not freed when doing `SeqDestroy()`

cf-testd/cf-testd.c
olehermanse: It's not. From the looks of it, `row` and the elements in `report` only point to different parts of the buffer allocated by the string writer. (`report_data`).

cf-testd/cf-testd.c
```diff
@@ -632,8 +633,32 @@ static void *CFTestD_ServeReport(void *config_arg)
...
+        StringRef ts_ref = StringGetToken(report_data, report_data_len, 0, "\n");
+        char *ts = (char *) ts_ref.data;
+        *(ts + ts_ref.len) = '\0';
+        SeqAppend(report, ts);
+
+        /* start right after the newline after the timestamp header */
+        char *position = ts + ts_ref.len + 1;
+        char *row;
+        size_t row_len;
+        while (CFTestD_GetRow(position, &row, &row_len))
```
olehermanse: I would prefer if you call it "report lines" instead of rows, to attempt some consistency: https://github.com/cfengine/nova/blob/master/REPORTING.md

(I know this has not been good in the past, and in CSV it's called rows).

olehermanse:
@vpodzime Looks okay, except for the `free()` @craigcomstock pointed out.

olehermanse:
@cf-bottom Can you please trigger a build in Jenkins for this PR?

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com/job/pr-pipeline/727/)

https://ci.cfengine.com/job/pr-pipeline/727/

vpodzime:
@cf-bottom Can you please trigger a build in Jenkins for this PR?

PR #3158, opened 2018-06-12
Bundle name is now part of the log message when aborting a bundle (3.10)

olehermanse:


olehermanse:
@nickanderson @craigcomstock @vpodzime Do you think it's appropriate to backport this log change back to 3.10 ?

atsaloli:
Yay! Thank you.  This will make our lives easier once we get onto a version that has this fix.  We haven't started using 3.12 yet.

PR #3163, opened 2018-06-19
Increase the connection backlog/queue size in cf-serverd

vpodzime:
man:tcp(7)).

Kernel has its own limit for this number (per socket) and it
defaults to 128. So let's use this maximal possible value,
there's nothing to lose for us. We have a capability to handle
many connections in threads (specified in 'body server control'
-> maxconnections), but if too many connections are attempted at
once, we need a big enough backlog/queue to give us enough time
to process them. Starting threads may just not be fast enough,
especially in combination with our bad way of counting threads
and other things using locks all the threads are fighting over
for.

### DIFF
--- a/cf-serverd/cf-serverd-functions.c
+++ b/cf-serverd/cf-serverd-functions.c
@@ -49,7 +49,7 @@
 #include <printsize.h>
 
 
-static const size_t QUEUESIZE = 50;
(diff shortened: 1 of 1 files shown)


vpodzime:
This is not supposed to be merged, it's a 3.10.1-based branch used for creating packages.

PR #3164, opened 2018-06-19
Increase the connection backlog/queue size in cf-serverd

vpodzime:
man:tcp(7)).

Kernel has its own limit for this number (per socket) and it
defaults to 128. So let's use this maximal possible value,
there's nothing to lose for us. We have a capability to handle
many connections in threads (specified in 'body server control'
-> maxconnections), but if too many connections are attempted at
once, we need a big enough backlog/queue to give us enough time
to process them. Starting threads may just not be fast enough,
especially in combination with our bad way of counting threads
and other things using locks all the threads are fighting over
for.


vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/786/)

cf-serverd/cf-serverd-functions.c
```diff
@@ -684,6 +688,26 @@ static void AcceptAndHandle(EvalContext *ctx, int sd)
...
+    if (queue_size_var != NULL)
+    {
+        long queue_size;
+        int ret = StringToLong(queue_size_var, &queue_size);
+        if ((ret == 0) && (queue_size > 0) && (queue_size <= MAX_LISTEN_QUEUE_SIZE))
+        {
+            return (size_t) queue_size;
+        }
+        Log(LOG_LEVEL_VERBOSE,
+            "$CF_SERVERD_LISTEN_QUEUE_SIZE = '%s' doesn't specify a valid number for listen queue size, "
```
craigcomstock: maybe INFO level? How often would this be called? Just every time cf-serverd is started from the command line or ?

vpodzime: Yeah, that's a good point. This only happens once. It should actually be a warning.

vpodzime:
The only failure in jenkins was a nested RHEL 4 VM that died before it even did anything.

PR #3165, opened 2018-06-20
3.10.x - Increase the connection backlog/queue size in cf-serverd

vpodzime:
man:tcp(7)).

Kernel has its own limit for this number (per socket) which is an
upper bound and it defaults to 128 [1]. So let's use this maximal
possible value as a default, there's nothing to lose for us. We
have a capability to handle many connections in threads
(specified in 'body server control' -> maxconnections), but if
too many connections are attempted at once, we need a big enough
backlog/queue to give us enough time to process them. Starting
threads may just not be fast enough, especially in combination
with our bad way of counting threads and other things using locks
all the threads are fighting over for.

[1] SOMAXCONN (defined in <sys/socket.h>), tunable on GNU/Linux
    via sysctl, other systems differ (e.g. 4096 on HP-UX)

And we should make this tunable, i.e. possible to change the
queue/backlog size without recompilation of the cf-serverd
binary. An environment variable is a simple and good enough
mechanism to do this (also being used in other similar places).

Ticket: CFE-2838
(cherry picked from commit 73acb17406b13f9a29e87c74165e466b1a792e22)

Conflicts:
  cf-serverd/cf-serverd-functions.c

### DIFF
--- a/cf-serverd/cf-serverd-functions.c
+++ b/cf-serverd/cf-serverd-functions.c
@@ -44,12 +44,16 @@
 #include <sysinfo.h>
 #include <time_classes.h>
 #include <connection_info.h>
+#include <string_lib.h>
 #include <file_lib.h>
 #include <loading.h>
 #include <printsize.h>
 
 
-static const size_t QUEUESIZE = 50;
+/* see man:listen(3) */
+#define DEFAULT_LISTEN_QUEUE_SIZE 128
+#define MAX_LISTEN_QUEUE_SIZE 2048
+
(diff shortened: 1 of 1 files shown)


vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/788/)

vpodzime:
This needs one more proper run through Jenkins, the previous one failed because of a spurious issue on a touchstone build and thus didn't really triggered the other builds.

vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/794/)

PR #3166, opened 2018-06-21
3.12.x - Increase the connection backlog/queue size in cf-serverd

vpodzime:
man:tcp(7)).

Kernel has its own limit for this number (per socket) which is an
upper bound and it defaults to 128 [1]. So let's use this maximal
possible value as a default, there's nothing to lose for us. We
have a capability to handle many connections in threads
(specified in 'body server control' -> maxconnections), but if
too many connections are attempted at once, we need a big enough
backlog/queue to give us enough time to process them. Starting
threads may just not be fast enough, especially in combination
with our bad way of counting threads and other things using locks
all the threads are fighting over for.

[1] SOMAXCONN (defined in <sys/socket.h>), tunable on GNU/Linux
    via sysctl, other systems differ (e.g. 4096 on HP-UX)

And we should make this tunable, i.e. possible to change the
queue/backlog size without recompilation of the cf-serverd
binary. An environment variable is a simple and good enough
mechanism to do this (also being used in other similar places).

Ticket: CFE-2838
(cherry picked from commit a8676e86517bf2c760517113c74ed95d4faae355)

### DIFF
--- a/cf-serverd/cf-serverd-functions.c
+++ b/cf-serverd/cf-serverd-functions.c
@@ -45,13 +45,17 @@
 #include <sysinfo.h>
 #include <time_classes.h>
 #include <connection_info.h>
+#include <string_lib.h>
 #include <file_lib.h>
 #include <loading.h>
 #include <printsize.h>
 
 #define WAIT_INCOMING_TIMEOUT 10
 
-static const size_t QUEUESIZE = 50;
+/* see man:listen(3) */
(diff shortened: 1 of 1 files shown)


vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/792/)

vpodzime:
jenkins all green except for an unrelated issue on one of the nested VMs

PR #3168, opened 2018-06-21
CFE-2844: Make sure 'augments' from def.json are applied last

vpodzime:
These are supposed to be most specialized and finest-granularity
policies (based on e.g. $(sys.flavor) and similar things). So
they should override values of any variables defined in 'vars' in
def.json if they defined the same variables, of course.

See the discussion in CFE-2741 for more details.

Changelog: CFE-2844: Apply augments after vars, classes and inputs in def.json
(cherry picked from commit 8bdc87c26f69e1886ac863bc9e62b68b6125f75e)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -92,6 +92,7 @@
 	  variables raises no security concerns and can be considered
 	  valid. (CFE-1915)
 	- Classes failsafe_fallback and bootstrap_mode are now reported by default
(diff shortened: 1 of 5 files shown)


vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/793/)

PR #3172, opened 2018-07-04
Better support for FHS-compliant installations

vpodzime:
So that we can make our way to modern distributions.


basvandervlies:
+1

vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/821/)

tests/acceptance/serverd-multi-versions.sh
```diff
@@ -49,7 +49,9 @@ function install(){
     rm -rf "$1"
     if test -z "$2"
     then
-        make install prefix=$PWD/$1/var/cfengine 2>&1 >/dev/null
+        ./configure --enable-debug --with-tokyocabinet --with-init-script --prefix=$PWD/$1/var/cfengine 2>&1 >/dev/null
+        make 2>&1 >/dev/null
+        make install 2>&1 >/dev/null
```
jimis: This will redirect stderr to stdout, but will not supress it.

tests/acceptance/serverd-multi-versions.sh
```diff
@@ -49,7 +49,9 @@ function install(){
     rm -rf "$1"
     if test -z "$2"
     then
-        make install prefix=$PWD/$1/var/cfengine 2>&1 >/dev/null
```
jimis: It is normal to have `prefix=` in the `make install` command. Also `destdir=` is even more common. So these should still work.

travis-scripts/script.sh
```diff
@@ -23,10 +23,10 @@ then
     # On osx the default gcc is actually LLVM
     export CC=gcc-7
     NO_CONFIGURE=1 ./autogen.sh
-    ./configure --enable-debug --prefix=$INSTDIR --with-init-script --with-lmdb=/usr/local/Cellar/lmdb  --with-openssl=/usr/local/opt/openssl
+    ./configure --enable-debug --prefix=$INSTDIR --bindir=$INSTDIR/var/cfengine/bin --with-init-script --with-lmdb=/usr/local/Cellar/lmdb  --with-openssl=/usr/local/opt/openssl
```
jimis: This looks weird. Maybe all you want to do is a plain `./configure` and then `make destdir=$INSTDIR install` ?

tests/acceptance/serverd-multi-versions.sh
vpodzime: What do you mean? It's redirected to ``/dev/null``, how more would you want to suppress it?

tests/acceptance/serverd-multi-versions.sh
vpodzime: This would only be possible if we started searching for our binaries in ``$PATH`` or something like that. Otherwise specifying ``prefix=`` in ``make install`` changes the paths too late for our code to adapt to it. Or do you have any suggestions how to make this work? Everything that comes to my mind is just worse than not supporting ``prefix=`` in ``make install``.

vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/822/)

vpodzime:
@jimis any suggestions/comments how to move on with this (see the above questions)? I have managed to make some nice and small builds in [a Copr repository](https://copr.fedorainfracloud.org/coprs/vpodzime/CFEngine/build/774577/) with these changes so I'd like to merge them and move on.

tests/acceptance/serverd-multi-versions.sh
Lex-2008: because that's how redirects work - they are evaluated left-to-right, and assign values, not pointers.

So, by default output goes like this:
1=/dev/stdout
2=/dev/stderr
After first assignment (`2>&1`) they are like this:
1=/dev/stdout
2=/dev/stdout
After second assignment (`>/dev/null`) they are like this:
1=/dev/null
2=/dev/stdout

To make it correct, you need to reorder them:
```
>/dev/null 2>&1
```

I've just tested it:
```
lex@flower:~$ cat test.sh
echo err >&2
lex@flower:~$ bash test.sh 2>&1 >/dev/null
err
lex@flower:~$ bash test.sh >/dev/null 2>&1
lex@flower:~$ _
```

libpromises/known_dirs.c
```diff
@@ -104,6 +107,21 @@ const char *GetWorkDir(void)
     return workdir == NULL ? GetDefaultWorkDir() : workdir;
 }
 
+const char *GetBinDir(void)
+{
+    const char *workdir = getenv("CFENGINE_TEST_OVERRIDE_WORKDIR");
+
+    if (workdir == NULL)
+    {
+        return GetDefaultBinDir();
```
Lex-2008: I'm not good at C, but I don't see this function defined anywhere

Lex-2008:
GetDefaultBinDir

libpromises/known_dirs.c
vpodzime: It's defined by a weird macro in that file. Took me a while to figure that out too and I consider myself quite good at C. :smile:

tests/acceptance/serverd-multi-versions.sh
vpodzime: Wow, I had no idea the order matters like this here! Thanks for the analysis, @Lex-2008! Fixing.

vpodzime: Oh, and we had it wrong even before my change.

jimis:
> @jimis any suggestions/comments how to move on with this (see the above questions)? I have managed to make some nice and small builds in a Copr repository with these changes so I'd like to merge them and move on.

@vpodzime Nice! I don't have specific fixes to suggest (did not review it again) so it's up to you in the end. The goal (getting a distro-compliant package) is solid so don't let the technicalities stop you. ;-)

jimis:
+1 code dollar to @Lex-2008 !

PR #3173, opened 2018-07-07
Added a log message for the case that the output file has ceased to exist by the time cf-execd tries to mail it.

sp1ff:
Admittedly rare, but the absence of this log statement made trouble-shooting an issue more difficult for me.

### DIFF
--- a/cf-execd/cf-execd-runner.c
+++ b/cf-execd/cf-execd-runner.c
@@ -583,6 +583,7 @@ static void MailResult(const ExecConfig *config, const char *file)
         struct stat statbuf;
         if (stat(file, &statbuf) == -1)
         {
+            Log(LOG_LEVEL_ERR, "Mail report: failed to stat file '%s' [errno: %d]", file, errno);
             return;
         }
 


cf-execd/cf-execd-runner.c
```diff
@@ -583,6 +583,7 @@ static void MailResult(const ExecConfig *config, const char *file)
         struct stat statbuf;
         if (stat(file, &statbuf) == -1)
         {
+            Log(LOG_LEVEL_DEBUG, "Mail report: failed to stat file '%s' [errno: %d]", file, errno);
```
vpodzime: I think this should be `LOG_LEVEL_ERR`. There's a similar place below logging that the file failed to open using the `ERR` level. And it makes sense to me that this should be reported not only when running in the `--debug` mode.

sp1ff:
@vpodzime Fair 'nuf-- updated.

sp1ff:
Sure NP-- squashed.

vpodzime:
> Sure NP-- squashed.

Hmm, it went from 2 commits to 5 commits including the original 2. That's not really how I imagined squashing. :smile:

sp1ff:
My bad-- I see one commit, now.

vpodzime:
> My bad-- I see one commit, now.

Amazing. Thanks!

vpodzime:
https://github.com/cfengine/core/pull/3184
https://github.com/cfengine/core/pull/3185

PR #3181, opened 2018-07-17
ENT-3959 Extend the test to cover more cases

nickanderson:


vpodzime:
@nickanderson the modified test seems to have failed in travis. Does it pass for you locally?

nickanderson:
@vpodzime  good, I was expecting it to fail the test. will note in the related issue.

PR #3187, opened 2018-07-23
ENT-3830: Added thread-safe stack data structure.

karlhto:
Has no tests as of currently. Just created PR for feedback on the current code.


libutils/stack.c
```diff
@@ -0,0 +1,129 @@
...
+static void DestroyRange(Stack *stack, int start, int end);
+static void ExpandIfNecessary(Stack *stack);
+
+Stack *StackNew(size_t initial_capacity, void (ItemDestroy) (void *item))
+{
+    Stack *stack = xmalloc(sizeof(Stack));
+
+    if (initial_capacity <= 0)
+    {
+        initial_capacity = 1;
```
vpodzime: I don't think this is a good default for initial capacity. Small stacks can/should be explicitly requested so I'd set this to at least `16`. That would also make sure that the expansion is not happening too often for the first few items added.

libutils/stack.c
```diff
@@ -0,0 +1,129 @@
...
+
+    ThreadLock(stack->lock);
+    ExpandIfNecessary(stack);
+
+    stack->data[++stack->top] = data;
+
+    ThreadUnlock(stack->lock);
+}
+
+int StackSize(Stack const *stack)
```
vpodzime: I think the word *size* is ambiguous -- it can mean the number of items as well as the capacity. But I don't really have a better short word for this. OTOH, I wouldn't mind naming the function like `StackGetNumItems`.

libutils/stack.c
```diff
@@ -0,0 +1,129 @@
...
+{
+    assert(stack);
+    return stack->top + 1;
+}
+
+static void DestroyRange(Stack *stack, int start, int end)
+{
+    if (stack->ItemDestroy)
+    {
+        for (int i = start; i < end; i++)
```
vpodzime: `start` and `end` should be checked first.

libutils/stack.h
```diff
@@ -0,0 +1,50 @@
...
+void *StackPop(Stack *stack);
+
+/**
+  @brief Get current size of stack.
+  @note On NULL stack, returns 0.
+  @param [in] stack The stack.
+  @return The amount of elements in the stack.
+  */
+int StackSize(Stack const *stack);
+
```
vpodzime: Would be nice to also have `StackCopy`.

libutils/stack.c
olehermanse: +1

libutils/stack.c
```diff
@@ -0,0 +1,129 @@
...
+{
+    assert(stack);
+
+    ThreadLock(stack->lock);
+    void *data = stack->data[stack->top];
+    stack->data[stack->top] = NULL;
+
+    if (stack->top > STACK_BOTTOM)
+    {
+        stack->top--;
```
olehermanse: I would clarify this to 

```
stack->top -= 1;
```

libutils/stack.c
```diff
@@ -0,0 +1,129 @@
...
+
+void *StackPop(Stack *stack)
+{
+    assert(stack);
+
+    ThreadLock(stack->lock);
+    void *data = stack->data[stack->top];
+    stack->data[stack->top] = NULL;
+
+    if (stack->top > STACK_BOTTOM)
```
olehermanse: You could consider using local variables for top and data. (Just make sure you assign correctly).

libutils/stack.c
```diff
@@ -0,0 +1,129 @@
...
+        free(stack);
+    }
+}
+
+void *StackPop(Stack *stack)
+{
+    assert(stack);
+
+    ThreadLock(stack->lock);
+    void *data = stack->data[stack->top];
```
olehermanse: please call this something else, like `element`, to avoid confusion with `stack->data`.

libutils/stack.c
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    ThreadUnlock(stack->lock);
+
+    return data;
+}
+
+void StackPush(Stack *stack, void *data)
+{
+    assert(stack);
```
olehermanse: +1 I would add `assert(stack)` to all the functions.

olehermanse:
@karlhto Looks good so far! I would add unit tests now, before you add more. They can be very useful, even if they don't test the multithreading in this case.

karlhto:
@olehermanse @vpodzime Please review new changes. Adding unit tests later today.

olehermanse:
@cf-bottom Could you test this in the jenkins pipeline?

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com/job/pr-pipeline/877/)

https://ci.cfengine.com/job/pr-pipeline/877/

olehermanse:
@cf-bottom Trigger another jenkins build for this PR.

karlhto:
I actually added that, but I didn't commit it yet

libutils/stack.c
```diff
@@ -0,0 +1,174 @@
...
+    ThreadLock(stack->lock);
+
+    top = &stack->top;
+    data = stack->data;
+    item = data[*top];
+    data[*top] = NULL;
+
+    if (*top > STACK_BOTTOM)
+    {
+        *top -= 1;
```
olehermanse: I don't think having top be an int* is very readable.

This is better, imo:

```

File: libutils/stack.c (hunk 2 above)
Comment (olehermanse) [reply libutils/stack.c]: I disagree, I think size is nicer. So up to you @karlhto .

**Comment by karlhto** [comment]
@vpodzime Since I don't necessarily know the size of the data the pointers point to, I assume you only want a shallow copy? The problem with that is that when either the original or the copied stack is destroyed, the other one will have a bunch of pointers to freed memory. Or is there something I'm missing?

**Comment by karlhto** [comment]
Another solution would be to just use a size parameter, or maybe a "sizeof" function parameter that returns the actual size of the element. Like strlen for strings, for instance:
```
Stack *s = StackNew(0, free);

StackPush(s, xstrdup("1"));
StackCopy(strlen);
```

**Comment by vpodzime** [comment]
> The problem with that is that when either the original or the copied stack is destroyed, the other one will have a bunch of pointers to freed memory.

That's what they call ``C``. We have to live with it.

**Comment by karlhto** [comment]
@olehermanse @vpodzime Should be done now. Please review.

**Comment by karlhto** [comment]
@cf-bottom jenkins please

**Comment by cf-bottom** [comment]
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com/job/pr-pipeline/903/)

https://ci.cfengine.com/job/pr-pipeline/903/

**Comment by karlhto** [comment]
@cf-bottom jenkins please

**Comment by cf-bottom** [comment]
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com/job/pr-pipeline/905/)

https://ci.cfengine.com/job/pr-pipeline/905/

**Comment by vpodzime** [comment]
A build including the exotics:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/908/)

File: libutils/stack.c [hunk 10]
```diff
@@ -0,0 +1,212 @@
...
+{
+    Stack *stack = xmalloc(sizeof(Stack));
+
+    if (initial_capacity <= 0)
+    {
+        initial_capacity = DEFAULT_CAPACITY;
+    }
+
+    pthread_mutex_t lock = PTHREAD_ERRORCHECK_MUTEX_INITIALIZER_NP;
+    stack->lock = lock;
```
vpodzime: I think these two lines can be squashed.

libutils/stack.c
```diff
@@ -0,0 +1,212 @@
...
+void StackDestroy(Stack *stack)
+{
+    if (stack != NULL)
+    {
+        ThreadLock(&stack->lock);
+        DestroyRange(stack, STACK_EMPTY, stack->size);
+        ThreadUnlock(&stack->lock);
+    }
+
+    StackSoftDestroy(stack);
```
vpodzime: I think there's a race condition here where something can push to the stack while it's being destroyed.

libutils/stack.c
```diff
@@ -0,0 +1,212 @@
...
+        free(stack->data);
+        free(stack);
+    }
+}
+
+void *StackPop(Stack *stack)
+{
+    assert(stack != NULL);
+
+    ThreadLock(&stack->lock);
```
vpodzime: I think this should check `if (stack->lock != NULL)` because of thread-safety with concurrent `Destroy()`.

libutils/stack.c
```diff
@@ -0,0 +1,212 @@
...
+}
+
+static void DestroyRange(Stack *stack, int start, int end)
+{
+    assert(stack != NULL);
+
+    if (stack->ItemDestroy)
+    {
+        assert(start >= STACK_EMPTY);
+        assert(end <= stack->capacity);
```
vpodzime: These checks won't be effective in the production-compiled code and yet they might be quite crucial.

libutils/stack.h
```diff
@@ -0,0 +1,97 @@
...
+Stack *StackNew(size_t initial_capacity, void (*ItemDestroy) ());
+
+/**
+  @brief Destroys the stack and frees the memory it occupies.
+  @param [in] stack The stack to destroy.
+  */
+void StackDestroy(Stack *stack);
+
+/**
+  @brief Frees the pointer to the data array and the stack pointer itself.
```
vpodzime: It's not really free`ing the pointers, it's freeing the memory.

libutils/stack.c
vpodzime: The only way to make this thread-safe would be to pass the lock as a separate argument to every function and keep it outside of the (private) `struct`. Otherwise the struct cannot be free'd while holding the lock. But having it like that is not nice either because of the lock not really being coupled with the structure it is supposed to be used with.
So I think we just need a `@warn` in the doxygen comments making it clear that the stack should never be destroyed while some other thread can still possibly be using it.

libutils/stack.c
```diff
@@ -0,0 +1,212 @@
...
+
+    return capacity;
+}
+
+bool StackIsEmpty(Stack *stack)
+{
+    assert(stack != NULL);
+
+    ThreadLock(&stack->lock);
+    bool empty = stack->size == STACK_EMPTY;
```
olehermanse:
```
const bool empty = (stack->size == STACK_EMPTY);
```

libutils/stack.c
```diff
@@ -0,0 +1,212 @@
...
+    new_stack->size = stack->size;
+
+    ThreadUnlock(&stack->lock);
+
+    return new_stack;
+}
+
+static void DestroyRange(Stack *stack, int start, int end)
+{
+    assert(stack != NULL);
```
olehermanse: Maybe add comments that these static functions assume that lock is already acquired.

libutils/stack.c
karlhto: I think this could be solved by just creating a new lock that is equal to the old one, and have the last thing you do being destroying the lock.

libutils/stack.c
```diff
@@ -0,0 +1,212 @@
...
+    stack->ItemDestroy = ItemDestroy;
+
+    return stack;
+}
+
+void StackDestroy(Stack *stack)
+{
+    if (stack != NULL)
+    {
+        ThreadLock(&stack->lock);
```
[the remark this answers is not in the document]
karlhto: What should it do in case of failure?
For instance, if it fails to unlock the thread after it has retrieved the data.

libutils/stack.c
karlhto: You could also have an if-check for if the stack exists after every ThreadLock, but then you'd have to hold the lock in a variable in every function.

libutils/stack.c
karlhto: This doesn't work. PTHREAD_ERRORCHECK_MUTEX_INITIALIZER_NP needs to be in the declaration, since it is an array initializer like {0}

libutils/stack.c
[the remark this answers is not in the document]
olehermanse: I think the more normal case would be that it fails to acquire the lock? Then it has to either return NULL for pop, false for push. But then there is no way to tell if you popped NULL or there was an error. The other approach would be to have a boolean return value for both, and an output parameter for pop. I think returning NULL is more clean.

Regarding unlock, I don't know. How can this happen, realistically? Is it valid to retry? Exit?

libutils/stack.c
vpodzime: Do you think it's safe to just assign it to something else?

libutils/stack.c
vpodzime: There would still be a race condition, AFAIK, because you cannot destroy the lock while holding it.

libutils/stack.c
karlhto: You're right. For some reason I thought pthread_mutex structs would work differently, but I definitely can't assign it like that.

libutils/stack.c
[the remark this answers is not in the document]
karlhto: Wouldn't it failing to acquire the lock be a result of a critical error? According to the documentation, the ways pthread_mutex_lock can fail are: 
- if the mutex does not exist
- if there is a deadlock, or the lock has already been locked by the calling thread

pthread_mutex_unlock:
- if the mutex does not exist
- if the calling thread does not currently own the lock

I'm not sure what is best practice here.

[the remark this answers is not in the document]
karlhto: What about keeping the mutexes locked? If you check out dbm_api.c, the closing functions just keep the threads locked until the program exits. Honestly, that doesn't look like the best solution, either.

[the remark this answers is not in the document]
karlhto: @olehermanse @vpodzime ^

libutils/mutex.c
```diff
@@ -51,30 +51,26 @@ int __ThreadLock(pthread_mutex_t *mutex,
...
     if (result != 0)
     {
-        /* Log() is blocking on mutexes itself inside malloc(), so maybe not
-         * the best idea here. */
-        Log(LOG_LEVEL_ERR,
+        fprintf(stderr,
             "Locking failure at %s:%d function %s! (pthread_mutex_lock: %s)",
             filename, lineno, funcname, GetErrorStrFromCode(result));
-        return false;
+        exit(255);
```
karlhto: This is probably not a good idea.

olehermanse: I think that's good. Considering the situations you mentioned, I can't think of a better resolution. I would prefer exit code 101 though. This is what cf-hub uses for unexpected / programming error.

libutils/stack.c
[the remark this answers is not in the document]
olehermanse: In that case, exiting seems good.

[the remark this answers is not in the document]
vpodzime: I agree with Ole that exiting would be appropriate in those cases.

karlhto:
@cf-bottom jenkins will probably fail

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/967/)

https://ci.cfengine.com/job/pr-pipeline/967/

karlhto:
@cf-bottom jenkins needs to run the acceptance tests as well

karlhto:
Started new build:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/)

vpodzime:
Analyzing *cf-serverd* I realized we will need one more function, something like:
```
bool StackPushReportCount(void *item, size_t *size)
```
i.e. a combination of `StackPush()` and `StackCount()`, but atomic and only acquiring the lock once.

karlhto:
@vpodzime Two things:

- Do you have an example of a function that is like that in the codebase already? It sounds simple enough, but just want to see if there is anything else I need.
- Could you check out the aix build of the build I started to check what went wrong? It seems like it failed in setup phase, and not due to changes in the code?

vpodzime:
> Do you have an example of a function that is like that in the codebase already? It sounds simple enough, but just want to see if there is anything else I need.

What do you mean by *"that is like that"*? A function that returns multiple things using out arguments? There are a couple. Or do you mean a function that would require this functionality I suggested? I don't think it's anywhere there, but I was thinking about some adapting code that would get more resources (e.g. threads) when the stack grows too much.

vpodzime:
> Why does it need to have bool, for instance? Wouldn't it be better to return the size instead?

Oh, that's just because I thought `StackPush()` returns `bool`. Which might not be a bad idea because it may fail, e.g. we could change the stack to have a size limit.

karlhto:
> What do you mean by "that is like that"? A function that returns multiple things using out arguments? There are a couple. Or do you mean a function that would require this functionality I suggested? I don't think it's anywhere there, but I was thinking about some adapting code that would get more resources (e.g. threads) when the stack grows too much.

Functions that have the functionality you suggested. I am not entirely sure how you want me to implement it.

> Oh, that's just because I thought StackPush() returns bool. Which might not be a bad idea because it may fail, e.g. we could change the stack to have a size limit.

I definitely agree with this, but I need suggestions for what such a size limit would be, and if it should be configurable.

vpodzime:
> I am not entirely sure how you want me to implement it.

I meant just pushing to the stack and returning it's new size/count/length.

vpodzime:
> I definitely agree with this, but I need suggestions for what such a size limit would be, and if it should be configurable.

We don't need to do everything now. :wink: This unlimited version is good for now, we will see what the future brings.

karlhto:
@vpodzime I have addressed your request 👌

karlhto:
@cf-bottom the jenkins

karlhto:
Started yet another build:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/)

libutils/stack.h
```diff
@@ -0,0 +1,106 @@
...
+  */
+size_t StackPushReportCount(Stack *stack, void *item);
+
+/**
+  @brief Get current number of items in stack.
+  @note On NULL stack, returns 0.
+  @param [in] stack The stack.
+  @return The amount of elements in the stack.
+  */
+size_t StackCount(Stack *stack);
```
vpodzime: This function as well as the few ones below should take `const Stack *stack` because they don't modify it.

libutils/mutex.c
```diff
@@ -43,39 +43,39 @@ pthread_mutex_t *cft_server_children = &MUTEXES[3]; /* GLOBAL_T */
...
                  const char *funcname, const char *filename, int lineno)
 
 {
     int result = pthread_mutex_lock(mutex);
 
     if (result != 0)
     {
-        /* Log() is blocking on mutexes itself inside malloc(), so maybe not
-         * the best idea here. */
-        Log(LOG_LEVEL_ERR,
```
vpodzime: I think the comment should stay in some form to explain why `Log` is not being used here.

libutils/mutex.c
```diff
@@ -43,39 +43,39 @@ pthread_mutex_t *cft_server_children = &MUTEXES[3]; /* GLOBAL_T */
...
+        fprintf(stderr,
             "Locking failure at %s:%d function %s! (pthread_mutex_lock: %s)",
```
vpodzime: This needs a newline at the end now.

libutils/mutex.h
```diff
@@ -36,9 +36,9 @@ extern pthread_mutex_t *cft_server_filter;
 #define ThreadLock(m)     __ThreadLock(m, __func__, __FILE__, __LINE__)
```
vpodzime: I don't think the `extern` `pthread_mutex_t` variables above this really belong to *libutils*. Maybe we should have something like *libpromises/locks.h*?

tests/unit/stack_test.c
```diff
@@ -0,0 +1,119 @@
...
+
+    Stack *new_stack = StackCopy(stack);
+
+    assert(new_stack != NULL);
+
+    char *old_str1 = StackPop(stack); char *new_str1 = StackPop(new_stack);
+    char *old_str2 = StackPop(stack); char *new_str2 = StackPop(new_stack);
+    char *old_str3 = StackPop(stack); char *new_str3 = StackPop(new_stack);
+
+    assert_int_equal(strcmp(old_str1, new_str1), 0);
```
vpodzime: Shouldn't these be the same pointers (i.e. `==`)?

tests/unit/stack_test.c
```diff
@@ -0,0 +1,119 @@
...
+    assert_int_equal(strcmp(old_str2, new_str2), 0);
+    assert_int_equal(strcmp(old_str3, new_str3), 0);
+
+    free(old_str1);
+    free(old_str2);
+    free(old_str3);
+
+    StackSoftDestroy(stack);
+    StackDestroy(new_stack);
```
vpodzime: Maybe check the state of `new_stack` between these two lines?

libutils/stack.c
```diff
@@ -0,0 +1,267 @@
...
+  @brief Destroys data in range.
+  @note Assumes that locks are acquired.
+  @param [in] stack Pointer to struct.
+  @param [in] start Start position to destroy from.
+  @param [in] end Where to stop.
+  */
+static void DestroyRange(Stack *stack, size_t start, size_t end)
+{
+    assert(stack != NULL);
+    if (start > stack->capacity || end > stack->capacity)
```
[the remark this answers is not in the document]
karlhto: Since start is unsigned, wouldn't that scenario be impossible anyway? It already checks for if it overflows.

libutils/stack.c
```diff
@@ -0,0 +1,267 @@
...
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to initialize mutex (pthread_mutex_init: %s)",
+            GetErrorStrFromCode(ret));
+        free(stack);
+        return NULL;
+    }
+
+    pthread_mutexattr_destroy(&attr);
```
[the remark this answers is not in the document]
karlhto: For errno to be correct, it should be done inside the scope of the if. Thanks!

libutils/mutex.h
karlhto: _libpromises/locks.h_ already exists. Do you want it to be put there instead?

libutils/stack.c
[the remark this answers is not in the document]
vpodzime: Yeah, but what if `STACK_EMPTY` becomes `1` at some point. I think we should either use the constant (macro) or get rid of it.

libutils/mutex.h
vpodzime: Hmm, those are different kinds of locks. Maybe *libpromises/global_mutex.h* then?

libutils/stack.c
```diff
@@ -0,0 +1,267 @@
...
+{
+    assert(stack != NULL);
+
+    ThreadLock(stack->lock);
+
+    Stack *new_stack = StackNew(stack->capacity, stack->ItemDestroy);
+    for (size_t i = 0; i < stack->size; i++)
+    {
+        new_stack->data[i] = stack->data[i];
+    }
```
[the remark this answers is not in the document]
karlhto: So `new_stack->data = xmemdup(stack->data, sizeof(void *) * stack->size);`?
I have no way of knowing the size of the data elements themselves without supplying a "size" function.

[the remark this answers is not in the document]
vpodzime: Yes, and of course not using `StackNew()`. I think this should still be a shallow copy, but the implementation should be more optimal than what a code without the knowledge of the internals could do (i.e. `xmemdup()`).

libutils/stack.c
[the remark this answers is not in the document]
karlhto: This will create a warning, though. Not sure we want that just because `STACK_EMPTY` might be `1` at some point. The functionality for the stack is also pretty basic, so I don't think that would be necessary.

libutils/stack.c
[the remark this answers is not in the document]
karlhto: Since you need a new mutex lock anyway, wouldn't it make more sense to just create a new object? I feel like it would lead to less duplicated code as well if you do that.

libutils/stack.c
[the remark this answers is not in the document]
vpodzime: Then I think it makes sense to get rid of the macro. I'd say it is more confusing than just `0` in this code.

libutils/stack.c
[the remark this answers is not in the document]
vpodzime: It's not such a big amount of code duplicated. And it may provide some nice performance benefits.

karlhto:
Other PRs blocked by this:
https://github.com/cfengine/nova/pull/1222
https://github.com/cfengine/enterprise/pull/423

Build:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/998/)

karlhto:
Started new build:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/1027/)

tests/unit/stack_test.c
vpodzime: That's a really comprehensive check of the stack. :smile: I meant something like checking it's size, capacity, maybe trying to push/pop. And even like this, it should at least be `assert(new_stack != NULL)`.

karlhto:
Maybe this is the final build? Who knows
[Build Status](https://ci.cfengine.com/job/pr-pipeline/1032/)

karlhto:
Is this the last build? Who knows
[Build Status](https://ci.cfengine.com/job/pr-pipeline/1035/)

olehermanse:
@vpodzime LGTM.

@karlhto Feel free to address smaller issues in follow up PRs.

olehermanse:
@karlhto Yes, don't cherry-pick.

olehermanse:
@karlhto FYI. In the future please make it abundantly clear what needs to be merged together. Also, try to make pull requests which are compatible with master, i.e. don't need to be merged together. When this is impossible, make the multi repo PRs as small as possible, so they are easy to get merged without too much noise. Edit the first comment to say:

> Merge with:
> cfengine/core#3187
> cfengine/enterprise#423
> cfengine/nova#1222

I will add this information to HACKING.md

PR #3188, opened 2018-07-24
Skip varibles with '#' in their lval (name) in variablesmatching()

vpodzime:
These are internal variables and should not be included in the result.

Ticket: CFE-2812

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -949,6 +949,15 @@ static StringSet *VariablesMatching(const EvalContext *ctx, VariableTableIterato
     Variable *v = NULL;
     while ((v = VariableTableIteratorNext(iter)))
     {
+        if (strchr(v->ref->lval, '#') != NULL)
+        {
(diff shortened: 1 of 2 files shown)


vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/871/)

libpromises/evalfunction.c
```diff
@@ -949,6 +949,15 @@ static StringSet *VariablesMatching(const EvalContext *ctx, VariableTableIterato
     Variable *v = NULL;
     while ((v = VariableTableIteratorNext(iter)))
     {
+        if (strchr(v->ref->lval, '#') != NULL)
```
olehermanse: Can `v->ref` be `NULL`?

vpodzime: No. We are assuming this in many other places (not only) in this function.

vpodzime:
I've improved the test a bit to make sure the change doesn't brake valid cases.
[Build Status](https://ci.cfengine.com/job/pr-pipeline/874/)

PR #3189, opened 2018-07-25
CFE-2826: Trigger promises.cf as the last step of bootstrap

vpodzime:
So that more complete data is available for the first report.

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -193,6 +193,7 @@ static const struct option OPTIONS[] =
     {"log-modules", required_argument, 0, 0},
     {"show-evaluated-classes", optional_argument, 0, 0 },
     {"show-evaluated-vars", optional_argument, 0, 0 },
+    {"skip-bootstrap-policy-run", no_argument, 0, 0 },
     {NULL, 0, 0, '\0'}
 };
 
@@ -221,6 +222,7 @@ static const char *const HINTS[] =
     "Enable even more detailed debug logging for specific areas of the implementation. Use together with '-d'. Use --log-modules=help for a list of available modules",
     "Show *final* evaluated classes, including those defined in common bundles in policy. Optionally can take a regular expression.",
     "Show *final* evaluated variables, including those defined without dependency to user-defined classes in policy. Optionally can take a regular expression.",
+    "Do not run policy as the last step of the bootstrap process",
     NULL
 };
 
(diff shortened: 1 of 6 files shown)


libpromises/failsafe.cf
```diff
@@ -69,6 +69,13 @@ bundle agent main
                     happen with the most up to date and complete
                     information.";
 
+      "Trigger Policy"
+        usebundle => failsafe_cfe_internal_trigger_policy,
```
nickanderson: So you think it will be best to do this from embedded policy than to have it done from the C?

vpodzime: I don't know yet. As I said, this is the first shot. But I think it makes sense to run `update.cf` and `promises.cf` in the same way.

vpodzime:
Seems to be working nicely. The only thing missing that I'm aware of is tests to make sure that `promises.cf` really is triggered by default during bootstrap.

vpodzime:
> @vpodzime Have you been able to see a difference in MP or similar?

Not really. But I don't quite know how to trigger the worst-case scenario. Things will/would be a lot clearer with MP gathering the data sooner.

vpodzime:
Added check to the deployment tests: https://github.com/cfengine/system-testing/pull/106

vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/927/)

cf-agent/cf-agent.c
```diff
@@ -587,6 +589,10 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
                 }
                 config->agent_specific.agent.show_evaluated_variables = xstrdup(optarg);
             }
+            else if (strcmp(OPTIONS[longopt_idx].name, "skip-bootstrap-policy-run") == 0)
```
olehermanse: `StringSafeEqual()`

libpromises/generic_agent.h
```diff
@@ -87,6 +87,7 @@ typedef struct
             char *bootstrap_port;
             char *bootstrap_ip;
             bool bootstrap_trust_server;
+            bool bootstrap_trigger_policy;
```
olehermanse: maybe invert this as well, so it's all skip, like the option? I dunno.

cf-agent/cf-agent.c
vpodzime: Yeah, but I prefer the similarity with the surrounding checks in this case.

libpromises/generic_agent.h
vpodzime: I was thinking the same, but then I saw we had a "constructor" that sets the defaults. So there's no technical reason to have it inverted and I think it is better readable like this.

vpodzime:
> I'm a little concerned that you haven't seen any difference

I am willing to believe it was just due to the MP cache.

PR #3211, opened 2018-08-15
Fix the off-by-one error in cf-runagent background process spawning

vpodzime:
New processes can be spawned while the number of spawned
processes is *less than* MAXCHILD.

Ticket: CFE-2873
Changelog: Title


cf-bottom:

PR #3218, opened 2018-08-20
Additional error reporting around safe links.

vpodzime:
This commit adds an additional log message (at level ERR) explaining
why we refuse to follow a symlink in certain conditions. I was
confused when it happened to me, and this error message would have
been a big help.

(cherry picked from commit 2c84b4e1b379a926cd5eab1056402ba090631e3d)


cf-bottom:

PR #3199, opened 2018-08-09
CFE-2850: Added function string_replace - 3.10.x

karlhto:
Cherry-picked from https://github.com/cfengine/core/pull/3178

PR #3223, opened 2018-08-24
CFE-2850 Added example for string_replace - 3.10.x

karlhto:
Tests replacing once, several and no occurences, as well as
ambiguous order.

Ticket: CFE-2850
Changelog: none

### DIFF
--- a/examples/string_replace.cf
+++ b/examples/string_replace.cf
@@ -1,0 +1,52 @@
+#  Copyright 2018 Northern.tech AS
+
+#  This file is part of Cfengine 3 - written and maintained by Northern.tech AS.
+
+#  This program is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU General Public License as published by the
+#  Free Software Foundation; version 3.
+
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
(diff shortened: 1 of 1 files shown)


karlhto:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1034/)

https://ci.cfengine.com/job/pr-pipeline/1034/

nickanderson:
?
I am confused. I thought string_replace was a new function that you added. New features can only go into new minor releases. eg 3.12.0, 3.13.0, 3.14.0

We cant add a new function in to 3.10.5

vpodzime:
> ?
> I am confused. I thought string_replace was a new function that you added. New features can only go into new minor releases. eg 3.12.0, 3.13.0, 3.14.0
> 
> We cant add a new function in to 3.10.5

Why not? I thought it was okay to add new functions so I merged https://github.com/cfengine/core/pull/3199 (and the other PR). Should those be reverted then? I think this new function is very useful and it won't be in an LTS release for a very long time if we don't add it to 3.12.1.

karlhto:
Considering it is a new feature, it does not modify or break old features. I can't really see any downsides by including it either, apart from that the documentation might not be correct for all of 3.10 if it is included in one of the later updates. Is that why it shouldn't be done?

olehermanse:
@karlhto @vpodzime One reason is to let new features stabilize, fixing any bugs before the next LTS.

karlhto:
Then it might be a good idea to reverse the changes from the 3.10 and 3.12 PRs that I made for the main functionality as well.

olehermanse:
@karlhto Absolutely, all or nothing. I don't have too strong of an opinion on this, I was just saying that might be a reason.

PR #3233, opened 2018-08-29
package_module body enhancements

vpodzime:


vpodzime:
This is still a WIP. Let's see if nothing breaks, let's test the new things then. Plus I also want to add things from [CFE-2103](https://tracker.mender.io/browse/CFE-2103).

Lex-2008:
@cf-bottom: can you request a build for me, please?

Lex-2008:
@cf-bottom: can you make a new build for me, please?

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1072/)

https://ci.cfengine.com/job/pr-pipeline/1072/

vpodzime:
Should be functionally-complete now, but I want to add some tests.

vpodzime:
I believe this is done now, but let's wait for @Lex-2008 to confirm it works for him too.

Lex-2008:
@vpodzime : it's not working for me, in https://github.com/cfengine/core/pull/3238 I'm adding some debug log messages trying to figure out WTF is going on. So far it looks like execution is going into `PipeIsReadWriteReady`, calls `ReadCompleted` (so calls `ReadFileEx`, too), but goes to neither of if/else branches after the calls to `ReadFileEx` and `SleepEx` in https://github.com/cfengine/enterprise/blob/master/libcfenterprise/win_pipe.c#L593

vpodzime:
# Note

Cherry-picking this will be a bit tricky because we cannot add the new attributes to the 3.12.x and 3.10.x branches, but we will probably need the internals to support the windows package module. IOW, **do not backport the parser changes**.

cf-agent/package_module.c
```diff
@@ -41,13 +42,16 @@
...
-static char *GetPackageModuleRealPath(const char *package_manager_name);
+static void GetPackageModuleExecInfo(const PackageModuleBody *package_module, char **exec_path,
+                                     char **script_path, char **script_exec_opts);
 static int NegotiateSupportedAPIVersion(PackageModuleWrapper *wrapper);
 
 
 void DeletePackageModuleWrapper(PackageModuleWrapper *wrapper)
 {
     free(wrapper->path);
+    free(wrapper->script_path);
```
olehermanse: These `Delete` functions usually work on `NULL` pointers, (doing nothing).

I would add that `if` here, or an `assert(wrapper != NULL)`.

cf-agent/package_module.c
```diff
@@ -58,21 +62,53 @@ PackageModuleWrapper *NewPackageModuleWrapper(PackageModuleBody *package_module)
...
+    {
+        if (access(wrapper->path, X_OK) != 0)
+        {
+            Log(LOG_LEVEL_ERR,
+                "can not find package wrapper in provided location '%s' "
+                "or access to file is restricted: %s",
+                wrapper->path, strerror(errno));
+            DeletePackageModuleWrapper(wrapper);
+            return NULL;
+        }
```
olehermanse: Add a comment or log message here; What does it mean when `script_path == NULL` and `access() == 0`?

cf-agent/package_module.c
```diff
@@ -58,21 +62,53 @@ PackageModuleWrapper *NewPackageModuleWrapper(PackageModuleBody *package_module)
...
+                "can not find the package wrapper script in provided location '%s' "
+                "or access to file is restricted: %s",
+                wrapper->script_path, strerror(errno));
+            DeletePackageModuleWrapper(wrapper);
+            return NULL;
+        }
+    }
+    else
+    {
+        if (access(wrapper->path, X_OK) != 0)
```
olehermanse: `else if` ?

cf-agent/package_module.c
vpodzime: There are log messages. What do you mean? The code here is just checking that files exist and have proper permissions set. If something is wrong, an error is logged and `NULL` is returned otherwise a proper `PackageModuleWrapper` is returned.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1153/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1153/

cf-agent/package_module.c
olehermanse: I just think these ifs are a little hard to decode (especially because of `wrapper->script_path` and `wrapper->path`). I'd add a comment along the lines of:

```
// Access for both script and interpreter is OK
```

Making variables for these two, might also improve readability:

```
const char *script_path = wrapper->script_path;
const char *interpreter_path = wrapper->path;
```

vpodzime: Done.

PR #3234, opened 2018-08-29
Add cf-info.sh diagnostic script

atsaloli:
cf-info was the precursor to cf-diag (see https://github.com/cfengine/core/blob/master/contrib/cf-diag/cf-diag.sh)

Purpose: facilitate diagnostics when investigating issues with CFEngine reporting.

It works by checking common pain points (e.g. read-only filesystem, old policy version) in large infrastructures.

This script is meant to be run on hosts managed by CFEngine by human
operators investigating problems with CFEngine reporting.

Intended for use with policy channels (see https://github.com/cfengine/core/tree/master/contrib/masterfiles-stage)

### DIFF
--- a/contrib/cf-info/cf-info.sh
+++ b/contrib/cf-info/cf-info.sh
@@ -1,17 +1,27 @@
 #!/bin/sh
 # Author: Mike Weilgart, 19 September 2016
+# Updated by: Aleksey Tsalolikhin, 30 Aug 2018
 echo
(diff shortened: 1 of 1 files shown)


cf-bottom:

atsaloli:
Excellent review!  Thank you for the encouragement and suggestions for improvement.

atsaloli:
:+1:  Thanks!

PR #3242, opened 2018-09-03
CFE-2850: Revert string_replace for 3.10

karlhto:
Due to not wanting to introduce new features to old versions, this needs to be reverted.


cf-bottom:

PR #3251, opened 2018-09-07
ENT-3831: Added new thread safe queue data structure

karlhto:
Can push, pop, block before popping if queue is empty, return
amount of elements, capacity, copy, etc. Can also be created
as a non-thread safe queue, if that is needed.

Ticket: ENT-3831
Changelog: None

### DIFF
--- a/libutils/mutex.c
+++ b/libutils/mutex.c
@@ -44,8 +44,6 @@ void __ThreadLock(pthread_mutex_t *mutex,
                 filename, lineno, funcname, GetErrorStrFromCode(result));
         fflush(stdout);
         fflush(stderr);
-        fsync(STDOUT_FILENO);
-        fsync(STDERR_FILENO);
         DoCleanupAndExit(101);
     }
 }
@@ -65,8 +63,50 @@ void __ThreadUnlock(pthread_mutex_t *mutex,
                 filename, lineno, funcname, GetErrorStrFromCode(result));
         fflush(stdout);
         fflush(stderr);
-        fsync(STDOUT_FILENO);
-        fsync(STDERR_FILENO);
         DoCleanupAndExit(101);
     }
+}
+
+int __ThreadWait(pthread_cond_t *pcond, pthread_mutex_t *mutex, int timeout,
+                    const char *funcname, const char *filename, int lineno)
+{
+    int result = 0;
+
+    if (timeout == THREAD_BLOCK_INDEFINITELY)
+    {
+        result = pthread_cond_wait(pcond, mutex);
+    }
+    else
+    {
+        struct timespec ts;
+        clock_gettime(CLOCK_REALTIME, &ts);
+
+        ts.tv_sec += timeout;
+        result = pthread_cond_timedwait(pcond, mutex, &ts);
+    }
+
+    if (result != 0)
+    {
+        if (result == ETIMEDOUT)
+        {
+            Log(LOG_LEVEL_DEBUG,
+                "Thread condition timed out at %s:%d function %s! "
+                "(pthread_cond_timewait): %s)",
+                filename, lineno, funcname, GetErrorStrFromCode(result));
+        }
+        else
+        {
+            /* Since Log blocks on mutexes, using it would be unsafe.
+               Therefore, we use fprintf instead */
+            fprintf(stderr,
+                    "Failed to wait for thread condition at %s:%d function "
+                    "%s! (pthread_cond_(wait|timewait)): %s)",
+                    filename, lineno, funcname, GetErrorStrFromCode(result));
+            fflush(stdout);
+            fflush(stderr);
+            DoCleanupAndExit(101);
+        }
+    }
+
+    return result;
 }
--- a/libutils/mutex.h
+++ b/libutils/mutex.h
@@ -27,12 +27,26 @@
 
 #include <platform.h>
 
-#define ThreadLock(m)     __ThreadLock(m, __func__, __FILE__, __LINE__)
-#define ThreadUnlock(m) __ThreadUnlock(m, __func__, __FILE__, __LINE__)
+#define THREAD_BLOCK_INDEFINITELY  -1
 
-void __ThreadLock(pthread_mutex_t *name,
+#define ThreadLock(m)       __ThreadLock(m, __func__, __FILE__, __LINE__)
+#define ThreadUnlock(m)   __ThreadUnlock(m, __func__, __FILE__, __LINE__)
+#define ThreadWait(m, n, t) __ThreadWait(m, n, t, __func__, __FILE__, __LINE__)
+
+void __ThreadLock(pthread_mutex_t *mutex,
                   const char *funcname, const char *filename, int lineno);
-void __ThreadUnlock(pthread_mutex_t *name,
+void __ThreadUnlock(pthread_mutex_t *mutex,
                     const char *funcname, const char *filename, int lineno);
+
+/**
+  @brief Function to wait for `timeout` seconds or until signalled.
+  @note Can use THREAD_BLOCK_INDEFINITELY to block until a signal is received.
+  @param [in] cond Thread condition to wait for.
+  @param [in] mutex Mutex lock to acquire once condition is met.
+  @param [in] timeout Seconds to wait for, can be THREAD_BLOCK_INDEFINITELY
+  @return 0 on success, -1 if timed out, exits if locking fails
+ */
+int __ThreadWait(pthread_cond_t *cond, pthread_mutex_t *mutex, int timeout,
+                 const char *funcname, const char *filename, int lineno);
 
 #endif
--- a/libutils/threaded_queue.c
+++ b/libutils/threaded_queue.c
@@ -1,0 +1,509 @@
+/*
+   Copyright 2018 Northern.tech AS
+
+   This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+   This program is free software; you can redistribute it and/or modify it
+   under the terms of the GNU General Public License as published by the
+   Free Software Foundation; version 3.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#include <alloc.h>
+#include <logging.h>
+#include <mutex.h>
+#include <pthread.h>
+#include <threaded_queue.h>
+
+
+#define EXPAND_FACTOR     2
+#define DEFAULT_CAPACITY 16
+
+/** @struct ThreadedQueue_
+  @brief An implementation of a thread safe queue based on a circular array
+
+  Can enqueue, dequeue and give various statistics about its contents, like
+  amount of elements, capacity and if it is empty. Has the ability to block
+  if queue is empty, waiting for new elements to be queued.
+  */
+struct ThreadedQueue_ {
+    pthread_mutex_t *lock;            /**< Thread lock for accessing data. */
+    pthread_cond_t *cond_non_empty;   /**< Blocking condition if empty     */
+    pthread_cond_t *cond_empty;       /**< Blocking condition if not empty */
+    void (*ItemDestroy) (void *item); /**< Data-specific destroy function. */
+    void **data;                      /**< Internal array of elements.     */
+    size_t head;                      /**< Current position in queue.      */
+    size_t tail;                      /**< Current end of queue.           */
+    size_t size;                      /**< Current size of queue.          */
+    size_t capacity;                  /**< Current memory allocated.       */
+};
+
+static void DestroyRange(ThreadedQueue *queue, size_t start, size_t end);
+static void ExpandIfNecessary(ThreadedQueue *queue);
+
+ThreadedQueue *ThreadedQueueNew(size_t initial_capacity,
+                                void (ItemDestroy) (void *item))
+{
+    ThreadedQueue *queue = xcalloc(1, sizeof(ThreadedQueue));
+
+    if (initial_capacity == 0)
+    {
+        initial_capacity = DEFAULT_CAPACITY;
+    }
+
+    pthread_mutexattr_t attr;
+    pthread_mutexattr_init(&attr);
+    // enables errorchecking for deadlocks
+    int ret = pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_ERRORCHECK);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to use error-checking mutexes for queue, "
+            "falling back to normal ones (pthread_mutexattr_settype: %s)",
+            GetErrorStrFromCode(ret));
+        pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_NORMAL);
+    }
+
+    queue->lock = xmalloc(sizeof(pthread_mutex_t));
+    ret = pthread_mutex_init(queue->lock, &attr);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to initialize mutex (pthread_mutex_init: %s)",
+            GetErrorStrFromCode(ret));
+        pthread_mutexattr_destroy(&attr);
+        free(queue->lock);
+        free(queue);
+        return NULL;
+    }
+
+    pthread_mutexattr_destroy(&attr);
+
+    queue->cond_non_empty = xmalloc(sizeof(pthread_cond_t));
+    ret = pthread_cond_init(queue->cond_non_empty, NULL);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to initialize thread condition (pthread_cond_init: %s)",
+            GetErrorStrFromCode(ret));
+        free(queue->lock);
+        free(queue->cond_non_empty);
+        free(queue);
+        return NULL;
+    }
+
+    queue->cond_empty = xmalloc(sizeof(pthread_cond_t));
+    ret = pthread_cond_init(queue->cond_empty, NULL);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to initialize thread condition "
+            "(pthread_cond_init: %s)",
+            GetErrorStrFromCode(ret));
+        free(queue->lock);
+        free(queue->cond_empty);
+        free(queue->cond_non_empty);
+        free(queue);
+        return NULL;
+    }
+
+    queue->capacity = initial_capacity;
+    queue->head = 0;
+    queue->tail = 0;
+    queue->size = 0;
+    queue->data = xmalloc(sizeof(void *) * initial_capacity);
+    queue->ItemDestroy = ItemDestroy;
+
+    return queue;
+}
+
+void ThreadedQueueDestroy(ThreadedQueue *queue)
+{
+    if (queue != NULL)
+    {
+        ThreadLock(queue->lock);
+        DestroyRange(queue, queue->head, queue->tail);
+        ThreadUnlock(queue->lock);
+
+        ThreadedQueueSoftDestroy(queue);
+    }
+}
+
+void ThreadedQueueSoftDestroy(ThreadedQueue *queue)
+{
+    if (queue != NULL)
+    {
+        if (queue->lock != NULL)
+        {
+            pthread_mutex_destroy(queue->lock);
+            free(queue->lock);
+        }
+
+        if (queue->cond_non_empty != NULL)
+        {
+            pthread_cond_destroy(queue->cond_non_empty);
+            free(queue->cond_non_empty);
+        }
+
+        if (queue->cond_empty != NULL)
+        {
+            pthread_cond_destroy(queue->cond_empty);
+            free(queue->cond_empty);
+        }
+
+        free(queue->data);
+        free(queue);
+    }
+}
+
+bool ThreadedQueuePop(ThreadedQueue *queue, void **item, int timeout)
+{
+    assert(queue != NULL);
+
+    ThreadLock(queue->lock);
+
+    if (queue->size == 0 && timeout != 0)
+    {
+        int res = 0;
+        do {
+            res = ThreadWait(queue->cond_non_empty, queue->lock, timeout);
+
+            if (res != 0)
+            {
+                /* Lock is reacquired even when timed out, so it needs to be
+                   released again. */
+                ThreadUnlock(queue->lock);
+                return false;
+            }
+        } while (queue->size == 0);
+        // Reevaluate predicate to protect against spurious wakeups
+    }
+
+    bool ret = true;
+    if (queue->size > 0)
+    {
+        size_t head = queue->head;
+        *item = queue->data[head];
+
+        queue->data[head++] = NULL;
+
+        head %= queue->capacity;
+        queue->head = head;
+        queue->size--;
+    } else {
+        ret = false;
+        *item = NULL;
+    }
+
+    if (queue->size == 0)
+    {
+        // Signals that the queue is empty for ThreadedQueueWaitEmpty
+        pthread_cond_broadcast(queue->cond_empty);
+    }
+
+    ThreadUnlock(queue->lock);
+
+    return ret;
+}
+
+size_t ThreadedQueuePopN(ThreadedQueue *queue,
+                         void ***data_array,
+                         size_t num,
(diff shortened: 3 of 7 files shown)


olehermanse:
@cf-bottom Please test this in jenkins :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1111/)

https://ci.cfengine.com/job/pr-pipeline/1111/

olehermanse:
IMO, it is better to introduce this as a new data structure `ThreadedQueue` first, then phase out the old one. Also, have you considered way to use the data structure without mutex? Like if I want to use queue in a single thread, and am worried about performance?

karlhto:
Yeah, you might be right. As for a non-mutex version, how would that best be done? Is there maybe a way to do it without adding if-checks before each locking? Probably not

karlhto:
@cf-bottom jenkins? exotics!

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1112/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1112/

olehermanse:
@karlhto You could add `if`s or separate functions.

libutils/queue.c
```diff
@@ -23,121 +23,386 @@
...
+Queue *QueueCopy(Queue const *queue)
+{
+    assert(queue != NULL);
+
+    ThreadLock(queue->lock);
+
+    Queue *new_queue = xmemdup(queue, sizeof(Queue));
+    new_queue->data = xmalloc(sizeof(void *) * queue->capacity);
+    memcpy(new_queue->data, queue->data,
+           sizeof(void *) * new_queue->capacity);
```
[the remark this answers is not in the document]
karlhto: There was an issue with this that I discovered when creating StackCopy. Reallocating a memdup'ed pointer will give a runtime error. At least in Valgrind

karlhto:
@cf-bottom jenkins with exotics, please.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1117/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1117/

karlhto:
Honestly a bit unsure about the ThreadLock wrapper functions, but I think it's nicer. What is not nicer is the locking not being uniform, like having to do `if (queue->lock)` instead of the functions when doing more than just locking.

olehermanse:
@karlhto This fails in tests, here is the relevant output:

```
02:24:34 ==================================================
02:24:34 Starting test: queue_test.c
02:24:34 ==================================================
02:24:34 test_basics: Starting test
02:24:34 test_basics: Test completed successfully.
02:24:34 test_destroy: Starting test
02:24:34 test_destroy: Test completed successfully.
02:24:34 All 2 tests passed
02:24:34 PASS: queue_test
02:24:34 Locking failure at threaded_queue.c:211 function ThreadedQueuePush! (pthread_mutex_lock: Resource deadlock avoided)==================================================
02:24:34 Starting test: threaded_queue_test.c
02:24:34 ==================================================
02:24:34 test_push_pop: Starting test
02:24:34 FAIL: threaded_queue_test
02:24:34 ==================================================
```

karlhto:
@olehermanse I already fixed that, just didn't trigger a new build.

@cf-bottom jenkins + exotics

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1119/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1119/

libutils/threaded_queue.c
```diff
@@ -0,0 +1,461 @@
...
+    assert(queue != NULL);
+
+    _ThreadedQueueLock(queue->lock);
+    void *item = _PopNonLock(queue);
+    _ThreadedQueueUnlock(queue->lock);
+
+    return item;
+}
+
+void *ThreadedQueuePopBlock(ThreadedQueue *queue)
```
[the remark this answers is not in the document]
vpodzime: It would also be great to have `ThreadedQueuePopN(queue, n, &n_popped)`, which would pop at most `n` elements at once (within a single lock-unlock "session"). If the items are easy to process, the locking overhead can be quite big if just one item is popped at a time.

tests/unit/threaded_queue_test.c
```diff
@@ -0,0 +1,171 @@
...
+    ThreadedQueuePush(queue, xstrdup("spam"));
+    ThreadedQueuePush(queue, xstrdup("spam"));
+    ThreadedQueuePush(queue, xstrdup("spam"));
+    ThreadedQueuePush(queue, xstrdup("spam"));
+    ThreadedQueuePush(queue, xstrdup("spam"));
+
+    assert_int_equal(ThreadedQueueCount(queue), 11);
+    assert_int_equal(ThreadedQueueCapacity(queue), 16);
+
+    ThreadedQueueDestroy(queue);
```
vpodzime: Would be nice to add comments to this test that would describe how the internal array looks like (or *is supposed to look like*) at those particular stages. IOW, explanations of what the particular blocks are testing.

libutils/threaded_queue.c
```diff
@@ -0,0 +1,461 @@
...
+    size_t capacity;                  /**< Current memory allocated.       */
+};
+
+static void _DestroyRange(ThreadedQueue *queue, size_t start, size_t end);
+static void _ExpandIfNecessary(ThreadedQueue *queue);
+
+static inline void *_PopNonLock(ThreadedQueue *queue);
+static inline void _PushNonLock(ThreadedQueue *queue, void *item);
+static inline void _ThreadedQueueLock(pthread_mutex_t *lock);
+static inline void _ThreadedQueueUnlock(pthread_mutex_t *lock);
```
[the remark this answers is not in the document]
karlhto: If you don't use a prefix like that, it isn't implicit that the function is only usable inside the scope of the file if you see it used inside another function. I think it's a nice convention, but I'm unsure of how much it is used in the rest of the codebase.

libutils/threaded_queue.c
```diff
@@ -0,0 +1,461 @@
...
+bool ThreadedQueueIsEmpty(ThreadedQueue const *queue)
+{
+    assert(queue != NULL);
+
+    _ThreadedQueueLock(queue->lock);
+    bool const empty = (queue->size == 0);
+    _ThreadedQueueUnlock(queue->lock);
+
+    return empty;
+}
```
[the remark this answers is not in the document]
karlhto: What would it do? Wait until it is empty, or wait until it is not empty?

karlhto:
Not entirely sure about what you mean with the review comment. The old queue is probably a lot slower due to cache misses and needing to allocate memory each time it adds a new element, since it is a linked list. Therefore it _does_ make sense to use a single-threaded version of this, but maybe what you mean is redoing the original queue struct to use arrays instead? Then you would remove the overhead of having to check if the lock exists each time you are trying to access the queue.

olehermanse:
> If you have a single thread that has a list/set/sequence of things to process, why would you not just use the Seq structure that we already have? I don't think that some artificial example of where a thread-unsafe FIFO structure would be needed is worth complicating the code.

I completely agree, but prefer a gradual transition. I.e. keep the old crappy unsafe queue, add this in a new file with a new name, and gradually revisit the places where the old queue is used, replacing it with the new queue and possibly making other fixes while reviewing that code.

libutils/threaded_queue.c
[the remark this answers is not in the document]
karlhto: >>> I think these two functions should be merged into a single one with the `timeout` argument determining what should happen -- return immediately (`0`), wait for `timeout` seconds or wait indefinitely (`-1`). And such a function needs to return `true/false` to indicate if it got a value or timed out.

Won't this inhibit how obvious it is to use? I would rather have two functions, where one returns immediately regardless, and the other takes in enums like `THREAD_BLOCK_INDEFINITELY` (-1), `THREAD_BLOCK_NONE` (0), and some other default value as a parameter.

Maybe I'm focusing on the wrong things?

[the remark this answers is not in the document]
vpodzime: I think we are all capable of reading the function's documentation if it's there. So I don't think it's a big deal if we only have one function. But if you want to add really straightforward functions like `PopWait()` and `TryPop()`, I think they should be based on a generic one that has this **properly documented** versatile `int timeout` parameter.

vpodzime:
> replacing it with the new queue and possibly making other fixes while reviewing that code

Sure, but do we have a place where the old crappy unsafe queue is used in a single thread and without locks? If so, should such a place use a queue/FIFO?

olehermanse:
> Sure, but do we have a place where the old crappy unsafe queue is used in a single thread and without locks? If so, should such a place use a queue/FIFO?

@vpodzime

A) I don't know without searching, which is sort of my point, it's safer to do this in steps, introduce new data structure (unused) and then start using and replacing.

B) After searching, the old Queue is used in 1 place, in nova. So either a simultaneous PR is needed, or this new queue must have a different name.

@karlhto Open a PR for nova (report-collector.c) for this to be mergeable, (or change the name of this new Queue).

karlhto:
@olehermanse This is already mergeable. I changed it from Queue to ThreadedQueue a while ago.

I'll push fixes tomorrow, but I need to know what you and @vpodzime decide upon. I agree with the doing it in steps part, and we can remove unused functions later.

libutils/mutex.c
```diff
@@ -69,3 +68,53 @@ void __ThreadUnlock(pthread_mutex_t *mutex,
...
+{
+    int result = 0;
+
+    if (timeout == THREAD_BLOCK_INDEFINITELY)
+    {
+        result = pthread_cond_wait(pcond, mutex);
+    }
+    else if (timeout == THREAD_BLOCK_NONE)
+    {
+        return 0;
```
vpodzime: This should still call `pthread_cond_timedwait()` and return a value saying if the condition was signalled or not.

libutils/mutex.h
```diff
@@ -27,12 +27,21 @@
 
 #include <platform.h>
 
+enum {
+    THREAD_BLOCK_INDEFINITELY = -1,
+    THREAD_BLOCK_NONE
```
vpodzime: I think these two should be macros.

libutils/mutex.c
```diff
@@ -69,3 +68,53 @@ void __ThreadUnlock(pthread_mutex_t *mutex,
         exit(101);
     }
 }
+
+int __ThreadWait(pthread_cond_t *pcond, pthread_mutex_t *mutex, int timeout,
```
vpodzime: Please add a documentation comment to this function and refer to the two important values of `timeout`.

libutils/threaded_queue.c
```diff
@@ -0,0 +1,393 @@
...
+    size_t tail;                      /**< Current end of queue.           */
+    size_t size;                      /**< Current size of queue.          */
+    size_t capacity;                  /**< Current memory allocated.       */
+};
+
+static void _DestroyRange(ThreadedQueue *queue, size_t start, size_t end);
+static void _ExpandIfNecessary(ThreadedQueue *queue);
+
+static inline void *_PopNonLock(ThreadedQueue *queue);
+static inline void _PushNonLock(ThreadedQueue *queue, void *item);
```
vpodzime: I think we don't need these two functions (and a "mode" of the structure that doesn't care about the locks). And they should not start with an underscore. They are `static`, that says it all.

tests/unit/threaded_queue_test.c
vpodzime: I still think these comments would be useful.

libutils/mutex.c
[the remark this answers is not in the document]
karlhto: If you look at the other functions in mutex.c and mutex.h, they are like that as well due to having a macro that fills in `__line__`, `__func__`, etc.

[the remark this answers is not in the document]
olehermanse: @karlhto Okay, but you are using it for other functions as well which are just static.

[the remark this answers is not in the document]
olehermanse: `_DestroyRange`, `_ExpandIfNecessary`,  `_PopNonLock` etc.

libutils/threaded_queue.c
```diff
@@ -0,0 +1,393 @@
...
+    {
+        return;
+    }
+
+    if (queue->ItemDestroy && queue->size > 0)
+    {
+        queue->ItemDestroy(queue->data[start]);
+
+        // in case start == end, start at second element in range
+        for (size_t i = start + 1; i != end; i++)
```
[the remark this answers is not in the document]
karlhto: Start is destroyed outside the loop. I do it this way since `queue->tail` can be the same as `queue->head`. I think it makes more sense that end is excluded, since it then makes sense to pass in just `queue->head` and `queue->tail`.

libutils/threaded_queue.c
```diff
@@ -0,0 +1,393 @@
...
+  @param [in] queue Pointer to struct.
+  @param [in] start Start position to destroy from.
+  @param [in] end Where to stop.
+  */
+static void _DestroyRange(ThreadedQueue *queue, size_t start, size_t end)
+{
+    assert(queue != NULL);
+    if (start > queue->capacity || end > queue->capacity)
+    {
+        return;
```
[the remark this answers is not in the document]
karlhto: This function already makes lots of assumptions about the data inside the internal array. Maybe this should be inside an `assert()` call, since it should _never_ be called with start and end indexes outside the capacity?

libutils/mutex.c
karlhto: I'm not sure if I agree. That will add unnecessary calls if you want to run the function without waiting for a signal.

Also, I think the signal is only sent to threads that are waiting, which means that it will never be in a situation where it will return that it has been signalled. Correct me if I'm wrong!

libutils/threaded_queue.c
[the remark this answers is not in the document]
olehermanse: @vpodzime likes ADA, so he might prefer that you do bounds checking at runtime ;)

karlhto:
@olehermanse @vpodzime What about just having one function for pushing that returns the current size, instead of both `ThreadedQueuePush` and `ThreadedQueuePushReportCount`? I feel like that is logical behaviour for a push function.

olehermanse:
@karlhto I think that's fine, but it sounds like it would break the compatibility with the other Queue implementation?

karlhto:
How? You don't have to use the return value for anything.
`QueueEnqueue(call_queue, contact);` -> `ThreadedQueuePush(call_queue, contact);`

libutils/mutex.c
karlhto: ...I guess that option then becomes completely useless, just like not running the function at all. I think I will remove it and only use the `THREAD_BLOCK_INDEFINITELY`.

karlhto:
@cf-bottom jenkins and exotics, please! 💯

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1305/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1305/

olehermanse:
> How? You don't have to use the return value for anything.
> `QueueEnqueue(call_queue, contact);` -> `ThreadedQueuePush(call_queue, contact);`

Makes sense, I thought the existing function had a boolean return value or something similar. Note that I said, "sounds like", because I didn't look up the source code before writing ;)

karlhto:

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1318/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1318/

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1335/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1335/

karlhto:
@olehermanse @vpodzime Addressed most of your concerns. The man page suggested I checked for spurious wakeups with `pthread_cond_wait`, which means I have to reevaluate the predicate whenever it wakes up. I am unsure if the way I did it is safe, since the variables the predicates check for may be changed by another thread that is waiting for lock acquisition and acquires it before the "waiting" thread.

```
do {
    res = ThreadWait(queue->cond_empty, queue->lock, timeout);
    if (res != 0)
    {
        return false;
    }
} while (queue->size != 0);
// Reevaluate predicate to protect against spurious wakeups
```

Also, tests are needed for thread-specific stuff before this is production ready.

karlhto:
...also, HP-UX does not have `CLOCK_MONOTONIC`, which might be a problem if I want to accurately time something, since `CLOCK_REALTIME` is not as accurate.
As a temporary solution I defined `CLOCK_MONOTONIC` as `CLOCK_REALTIME` if the OS is HP-UX.

karlhto:
@olehermanse @vpodzime I have now added tests for threads, although it's possible they should be grouped in the stress testing directory instead.

The first thread test spawns threads that block until there are elements in the queue, where the elements are added by spawned threads that push to the queue. The second test fills the queue, spawns a thread that waits until the queue is empty, then empties the queue. The thread that was waiting then adds an element to the queue, which is read by the main thread.

karlhto:
@cf-bottom Jenkins! Exotics! Woo

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1380/)

https://ci.cfengine.com/job/pr-pipeline/1380/

karlhto:
I guess we're trying again
@cf-bottom Jenkins with exotics, please!

@olehermanse could you remove the previous attempt?

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1386/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1386/

karlhto:
@cf-bottom jenkins with exotics, please!

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1412/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1412/

libutils/threaded_queue.c
```diff
@@ -0,0 +1,504 @@
...
+            res = ThreadWait(queue->cond_non_empty, queue->lock, timeout);
+
+            // If ThreadWait has timed out, thread not acquired
+            if (res != 0)
+            {
+                ThreadUnlock(queue->lock);
+                *data_array = NULL;
+                return 0;
+            }
+        } while (queue->size == 0);
```
karlhto: I think it's possible that predicate rechecking can result in race conditions, since another asynchronous operation might empty the queue before this thread acquires the lock, which will result in the thread just continuing to wait.

karlhto:
@olehermanse @vpodzime Please review again now. Pretty sure it should run fine on all the test platforms.

karlhto:
@cf-bottom Jenkins with exotics

vpodzime:
@cf-bottom Jenkins with exotics

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1433/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1433/

libutils/mutex.c
```diff
@@ -70,3 +70,49 @@ void __ThreadUnlock(pthread_mutex_t *mutex,
...
+            /* Since Log blocks on mutexes, using it would be unsafe.
+               Therefore, we use fprintf instead */
+            fprintf(stderr,
+                    "Failed to wait for thread condition at %s:%d function "
+                    "%s! (pthread_cond_(wait|timewait)): %s)",
+                    filename, lineno, funcname, GetErrorStrFromCode(result));
+            fflush(stdout);
+            fflush(stderr);
+            fsync(STDOUT_FILENO);
+            fsync(STDERR_FILENO);
```
vpodzime: @craigcomstock you have seen some issues with things like this, haven't you?

craigcomstock: Oh yes. I had introduced fflush() and fsync() previously and it brought some builds/tests to a halt with apparent deadlocks maybe between processes or threads. I can't find the exact code but I think it was in logging.c LogToConsole(). I think that calling fflush() is sufficient where fsync() can cause deadlocks.

karlhto:
@cf-bottom Jenkins with exotics, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1466/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1466/

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1468/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1468/

karlhto:
@olehermanse Could you cancel the last build of the two above? :s

libutils/threaded_queue.c
karlhto: > But the lock is acquired as part of the ThreadWait() call, right? So I don't think this point can ever be reached while not holding the lock.

I meant that: even though the thread has been signalled, the lock might be acquired by another thread before the signalled thread acquires it, which means that the predicate might be changed before this thread ever gets to check it, resulting in it just looping and waiting again. I suppose that's fine for ThreadedQueuePop, but it is probably _not_ fine for ThreadedQueueWaitEmpty, since you might have several threads that want to populate the queue when it has become empty.
I don't think there's a better way to protect against random wake-ups, so it will have to do.

karlhto:
@cf-bottom Jenkins with exotics, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1475/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1475/

karlhto:
Maybe it will actually build on all instances now:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/1519/)

https://ci.cfengine.com/job/pr-pipeline/1519/

karlhto:
@olehermanse Any idea why all my builds end up being aborted?

olehermanse:
@karlhto It looks like the `queue_test` is hanging on RHEL 7. @vpodzime can confirm / elaborate.

vpodzime:
> @karlhto It looks like the `queue_test` is hanging on RHEL 7. @vpodzime can confirm / elaborate.

Yes, somehow. It actually looks like the test finishes and then it all hangs, but I didn't inspect it too much.

karlhto:
`threaded_queue_test` is executed right after, so that's probably the cause, then.

karlhto:
@cf-bottom Jenkins with exotics

Removed an unnecessary macro that will most likely not change anything. (:

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1530/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1530/

karlhto:
@cf-bottom Jenkins with exotics, please.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1573/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1573/

olehermanse:
@cf-bottom I want to see another green build in jenkins, with exotics ;)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1585/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1585/

libutils/threaded_queue.c
vpodzime: > I meant that: even though the thread has been signalled, the lock might be acquired by another thread before the signalled thread acquires it

The thread *being signalled* and the thread *acquiring the lock* happens at the same time, AFAICT.

libutils/threaded_queue.h
```diff
@@ -0,0 +1,132 @@
...
+  */
+void ThreadedQueueSoftDestroy(ThreadedQueue *queue);
+
+/**
+  @brief Returns and removes the first element of the queue.
+  @note If queue is empty, blocks for `timeout` seconds or until signalled.
+        If THREAD_WAIT_INDEFINITELY is specified, waits forever until signal
+        is given. If it times out, it returns false.
+  @param [in] queue The queue to pop from.
+  @param [in] item The item at the first poisition in the queue.
```
vpodzime: Should be `[out]`, right?

libutils/threaded_queue.h
```diff
@@ -0,0 +1,132 @@
...
+  @param [in] timeout Timeout for blocking in seconds.
+  @return True on success, false if timed out or queue was empty.
```
vpodzime: *true*, it's a value not a random word here

libutils/threaded_queue.h
```diff
@@ -0,0 +1,132 @@
...
+bool ThreadedQueuePop(ThreadedQueue *queue, void **item, int timeout);
+
+/**
+  @brief Pops num elements from the queue into data_array, returns amount.
+  @note If queue is empty, blocks for `timeout` seconds or until signalled.
+        If THREAD_WAIT_INDEFINITELY is specified, waits forever until
+        signalled. If it times out, it returns 0 and sets *data_array to NULL.
+  @warning The pointer array will have to be freed manually.
+  @param [in] queue The queue to pop from.
+  @param [in] data_array Pointer to location to put popped elements.
```
vpodzime: Also `[out]`.

PR #3258, opened 2018-09-11
Refactored files_hashes.c

olehermanse:
Don't backport. I will make custom commits for 3.10 and 3.12 where this code is slightly different.


olehermanse:
@cf-bottom jenkins with exotics.

PR #3259, opened 2018-09-11
Made autogen default to brew openssl on mac (3.12)

olehermanse:


olehermanse:
@Lex-2008 let's merge this if it passes travis only.

PR #3268, opened 2018-09-13
CFE-2640: Run OS X unit tests in travis

olehermanse:


travis-scripts/script.sh
```diff
@@ -36,7 +40,7 @@ then
     make CFLAGS="-Werror -Wno-pointer-sign" -k
 elif [ "$JOB_TYPE" = compile_and_unit_test ]
 then
-    make CFLAGS="-Werror -Wno-pointer-sign"  &&
+    make CFLAGS="-Werror -Wno-pointer-sign"
```
vpodzime: Why this change?

vpodzime:
> @vpodzime Do you know why we test for sparse file support?

No idea, doesn't look like a unit test at all to me.

travis-scripts/script.sh
olehermanse: It seemed unneccessary, because of `set -e`?

PR #3274, opened 2018-09-19
Make sure to return VersionCmpResult from RunCmpCommand

vpodzime:
Ticket: ENT-4094


cf-bottom:

PR #3282, opened 2018-09-25
Renamed variable which shadowed param

olehermanse:
Also created some local variables to avoid excessive derefs
and unreadable code.

Reported by LGTM:
https://lgtm.com/rules/2156240606/


olehermanse:
@cf-bottom jenkins.

PR #3286, opened 2018-09-26
CFE-2886: Changed functions in acl_posix to use Attributes *

olehermanse:
Reported by LGTM:
https://lgtm.com/rules/2163210742/

### DIFF
--- a/cf-agent/acl_posix.c
+++ b/cf-agent/acl_posix.c
@@ -48,13 +48,13 @@
 #ifdef HAVE_LIBACL
 
 static int CheckPosixLinuxAccessACEs(EvalContext *ctx, Rlist *aces, AclMethod method, const char *file_path,
-                                     Attributes a, const Promise *pp, PromiseResult *result);
+                                     const Attributes *a, const Promise *pp, PromiseResult *result);
 static int CheckPosixLinuxDefaultACEs(EvalContext *ctx, Rlist *aces, AclMethod method, AclDefault acl_default,
-                                      const char *file_path, Attributes a, const Promise *pp, PromiseResult *result);
-static int CheckPosixLinuxACEs(EvalContext *ctx, Rlist *aces, AclMethod method, const char *file_path, acl_type_t acl_type, Attributes a,
+                                      const char *file_path, const Attributes *a, const Promise *pp, PromiseResult *result);
+static int CheckPosixLinuxACEs(EvalContext *ctx, Rlist *aces, AclMethod method, const char *file_path, acl_type_t acl_type, const Attributes *a,
                                const Promise *pp, PromiseResult *result);
-static int CheckDefaultEqualsAccessACL(EvalContext *ctx, const char *file_path, Attributes a, const Promise *pp, PromiseResult *result);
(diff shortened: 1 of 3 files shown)


olehermanse:
@cf-bottom jenkins

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1227/)

https://ci.cfengine.com/job/pr-pipeline/1227/

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1228/)

https://ci.cfengine.com/job/pr-pipeline/1228/

cf-agent/acl_posix.c
```diff
@@ -63,13 +63,14 @@ static int ACECount(acl_t acl);
...
-PromiseResult CheckPosixLinuxACL(EvalContext *ctx, const char *file_path, Acl acl, Attributes a, const Promise *pp)
+PromiseResult CheckPosixLinuxACL(EvalContext *ctx, const char *file_path, Acl acl, const Attributes *a, const Promise *pp)
 {
+    assert(a != NULL);
     PromiseResult result = PROMISE_RESULT_NOOP;
 
     if (!CheckPosixLinuxAccessACEs(ctx, acl.acl_entries, acl.acl_method, file_path, a, pp, &result))
     {
-        cfPS(ctx, LOG_LEVEL_ERR, PROMISE_RESULT_FAIL, pp, &a, "Failed checking access ACL on %s", file_path);
+        cfPS(ctx, LOG_LEVEL_ERR, PROMISE_RESULT_FAIL, pp, a, "Failed checking access ACL on %s", file_path);
```
craigcomstock: So cfPS() has an assert(attr != NULL) but also a dereference ```attr->classes```. Should we be checking if attr != NULL "for real" and not just in a debug assert()? Previously this was a bit more safe because the ```Attributes a``` is a struct which cannot be NULL right?

vpodzime: > So cfPS() has an assert(attr != NULL) but also a dereference `attr->classes`. Should we be checking if attr != NULL "for real" and not just in a debug assert()?

Probably not a bad idea, it would be a shame if a logging function caused a *segfault*.

olehermanse: @craigcomstock The general idea is that this function should never be called with a NULL pointer. Callers should always make sure that they are not calling with a NULL pointer. The assert communicates that it is invalid to call this function with a NULL pointer. The same applies to the other pointers `pp` and `ctx`, as far as I can tell, but they are not so clearly communicated with `assert`. I really wish that C had a `non-null` pointer type, similar to references in C++.

When creating a lot of functions which take pointers it doesn't make sense to recheck if they are NULL in the beginning of every function. This pattern is pretty common (remember the openssl functions we looked at a couple of weeks ago).

_With all of that being said_, in this specific case, I think it's fine to add 1 (or 3) `if`'s to the beginning of `cfPS`, just in case.

craigcomstock: Cool. Yeah, was wondering what a good pattern was. It makes sense to me to have the asserts and if you are getting good test coverage during development then you are pretty well off. _if_ you have good test coverage of course ;)

PR #3296, opened 2018-10-01
[3.12.x] Read mustache-rendered files in text mode when comparing digest

vpodzime:
On Windows writing a file in text mode means Unix newlines ("\n")
are transparently replaced by Windows newlines ("\r\n") and wise
versa for reading. We are using the text mode when writing
mustache-rendered contents into files and thus they end up having
Windows newlines. So in order to get comparable results for
digests of a file (Windows newlines) and in-memory string
representation (Unix newlines) we need to use the text mode when
reading the file and calculating the digest.

Other uses of the HashFile() function don't seem to be restricted
to working with text files and don't compare in-memory string
representations with file contents so we should use binary mode
there.

Changelog: Title
Ticket: ENT-2526
(cherry picked from commit 1f26996fe7526d58afe046a5a79bb5253697c026)


cf-bottom:

PR #3301, opened 2018-10-04
ENT-4136: Added max param to FindDollarParen to prevent OOB read

olehermanse:


olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1293/)

https://ci.cfengine.com/job/pr-pipeline/1293/

olehermanse:
Green in jenkins (failure in vagrant step is unrelated).

libpromises/iteration.c
```diff
@@ -479,11 +487,13 @@ static bool ShouldAddVariableAsIterationWheel(
...
                         char *s, char c)
 {
+    assert(s != NULL);
     assert(c == '(' || c == '{');
 
     char closing_paren   = opposite(c);
     char *s_end    = strchrnul(s, closing_paren);
-    char *next_var = s + FindDollarParen(s);
+    const size_t s_max = strlen(s);
+    char *next_var = s + FindDollarParen(s, s_max);
```
craigcomstock: @olehermanse can you explain why strlen() is safe? I don't see how this code is safer if FindDollarParen() could segfault on an outside of buffer access the same way that strlen() would in the case of a non-null terminated value. Mysterious memory contents and allocation makes it unpredictable.

Reminded me of the fun "search four bytes at a time" which I think we've looked at before: https://sourceware.org/git/?p=glibc.git;a=blob;f=string/strlen.c;h=8ce131848710f0aa8fce0d9d12d6198cc6f0c613;hb=HEAD

olehermanse: @craigcomstock it isn't. It is exactly as safe/unsafe as it used to be. It is safe for the case which I was fixing though, where you already know the buffer size.

PR #3308, opened 2018-10-08
Added a unit test for empty strings in csv parser

olehermanse:
I wanted to confirm that the CSV parser creates empty strings,
and not NULL pointers. This works as I expected, so I added
a unit test.


cf-bottom:

PR #3311, opened 2018-10-09
CFE-2913: Fixed a memory leak which occured while loading augments files

olehermanse:

PR #3315, opened 2018-10-10
CFE-2913: Fixed a memory leak which occured while loading augments files (3.10)

olehermanse:
#3311


cf-bottom:

PR #3317, opened 2018-10-11
[3.12.x] Add contrib/cf-log (shows promise_summary.log with human times)

vpodzime:
contrib/cf-log/cf-log.pl shows promise_summary.log but converts
epoch times to human-readable.

This is convenient for handling reports of:
- "My application was having issues at 8:57 PM, did CFEngine break it?"
- "Why, no, CFEngine wasn't running then, it ran at 8:59 PM (which
  I can quickly tell by running cf-log.pl)."

(cherry picked from commit b7da0a5a4c381235797236798c24f103b5492a96)


cf-bottom:

PR #3326, opened 2018-10-11
CFE-2919: ensure suid log has proper permissions

craigcomstock:


cf-agent/cf-agent.c
```diff
@@ -82,6 +82,8 @@
 #include <package_module.h>
 #include <string_lib.h>
 #include <cfnet.h>
+#include <sys/types.h>			/* checking umask on writing setxid log */
+#include <sys/stat.h>			/* checking umask on writing setxid log */
```
craigcomstock: fix indention

olehermanse:
This pull request **fixes 1 alert** when merging 52f5a188710a4233637a891a5c2a57b6341a39ef into f0a46467a2086488a54e498a1d3a93c7381ff6e8 - [view on LGTM.com](https://lgtm.com/projects/g/cfengine/core/rev/pr-625a304554bb5897a223cdd411d8469edb8b42b3)

**fixed alerts:**

* 1 for File created without restricting permissions

---

*Comment posted by [LGTM.com](https://lgtm.com)*

olehermanse:
This pull request **fixes 1 alert** when merging eeddf86aa885480a425ed27089ba98e8ec5c304b into cd4562ffbef00f85aa3ec05a164c8548c5ab5837 - [view on LGTM.com](https://lgtm.com/projects/g/cfengine/core/rev/pr-3040deb65c027ad6b4e50454359fb88ca65a00d0)

**fixed alerts:**

* 1 for File created without restricting permissions

---

*Comment posted by [LGTM.com](https://lgtm.com)*

cf-agent/cf-agent.c
```diff
@@ -796,8 +798,12 @@ static void ThisAgentInit(void)
 */
 
     snprintf(filename, CF_BUFSIZE, "%s/cfagent.%s.log", GetLogDir(), VSYSNAME.nodename);
+    ToLowerStrInplace(filename);
     MapName(filename);
 
+    mode_t current_umask;
+    umask(current_umask = umask(0777));
```
olehermanse: Add either an explaining comment, or a function `nonatomic_getumask()`.

olehermanse: If you don't create a function, I would prefer something like this:
```C
    const mode_t current_umask = umask(0777); // Gets and changes umask
    umask(current_umask);  // Restores umask
```

cf-agent/verify_files_utils.c
```diff
@@ -3243,14 +3244,22 @@ static void SaveSetxid(bool modified)
...
         snprintf(filename, CF_BUFSIZE, "%s/cfagent.%s.log", GetLogDir(), VSYSNAME.nodename);
+        ToLowerStrInplace(filename);
         MapName(filename);
 
         PurgeItemList(&VSETXIDLIST, "SETUID/SETGID");
 
         Item *current = RawLoadItemList(filename);
         if (!ListsCompare(VSETXIDLIST, current))
         {
+            mode_t oldmode = umask(022);
```
olehermanse: Add a comment explaining why `022` is correct.

cf-agent/verify_files_utils.c
```diff
@@ -3243,14 +3244,22 @@ static void SaveSetxid(bool modified)
...
+            Log(LOG_LEVEL_DEBUG, "Updating setxidlist at '%s', current umask is %o", filename, oldmode);
+            if(safe_chmod(filename, 0600))
```
olehermanse: Space after `if`.

olehermanse:
This pull request **fixes 1 alert** when merging 6df06db60f70575818917d30ea8978211707e7a6 into 3fb84b7d35b86e75b20d8616ca275cb8130714f6 - [view on LGTM.com](https://lgtm.com/projects/g/cfengine/core/rev/pr-4eb9fccd9c2b86770add6d24decfc28d99cf7515)

**fixed alerts:**

* 1 for File created without restricting permissions

---

*Comment posted by [LGTM.com](https://lgtm.com)*

cf-agent/verify_files_utils.c
```diff
@@ -3243,14 +3244,22 @@ static void SaveSetxid(bool modified)
...
         snprintf(filename, CF_BUFSIZE, "%s/cfagent.%s.log", GetLogDir(), VSYSNAME.nodename);
+        ToLowerStrInplace(filename);
         MapName(filename);
 
         PurgeItemList(&VSETXIDLIST, "SETUID/SETGID");
 
         Item *current = RawLoadItemList(filename);
         if (!ListsCompare(VSETXIDLIST, current))
         {
+            mode_t oldmode = umask(077); // This setxidlist file must only be accesible by root
```
olehermanse: Why did you change it to 077?

craigcomstock: I was looking at the bug report and it asked for 0600 permissions.

cf-agent/verify_files_utils.c
```diff
@@ -3243,14 +3244,22 @@ static void SaveSetxid(bool modified)
...
+            Log(LOG_LEVEL_DEBUG, "Updating setxidlist at '%s', current umask is %o, will create/update setxidlist using 0600.", filename, oldmode);
+            if (safe_chmod(filename, 0600)) // need to chmod in case the file already exists
+            {
```
craigcomstock: @olehermanse I am now wondering why I bother with safe_chmod() since the RawSaveItemList() call will always write a new file with the current umask. I tested this locally and things work well without safe_chmod().

olehermanse:
This pull request **fixes 1 alert** when merging ab229990a118469285cf014cae3fccd0444cbaae into 3fb84b7d35b86e75b20d8616ca275cb8130714f6 - [view on LGTM.com](https://lgtm.com/projects/g/cfengine/core/rev/pr-ea0ed3729405036305168eefbe12a3d627552354)

**fixed alerts:**

* 1 for File created without restricting permissions

---

*Comment posted by [LGTM.com](https://lgtm.com)*

cf-agent/verify_files_utils.c
craigcomstock: @olehermanse @nickanderson Does this look OK to you? umask set to 077 to achieve 0600 in the end.

cf-agent/cf-agent.c
craigcomstock: done

cf-agent/verify_files_utils.c
craigcomstock: removed code, not needed since RawSaveItemList() creates a new file each time anyway.

cf-agent/verify_files_utils.c
olehermanse: @vpodzime @craigcomstock As I understand it umask will mask permissions using bitwise AND, so:
```
const int restore = umask(0700);
FILE *f = fopen("file.txt", "w");
umask(restore);
```

Will create a file with permissions `0600`, since `0666` is the default, and the zeroes in the mask will cancel out the 2 last sixes.

vpodzime:
```
Specifically, permissions in the umask are turned off from the mode argument to open(2) and mkdir(2).
```
(from man:umask(2))

craigcomstock: @olehermanse as @vpodzime mentioned, from reading umask(2) and open(2) I see resulting permissions being `mode & ~umask` so zeroes in a mask will allow permissions through and sevens in a mask will block them out.

On my debian I have umask as 0022 (from command line command) and when I create a file with fopen() I get 644 as a result. If I do umask(0700) I get 0066 which is not what I want. I think 077 is right as it will always block out any permissions for group and other and cause the system call's mode parameter to affect only the user permissions.

nickanderson:
Useful: https://wintelguy.com/umask-calc.pl

```
Results :
Mask (Numeric Notation) 377
Allowed Permissions r--------

Examples:
umask 377
umask u=r,go=
```

craigcomstock:
@nickanderson @olehermanse @vpodzime approve the PR now?

cf-agent/cf-agent.c
```diff
@@ -84,6 +84,8 @@
 #include <string_lib.h>
 #include <cfnet.h>
 #include <repair.h>
+#include <sys/types.h>                  /* checking umask on writing setxid log */
```
vpodzime: Please add a proper title to the commit message before the long description.

cf-agent/cf-agent.c
```diff
@@ -796,8 +798,12 @@ static void ThisAgentInit(void)
 */
 
     snprintf(filename, CF_BUFSIZE, "%s/cfagent.%s.log", GetLogDir(), VSYSNAME.nodename);
+    ToLowerStrInplace(filename);
```
vpodzime: I don't quite understand what these `ToLower` calls are doing in this commit which is supposed to just make sure the permissions are correct. And shouldn't this be done in `MapName()` if we need it?

craigcomstock: The policy that manages this log file assumes lower case but the code did not. See https://github.com/cfengine/masterfiles/blob/master/inventory/any.cf#L204 for one example. @nickanderson I think you mentioned this to me. Is this really a problem? If so, I will move this change to another ticket/PR. :)

cf-agent/cf-agent.c
craigcomstock: Ack. :)

cf-agent/cf-agent.c
vpodzime: Thanks for the explanation. Let's see what @nickanderson says, but it should definitely be part of that particular commit.

nickanderson: From some quick research, it seems that when this policy was introduced the understanding was the log file was created from hostname -f and that command always returns lowercase.

```
When the agent observes a file that is setuid it is logged to $(sys.workdir)/cfagent.$(hostname -f).log.

This change introduces inventory for these observed files. For each file in the log that is setuid inventory of that file is tagged with setuid Files. Additionally separate inventory is done for each root owned setuid file as root owned setuid Files. This policy also handles clearing of files that were once setuid but are no longer found or are no longer setuid from the log since there is no pre-existing automatic processes to clear entries from the log.
```

So the policy looks for candidates that have a matching alll lowercase form. If this ToLower makes sure that the log file name is lowercase then it makes sense to me. Perhaps we were relying on an unenforced expectation before.

```
      "setuid_log_path"
        comment => "We select the file that matches the downcased version of the
                    hostname since sys.fqhost always returns lower case",
        string => "$(candidates)",
        if => strcmp( "$(sys.workdir)/cfagent.$(sys.fqhost).log",
string_downcase($(candidates)));
```

cf-agent/verify_files_utils.c
nickanderson: I agree. Based on my research it should be 077 umask to ensure that files are read/write by owner, directories are read/write/execute by owner, and nothing else.

```
    umask    077
    File Permissions    rw-------
    Directory Permissions    rwx------
```
http://www.webune.com/forums/umask-calculator.html

craigcomstock:
Ok. Squashed and edited commit to include title and then description as per @vpodzime 's comment.

olehermanse:
This pull request **fixes 1 alert** when merging 26dbf4c237244aa3027cfd99b3e6060a5ff21452 into 6dafc4082363c8902eb17e5f2c25945e0c776b21 - [view on LGTM.com](https://lgtm.com/projects/g/cfengine/core/rev/pr-19974c34634975cb51027426fc7f2eb1191c8be5)

**fixed alerts:**

* 1 for File created without restricting permissions

---

*Comment posted by [LGTM.com](https://lgtm.com)*

olehermanse:
Replaced by https://github.com/cfengine/core/pull/3629

PR #3335, opened 2018-10-12
CFE-2000 Test that classes can be defined without constraints

nickanderson:


cf-bottom:

PR #3339, opened 2018-10-12
CFE-2715: Added a simple valgrind check to travis

olehermanse:


cf-bottom:

olehermanse:
It looks like this is working. <3 I will do some more testing to make sure that it also fails when expected to.

PR #3346, opened 2018-10-15
￼ Build only master and version pushed branches

Lex-2008:


cf-bottom:

.travis.yml
```diff
@@ -5,6 +5,12 @@ addons:
     paths:
       - artifacts/
 
+# Build on pushes only to given branches
+branches:
+  only:
+    - master
+    - /3\.\d+\.x/
```
olehermanse: I think explicit would be better, just list out `3.10.x` and `3.12.x`. It is much more clear, even if we have to update it once every 18 months.

PR #3347, opened 2018-10-15
Disable codecov comments

olehermanse:
I'm tired of the spam. The comment is huge. Because of
fluctuations in coverage it has far more noise than useful
information.


cf-bottom:

PR #3350, opened 2018-10-16
The pgsql resource agent that supports our HA setup

vpodzime:
For the HA setup we need a resource agent that supports
PostgreSQL 10 on RHEL 6. The upstream [1] only provides versions
that are either compatible with RHEL 6 or supporting PostgreSQL
10. Hence adding our combined RA to contrib.

[1] https://github.com/ClusterLabs/resource-agents/


cf-bottom:

PR #3359, opened 2018-10-18
Cast address in iteration_test to avoid warning

olehermanse:


cf-bottom:

PR #3361, opened 2018-10-18
Respect windows newlines when splitting data in PipeReadData()

vpodzime:
This needs to do different things on Windows and UNIX-like systems.

Add a function to split string on newlines to do so because it is
not trivial because the "\r\n" 2-byte sequence
represents a newline on Windows.

Ticket: ENT-3716
Changelog: None

### DIFF
--- a/libpromises/pipes.c
+++ b/libpromises/pipes.c
@@ -119,7 +119,14 @@ Rlist *PipeReadData(const IOData *io, int pipe_timeout_secs, int pipe_terminatio
     }
 
     char *read_string = BufferClose(data);
-    Rlist *response_lines = RlistFromSplitString(read_string, '\n');
+
+#ifdef __MINGW32__
+    bool detect_crlf = true;
+#else
+    bool detect_crlf = false;
+#endif
+
+    Rlist *response_lines = RlistFromStringSplitLines(read_string, detect_crlf);
     free(read_string);
 
     return response_lines;
--- a/libpromises/rlist.c
+++ b/libpromises/rlist.c
@@ -980,9 +980,24 @@ void RlistDestroyEntry(Rlist **liststart, Rlist *entry)
  * than the length of the new entry in <to>.  The new entry is
  * prepended; the caller can reverse <to> once built.
  */
-static size_t SubStrnCopyChr(Rlist **to, const char *from, char sep)
+static size_t SubStrnCopyChr(Rlist **to, const char *from, char sep, char lstrip)
 {
     assert(from && from[0]);
+    size_t offset = 0;
+
+    while (lstrip != '\0' && from[0] == lstrip && from[0] != '\0')
+    {
+        /* Skip over all instances of the 'lstrip' character (e.g. ' ') if
+         * specified */
+        from++;
+        offset++;
+    }
+    if (from[0] == '\0')
+    {
+        /* Reached the end already so there's nothing to add to the result list,
+           just tell the caller how far they can move. */
+        return offset;
+    }
 
     const char *end = from;
     size_t escapes = 0;
@@ -1015,7 +1030,7 @@ static size_t SubStrnCopyChr(Rlist **to, const char *from, char sep)
     /* Prepend to the list and reverse when done, costing O(len),
      * instead of appending, which costs O(len**2). */
     RlistPrependRval(to, RvalCopyScalar((Rval) { copy, RVAL_TYPE_SCALAR }));
-    return consume;
+    return offset + consume;
 }
 
 Rlist *RlistFromSplitString(const char *string, char sep)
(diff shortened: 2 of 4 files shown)


vpodzime:
I know this is not perfect, but I'm sure it still covers 100 % of the cases we need. Also, no tests are added because we are not running tests on Windows and on other system the new function is just calling an older function which we have tests for.

vpodzime:
@Lex-2008 I will take your ACK as "it fixed my issues with newlines in package module on Windows".

Lex-2008:
https://ci.cfengine.com/job/testing-pr/1320/label=PACKAGES_i386_mingw/console
```
15:46:22 rlist.c: In function 'RlistFromStringSplitLines':
15:46:22 rlist.c:1093:27: error: 'sep' undeclared (first use in this function)
15:46:22              assert(*sp == sep && (sp == string || sp[-1] != '\\'));
15:46:22                            ^
15:46:22 rlist.c:1093:27: note: each undeclared identifier is reported only once for each function it appears in
15:46:22 Makefile:886: recipe for target 'rlist.lo' failed
```

vpodzime:
@olehermanse would appreciate your look too. Thx!

libpromises/pipes.c
```diff
@@ -119,7 +119,7 @@ Rlist *PipeReadData(const IOData *io, int pipe_timeout_secs, int pipe_terminatio
     }
 
     char *read_string = BufferClose(data);
-    Rlist *response_lines = RlistFromSplitString(read_string, '\n');
+    Rlist *response_lines = RlistFromStringSplitLines(read_string);
```
olehermanse: I would prefer it if `crlf` was a boolean argument to this functions, so you can test both modes on unix platforms.

libpromises/rlist.c
```diff
@@ -1044,6 +1059,56 @@ Rlist *RlistFromSplitString(const char *string, char sep)
...
+ * @return: an #Rlist where items are the individual lines **without** the
+ *          trailing newline character(s)
+ * @note: Free the result with RlistDestroy()
+ * @warning: This function doesn't work properly if @string uses "\r\n" newlines
+ *           and contains '\r' characters that are not part of any "\r\n"
+ *           sequence because it first splits @string on '\r'.
+ */
+Rlist *RlistFromStringSplitLines(const char *string)
+{
+#ifndef __MINGW32__
```
olehermanse: Please make this into an if based on argument instead, so we can test more easily.

vpodzime: Makes sense.

libpromises/rlist.c
```diff
@@ -1044,6 +1059,49 @@ Rlist *RlistFromSplitString(const char *string, char sep)
...
+ * @return: an #Rlist where items are the individual lines **without** the
+ *          trailing newline character(s)
+ * @note: Free the result with RlistDestroy()
+ * @warning: This function doesn't work properly if @string uses "\r\n" newlines
+ *           and contains '\r' characters that are not part of any "\r\n"
+ *           sequence because it first splits @string on '\r'.
+ */
+Rlist *RlistFromStringSplitLines(const char *string, bool detect_crlf)
+{
+    if (!detect_crlf || (strstr(string, "\r\n") == NULL))
```
olehermanse: `strstr` should never be called with `NULL` as one of the arguments.

You already have 

```
if (string == NULL)
```

further down. I suggest moving this up to the top, or adding asserts on top instead.

vpodzime: Another good catch.

libpromises/rlist.c
```diff
@@ -1044,6 +1059,50 @@ Rlist *RlistFromSplitString(const char *string, char sep)
...
+    }
+
+    /* else we split on '\r' just like RlistFromSplitString() above, but
+     * strip leading '\n' in every chunk, thus effectively split on \r\n. See
+     * the warning in the function's documentation.*/
+    Rlist *liststart = NULL;
+
+    for (const char *sp = string; *sp != '\0';)
+    {
+        sp += SubStrnCopyChr(&liststart, sp, '\r', '\n');
```
olehermanse: This looks a little scary, but I guess the only way `SubStrnCopyChr` can return `0` is when `*sp == '\0'`.

vpodzime: That's a belief we share.

libpromises/pipes.c
```diff
@@ -119,7 +119,14 @@ Rlist *PipeReadData(const IOData *io, int pipe_timeout_secs, int pipe_terminatio
     }
 
     char *read_string = BufferClose(data);
-    Rlist *response_lines = RlistFromSplitString(read_string, '\n');
+
+#ifdef __MINGW32__
+    bool detect_crlf = true;
```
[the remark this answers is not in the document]
vpodzime: I copied it over from some other place. Does it really matter to us when we are only building in one environment? Would be great to have native Windows builds and those would suffer from this, of course. But I think that's an agenda for a separate "Windows-related ifdefs revision" PR.

PR #3362, opened 2018-10-19
Add support for CIFS and Panasas panfs file systems to storage promise type

favoritelotus:
I'd like to share my patch which adds support for CIFS and Panasas panfs file systems to storage promise type.

The patch is attached in zip archive.
[panfs_cifs.patch.zip](https://github.com/cfengine/core/files/2496667/panfs_cifs.patch.zip)

PR #3365, opened 2018-10-19
Add CIFS and Panasas support to storage promise type

nickanderson:
The PR this code came from: https://github.com/cfengine/core/pull/3362
Shared by https://github.com/favoritelotus

O-pened in new PR because PR was not clean.

Changelog: TODO

### DIFF
--- a/cf-agent/nfs.c
+++ b/cf-agent/nfs.c
@@ -47,6 +47,7 @@
 static int FSTAB_EDITS = 0; /* GLOBAL_X */
 static Item *FSTABLIST = NULL; /* GLOBAL_X */
 
+void GetHostAndSource(char *buf, char *host, char *source);
 static void AugmentMountInfo(Seq *list, char *host, char *source, char *mounton, char *options);
 static int MatchFSInFstab(char *match);
 static void DeleteThisItem(Item **liststart, Item *entry);
@@ -126,13 +127,36 @@ static const char *const VMOUNTOPTS[] =
     [PLATFORM_CONTEXT_ANDROID] = "defaults",         /* android */
 };
 
+
+void GetHostAndSource(char *buf, char *host, char *source)
+/* Extracts from buffer host & source of cifs and panfs network systems */
+{
+    int i = 0, i2 = 0, bs = 0;
+
+    // copy host
+    while(bs != 3)
+    {
+        if (buf[i] == '/')
+            bs++;
+        host[i] = buf[i];
+        i++;
+    }
+    i--;
+    host[i] = '\0';
+
+    // copy source
+    while (buf[i] != ' ')
+        source[i2++] = buf[i++];
+    source[i2] = '\0';
+}
+
 bool LoadMountInfo(Seq *list)
 /* This is, in fact, the most portable way to read the mount info! */
 /* Depressing, isn't it? */
 {
     FILE *pp;
     char buf1[CF_BUFSIZE], buf2[CF_BUFSIZE], buf3[CF_BUFSIZE];
-    int i, nfs = false;
+    int i, nfs = false, panfs = false, cifs = false;
 
     for (i = 0; VMOUNTCOMM[VSYSTEMHARDCLASS][i] != ' '; i++)
     {
@@ -173,9 +197,17 @@ bool LoadMountInfo(Seq *list)
             }
         }
 
-        if (strstr(vbuff, "nfs"))
+        if (strstr(vbuff, "panfs"))
+        {
+            panfs = true;
+        }
+        else if (strstr(vbuff, "nfs"))
         {
             nfs = true;
+        }
+        else if (strstr(vbuff, "cifs"))
+        {
+            cifs = true;
         }
 
         // security note: buff is CF_BUFSIZE, so that is the max that can be written to buf1, buf2 or buf3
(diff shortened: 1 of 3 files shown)


cf-bottom:

favoritelotus:
> @favoritelotus It would be great if you could share any other information about your usage with this as well.

Hi Nick,

Sure. I have prepared example policy files with the comments for both cifs and panfs. You will find them in the attached zip archive.
[examples.zip](https://github.com/cfengine/core/files/2498177/examples.zip)

Regards,
Pavel.

cf-agent/nfs.c
```diff
@@ -126,13 +127,36 @@ static const char *const VMOUNTOPTS[] =
     [PLATFORM_CONTEXT_ANDROID] = "defaults",         /* android */
 };
 
+
+void GetHostAndSource(char *buf, char *host, char *source)
```
vpodzime: `buf` should be `const char *`.

cf-agent/nfs.c
```diff
@@ -126,13 +127,36 @@ static const char *const VMOUNTOPTS[] =
...
+/* Extracts from buffer host & source of cifs and panfs network systems */
+{
+    int i = 0, i2 = 0, bs = 0;
+
+    // copy host
+    while(bs != 3)
```
vpodzime: Would be great to add an example of the data this should process.

cf-agent/nfs.c
```diff
@@ -223,7 +255,7 @@ bool LoadMountInfo(Seq *list)
 
 
 #if defined(__sun) || defined(__hpux)
-        if (IsAbsoluteFileName(buf3))
+        if (IsAbsoluteFileName(buf3) && !cifs)
```
vpodzime: Why are these checks needed?

cf-agent/nfs.c
```diff
@@ -126,13 +127,36 @@ static const char *const VMOUNTOPTS[] =
     [PLATFORM_CONTEXT_ANDROID] = "defaults",         /* android */
 };
 
+
+void GetHostAndSource(char *buf, char *host, char *source)
+/* Extracts from buffer host & source of cifs and panfs network systems */
+{
+    int i = 0, i2 = 0, bs = 0;
```
olehermanse: So `bs` is used to count slashes, and `i2` is used as an index for `source`. I would call these variables `slashes` and `source_index`, respectively.

cf-agent/nfs.c
```diff
@@ -567,7 +615,19 @@ PromiseResult VerifyMount(EvalContext *ctx, char *name, const Attributes *a, con
     PromiseResult result = PROMISE_RESULT_NOOP;
     if (!DONTDO)
     {
-        snprintf(comm, CF_BUFSIZE, "%s -o %s %s:%s %s", CommandArg0(VMOUNTCOMM[VSYSTEMHARDCLASS]), opts, host, rmountpt, mountpt);
+
+        if (strcmp(a.mount.mount_type, "panfs") == 0)
```
olehermanse: `a` has been changed to a pointer, so this needs to be updated to `a->mount`. (Maybe @nickanderson 's mistake while cherry-picking.

favoritelotus:
Hello Vratisalv,

The variable bs stands for back slash. Inside the loop, the variable simply
counts slashes to find where the host name ends and source begins when a
CIFS share is mounted as //server/source. The same applies for panfs, which
has a similar mount syntax, e.g. panfs://server/source.

Since CIFS mounts start with two slashes, without the check "if
(IsAbsoluteFileName(buf3) && !cifs)" they will be interpreted as local
mounts and transformed to  localhost://server/host.

Regards,
Pavel.

cf-agent/nfs.c
```diff
@@ -47,6 +47,7 @@
 static int FSTAB_EDITS = 0; /* GLOBAL_X */
 static Item *FSTABLIST = NULL; /* GLOBAL_X */
 
+void GetHostAndSource(char *buf, char *host, char *source);
```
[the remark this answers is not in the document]
nickanderson: @favoritelotus  Do you plan to come back around and make fix-ups per these suggestions?

vpodzime:
Nice! Thanks for the detailed explanation! Could you please address the code issues pointed out in the reviews?

olehermanse:
@nickanderson @vpodzime It looks like @favoritelotus is not responding. We should either fix this one ourself, or close it.

favoritelotus:
Hello,
Sorry for not replying long. I don't have much free time to 
I just want to share my  new patch, which fixes  some bugs related to sigfault caused by Panasas mounts. The patch was debugged, tested and finally put into production. 
We have been using successfully CFengine 3.12.0 with my patch in production for three mounts on 194 servers. The patch is attached with 

[panfs-cifs-patch.zip](https://github.com/cfengine/core/files/3694422/panfs-cifs-patch.zip)

sample policy file.

olehermanse:
@favoritelotus Thank you for your reply, I'll take over this one. See this PR for further progress: https://github.com/cfengine/core/pull/3893

favoritelotus:
Hello Ole,

Thank you. I will release another patch soon, which adds support for SSHFS
to the storage promise type.

Regards,
Pavel.

PR #3369, opened 2018-10-22
TLS 1.3 support

vpodzime:
Merge together:
https://github.com/cfengine/buildscripts/pull/454

This requires OpenSSL 1.1.1 so it only has a chance to compile and pass tests when combined with https://github.com/cfengine/buildscripts/pull/454.

### DIFF
--- a/cf-serverd/server_tls.c
+++ b/cf-serverd/server_tls.c
@@ -100,27 +100,22 @@ bool ServerTLSInitialize(RSA *priv_key, RSA *pub_key, SSL_CTX **ssl_ctx)
      * CFEngine is not a web server so it does not need to support many
      * ciphers. It only allows a safe but very common subset by default,
      * extensible via "allowciphers" in body server control. By default
-     * it allows:
+     * the server side allows:
      *
      *     AES256-GCM-SHA384: most high-grade RSA-based cipher from TLSv1.2
      *     AES256-SHA: most backwards compatible but high-grade, from SSLv3
+     *     TLS_AES_256_GCM_SHA384: most high-grade RSA-based cipher from TLSv1.3
+     *
+     * Client side is using the OpenSSL's defaults by default.
      */
     const char *cipher_list = SERVER_ACCESS.allowciphers;
     if (cipher_list == NULL)
     {
-        cipher_list ="AES256-GCM-SHA384:AES256-SHA";
+        cipher_list = "AES256-GCM-SHA384:AES256-SHA:TLS_AES_256_GCM_SHA384";
     }
 
-    Log(LOG_LEVEL_VERBOSE,
-        "Setting cipher list for incoming TLS connections to: %s",
-        cipher_list);
-
-    ret = SSL_CTX_set_cipher_list(*ssl_ctx, cipher_list);
-    if (ret != 1)
+    if (!TLSSetCipherList(*ssl_ctx, cipher_list))
     {
-        Log(LOG_LEVEL_ERR,
-            "No valid ciphers in cipher list: %s",
-            cipher_list);
         goto err;
     }
 
--- a/cf-serverd/server_tls.c
+++ b/cf-serverd/server_tls.c
@@ -111,7 +111,11 @@ bool ServerTLSInitialize(RSA *priv_key, RSA *pub_key, SSL_CTX **ssl_ctx)
     const char *cipher_list = SERVER_ACCESS.allowciphers;
     if (cipher_list == NULL)
     {
+#ifdef SSL_OP_NO_TLSv1_3        /* defined if TLS 1.3 is supported */
         cipher_list = "AES256-GCM-SHA384:AES256-SHA:TLS_AES_256_GCM_SHA384";
+#else
+        cipher_list = "AES256-GCM-SHA384:AES256-SHA";
+#endif
     }
 
     if (!TLSSetCipherList(*ssl_ctx, cipher_list))
--- a/libcfnet/tls_generic.c
+++ b/libcfnet/tls_generic.c
@@ -855,9 +855,12 @@ void TLSSetDefaultOptions(SSL_CTX *ssl_ctx, const char *min_version)
                 " however because of old OpenSSL version it is set to: %s",
                 min_version, compiletime_min_version);
         }
+        else
+        {
+            Log(LOG_LEVEL_VERBOSE,
+                "Setting minimum acceptable TLS version: 1.1");
+        }
         options |= SSL_OP_NO_TLSv1;
-        Log(LOG_LEVEL_VERBOSE,
-            "Setting minimum acceptable TLS version: 1.1");
 
     }
     else if (StringSafeEqual(min_version, "1.2"))
@@ -869,9 +872,12 @@ void TLSSetDefaultOptions(SSL_CTX *ssl_ctx, const char *min_version)
                 " however because of old OpenSSL version it is set to: %s",
                 min_version, compiletime_min_version);
         }
+        else
+        {
+            Log(LOG_LEVEL_VERBOSE,
+                "Setting minimum acceptable TLS version: 1.2");
+        }
         options |= SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1;
-        Log(LOG_LEVEL_VERBOSE,
-            "Setting minimum acceptable TLS version: 1.2");
     }
     else if (StringSafeEqual(min_version, "1.3"))
     {
@@ -883,9 +889,12 @@ void TLSSetDefaultOptions(SSL_CTX *ssl_ctx, const char *min_version)
                 " however because of old OpenSSL version it is set to: %s",
                 min_version, compiletime_min_version);
         }
+        else
+        {
+            Log(LOG_LEVEL_VERBOSE,
+                "Setting minimum acceptable TLS version: 1.3");
+        }
         options |= SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1 | SSL_OP_NO_TLSv1_2;
-        Log(LOG_LEVEL_VERBOSE,
-            "Setting minimum acceptable TLS version: 1.3");
     }
     else
     {
--- a/libcfnet/tls_generic.c
+++ b/libcfnet/tls_generic.c
@@ -39,6 +39,43 @@
 
 #include <assert.h>
 
+/* known TLS versions */
+enum tls_version {
+    TLS_1_0 = 0,
+    TLS_1_1 = 1,
+    TLS_1_2 = 2,
+    TLS_1_3 = 3,
+};
+#define TLS_LAST TLS_1_3
+
+/* determine the highest TLS version supported by the available/used version of
+ * OpenSSL */
+#if defined(SSL_OP_NO_TLSv1_3)
+#define HAVE_TLS_1_3
+#define TLS_HIGHEST_SUPPORTED TLS_1_3
+#elif defined(SSL_OP_NO_TLSv1_2)
+#define TLS_HIGHEST_SUPPORTED TLS_1_2
+#elif defined(SSL_OP_NO_TLSv1_1)
+#define TLS_HIGHEST_SUPPORTED TLS_1_1
+#else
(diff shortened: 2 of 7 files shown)


cf-bottom:

libcfnet/tls_generic.c
```diff
@@ -868,6 +873,20 @@ void TLSSetDefaultOptions(SSL_CTX *ssl_ctx, const char *min_version)
         Log(LOG_LEVEL_VERBOSE,
             "Setting minimum acceptable TLS version: 1.2");
     }
+    else if (strcmp(min_version, "1.3") == 0)
+    {
+        if (strcmp(compiletime_min_version, "1.0") == 0 ||
+            strcmp(compiletime_min_version, "1.1") == 0 ||
+            strcmp(compiletime_min_version, "1.2") == 0)
```
olehermanse: Use `StringSafeEqual` instead?

libcfnet/tls_generic.c
```diff
@@ -907,3 +926,75 @@ void TLSSetDefaultOptions(SSL_CTX *ssl_ctx, const char *min_version)
...
+        {
+            StrCat(cipher_suites, max_len, &cipher_suites_len, spec_ref.data, spec_ref.len + 1);
+        }
+        else
+        {
+            StrCat(ciphers, max_len, &ciphers_len, spec_ref.data, spec_ref.len + 1);
+        }
+    }
+
+    int ret;
```
olehermanse: Minor nitpick, but I would prefer if you moved this into the scopes it is used.

cf-serverd/server_tls.c
```diff
@@ -100,27 +100,22 @@ bool ServerTLSInitialize(RSA *priv_key, RSA *pub_key, SSL_CTX **ssl_ctx)
...
+     * Client side is using the OpenSSL's defaults by default.
      */
     const char *cipher_list = SERVER_ACCESS.allowciphers;
     if (cipher_list == NULL)
     {
-        cipher_list ="AES256-GCM-SHA384:AES256-SHA";
+        cipher_list = "AES256-GCM-SHA384:AES256-SHA:TLS_AES_256_GCM_SHA384";
     }
 
-    Log(LOG_LEVEL_VERBOSE,
```
estenberg: Why was this removed? I think a verbose output of which ciphers can be used is good to know, especially if any vulnerability is found in any of the later (compliance).

libcfnet/tls_generic.c
```diff
@@ -868,6 +873,20 @@ void TLSSetDefaultOptions(SSL_CTX *ssl_ctx, const char *min_version)
...
     }
+    else if (strcmp(min_version, "1.3") == 0)
+    {
+        if (strcmp(compiletime_min_version, "1.0") == 0 ||
+            strcmp(compiletime_min_version, "1.1") == 0 ||
+            strcmp(compiletime_min_version, "1.2") == 0)
+        {
+            Log(LOG_LEVEL_WARNING, "Minimum requested TLS version is %s,"
+                " however because of old OpenSSL version it is set to: %s",
+                min_version, compiletime_min_version);
```
estenberg: I don't think this should be allowed (potential downgrade attack).

If I specify TLS 1.3 to be used (e.g. for compliance reasons), I don't want CFEngine to simply pick a different version. It should stop, and let me choose whether to upgrade my openssl or to compromise and downgrade to 1.2. But this should be done explicitly.

libcfnet/tls_generic.c
```diff
@@ -818,17 +819,21 @@ void TLSSetDefaultOptions(SSL_CTX *ssl_ctx, const char *min_version)
     }
 #endif
 
-    const char *compiletime_min_version = "1.2";
+    const char *compiletime_min_version = "1.3";
```
estenberg: I would recommend enabling this as a transition across releases.
So  the default TLS in the first release is 1.2, but 1.3 is supported and encouraged.
In the next release, 1.3 is the default, as most CFEngine clients/servers support 1.3 at this point.

This relates to my comment below, as I would also not recommend to allow the client/server to simply pick a lower TLS version as its discretion.

cf-serverd/server_tls.c
vpodzime: It still there, I just moved it to `TLSSetCipherList()`.

libcfnet/tls_generic.c
vpodzime: I agree, I was just following the same behavior that was used for older versions. Shall I change this in all the above versions too? The same logic applies there. However, using such logic only for 1.3 doesn't really make much sense to me.

estenberg: I think it should be changed for all versions. Though it is a bit interesting that it used to be like this too - can we find out some info on why? Are there commit logs or can we reach out to the developer doing it?

vpodzime: @nickanderson does the above ring any bells?

nickanderson: I will reach out and ask @jimis for any historical context.

jimis: commit that introduced it: 2719ce18c2cd6ba2f192b013db11be003675cf76

If I recall, this code exists to warn if for example you are trying to turn off TLS 1.2, but the openssl library does not even have defined `SSL_OP_NO_TLSv1_2`. Apparently you can't turn off TLS 1.2 if the openssl library thinks that 1.1 is the most current TLS version.

The above commit introduces a dummy (of value 0x0) define that will apparently not turn off anything, but at least the program will compile and keep accepting connections, and the important WARNING will be printed prompting the sysadmin to recompile with recent openssl.

nickanderson:
Thanks for the history @jimis, @vpodzime is it helpful?

libcfnet/tls_generic.c
vpodzime: > Apparently you can't turn off TLS 1.2 if the openssl library thinks that 1.1 is the most current TLS version.

That makes sense (I checked that the `NO_TLSv` macros have been being added to OpenSSL together with the support for the same particular version of TLS), but then I don't understand the logic of this:
```
#ifndef  SSL_OP_NO_TLSv1_1
#   define SSL_OP_NO_TLSv1_1 0x0                                /* nop */
    compiletime_min_version = "1.1";
```
which translates to *"if OpenSSL doesn't know about TLS 1.1, we set the minimal version to 1.1"*. And if user requests 1.2. or newer as the minimal version, we tell them 1.1 is the minimal version being used because of old OpenSSL (which doesn't support that version).

Shouldn't the user in such a case get a warning (or maybe an error) that the minimal requested version is higher than the maximal version supported by their OpenSSL library?

nickanderson: It does seem like we should be erroring that up to the user, or in this case, warning since the "error" is handled.

vpodzime:
@nickanderson @estenberg please see the refactoring commit I added here. It contains some reasoning/explanation in the commit message too.

vpodzime:
I'll check the failures in *cf-net* tests tomorrow.

vpodzime:
> I'll check the failures in _cf-net_ tests tomorrow.

Looks like it was just a spurious error.

vpodzime:
@nickanderson suggested we also add a *feature* hard class for TLS 1.3. I'll do that tomorrow. Should be easy and small so reviewers are still welcome to review what we already have.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1672/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1672/

libcfnet/tls_generic.c
```diff
@@ -819,91 +856,59 @@ void TLSSetDefaultOptions(SSL_CTX *ssl_ctx, const char *min_version)
...
+                        min_version, tls_version_strings[TLS_LOWEST_REQUIRED]);
+                    min_tls_version = TLS_LOWEST_REQUIRED;
+                }
+                else if (v > TLS_HIGHEST_SUPPORTED)
+                {
+                    Log(LOG_LEVEL_WARNING, "Minimum requested TLS version is %s,"
+                        " but maximum version supported by OpenSSL is %s."
+                        " Using the maximum supported version.",
+                        min_version, tls_version_strings[TLS_HIGHEST_SUPPORTED]);
+                    min_tls_version = TLS_HIGHEST_SUPPORTED;
```
olehermanse: Much more clear, thanks!

libcfnet/tls_generic.c
```diff
@@ -819,91 +856,59 @@ void TLSSetDefaultOptions(SSL_CTX *ssl_ctx, const char *min_version)
...
-        else
+        if (!found)
         {
-            Log(LOG_LEVEL_VERBOSE,
-                "Setting minimum acceptable TLS version: 1.3");
+            Log(LOG_LEVEL_WARNING,
+                "Unrecognized requested minimum TLS version '%s',"
+                " using the minimum required version %s.",
+                min_version, tls_version_strings[TLS_LOWEST_REQUIRED]);
+            min_tls_version = TLS_LOWEST_REQUIRED;
```
olehermanse: So this is the case where I would like it to default to `TLS_HIGHEST_SUPPORTED`. (But yes, let's keep it like this for the current release).

libcfnet/tls_generic.c
```diff
@@ -983,7 +988,7 @@ bool TLSSetCipherList(SSL_CTX *ssl_ctx, const char *cipher_list)
         }
     }
 
-#ifdef SSL_OP_NO_TLSv1_3        /* defined if TLS 1.3 is supported */
+#ifdef HAVE_TLS_1_3
```
olehermanse: Much better.

libcfnet/tls_generic.c
vpodzime: Yeah, I also think we could change the default minimum version to TLS 1.2 which is a general recommendation.

libcfnet/tls_generic.c
```diff
@@ -907,3 +908,74 @@ void TLSSetDefaultOptions(SSL_CTX *ssl_ctx, const char *min_version)
...
 }
+
+bool TLSSetCipherList(SSL_CTX *ssl_ctx, const char *cipher_list)
+{
+    assert(ssl_ctx);
+
+    if (cipher_list == NULL)
+    {
+        Log(LOG_LEVEL_VERBOSE, "Using the OpenSSL's default cipher list");
+        /* nothing more to do */
```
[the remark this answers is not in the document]
vpodzime: Well, there's a magic string `"DEFAULT"` so let's see if it works. I'm adding an extra commit to try it out.

libcfnet/tls_generic.c
```diff
@@ -819,91 +856,59 @@ void TLSSetDefaultOptions(SSL_CTX *ssl_ctx, const char *min_version)
...
-        || StringSafeEqual(min_version, "1")
-        || StringSafeEqual(min_version, "1.0"))
-    {
-        /* Do nothing, that's our default setting */
-        Log(LOG_LEVEL_VERBOSE,
-            "Setting minimum acceptable TLS version: 1.0");
-    }
-    else if (StringSafeEqual(min_version, "1.1"))
+    enum tls_version min_tls_version = TLS_LOWEST_REQUIRED;
+    if (min_version != NULL && !StringSafeEqual(min_version, ""))
```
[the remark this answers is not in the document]
vpodzime: Yup!

vpodzime:

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1681/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1681/

libcfnet/tls_generic.c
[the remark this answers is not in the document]
vpodzime:
```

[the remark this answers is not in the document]
vpodzime: Dropping the commit.

vpodzime:

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1682/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/1682/

PR #3372, opened 2018-10-23
ENT-4136: Made SetJoin and StringSetJoin optionally copy elements

olehermanse:
This fixes some small memory leaks in argument parsing
for cf-agent, cf-execd, cf-promises, and cf-serverd.
These leaks were not so serious, as they only happen
once, during startup (they don't increase over time).
It also fixes a memory leak in set_test.

This part of the code is also not performance critical,
so the extra copying should be no problem.

Before this change it was impossible to use SetJoin
without leaking memory. This is because SetJoin reused
elements from second argument, so you end up with 2 sets
with pointers to the same data. Sets only have keys, so there
is no soft destroy function.

Now, if you use the copy function parameter to SetJoin, for
example with xstrdup, both sets can be destroyed using normal
SetDestroy.

StringSetJoin could be made to always use xstrdup, however,
I like the current approach (pass by arg) better because:

1) It is the same as SetJoin
2) It is more explicit, it is more clear what this does:

   StringSetJoin(a, b, xstrdup);

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -449,8 +449,8 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
                 }
                 else
                 {
-                    StringSetJoin(config->heap_soft, defined_classes);
-                    free(defined_classes);
+                    StringSetJoin(config->heap_soft, defined_classes, xstrdup);
+                    StringSetDestroy(defined_classes);
                 }
             }
             break;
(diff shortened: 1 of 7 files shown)


cf-bottom:

olehermanse:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1508/)

https://ci.cfengine.com/job/pr-pipeline/1508/

PR #3374, opened 2018-10-23
[3.10.x] Respect windows newlines when splitting data in PipeReadData()

vpodzime:
This needs to do different things on Windows and UNIX-like systems.

Add a function to split string on newlines to do so because it is
not trivial because the "\r\n" 2-byte sequence
represents a newline on Windows.

Ticket: ENT-3716
Changelog: None
(cherry picked from commit aad945ff22e2f84cb601810db9df409f06aca8da)


cf-bottom:

PR #3388, opened 2018-10-26
Disable codecov comments (3.12)

olehermanse:
I'm tired of the spam. The comment is huge. Because of
fluctuations in coverage it has far more noise than useful
information.


cf-bottom:

PR #3389, opened 2018-10-26
CFE-2931: Fixed an issue with splay time in cf-execd

olehermanse:


nickanderson:
My manual testing failed. I haven't yet reproduced it. I don't know if its a good or bad change.

PR #3391, opened 2018-10-29
Asserts and checks for TYPED_MAP_DEFINE

vpodzime:
This macro defines a typed map with many functions. Most of these
functions dereference their 'map' argument so they should assert
it's not NULL. The two exceptions are the `Prefix##MapDestroy()`
and `Prefix##MapSoftDestroy()` functions which should do nothing
when given `NULL`.


cf-bottom:

PR #3404, opened 2018-11-05
ENT-4136: Fixed memory leaks in unit tests

olehermanse:


olehermanse:
@cf-bottom jenkins

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1640/)

https://ci.cfengine.com/job/pr-pipeline/1640/

libutils/sequence.c
```diff
@@ -553,6 +553,7 @@ Seq *SeqStringDeserialize(const char *const serialized)
             || NULL == (new_str = ValidDuplicate(src, length))
             || src[length] != '\n')
         {
+            free(new_str);
```
karlhto: This can result in an invalid free if any of the first three if's hit.

olehermanse: @karlhto no, look further up at the diff. I initialized it to NULL. `free(NULL)` is valid.

olehermanse: From Linux man page:

> If ptr is NULL, no operation is performed.

From C99 standard:

> If ptr is a null pointer, no action occurs.

karlhto: Alright :+1:

PR #3410, opened 2018-11-12
ENT-3858: Travis: Run unit tests with ASAN

olehermanse:
AddressSanitizer (ASAN) adds instrumentation to binaries
to detect memory leaks, buffer overflows etc.

It is similar to valgrind, but detects other kinds of problems,
and is much more performant (but a makes it a little harder
to build).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ matrix:
   include:
     - os: linux
       sudo: required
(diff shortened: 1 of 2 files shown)


olehermanse:
@vpodzime please review

olehermanse:
Wait for #3411

PR #3414, opened 2018-11-13
Changelog for 3.12.x

Lex-2008:


cf-bottom:

ChangeLog
```diff
@@ -1,3 +1,35 @@
+3.12.1:
+	- Add commit title to changelog by default, if it has ticket number
+	  (ENT-3638)
```
olehermanse: This change isn't really user facing, and can be removed imo.

ChangeLog
```diff
@@ -1,3 +1,35 @@
...
+	- Class names set by module protocol are automatically canonified
+	  (CFE-2877, CFE-2887)
+	- Correct log level for data_readstringarray* (CFE-2922)
+	- Eliminated error messages caused by attempting to kill expired processes
+	  (CFE-2824)
+	- Extend the test to cover more cases (ENT-3959)
```
olehermanse: This doesn't make sense on it's own, and can probably be removed.

ChangeLog
```diff
@@ -1,3 +1,35 @@
...
+	- Fixed memory leak in lmdb cleanup (CFE-2918)
+	- Fixed memory leaks in cf-agent during bootstrap (CFE-2918)
+	- Fixed memory leaks in variablesmatching() and findfiles() (CFE-2918)
+	- Fixed missing class with mustache templates in warn_only mode (CFE-2600)
+	- Fixed small memory leak in cf-serverd (CFE-2918)
+	- Fixed valgrind warning about lmdb writing uninitialized memory (CFE-2591)
+	- Read mustache-rendered files in text mode when comparing digest (ENT-2526)
+	- Reload persistent classes on config reload in cf-execd and cf-serverd
+	  (CFE-2857)
+	- Test that classes can be defined without constraints (CFE-2000)
```
olehermanse: Also a test and can probably be removed.

ChangeLog
Lex-2008: well, it concerns those who are committing:
https://github.com/cfengine/core/blob/master/HACKING.md#changelog-entries

ChangeLog
```diff
@@ -1,3 +1,35 @@
...
+	- Refactor use of atexit to use custom cleanup function instead. On Windows
+	  atexit() unloads DLLs before and/or during atexit functions being called
+	  which cases bad behavior. (ENT-3756)
```
olehermanse: I would reword this one, not really a "refactor" per se. Something like: started using custom cleanup functions instead of `atexit()` which was causing race conditions related to DLL unloading on windows.

Lex-2008:
reworded!

ChangeLog
olehermanse: > When a new feature or a bugfix is being merged, it is often necessary to be accompanied by a proper entry in the ChangeLog file.

> Keep in mind that changelog entries should be written in a way that is understandable by non- programmers. This means that references to implementation details are not appropriate, leave this for the non-changelog part of the commit message. It is the behavior change which is important. This implies that refactorings that have no visible effect on behavior don't need a changelog entry.

IMO, and based on the quotes above, you should edit HACKING.md with this change, and not have it in the Changelog. A change in how changelogs are generated is A) not a bugfix or feature, B) not relevant to users/non-programmers.

Lex-2008: agreed

PR #3421, opened 2018-11-15
Update man pages to remove harded 'agent' on verbose option

atsaloli:
I ran into this in https://support.northern.tech/hc/en-us/requests/4073

man page for cf-runagent said

-v

Output verbose information about the behaviour of the agent

but it didn't; it output verbose information about the behavior of cf-runagent.

Nick Anderson suggested I update the man pages by editing these C files,
so here you go.


cf-bottom:

atsaloli:
Thanks @vpodzime

atsaloli:
:+1:

PR #3435, opened 2018-11-28
ENT-4295: Skip parsing of idle processes in ps output on AIX (3.12)

olehermanse:


cf-bottom:

PR #3456, opened 2018-12-05
Allow sourcing the init script from state restore

vpodzime:
This way we can use the functions from the init script in the
state restoration process where we need to start daemons.

Ticket: ENT-4313


cf-bottom:

vpodzime:
Merge together:
https://github.com/cfengine/buildscripts/pull/502

vpodzime:
Note: This is only needed on 3.10.x for Redis and I think we should not do the same on master.

PR #3467, opened 2018-12-13
ENT-4313: Introduce function to clean up logging contexts (3.12)

olehermanse:
This can be used to eliminate potential small memory leaks
in multi-threaded programs, like cf-hub and cf-serverd.

### DIFF
--- a/libutils/logging.c
+++ b/libutils/logging.c
@@ -66,6 +66,19 @@ LoggingContext *GetCurrentThreadContext(void)
     return lctx;
 }
 
+void LoggingFreeCurrentThreadContext(void)
+{
+    pthread_once(&log_context_init_once, &LoggingInitializeOnce);
+    LoggingContext *lctx = pthread_getspecific(log_context_key);
+    if (lctx == NULL)
+    {
+        return;
+    }
+    // lctx->pctx is usually stack allocated and shouldn't be freed
+    FREE_AND_NULL(lctx);
+    pthread_setspecific(log_context_key, NULL);
+}
+
 void LoggingSetAgentType(const char *type)
 {
     strlcpy(AgentType, type, sizeof(AgentType));
--- a/libpromises/eval_context.c
+++ b/libpromises/eval_context.c
@@ -1007,11 +1007,13 @@ void EvalContextDestroy(EvalContext *ctx)
         free(ctx->launch_directory);
         free(ctx->entry_point);
 
+        // Freeing logging context doesn't belong here...
         {
             LoggingPrivContext *pctx = LoggingPrivGetContext();
             free(pctx);
             LoggingPrivSetContext(NULL);
         }
+        LoggingFreeCurrentThreadContext();
 
         EvalContextDeleteIpAddresses(ctx);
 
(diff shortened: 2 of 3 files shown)


olehermanse:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1802/)

https://ci.cfengine.com/job/pr-pipeline/1802/

libutils/logging.c
```diff
@@ -66,6 +66,19 @@ LoggingContext *GetCurrentThreadContext(void)
...
+void LoggingFreeCurrentThreadContext(void)
+{
+    pthread_once(&log_context_init_once, &LoggingInitializeOnce);
+    LoggingContext *lctx = pthread_getspecific(log_context_key);
+    if (lctx == NULL)
+    {
+        return;
+    }
+    // lctx->pctx is usually stack allocated and shouldn't be freed
+    FREE_AND_NULL(lctx);
```
craigcomstock: I was wondering why `free()` given in the call `pthread_key_create(&log_context_key, &free)` wouldn't take care of cleanup at the right time aka at thread termination time. But it seems LoggingFreeCurrentThreadContext() is called at a different time?

But I haven't yet taken the time to understand exactly how EvalContextDestroy() is used (most likely during teardown/shutdown/termination of some sort) and how the EvalContext and LoggingContext are related.

So can you explain some of this to me? Or explain why your change is correct?

For example, cf-agent main() calls GenericAgentFinalize(ctx, config) right before returning/exiting. So wouldn't the free() function registered as destr function for log_context_key be called?

Obviously the fact that you are fixing this code means that something wasn't quite right.

Sorry. I know this is in a cherry-pick. Feel free to comment here or in slack.

olehermanse: @craigcomstock Logging contexts have very little to do with evaluation contexts, which is why my comment says it should be moved. `GenericAgentFinalize` might be appropriate, yes.

I had not considered the pthread destructor argument ( `&free` ). After reading the pthread manpages I'm more confused.

I'm going by the stack traces in valgrind output (You can see them in the JIRA ticket):
```
==7161== 24 bytes in 1 blocks are definitely lost in loss record 1,343 of 2,799
==7161==    at 0x4C31B25: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==7161==    by 0x4ED395D: xcalloc (alloc.c:48)
==7161==    by 0x4EDD45E: GetCurrentThreadContext (logging.c:61)
==7161==    by 0x4EDD7CA: VLog (logging.c:268)
==7161==    by 0x4EDDAB5: Log (logging.c:353)
==7161==    by 0x40B51C: CollectCallLoop (cf-hub-collect-call.c:149)
==7161==    by 0x5A3F6DA: start_thread (pthread_create.c:463)
==7161==    by 0x5D7888E: clone (clone.S:95)
```

craigcomstock: So based on that output it would seem that the xcalloc() is not being "freed". Possibly because some other code somewhere is setting lctx to NULL. That would be my first guess. Since in the man page we have: 
```The destr_function is not called if that value is NULL.```

PR #3471, opened 2018-12-18
ENT-3832: Thread-safe double ended queue

karlhto:
Currently WIP. Functionality is fine, but a lot of the code could be standardised a lot more.

Also needs tests.

@olehermanse or someone else, add WIP, please.

### DIFF
--- a/libutils/threaded_deque.c
+++ b/libutils/threaded_deque.c
@@ -1,0 +1,641 @@
+/*
+   Copyright 2019 Northern.tech AS
+
+   This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+   This program is free software; you can redistribute it and/or modify it
+   under the terms of the GNU General Public License as published by the
+   Free Software Foundation; version 3.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#include <alloc.h>
+#include <logging.h>
+#include <mutex.h>
+#include <pthread.h>
+#include <threaded_deque.h>
+
+
+#define EXPAND_FACTOR     2
+#define DEFAULT_CAPACITY 16
+
+/** @struct ThreadedDeque_
+  @brief An implementation of a thread safe deque based on a circular array
+
+  Can push left, push right and give various statistics about its contents,
+  like amount of elements, capacity and if it is empty. Has the ability to
+  block if deque is empty, waiting for new elements to be pushed.
+  */
+struct ThreadedDeque_ {
+    pthread_mutex_t *lock;            /**< Thread lock for accessing data. */
+    pthread_cond_t *cond_non_empty;   /**< Blocking condition if empty     */
+    pthread_cond_t *cond_empty;       /**< Blocking condition if not empty */
+    void (*ItemDestroy) (void *item); /**< Data-specific destroy function. */
+    void **data;                      /**< Internal array of elements.     */
+    size_t left;                      /**< Current position in deque.      */
+    size_t right;                     /**< Current end of deque.           */
+    size_t size;                      /**< Current size of deque.          */
+    size_t capacity;                  /**< Current memory allocated.       */
+};
+
+static void DestroyRange(ThreadedDeque *deque, size_t start, size_t end);
+static void ExpandIfNecessary(ThreadedDeque *deque);
+
+ThreadedDeque *ThreadedDequeNew(size_t initial_capacity,
+                                void (ItemDestroy) (void *item))
+{
+    ThreadedDeque *deque = xcalloc(1, sizeof(ThreadedDeque));
+
+    if (initial_capacity == 0)
+    {
+        initial_capacity = DEFAULT_CAPACITY;
+    }
+
+    pthread_mutexattr_t attr;
+    pthread_mutexattr_init(&attr);
+    // enables errorchecking for deadlocks
+    int ret = pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_ERRORCHECK);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to use error-checking mutexes for deque, "
+            "falling back to normal ones (pthread_mutexattr_settype: %s)",
+            GetErrorStrFromCode(ret));
+        pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_NORMAL);
+    }
+
+    deque->lock = xmalloc(sizeof(pthread_mutex_t));
+    ret = pthread_mutex_init(deque->lock, &attr);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to initialize mutex (pthread_mutex_init: %s)",
+            GetErrorStrFromCode(ret));
+        pthread_mutexattr_destroy(&attr);
+        free(deque->lock);
+        free(deque);
+        return NULL;
+    }
+
+    pthread_mutexattr_destroy(&attr);
+
+    deque->cond_non_empty = xmalloc(sizeof(pthread_cond_t));
+    ret = pthread_cond_init(deque->cond_non_empty, NULL);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to initialize thread condition (pthread_cond_init: %s)",
+            GetErrorStrFromCode(ret));
+        free(deque->lock);
+        free(deque->cond_non_empty);
+        free(deque);
+        return NULL;
+    }
+
+    deque->cond_empty = xmalloc(sizeof(pthread_cond_t));
+    ret = pthread_cond_init(deque->cond_empty, NULL);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to initialize thread condition "
+            "(pthread_cond_init: %s)",
+            GetErrorStrFromCode(ret));
+        free(deque->lock);
+        free(deque->cond_empty);
+        free(deque->cond_non_empty);
+        free(deque);
+        return NULL;
+    }
+
+    deque->capacity = initial_capacity;
+    deque->left = 0;
+    deque->right = 0;
+    deque->size = 0;
+    deque->data = xmalloc(sizeof(void *) * initial_capacity);
+    deque->ItemDestroy = ItemDestroy;
+
+    return deque;
+}
+
+void ThreadedDequeDestroy(ThreadedDeque *deque)
+{
+    if (deque != NULL)
+    {
+        DestroyRange(deque, deque->left, deque->right);
+
+        ThreadedDequeSoftDestroy(deque);
+    }
+}
+
+void ThreadedDequeSoftDestroy(ThreadedDeque *deque)
+{
+    if (deque != NULL)
+    {
+        if (deque->lock != NULL)
+        {
+            pthread_mutex_destroy(deque->lock);
+            free(deque->lock);
+        }
+
+        if (deque->cond_non_empty != NULL)
+        {
+            pthread_cond_destroy(deque->cond_non_empty);
+            free(deque->cond_non_empty);
+        }
+
+        if (deque->cond_empty != NULL)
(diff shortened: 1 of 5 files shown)


olehermanse:
@karlhto I changed the title, please update commit also, it is an ENT ticket, not CFE.

olehermanse:
This pull request **introduces 1 alert** when merging a02724962b529426422dc1c6d5bcb2d7010976ea into 414a4a633437f45a113d1eb6372d98076b5914f3 - [view on LGTM.com](https://lgtm.com/projects/g/cfengine/core/rev/pr-91b392d82d70c604dcb01af7e3f55e2a4c24da08)

**new alerts:**

* 1 for For loop variable changed in body

---

*Comment posted by [LGTM.com](https://lgtm.com)*

olehermanse:
This pull request **introduces 1 alert** when merging 2fffa2ffaba32d52588f8bbb9c766c2b4097110a into 414a4a633437f45a113d1eb6372d98076b5914f3 - [view on LGTM.com](https://lgtm.com/projects/g/cfengine/core/rev/pr-1c4ecbaa171b418f73df437173a19573267f53d2)

**new alerts:**

* 1 for For loop variable changed in body

---

*Comment posted by [LGTM.com](https://lgtm.com)*

karlhto:
@cf-bottom Jenkins with exotics, please.

karlhto:
@olehermanse @vpodzime This is no longer WIP, and is ready for review.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2077/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/2077/

tests/unit/Makefile.am
```diff
@@ -90,7 +90,7 @@ libdb_la_SOURCES += ../../libpromises/cf3globals.c
 endif
 libdb_la_LIBADD = libstr.la ../../libutils/libutils.la
 #libdb_la_CPPFLAGS = $(LMDB_CPPFLAGS) $(TOKYOCABINET_CPPFLAGS) $(QDBM_CPPFLAGS)
-# Make sure that source files are compiled to separate object files 
+# Make sure that source files are compiled to separate object files
```
karlhto: Uh. I have configured vim to remove spaces at the end of lines when saving. Should I exclude this from the PR?

karlhto:
Jenkins error seems unrelated

tests/unit/Makefile.am
vpodzime: Ideally, put changes like this into a separate commit like "whitespace cleanup".

olehermanse: "Always" use `git add -p`. Then you are concious about changes like this. You can either skip it, or make it a separate commit.

vpodzime: Or `git commit -p`.

vpodzime: But definitely not a big deal in this case with just one line changed.

olehermanse:
@cf-bottom run in jenkins again with exotics.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2171/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/2171/

libutils/threaded_deque.c
```diff
@@ -0,0 +1,641 @@
...
+                    return false;
+                }
+            } while (deque->size != 0);
+            // Reevaluate predicate to protect against spurious wakeups
+        }
+        else
+        {
+            ret = false;
+        }
+    }
```
vpodzime: I think this would be nicer if it was restructured to deal with the "easy" cases first. So things like
```
if (deque->size == 0)
{
    ThreadUnlock(deque->lock);
    return true;
}
```
and similarly for `timeout`.

libutils/threaded_deque.c
```diff
@@ -0,0 +1,641 @@
...
+        do
+        {
+            deque->ItemDestroy(deque->data[i++]);
+            i %= deque->capacity;
+        } while (i != end);
+    }
+}
+
+/**
+  @brief Either expands capacity of dedeque, or shifts tail to beginning.
```
vpodzime: *dedeque*

libutils/threaded_deque.c
```diff
@@ -0,0 +1,641 @@
...
+    {
+        if (deque->right <= deque->left)
+        {
+            size_t old_capacity = deque->capacity;
+
+            deque->capacity *= EXPAND_FACTOR;
+            deque->data = xrealloc(deque->data,
+                                   sizeof(void *) * deque->capacity);
+
+            memmove(deque->data + old_capacity, deque->data,
```
vpodzime: Aren't the arguments here in a wrong order? Also, is the number of items to be moved really always given by `deque->right`?

libutils/threaded_deque.h
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+  @brief Destroys the deque and frees the memory it occupies.
+  @warning ThreadedDeque should only be destroyed if all threads are joined
+  @param [in] deque  The deque to destroy.
+  */
+void ThreadedDequeDestroy(ThreadedDeque *deque);
+
+/**
+  @brief Frees the memory allocated for the data pointer and the struct itself.
```
vpodzime: *memory allocated for the data pointer*? *pointer***s** maybe?

tests/unit/threaded_deque_test.c
```diff
@@ -0,0 +1,426 @@
...
+
+static void test_push_pop(void)
+{
+    // Initialised with DEFAULT_CAPACITY = 16
+    ThreadedDeque *deque = ThreadedDequeNew(0, free);
+    // |^---------------|
+    ThreadedDequePushLeft(deque, xstrdup("1"));
+    // |<-------------->|
+    ThreadedDequePushRight(deque, xstrdup("2"));
+    // |x<------------->|
```
vpodzime: This doesn't look like a result of `PushRight` to me.

libutils/threaded_deque.c
karlhto: Heh, some of this is directly from `s/queue/deque/g`.

karlhto:
Thanks for thorough reviews! :)

tests/unit/threaded_deque_test.c
karlhto: Whoops. Thanks for noticing!

libutils/threaded_deque.c
karlhto: This case only happens when the right index is lower than the left index, in which case we need to make sure we have a continuous block of memory.

`memmove` takes parameters `dest`, `src`, `size`, in that order.

The first position is the position it will be moved to, which is at the first position of the newly allocated memory, at offset `old_capacity` in `deque->data`.

libutils/threaded_deque.c
```diff
@@ -0,0 +1,641 @@
...
+        free(deque);
+    }
+}
+
+bool ThreadedDequePopLeft(ThreadedDeque *deque,
+                          void **item,
+                          int timeout)
+{
+    assert(deque != NULL);
+
```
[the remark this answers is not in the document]
karlhto: Wasn't there a discussion about this is the `ThreadedQueue` PR? I thought we decided that the data structures should be able to hold `NULL` items. I also have a test for pushing `NULL` that actually pushes it as a result of that discussion.

libutils/threaded_deque.c
karlhto: I assume you mean that the function should be "flatter" as a result, right?

tests/unit/threaded_deque_test.c
karlhto: Wait, yes it is! You push to the right index.

karlhto: It looks confusing, I know.

libutils/threaded_deque.c
```diff
@@ -0,0 +1,641 @@
...
+    return deque;
+}
+
+void ThreadedDequeDestroy(ThreadedDeque *deque)
+{
+    if (deque != NULL)
+    {
+        ThreadLock(deque->lock);
+        DestroyRange(deque, deque->left, deque->right);
+        ThreadUnlock(deque->lock);
```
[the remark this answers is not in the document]
karlhto: I agree, and we talked about this in the previous PR as well.

This function should only be called when you are sure all the threads are joined and the program is about to die.

karlhto:
@olehermanse @vpodzime I have addressed you concerns.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2183/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/2183/

libutils/threaded_deque.c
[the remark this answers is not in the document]
vpodzime: Yeah, we discussed this in the `Queue` case. There's no good way to make these things thread-safe without a lock that would be outside of the `Queue/Dequeue` structure which is something we want to avoid.

libutils/threaded_deque.c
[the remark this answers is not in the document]
vpodzime: There's a difference between `item` and `*item` though. ;-)

libutils/threaded_deque.c
vpodzime: Exactly.

libutils/threaded_deque.c
```diff
@@ -0,0 +1,643 @@
...
+        return true;
+    }
+
+    int ret = true;
+
+    if (timeout == 0)
+    {
+        ret = false;
+    }
```
vpodzime: This can also just unlock and return `false`.

libutils/threaded_deque.c
vpodzime: I see. So it's moving the "overflow" items that are at the beginning of the data segment, but are actually at the end (right side) of the queue in the "abstract picture". Could you please add an explanatory comment above this call?

tests/unit/threaded_deque_test.c
vpodzime: So you push to the right and the item is on the left? :confused:

karlhto: It pushes to the _right end_ of the queue. This means that since the array has wrapped around by first pushing to the left, it looks as if you are pushing on the left, I guess?

If you had only do pushes right, it would look like this:
```
|><--------------|
|>x<-------------|
```
instead of:
```
|<-------------->|
|x<------------->|
```

karlhto:

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2187/)

(with exotics)

https://ci.cfengine.com/job/pr-pipeline/2187/

tests/unit/threaded_deque_test.c
vpodzime: OK, analyzed this with @olehermanse and it makes sense. Maybe the legend should explicitly say that `<` and `>` are positions where the `Pop()` would take its item from not where `Push()` would push its item. That would also make it clear that `<` is actually also `x` (occupied memory).

tests/unit/threaded_deque_test.c
```diff
@@ -0,0 +1,426 @@
...
+ *      x : used memory                 *
+ *      - : unused memory               */
+
+static void test_push_pop(void)
+{
+    // Initialised with DEFAULT_CAPACITY = 16
+    ThreadedDeque *deque = ThreadedDequeNew(0, free);
+    // |^---------------|
+    ThreadedDequePushLeft(deque, xstrdup("1"));
+    // |<-------------->|
```
olehermanse: @karlhto I still find this thoroughly confusing. After 1 push, shouldn't `left == right`, since a PopLeft or a PopRight would both pop the same index?

karlhto: Alright, so:

The right index is the first empty element in a continuous block of data (that can wrap around the allocated memory area). I made it this way so I wouldn't have to have a special case when `size` is `0`, where after 

I also think doing it differently will mess with the internals of the array. It makes sense to me that you should end up at different indices when popping from the left and from the right, e.g.:
```c
/* Starting position (with one element in the data structure), where:
 *     < is right index (first empty element)
 *     - is empty
 *     > is left index (first non-empty element)
 *     ^ is both indexes when array is empty
 */
|<-------------->|

// Pop left:
|^---------------|

// Pop right
|---------------^|
```

Since we already have the `size` parameter, it is not a big deal to change.

vpodzime: I think the legend is missing the information that `<` is actually a first **empty** element while `>` is a **non-empty** element.

vpodzime:
I'd suggest not wasting more time on the nitpicks here. @karlhto please add the empty/non-empty info to the legend and let's merge this. I am now not even sure we will need this structure for something.

karlhto:
@cf-bottom Jenkins with exotics (actual last time)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2256/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2256/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2256/

PR #3481, opened 2019-01-02
Use stdout in LogToConsole

amousset:


cf-bottom:

karlhto:
I think removing `output_file` and explicitly using `stdout` is more descriptive. Doesn't seem like it was intended that the value would change, at least according to this:
```
    FILE *output_file = stdout; // Messages should ALL go to stdout else they are disordered
```

@olehermanse @vpodzime Either of you know? Seems like the variable was added by Mark in 2014.

vpodzime:
> @olehermanse @vpodzime Either of you know?

No idea. Looks like over-engineering to me. Let's get rid of it.

nickanderson:
Ping @olehermanse. @amousset has pushed since your last comment.

PR #3494, opened 2019-01-18
Keep read data when PipeIsReadWriteReady returns negative result

vpodzime:
It also might happen when pipe is closed (child process exited)

Ticket: ENT-3719
Changelog: none

Merge together: https://github.com/cfengine/enterprise/pull/486


cf-bottom:

PR #3495, opened 2019-01-21
Add cf-doc tool to generate policy documentation

mikeweilgart:
The creation of this tool was prompted by the discussion on the help-cfengine list:

https://groups.google.com/forum/#!topic/help-cfengine/_0urLa5EyDo

The script documentation is inline as a usage message (visible with "--help").
To quote the first paragraph:

cf-doc will extract particular meta tags and meta promises
from CFEngine policy and will output a list of them in
Markdown format, to document and index a codebase of
CFEngine policy.


cf-bottom:

contrib/cf-doc/get-meta-promises.jq
```diff
@@ -0,0 +1,3 @@
+#!/bin/jq -f
+
+.bundles[]|{bundle: .name,file: .sourcePath,toBeFiltered: (.promiseTypes[]|select(.name == "meta")|.contexts[]|{context: .name,promises: .promises[]})}|{bundle,file,context: .toBeFiltered.context,promiser: .toBeFiltered.promises.promiser,metacomment: (.toBeFiltered.promises.attributes[]|select(.lval == "string")|.rval.value),linenumber: (.toBeFiltered.promises.attributes[]|select(.lval == "string")|.line)}
```
Lex-2008: Is it possible to add some kind of comment explaining what it's doing? Sample input/output and maybe text version, like:
> for each item in "bundles", build object with name and file and temporary toBeFiltered object holding meta information (object with name and array of promises for each item in "contexts" build).

(if I read it right, there might be some intermediate result useful for understanding this command)
> Then, take resulting bundle objects and reconstruct them to extract context and promiser from the toBeFiltered object together with metacomment which is `.rval.value` of `promises.attribute` element with `.lval == "string"` and a linenumber where it happened.

Maybe something more human readable is possible...

contrib/cf-doc/get-meta-promises.jq
```diff
@@ -1,3 +1,3 @@
 #!/bin/jq -f
 
-.bundles[]|{bundle: .name,file: .sourcePath,toBeFiltered: (.promiseTypes[]|select(.name == "meta")|.contexts[]|{context: .name,promises: .promises[]})}|{bundle,file,context: .toBeFiltered.context,promiser: .toBeFiltered.promises.promiser,metacomment: (.toBeFiltered.promises.attributes[]|select(.lval == "string")|.rval.value),linenumber: (.toBeFiltered.promises.attributes[]|select(.lval == "string")|.line)}
+.bundles[]|{bundle: .name,file: .sourcePath} + (.promiseTypes[]|select(.name == "meta")|.contexts[]|{context: .name} + (.promises[]|{promiser} + (.attributes[]|select(.lval == "string")|{metacomment: .rval.value,linenumber: .line})))
```
Lex-2008: The command might still benefit from some explanation. Wrapping command into lines, if possible, might also help readability.

nickanderson:
A long time ago https://tracker.mender.io/browse/CFE-164 there was some minimal work done on a tool called cf-doc. It's intended use was a bit different. As I recall, it was cli access to documentation. So you could search for a promise type, attribute name, function etc ... and jump to the spot in info docs. The docs moved away from texinfo to markdown shortly after that and it never moved forward.

So, my only minor comment is that cf-doc makes me think of looking up docs about cfengine (not necessarily custom policy). I do think this would be a good tool to put into contrib, but I would consider a name change to something that reflects extraction of docs from policy. That being said, there is no existing cf-doc, so I am probably one of the few people that would find the name a bit confusing.

contrib/cf-doc/cf-doc.sh
```diff
@@ -0,0 +1,160 @@
...
+        are as defined in the "collection" variable in this script.
+        For further details on this see the inline documentation
+        in the supporting script 'extract-cf-meta.jq'.
+
+OPTIONS
+        -f FILE         File to pass to cf-promises for parsing.
+                        If not specified cf-promises will use its own
+                        default of /var/cfengine/inputs/promises.cf
+
+        -p VERSION      Version of the policy that is being checked,
```
nickanderson: The version of the MPF policy? Perhaps it could try to derive the default from the MPF directly.
https://github.com/cfengine/masterfiles/blob/master/promises.cf.in#L52
But, I don't think there is yet a variable that exposes this. I think I would call it sys.policy_entry_version if we added one.

mikeweilgart: Really the purpose is to allow linking directly to the line of code in GitLab or GitHab or other version control system.  Just exactly like the link you just wrote.  :)  It contains "master" and hence the default value of master.

contrib/cf-doc/format-cf-markdown.jq
```diff
@@ -0,0 +1,26 @@
...
+#
+# Author: Mike Weilgart
+# Date: 18 January 2019
+# Purpose: Format specific json object into a pretty markdown list.
+#
+# The variables are expected to be set before this script runs, e.g.:
+#
+#   cat input.json |
+#   jq --arg url_prefix "https://example.com/cfengine/masterfiles/blob/" \
+#     --arg version "master" -r -f thisscript
```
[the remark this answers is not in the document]
mikeweilgart: Hi Lex, thanks for reviewing this!  Could you please review the final result—the two files in the merge request?

If the development work is squashed for the merge it wouldn't break my heart, though I like preserving history.  But for review purposes, everything you need is right there in the final diff (addition of two files).

mikeweilgart:
@nickanderson regarding the minimal work on the prior tool in the line of documenting the CFEngine language itself, akin to perldoc—this tool is intended for generating documentation for CFEngine policy or libraries, akin to go doc.  The tool go doc is for extraction of documentation from go code, not for information on the syntax of the go language itself; that's where I got the name from.

> I would consider a name change to something that reflects extraction of docs from policy.

Hopefully the etymology I just provided helps.  :)  It's true that "go doc" and "perldoc" reflect very different things, but so long as the documentation is clear about what *this* tool does, I think it's fine with the name cf-doc.

@vpodzime this is my development branch with all of the development work.  Since there are only two files being added, all of the intermediate work can be elided (squashed) and only the final result merged into the cfengine/core repo.

If there's no option here for "squash and merge" (GitLab has that feature, I don't know about GitHub) then I can push a squashed version so it's just one commit to be merged.  Please advise?

(In this case I would keep the history in my own fork anyway, though; there are some interim results that others might find useful if they're tinkering with CFEngine and jq.)

olehermanse:
> If there's no option here for "squash and merge" (GitLab has that feature, I don't know about GitHub) then I can push a squashed version so it's just one commit to be merged. Please advise?

> (In this case I would keep the history in my own fork anyway, though; there are some interim results that others might find useful if they're tinkering with CFEngine and jq.)

@mikeweilgart That works. GitHub has `Squash and Merge`, but we disabled it. We prefer if you squash your commits locally. (We disabled rebase & merge, and squash & merge, because it redoes the commits on githubs servers, changing the commit SHA and removing any GPG signatures). Otherwise, I can temporarilty enable it to merge this PR.

mikeweilgart:
@olehermanse I've squashed my commits and pushed.  I also rebased to master, since it had advanced.

PR #3502, opened 2019-02-05
Enable all LGTM C/++ queries

olehermanse:


olehermanse:
@vpodzime According to the folks at LGTM, changes like this won't be reflected in PR builds. Let's merge and see?

PR #3503, opened 2019-02-05
pkg-config

hdatma:
Problem: configuration failure on macos
Solution: use of pkg-config


hdatma:
Explanation of Travis error.

I used "which -s".
>   -s      No output, just return 0 if all of the executables are found, or 1 if some were not found.

PR #3507, opened 2019-02-05
Change /var/cfengine/backup/ to backups/

mikeweilgart:
The default location for file backups by agent file promises
is /var/cfengine/backups/.  It would be nicer not to have
both backup/ and backups/ (confusing).

### DIFF
--- a/cf-check/backup.c
+++ b/cf-check/backup.c
@@ -26,7 +26,7 @@ int backup_files(Seq *filenames)
 const char *create_backup_dir()
 {
     static char backup_dir[PATH_MAX];
-    const char *const backup_root = "/var/cfengine/backup/";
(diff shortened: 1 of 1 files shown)


cf-bottom:

nickanderson:
I really don't have an opinion on this change.

In core, there is no default setting for body agent control default_repository https://docs.cfengine.com/docs/3.12/reference-components-cf-agent.html#default_repository

In the MPF, defining the mpf_control_agent_default_repository class will cause these backups to be placed in `$(sys.workdir)/backups`

olehermanse:
Thanks @mikeweilgart. I'll accept this, since it is a slight improvement, but I will probably follow up and change it to `/var/cfengine/backups/state`.

olehermanse:
**Cherry-pick:** #3555

PR #3512, opened 2019-02-19
cf-remote: --hub and --bootstrap now accept more than one IP/hostname

olehermanse:
Useful if you want to install multiple hubs in one command.
This is a more niche use case, so I won't change the name or
description of the options. (99% of users will want 1 hub and 1
bootstrap IP).


olehermanse:
@vpodzime @craigcomstock Please review.

PR #3516, opened 2019-02-20
CFE-3008 Message about invalid class characters from module protocol moved to VERBOSE (master)

nickanderson:


cf-bottom:

PR #3519, opened 2019-02-20
CFE-3008 Message about invalid class characters from module protocol moved to VERBOSE (3.7.x)

nickanderson:
In CFE-2887 the module protocol began automatically canonifying classes
that were defined in order to align with the behaviour of policy. This
change moves a warning about the class containing invalid characters to
VERBOSE as to align with policy.

(cherry picked from commit 829afc3d3b786cc3b5d8c469f1023d589b9dc33c)


cf-bottom:

PR #3528, opened 2019-03-04
Avoid swallowing the number part of YAML string value

vpodzime:
The JsonParseAsNumber() tries to parse the given string as a
number, *and also moves the pointer to the point it failed to
continue doing so*. To use it as a check function we need to give
it a pointer it can mangle instead of the pointer we plan to
continue using.

Ticket: CFE-2033
Changelog: Fix parsing YAML values starting with numbers

### DIFF
--- a/libutils/json-yaml.c
+++ b/libutils/json-yaml.c
@@ -111,7 +111,8 @@ static JsonElement* JsonParseYamlScalarValue(yaml_event_t *event)
          JSON_YAML_SCALAR_TAG_IS(tag, YAML_FLOAT_TAG)))
     {
         JsonElement *tobuild;
-        if (JSON_PARSE_OK == JsonParseAsNumber(&value, &tobuild))
+        const char *end_of_num_part = value;
+        if (JSON_PARSE_OK == JsonParseAsNumber(&end_of_num_part, &tobuild))
         {
             return tobuild;
         }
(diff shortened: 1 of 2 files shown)


cf-bottom:

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2098/)

https://ci.cfengine.com/job/pr-pipeline/2098/

olehermanse:
@vpodzime If `value` was a `*const` pointer this would be a warning ;) `*const` for president.

```
$ cat test.c 
void edit_pointer(const char **address)
{
  *address = 0x0;
}

int main(int argc, char **argv)
{
  const char *const str = "Hello";
  edit_pointer(&str);
  return 0;
}
$ gcc test.c -o test
test.c:9:16: warning: passing 'const char *const *' to parameter of type 'const char **' discards qualifiers
      [-Wincompatible-pointer-types-discards-qualifiers]
  edit_pointer(&str);
               ^~~~
test.c:1:32: note: passing argument to parameter 'address' here
void edit_pointer(const char **address)
                               ^
1 warning generated.
$ 
```

vpodzime:
> @vpodzime If `value` was a `*const` pointer this would be a warning ;) `*const` for president.

`const char *value = (const char *) event->data.scalar.value;` -- close :)

PR #3529, opened 2019-03-04
cf-remote: Better command output and error handling

olehermanse:
This preserves formatting on the first line,
for example in cf-agent output.


olehermanse:
@vpodzime Please review.

PR #3533, opened 2019-03-07
Avoid swallowing the number part of YAML string value

vpodzime:
The JsonParseAsNumber() tries to parse the given string as a
number, *and also moves the pointer to the point it failed to
continue doing so*. To use it as a check function we need to give
it a pointer it can mangle instead of the pointer we plan to
continue using.

Ticket: CFE-2033
Changelog: Fix parsing YAML values starting with numbers
(cherry picked from commit 48619fc0f3bf1dab31c02b9425d88e8ada9e35c3)


cf-bottom:

PR #3545, opened 2019-03-21
Fix a segfault in 'cf-promises -p json-full'

vpodzime:
When adding promise attributes there may be none in which case
"rval" is an empty array not an object, like in this example::

  {
    "attributes": [
      {
        "line": 202,
        "lval": "data",
        "rval": []
      }
    ],
    "line": 202,
    "promiser": "tests_passed"
  }

AttributeValueToJson() can return an array as well as an object
so we need to take that into account when appending its result.

Ticket: CFE-3019
Changelog: Title

### DIFF
--- a/libpromises/policy.c
+++ b/libpromises/policy.c
@@ -1642,6 +1642,11 @@ static Buffer *EscapeQuotes(const char *raw, Buffer *out)
     return out;
 }
 
+/**
+ * Converts the given attribute rval to a JSON object.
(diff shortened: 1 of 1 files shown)


cf-bottom:

olehermanse:
@vpodzime Thanks for fixing. We should add a test which runs `cf-promises -p json-full` on `promises.cf` and `update.cf`

PR #3548, opened 2019-03-21
CFE-3023: Retry SSL_read on SSL_ERROR_WANT_READ (3.10)

nickanderson:
Changelog: Title

When the client gets SSL_ERROR_WANT_READ, we should re-try instead of shutting
down the whole session.

(cherry picked from commit c72d1c65eef399faa2a5b5cc65fe95f48bcac3a5)


vpodzime:
Superseded by #3618

PR #3550, opened 2019-03-21
ENT-4497: Fixed inconsistent NULL checks (3.10)

olehermanse:
In all cases, if the functions would return
NULL previously it would result in undefined
behavior / crashes (deref NULL pointer). So
logging errors and/or stopping agent should
be safe changes.

Reported by LGTM:
https://lgtm.com/rules/2151900540/


olehermanse:
Also fixes: https://tracker.mender.io/browse/ENT-4497

PR #3556, opened 2019-03-22
Reorganized and changed contribution and code style guidelines

olehermanse:
Tried to make them more friendly and useful to new contributors:

* More friendly language
* Easier to read
* Organized into sections
* More concise/direct
* More examples
* Removed some noise

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -27,31 +27,35 @@ in order to make it to the first minor release.
 Pull Requests
 -------------------------------------------------
 
-### Top reasons pull-requests are rejected or delayed
+### Checklist
 
-* Code does not follow style guidlines. See [Coding Style](#coding-style).
+When submitting your pull request, please make sure you:
 
-* Pull request addresses several disparate issues. In general, smaller
-pull-requests are better because they are easier to review and stay mergeable
-longer.
+* Follow our [Coding Style](#coding-style).
 
-* Big feature is added, but it is not configurable in compile-time.
-We are striving to keep CFEngine lightweight and fast, so big new
-features should be possible to disable with
-`./configure --disable-feature` and linking to new libraries
-should be optional with `./configure --without-libfoo`.
+* Address only one issue/feature.
+  Smaller pull requests are generally better.
 
-* Messy commit log. Tidy up the commit log by squashing commits.
+* Add tests.
+  C functions should have unit tests.
+  Promise types, attributes, and functions should have acceptance tests.
 
-* Missing ChangeLog description in commit message, which is mandatory
-for new features or bugfixes to be accepted.
-See [ChangeLog Entries](#changelog-entries) for details.
+* Pay attention to review comments and CI results.
+  We have code analysis and tests which run when you submit your PR.
+  You may have to make some changes.
 
-* Code is out-of-date, does not compile, or does not pass all tests. Again,
-focused and small pull-requests are better.
+* Check that you don't have any merge conflicts.
+  (Rebase on master or target branch).
 
-* No attached test case. Normally, all new code needs test cases. This means a
-functional test runnable with `make check`.
+* Tidy up the commit log by squashing commits.
+  Each commit should be a valid change and make sense on its own.
+  Usually a Pull Request will only have one commit.
+
+* Add a [ChangeLog Entry](#changelog-entries) to the commit.
+
+* Add a configure option if appropriate:
+  `./configure --disable-big-feature` or `./configure --without-libfoo`.
+
 
 ### Multi-repo pull requests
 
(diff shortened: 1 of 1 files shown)


cf-bottom:

HACKING.md
```diff
@@ -38,7 +38,7 @@ When submitting your pull request, please make sure you:
 
 * Add tests.
   C functions should have unit tests.
-  Promise types, attributes and functions should have acceptance tests.
+  Promise types, attributes, and functions should have acceptance tests.
```
vpodzime: Why this comma? (I'm no expert on English comma-grammar, but it feels weird here)

HACKING.md
```diff
@@ -79,96 +79,197 @@ The bigger PR will likely take some time to get reviewed, and discussed, while s
...
-  In order to improve grep-ability. If they do not fit on a single line,
-  try breaking on punctuation. In worst case scenario, you are allowed
-  to surpass the 78 columns limit.
+### Formatting / Whitespace
 
-  Bad:
+* Feel free to use `clang-format` to format new code you submit.
+  (Our configuration is in `.clang-format`).
+* 4 spaces per indentation, no tabs.
+* Fold new code at 78 columns.
```
vpodzime: I still hate this.

HACKING.md
```diff
@@ -79,96 +79,197 @@ The bigger PR will likely take some time to get reviewed, and discussed, while s
...
+      } Point;
+      ```
+      Don't add a trailing `_t` or similar for typedefs.
+    * Structs and functions which modify them are similar to classes and methods in other languages.
+        * Method names should start with the struct name.
+        * The first parameter should be the "self" pointer:
+          ```c
+          void SeqAppend(Seq *seq, void *item);
+          ```
+* Document using Doxygen (within reason), preferably in the `.c` files,
```
vpodzime: Looks like the markup is wrong here (it's rendered in italics instead of as a bullet point). Also, I don't really agree with this. I hate to open `.c` files only to read what a particular function does. I wish we had all the doc comments in the header files.

HACKING.md
```diff
@@ -79,96 +79,197 @@ The bigger PR will likely take some time to get reviewed, and discussed, while s
...
+not the header files.
+
+
+### Safety / Correctness
+
+* Minimize use of global variables.
+* Use pure functions when possible:
+    * Functions with no / few side effects are easier to reuse.
```
vpodzime: I'd also mention avoiding use of global variables.

HACKING.md
```diff
@@ -79,96 +79,197 @@ The bigger PR will likely take some time to get reviewed, and discussed, while s
...
+    ```c
+    if (data != NULL)
+    ```
+    * Have the literal (constant) to the right side, so it reads well:
+    ```c
+    if (age >= 33)
+    ```
+    * Compare to the correct "type" literal:
+    ```c
+    if (character == '\0' && integer == 0)
```
vpodzime: This is also mentioned above.

HACKING.md
olehermanse: Let's do a separate discussion or PR for changing it then :)

HACKING.md
karlhto: I agree with having documentation in the header files. Long documentation entries can clutter the code, and makes navigating the file slower.

The exception is for `static` functions obviously.

HACKING.md
```diff
@@ -27,31 +27,35 @@ in order to make it to the first minor release.
...
 
-* Code is out-of-date, does not compile, or does not pass all tests. Again,
-focused and small pull-requests are better.
+* Check that you don't have any merge conflicts.
+  (Rebase on master or target branch).
+
+* Tidy up the commit log by squashing commits.
+  Each commit should be a valid change and make sense on its own.
+  Usually a Pull Request will only have one commit.
+
```
craigcomstock: separate whitespace commits?

HACKING.md
olehermanse: This renders ok on github. Just a problem with their syntax highlighter.

Let's discuss these things in a new PR.

HACKING.md
olehermanse: Thanks.

HACKING.md
olehermanse: There is a separate whitespace commit, but I squashed the 2 other commits so this makes more sense.

olehermanse:
Addressed review comments. GitHub's markdown syntax highlighter is not working so well, use the markdown preview instead.

olehermanse:
@vpodzime @karlhto Let's open new pull requests about changing some of the guidelines we don't agree with.

HACKING.md
craigcomstock: No I mean in the text specify that PRs should include a separate commit for whitespace changes.

PR #3557, opened 2019-03-22
Fix a segfault in 'cf-promises -p json-full' (3.12)

vpodzime:
When adding promise attributes there may be none in which case
"rval" is an empty array not an object, like in this example::

  {
    "attributes": [
      {
        "line": 202,
        "lval": "data",
        "rval": []
      }
    ],
    "line": 202,
    "promiser": "tests_passed"
  }

AttributeValueToJson() can return an array as well as an object
so we need to take that into account when appending its result.

Ticket: CFE-3019
Changelog: Title
(cherry picked from commit 98e667b7beb9537daf0585a79aae86d47f260afb)

### DIFF
--- a/libpromises/policy.c
+++ b/libpromises/policy.c
@@ -1590,6 +1590,11 @@ static Buffer *EscapeQuotes(const char *raw, Buffer *out)
     return out;
 }
 
+/**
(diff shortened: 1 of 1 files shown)


cf-bottom:

PR #3567, opened 2019-03-25
Fix class propagation for Mustache templates

pks-t:
When calling a bundle that is expected to act on a Mustache template,
then its classes are not properly being propagated to the calling
bundle. E.g. with the following promises, only "Sub: repaired" will be
printed if the "example.target" file promise has been repaired.

```
    bundle agent sub
    {
    files:
      "example.target"
        create => "true",
        edit_template => "example.template",
        classes => if_repaired("template_repaired");

    reports:
      template_repaired::
        "Sub: repaired";
    }

    bundle agent main
    {
    methods:
      "any" usebundle => sub(),
        classes => if_repaired("template_repaired");

    reports:
      template_repaired::
        "Main: repaired";
    }
```

The root issue here is how rendering of Mustache templates and file edit
contexts work. At the end of every file promise, the function
`FinishEditContext` will be called that has the responsibility of
properly setting the promise result. Even if the file promise announces
a change, the function will re-adjust that result to only report a
change if the following conditions are met:

    - there is an edit context
    - that edit context has at least one edit
    - edited contents do not match the target file content

While "normal" edits do meet those criteria, a Mustache based edit does
not increment the number of edits. Thus, the initial result of "CHANGE"
will be reverted to "NOOP" and no classes will be propagated.

Fix this issue by properly adjusting the `num_edits` counter when a
Mustache template leads to a "CHANGE"d promise. As this will cause
`FinishEditContext` to try to apply the changed results, we now have to
populate the list of changes `file_start` with the new lines. As a
result, we can now completely rely on the updating logic of
`FinishEditContext` and thus remove the equivalent logic in
`RenderMustacheTemplate`.

CFE-1770: Mustache template repair of existing file not propagated to client bundle
CFE-3010: Mustache promise changes do not propagate when used as part of another bundle
Signed-off-by: Patrick Steinhardt <ps@pks.im>


cf-bottom:

cf-agent/verify_files.c
```diff
@@ -639,29 +607,23 @@ static PromiseResult RenderTemplateMustache(EvalContext *ctx, const Promise *pp,
...
         HashString(BufferData(output_buffer), BufferSize(output_buffer), rendered_output_digest, CF_DEFAULT_DIGEST);
+
         if (!HashesMatch(existing_output_digest, rendered_output_digest, CF_DEFAULT_DIGEST))
         {
-            if (a.transaction.action == cfa_warn || DONTDO)
-            {
-                cfPS(ctx, LOG_LEVEL_WARNING, PROMISE_RESULT_WARN, pp, &a,
-                     "Need to update rendering of '%s' from mustache template '%s' but policy is dry-run",
-                     pp->promiser, message);
-                result = PromiseResultUpdate(result, PROMISE_RESULT_WARN);
```
vpodzime: This `DONTDO` check and related things have to be preserved.

cf-agent/verify_files.c
```diff
@@ -639,29 +607,23 @@ static PromiseResult RenderTemplateMustache(EvalContext *ctx, const Promise *pp,
...
+            char *d = (char *) BufferData(output_buffer), *s, *last;
```
vpodzime: Please declare/define variables on separate lines.

cf-agent/verify_files.c
```diff
@@ -639,29 +607,23 @@ static PromiseResult RenderTemplateMustache(EvalContext *ctx, const Promise *pp,
...
+
+            while ((s = strtok_r(d, "\n", &last)) != NULL) {
```
vpodzime: Shouldn't this split the rendered result into lines and populate the `edcontext->file_start` list with all the lines? Why is it just appending the last line there?

cf-agent/verify_files.c
vpodzime: Also please give the variables more descriptive names.

cf-agent/verify_files.c
pks-t: These shouldn't be necessary anymore. With these changes, we're now completely relying on `FinishEditContext` to apply the file edits. And as `FinishEditContext` does already handle "DONTDO" and "cfa_warn", we shouldn't have to replicate these checks.

cf-agent/verify_files.c
pks-t: Will do.

cf-agent/verify_files.c
pks-t: Well, originally I tried to simply append the complete output buffer. But as `SaveItemListAsFile` appends line endings to each item, it will result in an additional "\n" at the end of the edited file. That's why I decided to try to split the output buffer by newlines and append each single line via `AppendItem` (note that it's not only the last line, but I iterate through all the lines and append them as item). Furthermore, this should also have the side effect of correctly honoring `NewLineMode`

pks-t:
> Thanks for working on this! Unfortunately, the proposed changes seem to break almost all our mustache-rendering tests so there's clearly something going wrong.

That's curious. I obviously tested those changes before submitting them, and I didn't get any errors at all. Huh. Maybe the test suite was broken due to me using "--gainroot=env", like proposed by `testall --help`.

pks-t:
After tearing my hair out for another hour I finally decided that untangling the current code would probably be too much of a hassle. So I've changed gears and switched to the simplest possible fix instead. It now passes both my new test as well as all existing ones.

cf-agent/verify_files.c
vpodzime: :+1: thanks for the explanation.

cf-agent/verify_files.c
vpodzime: > note that it's not only the last line, but I iterate through all the lines and append them as item

Wow, I don't know how I managed to overlook the `while`, sorry for the noise. Maybe I got confused by the name of the variable `last`.

vpodzime:
> Maybe the test suite was broken due to me using "--gainroot=env", like proposed by `testall --help`.

That's quite possible, I think it at least skips some tests. To be honest, we are all using PRs to run tests on a WIP code in Travis and Jenkins. :-)

vpodzime:
> After tearing my hair out for another hour I finally decided that untangling the current code would probably be too much of a hassle. So I've changed gears and switched to the simplest possible fix instead. It now passes both my new test as well as all existing ones.

**THANKS** for the extra effort! I've myself ended in a similar situation whenever I touched the mustache-rendering code. A think this minimal fix is the best way to go.

vpodzime:

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2230/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2230/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2230/

vpodzime:
> Build Status

There were some issues with spawning a Debian 4 VM, but everything is green!

pks-t:
Thanks for merging!

PR #3571, opened 2019-03-26
cf-remote: Fixed typo in output

olehermanse:


olehermanse:
Spotted by @craigcomstock

PR #3574, opened 2019-03-28
Set policy->release_id to "failsafe"/"bootstrap" when running failsafe.cf

vpodzime:
This makes it clear in the logs that the policy run was running
the failsafe policy to either bootstrap or try to recover.

Ticket: CFE-3031
Changelog: Title


cf-bottom:

vpodzime:
@olehermanse, @nickanderson, what do you think about backporting this to 3.12.x and 3.10.x?

olehermanse:
If you have good reasons to backport it (users need it, or you need it to fix a bug) go for it.

nickanderson:
+1

PR #3577, opened 2019-03-28
Fix class propagation for Mustache templates

vpodzime:
When calling a bundle that is expected to act on a Mustache template,
then its classes are not properly being propagated to the calling
bundle. E.g. with the following promises, only "Sub: repaired" will be
printed if the "example.target" file promise has been repaired.

```
    bundle agent sub
    {
    files:
      "example.target"
        create => "true",
        edit_template => "example.template",
        classes => if_repaired("template_repaired");

    reports:
      template_repaired::
        "Sub: repaired";
    }

    bundle agent main
    {
    methods:
      "any" usebundle => sub(),
        classes => if_repaired("template_repaired");

    reports:
      template_repaired::
        "Main: repaired";
    }
```

The root issue here is how rendering of Mustache templates and file edit
contexts work. At the end of every file promise, the function
`FinishEditContext` will be called that has the responsibility of
properly setting the promise result. A "CHANGE" will only be propagated
in case where there has been at least one edit in the current editing
context and if the target file's contents do not match the edited
contents.

The Mustache rendering code is incompatible with this as it will never
increment the `num_edits` variable of the editing cntext. Fixing this is
not going to work either, though, as that would trigger the updating
logic of `FinishEditContext` that causes it to write all changed lines
to the target file. As this is handled by `RenderTemplateMustache`
already, this would essentially wipe out the target file's contents.

To fix this, we instead introduce a new field into the editing context
which will track whether the target files has been rewritten already.
If so, `FinishEditContext` will know to propagate the "CHANGE".

CFE-1770: Mustache template repair of existing file not propagated to client bundle
CFE-3010: Mustache promise changes do not propagate when used as part of another bundle
Signed-off-by: Patrick Steinhardt <ps@pks.im>
(cherry picked from commit 07c48ccee21f39618f753743beafe309f02c2650)


cf-bottom:

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2248/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2248/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2248/

PR #3590, opened 2019-04-08
Do not warn about multiple cf-hub processes

vpodzime:
cf-hub is forking and also it is fine to run it multiple times in
parallel as long as just one instance runs the main loop. The

 "Warning: Multiple cf-hub processes present. Should restart CFEngine."

message is just confusing in such situations.


cf-bottom:

PR #3592, opened 2019-04-09
cf-info.sh - fixed issue with update.cf and promises.cf calculations

nicholasa78:
cf-info.sh assumed the last 2 runs in promise_summary.log were for update.cf and promises.cf.   Added logic to specify the last update.cf run and the last non-update.cf run to be used instead.

Also changed the "Last cf-agent run started x minutes ago and lasted y seconds" code to only show the last non-update.cf run.

### DIFF
--- a/contrib/cf-info/cf-info.sh
+++ b/contrib/cf-info/cf-info.sh
@@ -28,9 +28,11 @@ echo
 echo "Host key:"
 /var/cfengine/bin/cf-key -p /var/cfengine/ppkeys/localhost.pub
 echo
-awk -F '[:,]' -v time="$(date +%s)" 'END {printf "Last cf-agent run started %d minutes ago and lasted %d seconds\n", (time - $1)/60, $2 - $1}' /var/cfengine/promise_summary.log
+awk -F '[:,]' -v time="$(date +%s)" '!/update.cf/ {saved = $0} END {$0 = saved; printf "Last cf-agent run started %d minutes ago and lasted %d seconds\n", (time - $1)/60, $2 - $1}' /var/cfengine/promise_summary.log
 echo
-tail -n 2 /var/cfengine/promise_summary.log | sed 's/.*Total promise compliance: \([^.]*\).*/\1/;1s/^/update.cf compliance:   /;2s/^/promises.cf compliance: /'
+# h = hold that line for later use; g = get the line last written to the "hold space"; $ = do this only when last line is reached.
+sed -n '/update.cf/  h; $ { g; s/.*Total promise\( compliance: [^.]*\).*/\1/; s/^/  update.cf/p; }' /var/cfengine/promise_summary.log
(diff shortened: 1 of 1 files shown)


cf-bottom:

atsaloli:
The situation we were running into, a fraction of the time cf-info would swap the promises.cf ourcomes with the update.cf outcomes; that is to say, if the last 2 entries in promise summary log were for promises.cf and update.cf (as opposed to update.cf and promises.cf, which is what you usually see), cf-info would report them as "update.cf" and "promises.cf" (while actually showing promises.cf and update.cf) which was confusing.  Nick's pull request fixes that.

atsaloli:
:+1:

atsaloli:
We are going to update this pull request, we have a better (more portable) way of doing this check.

nicholasa78:
It has been pointed out that using tac may not be the most universal way to accomplish this. I've made changes to use sed and awk to be more POSIX compliant

contrib/cf-info/cf-info.sh
```diff
@@ -28,9 +28,11 @@ echo
 echo "Host key:"
 /var/cfengine/bin/cf-key -p /var/cfengine/ppkeys/localhost.pub
 echo
-awk -F '[:,]' -v time="$(date +%s)" 'END {printf "Last cf-agent run started %d minutes ago and lasted %d seconds\n", (time - $1)/60, $2 - $1}' /var/cfengine/promise_summary.log
+grep -v update.cf  /var/cfengine/promise_summary.log | awk -F '[:,]' -v time="$(date +%s)" 'END {printf "Last cf-agent run started %d minutes ago and lasted %d seconds\n", (time - $1)/60, $2 - $1}'
```
vpodzime: Can the regex matching be done in `awk` too?

nickanderson: It's possible for there to be other entries, so if you are only interested in promises.cf, perhaps:

```
awk -F '/CFEngine Promises.cf/[:,]' -v time="$(date +%s)" 'END {printf "Last cf-agent run started %d minutes ago and lasted %d seconds\n", (time - $1)/60, $2 - $1}' /var/cfengine/promise_summary.log 
Last cf-agent run started 4 minutes ago and lasted -1554913420 seconds
```

mikeweilgart:
Initially I thought the grep solution was a lot simpler than it would be to combine it all into awk, but on second look it seems doable:

    awk -F '[:,]' -v time="$(date +%s)" '!/update.cf/ {saved = $0} END {$0 = saved; printf "Last cf-agent run started %d minutes ago and lasted %d seconds\n", (time - $1)/60, $2 - $1}' /var/cfengine/promise_summary.log

@nicholasa78, could you make this update?

nicholasa78:
Made the changes suggested to remove grep and use awk for the last non-update.cf cf-agent run.

atsaloli:
Thanks, Vrata!

-- 
Achieve real learning.  Email training@verticalsysadmin.com.

PR #3596, opened 2019-04-12
proofreading changes

craigcomstock:


craigcomstock:
Am reading through this entire document just to keep it fresh in my mind and will add any bits I think need fixing. One thing I wonder about is spacing around '/'. I don't think we need spaces and should be consistent. Such as "cannot / should" versus "cannot/should". Oxford style guide doesn't mention this directly but in its explanatory text uses no spaces consistently.

CONTRIBUTING.md
```diff
@@ -270,9 +270,9 @@ For other integer types without a format, cast `signed` types to `intmax_t` and
 Logging Conventions
 -----------------------------------
 
-CFEngine outputs messages about what its doing using the `Log()` function. It
-takes a `LogLevel` enum mapping closely to syslog priorities. Please try to do
-the following when writing output messages.
+CFEngine outputs messages about what it's doing using the `Log()` function. It
+takes a `LogLevel` enum mapping closely related to syslog priorities. Please try
```
olehermanse: ~related~

I prefer how it was before, but you can maybe add a comma after `enum`.

craigcomstock: The comma helps me. Otherwise I was reading it as "enum mapping" instead of an "enum (which is mapping) ..." :+1:

craigcomstock:
started this as an edit in github.com but would rather have as a branch from my own account.

PR #3603, opened 2019-04-15
Fix style for prep section in classfiltercsv() example

nickanderson:
This aligns the style with that used in peerleaders() which does render
correctly in docs.


cf-bottom:

PR #3604, opened 2019-04-16
[3.12.x] Make sure map->load is 0 at the end of HashMapClear()

vpodzime:
Otherwise the size is unchanged and the map thinks it still has
all the items that were there before. The items are gone, but
when new ones are inserted, the table sooner or later hits the
threshold and grows automatically. Repeat this process of "insert
a couple hundred items, clear, repeat" multiple times and the
table grows and grows...and grows.

Let's decrement map->load everytime we remove an item and assert
that it reaches 0 when all items are removed.

Ticket: CFE-3032
Changelog: Fix growing memory footprint of daemons
(cherry picked from commit 09e5d244793d80ec6e953ae6f57f045eda30ab0d)

### DIFF
--- a/libutils/hash_map.c
+++ b/libutils/hash_map.c
@@ -199,6 +199,7 @@ static void FreeBucketListItem(HashMap *map, BucketListItem *item)
     map->destroy_key_fn(item->value.key);
     map->destroy_value_fn(item->value.value);
     free(item);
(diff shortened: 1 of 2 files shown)


cf-bottom:

vpodzime:
Jenkins:

PR #3624, opened 2019-04-25
LGTM: Detect non-bool functions which return true or false

olehermanse:
This custom query looks for functions with return type other than
`int` and `bool`, and alerts if they return `true` or `false`.

The severity can be reduced later, I just want to test this.
It can also be expanded to catch int functions, but then
we will have a lot of alerts.


cf-bottom:

olehermanse:
@vpodzime I don't think these things can be tested before merge. Please approve and merge as there are no C changes in this PR.

vpodzime:
This is really nice!

PR #3625, opened 2019-04-26
LGTM: Use macro name (false) instead of value (0) in alert

olehermanse:


cf-bottom:

PR #3626, opened 2019-04-26
LGTM: Detect non-bool functions which return bool variable

olehermanse:
Since logical expressions and `true`/`false` macros are `int`
type, this query only detects functions which return a `bool`
variable (or a value typecast to `bool`).


cf-bottom:

PR #3627, opened 2019-04-26
See that Tom doesn't comment.

olehermanse:


olehermanse:
Confirmed:

```
Q: "...python3 -m tom ": [INFO] Looking at: See that Tom doesn't comment. (cfengine/core/pulls/3627)
Q: "...python3 -m tom ": [INFO] Someone already assigned a reviewer, I won't disturb
```

PR #3631, opened 2019-04-30
LGTM: Query for dereferencing a pointer argument without NULL-comparison

olehermanse:
This query has 6000 alerts currently, but I still think it will be valuable for new Pull Requests.

### DIFF
--- a/.lgtm/cpp-queries/missing-argument-null-check.c
+++ b/.lgtm/cpp-queries/missing-argument-null-check.c
@@ -1,0 +1,29 @@
+#include <assert.h>
+
+typedef struct {
+    char *string;
+} StructWithString;
+
+void bad_deref(StructWithString *data)
+{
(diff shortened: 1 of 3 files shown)


.lgtm/cpp-queries/missing-argument-null-check.c
```diff
@@ -0,0 +1,29 @@
...
+
+void bad_deref(StructWithString *data)
+{
+    char *string = data->string;
+}
+
+void good_with_assert(StructWithString *data)
+{
+    assert(data != NULL);
+    char *string = data->string;
```
vpodzime: I think we should also have `good_with_if` because in many cases that's the best thing to do.

olehermanse: Good point.

PR #3635, opened 2019-04-30
Fix ShellCommandReturnsZero() return value in error path

rikardfalkeborn:
Returning non-zero from a function returning bool is the same as
returning true. If there is an error in the child process, the function
should return false to indicate failure.

### DIFF
--- a/libpromises/unix.c
+++ b/libpromises/unix.c
@@ -244,7 +244,7 @@ bool ShellCommandReturnsZero(const char *command, ShellType shell)
         {
             if (errno != EINTR)
             {
-                return -1;
+                return false;
             }
         }
 


olehermanse:
@cf-bottom jenkins

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2413/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2413/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2413/

olehermanse:
Thanks for your contribution @rikardfalkeborn. Out of curiosity, how did you find this?

rikardfalkeborn:
Short  story: I used cppcheck. (I should probably have mentioned this in the commit message)

Longer story: As part of testing cppcheck, both latest release and latest master is run on Debian packages to catch regressions and find other issues. I occasionally contribute to cppcheck (I wrote the checker that flagged this) and went through the watnings it triggered to see if it found something uuseful and figured I might as well fix this one.

There are a couple more in https://github.com/cfengine/core/blob/89ddee88a13906fac657e89e7a2bf6fd13ccfb55/libpromises/verify_vars.c#L519 that I didn't fix because I had a bit of a brain fart when I read the code the first time., but those look wrong  too.

olehermanse:
@rikardfalkeborn Thanks for the explanation :rocket:

I added a rule to our code analysis CI (LGTM) here: 
https://github.com/cfengine/core/pull/3636

It found the same problem in 22 other places, so I will fix those as well.

olehermanse:
Thank you for your contribution @rikardfalkeborn

PR #3648, opened 2019-05-07
LGTM: Removed problematic & in qhelp

olehermanse:
https://lgtm.com/projects/g/cfengine/core/queries/logs/?pack=com.lgtm.github_apps%2Fcfengine%2Fcore%2Fcpp-queries

### DIFF
--- a/.lgtm/cpp-queries/bool-type-mismatch-return.qhelp
+++ b/.lgtm/cpp-queries/bool-type-mismatch-return.qhelp
@@ -30,7 +30,7 @@ A boolean macro - <code>true</code> or <code>false</code>
 A comparison - <code>==</code> or <code>!=</code> (or less than / greater than variants).
 </li>
 <li>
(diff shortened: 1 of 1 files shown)


.lgtm/cpp-queries/bool-type-mismatch-return.qhelp
```diff
@@ -30,7 +30,7 @@ A boolean macro - <code>true</code> or <code>false</code>
 A comparison - <code>==</code> or <code>!=</code> (or less than / greater than variants).
 </li>
 <li>
-A logical operation - <code>&&</code> or <code>||</code>
+A logical operation - AND or OR.
```
vpodzime: You could use `&amp;`.

olehermanse: I'm paranoid at this point, will add it back later.

PR #3660, opened 2019-05-08
Only use 'cmp_to_key' in determine_version if it exists

vpodzime:
Otherwise do things the old way using 'cmp=' because if
'cmp_to_key' doesn't exist, we are running an old version of
Python which supports 'cmp='.

This fixes running determine_version.py on CentOS 6 which has
Python 2.6.


olehermanse:
@vpodzime Feel free to merge and cherry-pick if/when appropriate.

PR #3666, opened 2019-05-13
augments file can handle cfengine class expressions

basvandervlies:
Changelog: Title
Ticket: CFE-2954

### DIFF
--- a/libpromises/generic_agent.c
+++ b/libpromises/generic_agent.c
@@ -197,7 +197,7 @@ Policy *SelectAndLoadPolicy(GenericAgentConfig *config, EvalContext *ctx, bool v
     return policy;
 }
 
-static bool CheckContextClassmatch(EvalContext *ctx, const char* class_str)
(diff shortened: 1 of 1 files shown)


cf-bottom:

nickanderson:
If I understand correctly, this patch looks to match [this comment](https://tracker.mender.io/browse/CFE-2971?focusedCommentId=94458&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-94458) in CFE-2971 which is the ticket opened to add functionality so that Class expressions can be used in augments 'classes' values.

I think we need some test cases excercising the regex only, class only, and mixed use behavior.

olehermanse:
@basvandervlies Sorry for taking over this one, but we are in the middle of building/testing/releasing 3.12.2. I want to get this merged ASAP, so I opened a new PR. See: https://github.com/cfengine/core/pull/3667

PR #3669, opened 2019-05-14
ifelse acceptance improvements

basvandervlies:
Changelog: title


cf-bottom:

basvandervlies:
```
Note: As a general rule function evaluation is skipped when undefined variables are used. However this function has special behavior when exactly three arguments are used, allowing it to be evaluated even if it contains undefined variables. For example:
```

will close it

PR #3672, opened 2019-05-14
CFE-2954: Augments files can now handle class expressions by appending '::' (3.12)

olehermanse:
A condition in an augments file is treated as a class expression
if it ends in `::`. Otherwise it is treated as a regular
expression.

Previously there was a bug, where the conditions in augments files
were treated as both class expressions and regular expressions.
This was ambiguous, and problematic since something like
`cfengine.nginx.enabled` as a regex would match a class called
`cfengine_nginx_enabled`, while as a class expression would match
`cfengine && nginx && enabled`.

This was fixed in 3.12.1 by treating it as a regex only.
(Documented and intended behavior).

Some users still wanted a way to use class expression, so the `::`
suffix solves the problem. For the example above, you can use this
in the augments file:

`cfengine.nginx.enabled::`

Since classes cannot end with `:`, this shouldn't be a problem,
it is not useful to have regular expressions ending with `::` in
this context.

Original commit made by @basvandervlies(bas.vandervlies@surfsara.nl).
Reworked by @olehermanse(ole@northern.tech).

### DIFF
--- a/libpromises/generic_agent.c
+++ b/libpromises/generic_agent.c
@@ -197,12 +197,34 @@ Policy *SelectAndLoadPolicy(GenericAgentConfig *config, EvalContext *ctx, bool v
     return policy;
 }
 
-static bool CheckContextClassmatch(EvalContext *ctx, const char* class_str)
+static bool CheckContextClassmatch(EvalContext *ctx, const char *class_str)
 {
+    if (StringEndsWith(class_str, "::")) // Treat as class expression, not regex
+    {
+        const size_t length = strlen(class_str);
+        if (length <= 2)
+        {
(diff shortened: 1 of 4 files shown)


olehermanse:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2510/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2510/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2510/

basvandervlies:
thanks

PR #3699, opened 2019-06-12
CONTRIBUTING.md: Improved instructions for commit messages

olehermanse:
Let's try to make our changelogs nicer.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,6 +34,7 @@ When submitting your pull request, please make sure you:
 
 * Address only one issue/feature.
   Smaller pull requests are generally better.
+  Usually a Pull Request will have only one commit.
 
 * Add tests.
   C functions should have unit tests.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -75,6 +90,133 @@ If the change is needed for a bigger feature, make that feature a separate pull
...
+Most new feature and fixes should have an entry in the ChangeLog.
+We have an automatic way of generating them before the release, by properly formatting *commit messages*.
+
+* To write arbitrary message in the ChangeLog:
+  `Changelog: <message>`
+* To use the commit title line in the ChangeLog:
+  `Changelog: Title`
+* To use the entire commit message in the ChangeLog:
+  `Changelog: Commit`
+
```
nickanderson: I also use Changelog: None to explicitly not have a changelog entry, and have that entry not come up inthe automatic changelog generator. Is that wrong/unnecessary?

craigcomstock: Yes, please clarify this. I was using this practice as well.

olehermanse: Added.

PR #3711, opened 2019-06-25
SetUmask function now correctly applies the given umask

olehermanse:
My mistake. Luckily, it was only ever used with 0077 anyway,
so no change in behavior.

### DIFF
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -466,8 +466,8 @@ Seq *ListDir(const char *dir, const char *extension)
 
 mode_t SetUmask(mode_t new_mask)
 {
-    const mode_t old_mask = umask(0077);
-    Log(LOG_LEVEL_DEBUG, "Set umask to 0077, was %o", old_mask);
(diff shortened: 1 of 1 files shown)


libutils/file_lib.c
```diff
@@ -466,8 +466,8 @@ Seq *ListDir(const char *dir, const char *extension)
 
 mode_t SetUmask(mode_t new_mask)
 {
-    const mode_t old_mask = umask(0077);
-    Log(LOG_LEVEL_DEBUG, "Set umask to 0077, was %o", old_mask);
```
oleorhagen: Why not `777` ?

olehermanse: Because it's a mask. `007` always wears a mask to be sneaky.

olehermanse:
Cherry-picks:
https://github.com/cfengine/core/pull/3757
https://github.com/cfengine/core/pull/3756
https://github.com/cfengine/core/pull/3758

PR #3714, opened 2019-06-25
Attributes structs are now initialized using a macro

olehermanse:
The normal way of initializing a struct to zeroes is:

```
Attributes a = { 0 };
```

* If the struct has only simple types (int, enum, pointers), this
  works well.
* If the struct has other structs inside (not pointers to structs)
  it still works, compilers seem to 0 out everything correctly.
  * It does however trigger some warnings when compiling with
    `-Wextra`.

These warnings are useful in other scenarios, like here:

```
return (FnCallResult) { FNCALL_FAILURE, { 0 } };
```

Ensuring that you specify correct number of initial values,
and use `{` for structs, and don't use `{` for integers or
pointers. A kind of type checking, if you will.

Thus, to satisfy this warning, I added a macro for the
`Attributes` struct. It should be noted that this struct is the
only one which needs the special treatment (a macro). Most
structs are much simpler than this one.

There are a couple of other ways to do this:

* Initialize using memset
  * Would also zero out the struct padding
  * A function call seems hacky/unnecessary
* Initialize using a global static const
  * Static variables are always zero initialized
  * Also seems unnecessary

Changelog: None
Ticket: None
Signed-off-by: Ole Herman Schumacher Elgesem <ole@northern.tech>

### DIFF
--- a/libpromises/attributes.c
+++ b/libpromises/attributes.c
@@ -44,7 +44,7 @@ void ClearFilesAttributes(Attributes *whom)
 
 Attributes GetFilesAttributes(const EvalContext *ctx, const Promise *pp)
 {
-    Attributes attr = { {0} };
+    Attributes attr = ZeroAttributes;
 
 // default for file copy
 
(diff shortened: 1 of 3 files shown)


olehermanse:

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2720/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2720/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2720/

PR #3719, opened 2019-06-28
Added make list to list all make targets

nickanderson:


cf-bottom:

craigcomstock:
This might be interesting in case it catches some cases yours does not: https://github.com/scop/bash-completion/blob/master/completions/make

olehermanse:
@nickanderson Do you have an example of when/why this is useful?

nickanderson:
I couldn't remember the name of the target I was looking for (it was make check). I thought having a target to list the targets was useful. If you don't like then we can just reject.

olehermanse:
> I couldn't remember the name of the target I was looking for (it was make check). I thought having a target to list the targets was useful. If you don't like then we can just reject.

It's good, I just like to see some reasoning/motivation for making changes.

PR #3725, opened 2019-07-03
CFE-3091: Added debug_abort_if_reached macro

olehermanse:


libutils/condition_macros.h
```diff
@@ -3,10 +3,19 @@
 
 #include <assert.h>
 
+// This function contains macros to use for error checking, assertions,
```
vpodzime: *This file...*

libutils/condition_macros.h
```diff
@@ -17,4 +26,13 @@
 }
 #endif
 
+// Similar to assert_or_return, except you put it inside the if
+// body which handles the error in release builds
+#ifndef debug_abort_if_reached
+#define debug_abort_if_reached() { \
+        assert(false);             \
+    }                              \
+}
```
vpodzime: Extra `}` here?

olehermanse: Fixed, thanks!

libutils/condition_macros.h
olehermanse: Thanks, fixed!

libutils/condition_macros.h
```diff
@@ -17,4 +26,12 @@
 }
 #endif
 
+// Similar to assert_or_return, except you put it inside the if
+// body which handles the error in release builds
+#ifndef debug_abort_if_reached
+#define debug_abort_if_reached() { \
+    assert(false);                 \
```
karlhto: I guess the only problem here is that you aren't told exactly what condition is wrong when it fails. Adding a parameter to the macro would kind of remove the point of having it in the first place, however.

olehermanse: It should be placed inside an if, and you get a line number. It should be pretty clear.

olehermanse: I hope you get a line number. Good point, when I statrt using it I can do some testing and improve it if necessary.

olehermanse:
Cherry-picked in https://github.com/cfengine/core/pull/3825

PR #3727, opened 2019-07-08
Changed some functions to be bool-correct

karlhto:
Also removed some FIXME's from the code regarding return values.

Decided to work down the LGTM list.

### DIFF
--- a/cf-agent/files_editline.c
+++ b/cf-agent/files_editline.c
@@ -89,7 +89,7 @@ static int ReplacePatterns(EvalContext *ctx, Item *start, Item *end, const Attri
 static bool EditColumns(EvalContext *ctx, Item *file_start, Item *file_end, const Attributes *a, const Promise *pp, EditContext *edcontext, PromiseResult *result);
 static bool EditLineByColumn(EvalContext *ctx, Rlist **columns, const Attributes *a, const Promise *pp, EditContext *edcontext, PromiseResult *result);
 static bool DoEditColumn(Rlist **columns, const Attributes *a, EditContext *edcontext);
-static int SanityCheckInsertions(const Attributes *a);
+static bool SanityCheckInsertions(const Attributes *a);
 static bool SanityCheckDeletions(const Attributes *a, const Promise *pp);
 static bool SelectLine(EvalContext *ctx, const char *line, const Attributes *a);
 static bool NotAnchored(char *s);
@@ -1322,11 +1322,11 @@ static bool EditColumns(EvalContext *ctx, Item *file_start, Item *file_end, cons
 
 /***************************************************************************/
 
-static int SanityCheckInsertions(const Attributes *a)
+static bool SanityCheckInsertions(const Attributes *a)
 {
     long not = 0;
     long with = 0;
-    long ok = true;
+    bool ok = true;
     Rlist *rp;
     InsertMatchType opt;
     int exact = false, ignore_something = false;
--- a/cf-agent/verify_databases.c
+++ b/cf-agent/verify_databases.c
@@ -989,7 +989,6 @@ static void DeleteSQLColumns(char **name_table, char **type_table, int *size_tab
 
 static bool CheckSQLDataType(char *type, char *ref_type, const Promise *pp)
 {
-    // FIXME: This function always returns true
     static char *const aliases[3][2] =
         {
             {"varchar", "character@varying"},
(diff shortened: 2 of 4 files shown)


cf-bottom:

cf-agent/files_editline.c
```diff
@@ -1417,7 +1417,7 @@ static bool SanityCheckDeletions(const Attributes *a, const Promise *pp)
         {
             Log(LOG_LEVEL_ERR,
                 "Makes no sense to promise multi-line delete with not_matching. Cannot be satisfied for all lines as a block.");
-            // FIXME: This function always returns true (!)
+            return false;
```
vpodzime: Are you sure this doesn't change any behavior?

karlhto: The return value of the function is never evaluated, so it doesn't. It might be better to just set it to `void`, but shouldn't the "SanityCheck" return failure if it also gives an error? Who knows.

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2331/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2331/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2331/

karlhto:
@cf-bottom Jenkins again

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2338/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2338/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2338/

cf-agent/files_editline.c
olehermanse: @karlhto That's not true:

https://github.com/cfengine/core/blob/6403c3eb9784db75f87abcf99693e144c1d1a9b8/cf-agent/files_editline.c#L360

cf-agent/verify_databases.c
```diff
@@ -987,9 +987,8 @@ static void DeleteSQLColumns(char **name_table, char **type_table, int *size_tab
 
 /*****************************************************************************/
 
-static bool CheckSQLDataType(char *type, char *ref_type, const Promise *pp)
+static void CheckSQLDataType(char *type, char *ref_type, const Promise *pp)
 {
-    // FIXME: This function always returns true
```
olehermanse: @karlhto The reason why I added this FIXME was because I didn't want to do the change you're doing here (`bool` -> `void`). As you point out in the comments above, a `Check` function should return `true` or `false`.

As it is now, this function does nothing, except for some verbose logging and maybe some segfaulting if we're lucky. I think it's probably more appropriate to return false in the cases where the type mismatch was logged.

cf-agent/verify_packages.c
```diff
@@ -3557,7 +3556,7 @@ static bool PrependMultiLinePackageItem(
         }
     }
 
-    return false;
+    return true;
```
olehermanse: This one is never evaluated, but it seems correct to change it to `true`.

cf-agent/files_editline.c
olehermanse: I agree with the change though, but it will/might change behavior. So I would probably do it in a separate PR, and try to do some extra (manual) testing.

cf-agent/verify_databases.c
olehermanse: Please change this function to have a `bool ret` variable, returning false if one of the verbose logs were printed, and true if not.

olehermanse: (`CheckSQLDataType` return value is never checked, so there is no behavior change here either way.)

cf-agent/files_editline.c
karlhto: Whoops, think I thought about its "Insertions" variant instead. Thanks for noticing. :^)

olehermanse:
@cf-bottom jenkins

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2446/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2446/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2446/

olehermanse:
Green in jenkins!

PR #3729, opened 2019-07-08
CFE-3094: Package modules now hit network when package cache is first initialized (master)

nickanderson:


cf-bottom:

cf-agent/package_module.c
```diff
@@ -1590,6 +1590,10 @@ bool UpdateSinglePackageModuleCache(EvalContext *ctx,
             Log(LOG_LEVEL_VERBOSE,
                 "Forcing package list update due to missing database");
             force_update = true;
+            if (type == UPDATE_TYPE_LOCAL_UPDATES)
```
craigcomstock: looking at the yum module I see that list-updates (UPDATE_TYPE_UPDATES) would go online and list-updates-local (UPDATE_TYPE_LOCAL_UPDATES) should use local cache and not go online. So here you are forcing going online in the case that the local database is missing. Seems reasonable to me!

nickanderson: Yeah, that's exactly what I was trying to do here. Not sure if it's the right thing to do or not.

It's possible that the OS has a local cache of updates and that we could avoid hitting the network, but in the case the OS does NOT have a valid local cache, listing local may just fail. I am not enough of a wiz to see how to try local listing first, then fall back to network listing like the patch to the yum module.

https://github.com/cfengine/masterfiles/pull/1463

vpodzime: This looks OK to me. You don't need the `if`, I think, let's just do `type = UPDATE_TYPE_UPDATES` with a comment why we always want to do this in this special case.

nickanderson: I started with setting type unconditional, but I noticed that `UpdateSinglePackageModuleCache` gets called with type `UPDATE_TYPE_INSTALLED` (to get list of installed packages), `UPDATE_TYPE_UPDATES` (to hit the network for updates), and `UPDATE_TYPE_LOCAL_UPDATES` for listing package updates without hitting the network based on existing OS cache. So, if I set type unconditionally, then i would override `UPDATE_TYPE_INSTALLED` and it wouldn't update the list of local packages installed.


https://github.com/nickanderson/core/blob/CFE-3094/master/cf-agent/package_module.c#L202-L208

PR #3741, opened 2019-07-12
CFE-3099/3.10.x The @ character is now allowed in the key of classic arrays defined by the module protocol

nickanderson:


cf-bottom:

PR #3744, opened 2019-07-15
CFE-3104/master Improved hostrnage() function example

nickanderson:
Merge Together:
https://github.com/cfengine/documentation/pull/2178


nickanderson:
@vpodzime I am confident that it will since it does not have an `example_output` block, it should not be executed by any of the tests. Instead, I included a `static_example_output` block and in the docs PR I include the example code and example output block explicitly.

PR #3751, opened 2019-07-23
CFE-3082: cf-check code cleanup, better errors, man page and other options

olehermanse:
cf-check now supports the same options as the other CFEngine
binaries; log-level, verbose, debug, manpage, version, help.

There were some error messages before, but this commit ensures that
all commands (diagnose, repair, backup) both print errors and exit
with non-zero exit code, when no db files are found.

Also clarified log messages and code comments to be more clear.
The DB check in cf-agent and cf-execd does not fail if there are
no files.

### DIFF
--- a/cf-check/backup.c
+++ b/cf-check/backup.c
@@ -8,7 +8,7 @@
 
 #if defined(__MINGW32__) || !defined(LMDB)
 
-int backup_main(ARG_UNUSED int argc, ARG_UNUSED char **argv)
+int backup_main(ARG_UNUSED int argc, ARG_UNUSED const char *const *const argv)
 {
     Log(LOG_LEVEL_ERR,
         "cf-check backup not available on this platform/build");
@@ -103,7 +103,7 @@ int backup_files(Seq *filenames)
     return 0;
 }
 
-int backup_main(int argc, char **argv)
+int backup_main(int argc, const char *const *const argv)
 {
     Seq *files = argv_to_lmdb_files(argc, argv);
     const int ret = backup_files(files);
--- a/cf-check/backup.c
+++ b/cf-check/backup.c
@@ -106,6 +106,11 @@ int backup_files(Seq *filenames)
 int backup_main(int argc, const char *const *const argv)
 {
     Seq *files = argv_to_lmdb_files(argc, argv);
+    if (files == NULL || SeqLength(files) == 0)
+    {
+        Log(LOG_LEVEL_ERR, "No database files to back up");
+        return -1;
+    }
     const int ret = backup_files(files);
     SeqDestroy(files);
     return ret;
--- a/cf-check/backup.c
+++ b/cf-check/backup.c
@@ -8,14 +8,14 @@
 
 #if defined(__MINGW32__) || !defined(LMDB)
 
-int backup_main(int argc, char **argv)
+int backup_main(ARG_UNUSED int argc, ARG_UNUSED char **argv)
 {
     Log(LOG_LEVEL_ERR,
         "cf-check backup not available on this platform/build");
     return 1;
 }
 
-int backup_files(Seq *filenames)
+int backup_files(ARG_UNUSED Seq *filenames)
 {
     Log(LOG_LEVEL_INFO,
         "database backup not available on this platform/build");
--- a/cf-check/backup.h
+++ b/cf-check/backup.h
@@ -4,6 +4,6 @@
 #include <sequence.h>
 
 int backup_files(Seq *filenames);
-int backup_main(int argc, char **argv);
+int backup_main(int argc, const char *const *argv);
 
 #endif
--- a/cf-check/cf-check.c
+++ b/cf-check/cf-check.c
@@ -36,7 +36,7 @@ static void print_help()
         "\n");
 }
 
-int main(int argc, char **argv)
+int main(int argc, const char *const *const argv)
 {
     if (StringEndsWith(argv[0], "lmdump"))
     {
@@ -57,8 +57,8 @@ int main(int argc, char **argv)
     }
 
     const int cmd_argc = argc - 1;
-    char **cmd_argv = argv + 1;
-    char *command = cmd_argv[0];
+    const char *const *const cmd_argv = argv + 1;
+    const char *const command = cmd_argv[0];
 
     if (StringSafeEqual_IgnoreCase(command, "lmdump") ||
         StringSafeEqual_IgnoreCase(command, "dump"))
--- a/cf-check/cf-check.c
+++ b/cf-check/cf-check.c
@@ -4,6 +4,7 @@
 #include <repair.h>
 #include <string_lib.h>
 #include <logging.h>
+#include <man.h>
 
 static void print_version()
 {
@@ -36,7 +37,54 @@ static void print_help()
         "\n");
 }
 
-int main(int argc, const char *const *const argv)
+static const char *const CF_CHECK_SHORT_DESCRIPTION =
+    "Utility for diagnosis and repair of local CFEngine databases.";
+
+static const char *const CF_CHECK_MANPAGE_LONG_DESCRIPTION =
+    "cf-check does not evaluate policy or rely on the integrity of the\n"
+    "databases. It is intended to be able to detect and repair a corrupt\n"
+    "database.";
+
+// static const Description COMMANDS[] =
+// {
+//     {"help",     "Prints general help or per topic",
+//                  "cf-check help [command]"},
+//     {"diagnose", "Assess the health of one or more database files",
+//                  "cf-check diagnose"},
+//     {"backup",   "Copy database files to a timestamped folder",
+//                  "cf-check backup"},
+//     {"repair",   "Diagnose, then backup and delete any corrupt databases",
+//                  "cf-check repair"},
+//     {"dump",     "Print the contents of a database file",
+//                  "cf-check dump -a " WORKDIR "/state/cf_lastseen.lmdb"},
+//     {NULL, NULL, NULL}
+// };
+
+static const struct option OPTIONS[] =
+{
+    {"help",        optional_argument,  0, 'h'},
+    {"manpage",     no_argument,        0, 'M'},
+    {"version",     no_argument,        0, 'V'},
+    {"debug",       no_argument,        0, 'd'},
+    {"verbose",     no_argument,        0, 'v'},
+    {"log-level",   required_argument,  0, 'g'},
+    {"inform",      no_argument,        0, 'I'},
+    {NULL,          0,                  0, '\0'}
+};
+
+static const char *const HINTS[] =
+{
+    "Print the help message",
+    "Print the man page",
+    "Output the version of the software",
+    "Enable debugging output",
+    "Enable verbose output",
+    "Specify how detailed logs should be. Possible values: 'error', 'warning', 'notice', 'info', 'verbose', 'debug'",
+    "Enable basic information output",
+    NULL
+};
+
+int main(int argc, const char *const *argv)
 {
     if (StringEndsWith(argv[0], "lmdump"))
     {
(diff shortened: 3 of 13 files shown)


olehermanse:
**Example output:**

```
root@dev core $ cf-agent --skip-db-check=no --log-level info empty.cf
    info: Running internal DB (LMDB) consistency checks
    info: Status of '/var/cfengine/state/cf_state.lmdb': OK
    info: Status of '/var/cfengine/state/cf_lock.lmdb': OK
    info: All 2 databases healthy
    info: No corruption - nothing to do
root@dev core $ rm -f /v
vagrant/     var/         vmlinuz      vmlinuz.old
root@dev core $ rm -f /var/cfengine/state/*.lmdb
root@dev core $ cf-agent --skip-db-check=no --log-level info empty.cf
    info: Running internal DB (LMDB) consistency checks
    info: Skipping local database repair, no lmdb files
root@dev core $ cf-check repair
    info: No filenames specified, defaulting to .lmdb files in /var/cfengine/state
    info: Status of '/var/cfengine/state/cf_state.lmdb': OK
    info: Status of '/var/cfengine/state/cf_lock.lmdb': OK
    info: All 2 databases healthy
    info: No corruption - nothing to do
root@dev core $ rm -f /var/cfengine/state/*.lmdb
root@dev core $ cf-check repair
    info: No filenames specified, defaulting to .lmdb files in /var/cfengine/state
   error: No database files to repair
root@dev core $ echo $?
255
root@dev core $ cf-check repair
    info: No filenames specified, defaulting to .lmdb files in /var/cfengine/state
   error: No database files to repair
root@dev core $ echo $?
255
root@dev core $
```

olehermanse:
**Man page:**

```
CF-CHECK(8)                  System Administration                 CF-CHECK(8)

NAME
       cf-check  -  Utility  for  diagnosis and repair of local CFEngine data-
       bases.

SYNOPSIS
       cf-check [OPTION]...  [FILE]

DESCRIPTION
       cf-check does not evaluate policy or rely on the integrity of the data-
       bases.  It  is intended to be able to detect and repair a corrupt data-
       base.

OPTIONS
       --help, -h value
              Print the help message

       --manpage, -M
              Print the man page

       --version, -V
              Output the version of the software

       --debug, -d
              Enable debugging output

       --verbose, -v
              Enable verbose output

       --log-level, -g value
              Specify how detailed logs should be. Possible  values:  'error',
              'warning', 'notice', 'info', 'verbose', 'debug'

       --inform, -I
              Enable basic information output

CFENGINE
       CFEngine  provides  automated  configuration  management of large-scale
       computer systems. A system administrator describes the desired state of
       a  system using CFEngine policy code. The program cf-agent reads policy
       code and attempts to bring the current  system  state  to  the  desired
       state  described.  Policy  code  is  downloaded  by cf-agent from a cf-
       serverd daemon. The daemon cf-execd is responsible for running cf-agent
       periodically.
       Documentation for CFEngine is available at http://cfengine.com/documen-
       tation/.

PROMISE THEORY
       CFEngine is built on principles from promise theory, proposed  by  Mark
       Burgess  in  2004.  Promise  theory is a model of voluntary cooperation
       between individual, autonomous  actors  or  agents  who  publish  their
       intentions  to one another in the form of promises. A promise is a dec-
       laration of intent whose purpose is to increase  the  recipient's  cer-
       tainty  about  a  claim  of  past,  present  or future behaviour. For a
       promise to  increase  certainty,  the  recipient  needs  to  trust  the
       promiser, but trust can also be built on the verification that previous
       promises have been kept, thus trust plays a symbiotic relationship with
       promises.
       For     an    introduction    to    promise    theory,    please    see
       http://arxiv.org/abs/0810.3294/

AVAILABILITY
       cf-check is part of CFEngine.
       Binary packages may be downloaded from  http://cfengine.com/downloads/.
       The source code is available at http://github.com/cfengine/

BUGS
       Please see the public bug-tracker at http://bug.cfengine.com/.
       GitHub        pull-requests       may       be       submitted       to
       http://github.com/cfengine/core/.

SEE ALSO
       cf-promises(8),  cf-agent(8),  cf-serverd(8),   cf-execd(8),   cf-moni-
       tord(8), cf-runagent(8), cf-key(8)

AUTHOR
       Mark Burgess and Northern.tech AS

CFEngine                          2019-07-23                       CF-CHECK(8)

```

cf-check/cf-check.c
```diff
@@ -36,7 +37,54 @@ static void print_help()
...
+//                  "cf-check diagnose"},
+//     {"backup",   "Copy database files to a timestamped folder",
+//                  "cf-check backup"},
+//     {"repair",   "Diagnose, then backup and delete any corrupt databases",
+//                  "cf-check repair"},
+//     {"dump",     "Print the contents of a database file",
+//                  "cf-check dump -a " WORKDIR "/state/cf_lastseen.lmdb"},
+//     {NULL, NULL, NULL}
+// };
+
```
craigcomstock: Why the commented out COMMANDS[]?

olehermanse: It's currently unused, but ready to be used in an improved help menu or man page.

cf-check/backup.c
```diff
@@ -8,14 +8,14 @@
 
 #if defined(__MINGW32__) || !defined(LMDB)
 
-int backup_main(int argc, char **argv)
+int backup_main(ARG_UNUSED int argc, ARG_UNUSED const char *const *const argv)
```
karlhto: I think it would be clearer to follow the language specifications in regards to modifiers here (modifier _after_ type), even though we are pretty inconsistent regarding `const` modifiers in other parts of the codebase.

I would prefer `ARG_UNUSED char const *const *const argv`, though it's probably not a big deal.

cf-check/utilities.h
```diff
@@ -6,8 +6,8 @@
 // These functions should be moved to libutils once cf-check is
 // implemented and backported.
 
-Seq *argv_to_seq(int argc, char **argv);
+Seq *argv_to_seq(int argc, const char *const *argv);
```
karlhto: Maybe we should at some point move this function to `libutils`? Seems useful.

cf-check/backup.c
olehermanse: @karlhto What do you mean by "following the language specifications"? [The ISO C11 standard](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf) calls these qualifiers, and does not specify whether you should put them before or after the first _type-specifier_ in a _type-name_, both are allowed by the syntax definition.

However, they do consistently use my preferred style, placing the const before the first _type-specifier_, in all the examples:

```C
const char *c_cp;
const void *c_vp;
extern int atoi(const char *);
```

```C
typedef int A[2][3];
const A a = {{4, 5, 6}, {7, 8, 9}}; // array of array of const int
int *pi;
const int *pci;
```

I would argue that `const char *const *` is the most readable way. The first `const` is unambiguous, intuitively, since there is nothing to the left it must apply to the right. You don't have to know the "const rule of thumb" to understand this. The second `const` is put together with the pointer asterisk (no space between) to indicate that it applies to this pointer. Most examples in the C standard follow this:

```C
const int *ptr_to_constant;
int *const constant_ptr;
```

(But there are also a couple of examples that use `* const`).

[glibc also seems to mostly follow the same convention.](https://github.com/bminor/glibc/blob/master/posix/execv.c#L23)

cf-check/utilities.h
olehermanse: Yep. TODO.

cf-check/backup.c
karlhto: Well, I might remember incorrectly, but I was pretty sure at least ANSI C required qualifiers to come after the type. IMO it might be slightly less ambiguous, because then you don't have the `const` qualifiers behaving differently for pointers and types, but keeping it consistent is a better idea anyway.

> I would argue that const char *const * is the most readable way. The first const is unambiguous, intuitively, since there is nothing to the left it must apply to the right. You don't have to know the "const rule of thumb" to understand this.

I agree with this. :)

cf-check/repair.c
```diff
@@ -127,11 +128,24 @@ int repair_main(int argc, const char *const *const argv)
 
 int repair_default()
 {
+    // This function is used by cf-execd and cf-agent, not cf-execd
```
vpodzime: I think this line needs a consistency check.

cf-check/cf-check.c
```diff
@@ -53,12 +101,71 @@ int main(int argc, const char *const *const argv)
...
+                break;
+            }
+            default:
+            {
+                return EXIT_FAILURE;
+                break;
+            }
+        }
+    }
+    const char *const *const cmd_argv = &(argv[optind]);
```
vpodzime: I think this is worth adding as a comment:
> When ‘getopt_long’ has no more options to handle, it returns ‘-1’,
> and leaves in the variable ‘optind’ the index in ARGV of the next
> remaining argument.

vpodzime: Also, I think `argv + optind` would be better readable than the `&(a[i])` expression.

cf-check/backup.c
olehermanse: What I wrote above applies for both [ANSI C](https://www.pdf-archive.com/2014/10/02/ansi-iso-9899-1990-1/ansi-iso-9899-1990-1.pdf), as well as the K&R book. They don't say whether to put it before or after, the syntax definitions allow both, and they put it first in examples. Maybe you read it somewhere else (course material, blog post, etc.), but it doesn't seem to be part of any of the standards.

cf-check/repair.c
olehermanse: Haha, good catch. 

> Compilers don't read comments. Programmers don't read comments.

cf-check/cf-check.c
olehermanse: Alright.

cf-check/backup.c
karlhto: Alright. Thanks for clarifying. :)

cf-check/backup.h
```diff
@@ -4,6 +4,6 @@
 #include <sequence.h>
 
 int backup_files(Seq *filenames);
-int backup_main(int argc, char **argv);
+int backup_main(int argc, const char *const *argv);
```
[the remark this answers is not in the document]
olehermanse: @craigcomstock Generally, In C, the declaration(function signature usually in a .h file) and definition (implementation usually in a .c file) can have differences in parameter names and types, as long as the types are "compatible enough". This can be useful in some circumstances, and confusing in others.

**The declaration** gives you the information you need to know for calling the function. It gives the compiler adequate information to set up the arguments in registers / stack correctly without knowing how the function is implemented.

**The definition** is used for actually compiling the contents of the function, and enforcing the const restrictions withing the function scope, according to Cs rules.

So, consider this simple example:

```C
int double_it(int num);

int main(void)
{
  return double_it(3);
}

int double_it(const int n)
{
  // n = 2; // Nope, const..
  return 2 * n;
}
```

* The `double_it()` declaration here tells you that the function takes 1 integer (input) and returns 1 integer (output)
* You now know how to use the function within `main()`
* The `double_it()` definition has an added `const`, telling the compiler that the argument/local variable will not change within this function.

That's essentially it. The last asterisk in the pointer type refers to the local value, and I'm just declaring that it won't be assigned to in this function. It has 0 impact on the caller.

You can add it to the declaration, it makes no difference, but I prefer not to, because:

* If you just see the declaration, and you don't know about this stuff, it might be confusing why a function takes a const int (or const pointer).
* If somebody comes along later and removes it from the definition, it can still be in the declaration, without warnings. That makes it even more confusing.
* (I thought I've seen a warning for this once, essentially saying `this const does nothing`, but I can't remember where, maybe clang scan-build or something.)

The counterargument is of course that it can be confusing that the declaration and definition doesn't match, as evident by your question. So... tomato/potato.

**Note:** the param is called `n`/`num`. I don't recommend this, but I guess there are some _very rare_ situations where it's appropriate to have long, descriptive, names in the declaration and shorter more technical names in the definition. Maybe.

**Note 2:** Also, the function could have multiple definitions (inside `#ifdef`s). Some of them might want const, and other not. Different parameter names in the different definitions might also make sense.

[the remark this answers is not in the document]
craigcomstock: Perfect. Thank you for taking the time to explain so carefully. Very much appreciated. :1st_place_medal:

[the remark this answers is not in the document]
Lex-2008: Agreed. Few interesting points learned here!

PR #3752, opened 2019-07-23
CFE-3094: Package modules now hit network when package cache is first initialized (3.12.x)

nickanderson:


cf-bottom:

PR #3755, opened 2019-07-24
CFE-3089: New CFNet function OPENDIR

karlhto:
Creates a directory listing of a path on a remote host.

Ticket: CFE-3089
Changelog: None

### DIFF
--- a/cf-net/cf-net.c
+++ b/cf-net/cf-net.c
@@ -23,7 +23,7 @@
 */
 
 
-#define CF_NET_VERSION "0.1.1"
+#define CF_NET_VERSION "0.1.2"
 
 #include <platform.h>
 #include <libgen.h>
@@ -42,6 +42,8 @@
 #include <cf-windows-functions.h> // TODO: move this out of libpromises
 #include <known_dirs.h>           // TODO: move this 'out of libpromises
 #include <cleanup.h>
+#include <protocol.h>
+#include <sequence.h>
 
 #define ARG_UNUSED __attribute__((unused))
 
@@ -204,7 +206,9 @@ int main(int argc, char **argv)
     ret = CFNetRun(&opts, args, hostnames);  // Commands return exit code
     free(hostnames);
     CFNetOptionsClear(&opts);
-    return ret;
+
+    ret = (ret == 0) ? EXIT_SUCCESS : EXIT_FAILURE;
+    DoCleanupAndExit(ret);
 }
 
 //*******************************************************************
@@ -874,17 +878,12 @@ static int CFNetGet(ARG_UNUSED CFNetOptions *opts, const char *hostname, char **
     return failure ? -1 : 0;
 }
 
-static void PrintDirs(const Item *list)
+static void PrintDirs(const Seq *list)
 {
-    const Item *ip = list;
-
-    while (ip != NULL)
+    for (size_t i = 0; i < SeqLength(list); i++)
     {
-        // TODO: I hate this:
-        struct dirent *de = (struct dirent *) ip->name;
-        printf("%s\n", de->d_name);
-
-        ip = ip->next;
+        char *dir_entry = SeqAt(list, i);
+        printf("%s\n", dir_entry);
     }
 }
 
@@ -912,8 +911,14 @@ static int CFNetOpenDir(ARG_UNUSED CFNetOptions *opts, const char *hostname, cha
 
     const char *remote_path = args[1];
 
-    Item *items = RemoteDirList(remote_path, false, conn);
-    PrintDirs(items);
+    Seq *seq = ProtocolOpenDir(conn, remote_path);
+    if (seq == NULL)
+    {
+        return -1;
+    }
+
+    PrintDirs(seq);
+    SeqDestroy(seq);
     CFNetDisconnect(conn);
     return 0;
 }
--- a/libcfnet/protocol.c
+++ b/libcfnet/protocol.c
@@ -1,0 +1,90 @@
+/*
+   Copyright 2019 Northern.tech AS
+
+   This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+   This program is free software; you can redistribute it and/or modify it
+   under the terms of the GNU General Public License as published by the
+   Free Software Foundation; version 3.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#include <protocol.h>
+
+#include <client_code.h>
+#include <client_protocol.h>
+#include <definitions.h>
+#include <net.h>
+#include <string_lib.h>
+#include <tls_generic.h>
+
+Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
+{
+    assert(conn != NULL);
+    assert(path != NULL);
+
+    char buf[CF_MSGSIZE] = {0};
+    int tosend = snprintf(buf, CF_MSGSIZE, "OPENDIR %s", path);
+    if (tosend < 0 || tosend >= CF_MSGSIZE)
+    {
+        return NULL;
+    }
+
+    int ret = SendTransaction(conn->conn_info, buf, tosend, CF_DONE);
+    if (ret == -1)
+    {
+        return NULL;
+    }
+
+    Seq *seq = SeqNew(0, free);
+
+    int more = 1;
+    while (more != 0)
+    {
+        int len = ReceiveTransaction(conn->conn_info, buf, &more);
+        if (len == -1)
+        {
+            break;
+        }
+
+        if (BadProtoReply(buf))
+        {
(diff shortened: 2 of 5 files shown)


cf-bottom:

karlhto:
Note that the file and functions names are subject to change. I'd like input on how they should look. :^)

libcfnet/protocol.c
```diff
@@ -0,0 +1,83 @@
...
+
+    if (SendTransaction(conn->conn_info, buf, tosend, CF_DONE) == -1)
+    {
+        return NULL;
+    }
+
+    Seq *seq = SeqNew(0, free);
+
+    bool more = true;
+    while (more == true)
```
vpodzime: `while (more)`

libcfnet/protocol.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <cfnet.h>
+#include <sequence.h>
+
+/**
+ * Creates a sequence from a directory listing from the server.
+ *
+ * @param [in] conn  Connection to use.
+ * @param [in] path  Path to retrieve directory listing of
+ * @return A list
```
vpodzime: *A list* is not really too descriptive. :smile:

karlhto: Heh, I should have added WIP to this. I still need to add it in `cf-net` as well.

libcfnet/protocol.c
```diff
@@ -0,0 +1,88 @@
...
+
+        if (BadProtoReply(buf))
+        {
+            Log(LOG_LEVEL_INFO, "%s", buf + strlen("BAD: "));
+            break;
+        }
+
+        // FIXME Placeholder for the time being. There is probably a cleaner
+        //       and safer way to do this
+        for (int i = 0; buf[i] != '\0'; i += strlen(buf + i) + 1)
```
olehermanse: Please explain the logic here, what does the response look like? Also, does ReceiveTransaction guarantee that buf is `\0` terminated, even for untrusted / bad input? Or do you need to add checks for `CF_BUFSIZE` overflow?

libcfnet/protocol.c
```diff
@@ -0,0 +1,88 @@
...
+        {
+            if (strcmp(buf + i, CFD_TERMINATOR) == 0)
```
olehermanse: `StringSafeEqual()`.

libcfnet/protocol.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <cfnet.h>
+#include <sequence.h>
+
+/**
+ * Creates a sequence from a directory listing from the server.
+ *
+ * @param [in] conn  Connection to use.
+ * @param [in] path  Path to retrieve directory listing of
+ * @return A sequence containing the elements of the directory
```
olehermanse: What does elements mean? You should mention that they are strings.

cf-net/cf-net.c
```diff
@@ -873,17 +875,13 @@ static int CFNetGet(ARG_UNUSED CFNetOptions *opts, const char *hostname, char **
...
+static void PrintDirs(const Seq *list)
 {
-    const Item *ip = list;
-
-    while (ip != NULL)
+    for (size_t i = 0; i < SeqLength(list); i++)
     {
         // TODO: I hate this:
-        struct dirent *de = (struct dirent *) ip->name;
+        struct dirent *de = (struct dirent *)SeqAt(list, i);
```
olehermanse: They are strings, not dirent (!)

libcfnet/protocol.c
```diff
@@ -0,0 +1,88 @@
...
+    {
+        return NULL;
+    }
+
+    Seq *seq = SeqNew(0, free);
+
+    bool more = true;
+    while (more)
+    {
+        ret = ReceiveTransaction(conn->conn_info, buf, (int *)&more);
```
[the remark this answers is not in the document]
karlhto: I agree, and I was thinking of changing `ReceiveTransaction` to take bool instead.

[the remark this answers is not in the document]
olehermanse: @karlhto Actually, bool is (usually) 1 byte, so, this is dangerous. You might be overwriting bytes outside of the more variable in this case. So please make more an integer here, until we can change ReceiveTransaction.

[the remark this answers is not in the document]
karlhto: Welp, that's a good point.

karlhto:
@olehermanse Think this should be ready for review. It might be a good idea to add tests of some kind, though should probably be something more advanced than unit tests, right?

karlhto:
@cf-bottom Jenkins

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2726/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2726/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2726/

libcfnet/protocol.h
```diff
@@ -0,0 +1,66 @@
...
+ *         printf("%s\n", entry);
+ *     }
+ * @endcode
+ *
+ * In the protocol, this will look like this on the server side:
+ *     Received: OPENDIR masterfiles
+ *     Translated to: OPENDIR /var/cfengine/masterfiles
+ *     Sends string:
+ *     ".\0..\0cfe_internal\0cf_promises_release_id\0...
+ *      ...templates\0update.cf\0" CFD_TERMINATOR
```
olehermanse: Very nice explanation!

cf-net/cf-net.c
```diff
@@ -874,17 +876,12 @@ static int CFNetGet(ARG_UNUSED CFNetOptions *opts, const char *hostname, char **
...
 
-static void PrintDirs(const Item *list)
+static void PrintDirs(const Seq *list)
 {
-    const Item *ip = list;
-
-    while (ip != NULL)
+    for (size_t i = 0; i < SeqLength(list); i++)
     {
-        // TODO: I hate this:
```
vpodzime: Someone should be happy to see this go away... :)

cf-net/cf-net.c
```diff
@@ -912,8 +909,14 @@ static int CFNetOpenDir(ARG_UNUSED CFNetOptions *opts, const char *hostname, cha
 
     const char *remote_path = args[1];
 
-    Item *items = RemoteDirList(remote_path, false, conn);
-    PrintDirs(items);
+    Seq *seq = CfNetOpenDir(conn, remote_path);
+    if (seq == NULL)
+    {
+        return -1;
```
vpodzime: Is this value propagated as the exit code of the whole process? If yes, then it should not be `-1`, but rather `1` or something similar.

karlhto: `-1` is used throughout the rest of `cf-net` as well, though. I think a better change would be to check the result of functions and using correct exit code based on that. Seems arbitrary for a function to return `EXIT_FAILURE` for instance.

vpodzime: Cannot agree more. Just please do the change in a follow-up commit or create a task so that it gets fixed later.

libcfnet/protocol.c
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        /*
+         * Iterates over each string in the received transaction and appends
+         * it to the Seq list, until it either finds the CFD_TERMINATOR
+         * string, or reaches the end of the message.
+         */
+        for (int i = 0; i < len && buf[i] != '\0'; i += strlen(buf + i) + 1)
+        {
+            if (StringSafeEqual(buf + i, CFD_TERMINATOR))
```
olehermanse: It seems in other places this is doing `strncmp`:

```
if (strncmp(recvbuffer, CFD_TERMINATOR, strlen(CFD_TERMINATOR)) == 0)
```

So we are not looking for the NUL-byte, only that the bytes within `strlen(CFD_TERMINATOR)` are equal. You should probably use `StringSafeEqualN()`

karlhto: If anything, `sizeof(CFD_TERMINATOR) - 1` should be used. Kind of pointless to traverse the string twice every time the function is called (though I assume the compiler probably caches the value).

libcfnet/protocol.c
```diff
@@ -0,0 +1,90 @@
...
+#include <tls_generic.h>
+
+Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
+{
+    assert(conn != NULL);
+    assert(path != NULL);
+
+    char buf[CF_MSGSIZE] = {0};
+    int tosend = snprintf(buf, CF_MSGSIZE, "OPENDIR %s", path);
+    if (tosend < 0 || tosend >= CF_MSGSIZE)
```
karlhto: `tosend >= CF_MSGSIZE` is kind of redundant. `snprintf` can't exceed `CF_MSGSIZE` if it is limited by it.

olehermanse: @karlhto Depends on the implementation. BSD man page:

```
     The snprintf() and vsnprintf() functions will write at most size-1 of
     the characters printed into the output string (the size'th character
     then gets the terminating `\0'); if the return value is greater than or
     equal to the size argument, the string was too short and some of the
     printed characters were discarded.
```

```
     These functions return the number of characters printed (not including
     the trailing `\0' used to end output to strings), except for snprintf()
     and vsnprintf(), which return the number of characters that would have
     been printed if the size were unlimited (again, not including the final
     `\0').  These functions return a negative value if an error occurs.
```

karlhto: Alright, thanks for the explanation @olehermanse :)
This should now be ready for merge IMO.

olehermanse:
@cf-bottom let's run this through jenkins one last time, just in case.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2881/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2881/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2881/

vpodzime:
> Alright, I triggered a build:
> 
> [Build Status](https://ci.cfengine.com//job/pr-pipeline/2881/)

Green, except for one random failure in DTs.

PR #3777, opened 2019-08-02
Fix .xy floating point number parsing in eval()

vpodzime:
The first commit is just paving the way. Only the changes in the *libpromises/math.peg* file are relevant for review.

merge together:
https://github.com/cfengine/core/pull/3778


3rdparty/peg-0.1.15/src/peg.peg-c
```diff
@@ -1,4 +1,4 @@
-/* A recursive-descent parser generated by peg 0.1.13 */
+/* A recursive-descent parser generated by peg 0.1.15 */
```
nickanderson: so did this actually update the dependacny version? We should note that inthe dependancy tables in buildscripts I guess.

vpodzime: This is not a runtime dependency, not even a build dependency. It's only needed when you want to regenerate the `math.pc` file from the `math.peg` grammar (which we do manually).

And I only updated `peg` to a version that was clearly used before to (re)generate the `math.pc` file.

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2487/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2487/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2487/

3rdparty/peg-0.1.15/src/peg.peg-c
nickanderson: I see. thanks.

PR #3789, opened 2019-08-12
CFE-3116: Added test for promising permissions of a symlink

nickanderson:


cf-bottom:

PR #3798, opened 2019-08-20
CFE-3124: Adjusted template so that we know template rendered

nickanderson:


cf-bottom:

PR #3806, opened 2019-08-27
Added individual Makefiles to .gitignore

olehermanse:
This was causing problem with the following workflow:

- On master branch
- With libntech submodule
- libntech submodule has peg Makefile
- Run `git checkout 3.12.x`
- libntech folder is left in place (no submodule)
- The peg Makefile inside libntech is now ignored (not commited)
- Run `git clean -fXd`
- peg Makefile is deleted
- Run `git checkout master`
- libntech is now a submodule again, but peg Makefile is deleted

This change should be backported to make it work as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@
 *.o
 .deps
 .libs
-Makefile
 Makefile.in
 stamp-h1
 .dirstamp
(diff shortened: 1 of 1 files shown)


olehermanse:
Cherry-picks:
https://github.com/cfengine/core/pull/3815
https://github.com/cfengine/core/pull/3816

PR #3818, opened 2019-09-02
A small code fix in network info gathering

vpodzime:


vpodzime:

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2701/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2701/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2701/

olehermanse:
https://github.com/cfengine/core/pull/3819 was merged causing this to conflict, I believe you can just drop the commit which changes libntech commit SHA here.

vpodzime:
> #3819 was merged causing this to conflict, I believe you can just drop the commit which changes libntech commit SHA here.

Done.

libenv/unix_iface.c
```diff
@@ -1168,6 +1168,46 @@ JsonElement* GetProcFileInfo(EvalContext *ctx, const char* filename, const char*
 
 /*******************************************************************/
 
+static long GetInterfaceMTU(const char *iface)
```
olehermanse: `limitted` -> `limited` in commit message.

libenv/unix_iface.c
```diff
@@ -1168,6 +1168,46 @@ JsonElement* GetProcFileInfo(EvalContext *ctx, const char* filename, const char*
...
+{
+    char sys_file_path[PATH_MAX] = {0};
+    int ret = snprintf(sys_file_path, PATH_MAX, "/sys/class/net/%s/mtu", iface);
```
olehermanse: From man page:

> The snprintf() and vsnprintf() functions will write at most size-1 of the characters printed into the output string (the size'th character then gets the terminating '\0'); if the return value is greater than or equal to the size argument, the string was too short and some of the printed characters were discarded.  The output is always null-terminated, unless size is 0.

So as I understand it, you will (worst case) print `\0` to `sys_file_path[PATH_MAX]`, a buffer overflow of 1 byte.

According to:

https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html

PATH MAX includes the NUL byte:

> Maximum number of bytes in a pathname, including the terminating null character.

So I guess the correct thing is to do `PATH_MAX - 1` both here, and in the comparison below. For snprintf, both size param and return value should be treated as string length;

> These functions return the number of characters printed (not including the trailing '\0' used to end output to strings), except for snprintf() and vsnprintf(), which return the number of characters that would have been printed if the size were unlimited (again, not including the final '\0').

libenv/unix_iface.c
```diff
@@ -1217,6 +1258,28 @@ void GetNetworkingInfo(EvalContext *ctx)
                     default_route = route;
                 }
             }
+
+            /* fixup MTU -- use the interface's MTU if the route doesn't specify
+             * its own [CFE-3098] */
+            JsonElement *mtu = JsonObjectGet(route, "mtu");
+            if (mtu != NULL &&
```
olehermanse: What does it mean if `mtu == NULL`?

libenv/unix_iface.c
```diff
@@ -1217,6 +1258,28 @@ void GetNetworkingInfo(EvalContext *ctx)
...
+                JsonGetElementType(mtu) == JSON_ELEMENT_TYPE_PRIMITIVE &&
```
olehermanse: I'd prefer the `&&` operator on the start of lines here.

libenv/unix_iface.c
vpodzime: Good catch, thanks!

libenv/unix_iface.c
vpodzime: It means parsing of the info for the route failed. I think it's better to avoid working with it further here.

libenv/unix_iface.c
vpodzime: There's a very similar block of code right above this new one I added which uses this style. Maybe the whole file could be reformatted with clang-format in a separate commit, but I'm not using it.

vpodzime:
Fixed the bug with buffer size, answered your questions/comments, @olehermanse. Please re-review.

vpodzime:
@olehermanse The change to propagate the iface MTU value to the routes' info was rejected so I removed it from this PR. Let's just merge the small code fix.

PR #3823, opened 2019-09-09
CFE-3115 cf-remote uninstall

merlin-northern:
Usage:

./cf-remote --hub user@host uninstall;
./cf-remote --hosts user@host uninstall;

ChangeLog:title
Signed-off-by: Peter Grzybowski <peter@northern.tech>

### DIFF
--- a/contrib/cf-remote/cf_remote/main.py
+++ b/contrib/cf-remote/cf_remote/main.py
@@ -53,6 +53,8 @@ def run_command_with_args(command, args):
             version=args.version,
             demo=args.demo,
             call_collect=args.call_collect)
+    elif command == "uninstall":
+        commands.uninstall(args.hub, args.hosts)
     elif command == "packages":
         commands.packages(tags=args.args, version=args.version)
     elif command == "run":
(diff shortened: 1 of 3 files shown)


cf-bottom:

contrib/cf-remote/cf_remote/remote.py
```diff
@@ -204,3 +219,46 @@ def install_host(
             demo_lib.agent_run(data, connection=connection)
             demo_lib.disable_password_dialog(host)
         demo_lib.agent_run(data, connection=connection)
+
+
+def get_artifact_url(hub, data):
```
olehermanse: This function seems a little overkill for deletion, since you could just try the 3 possibilities. Something like:

```
dpkg --purge cfengine3-community || dpkg --purge cfengine3-nova-hub || dpkg --purge cfengine3
```

(Or you could even search for what cfengine is installed).
```

File: contrib/cf-remote/cf_remote/main.py [hunk 2]
```diff
@@ -66,9 +68,12 @@ def run_command_with_args(command, args):
 
 
 def validate_command(command, args):
-    if command in ["info", "sudo", "run"] and not args.hosts:
+    if command in ["info", "sudo", "run" ] and not args.hosts:
```
olehermanse: Very minor, but this added space is not wanted.

contrib/cf-remote/cf_remote/remote.py
olehermanse: Also, I think there is a flaw here. I want to uninstall my hosts like this:

```
cf-remote uninstall --hosts 1.2.3.4,2.2.3.4,3.3.3.4
```

But this will fail if one of those is a hub.

olehermanse:
Closing this, I have submitted a new version here:

https://github.com/cfengine/core/pull/3973

PR #3844, opened 2019-09-17
Fixed issue with including pcre.h on OpenBSD

olehermanse:
It seems this variable already had a value, with the include
path of the header, so overwriting it here was a bad idea.

We should probably fix the line below as well, but I don't
know what's appropriate, and this patch was enough to fix
my problems when building on OpenBSD.

Merge first: https://github.com/cfengine/libntech/pull/10

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -491,7 +491,7 @@ fi
 
 CF3_WITH_LIBRARY(pcre, [
   AC_CHECK_LIB(pcre, pcre_exec, [], [AC_MSG_ERROR(Cannot find PCRE)])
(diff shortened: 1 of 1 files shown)


olehermanse:
@cf-bottom jenkins, just in case.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2784/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2784/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2784/

PR #3851, opened 2019-09-20
Travis: Adapted valgrind.sh to be usable in masterfiles repo

olehermanse:
See: https://github.com/cfengine/masterfiles/pull/1517

### DIFF
--- a/travis-scripts/valgrind.sh
+++ b/travis-scripts/valgrind.sh
@@ -5,9 +5,23 @@ if [ -d /var/cfengine ]; then
     rm -rf /var/cfengine
 fi
 
+# Test assumes we start in core / masterfiles directory
+cd ../
+
+if [ ! -d core ]; then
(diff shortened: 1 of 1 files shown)


travis-scripts/valgrind.sh
```diff
@@ -5,9 +5,23 @@ if [ -d /var/cfengine ]; then
     rm -rf /var/cfengine
 fi
 
+# Test assumes we start in core / masterfiles directory
```
Lex-2008: I'd write `or` instead of `/` because everywhere in this file `/` is a path separator so this looks like `masterfiles` must be subdir of `core`. But that's just a nitpick.

olehermanse: https://github.com/cfengine/masterfiles/commit/16598847a1910d645b73ff81a65bbfcc8aa0ff0e
https://github.com/cfengine/core/pull/3853

PR #3856, opened 2019-09-24
Never use MDB_NOSYNC on AIX and Solaris

vpodzime:
The performance gain is not worth the instability and issues with
consistency on these legacy platforms.

Ticket: ENT-4002
Changelog: Local databases now use synchronous access on AIX and Solaris

### DIFF
--- a/libpromises/dbm_lmdb.c
+++ b/libpromises/dbm_lmdb.c
@@ -282,10 +282,16 @@ DBPriv *DBPrivOpenDB(const char *const dbpath, const dbid id)
     }
 
     unsigned int open_flags = MDB_NOSUBDIR;
+#if !defined(_AIX) && !defined(__sun)
+    /* The locks and lastseen (on hubs) DBs are heavily used and using
+     * MDB_NOSYNC increases performance. However, AIX and Solaris often suffer
+     * from some serious issues with consistency (ENT-4002) so it's better to
+     * sacrifice some performance there in favor of stability. */
     if (id == dbid_locks || (GetAmPolicyHub() && id == dbid_lastseen))
     {
         open_flags |= MDB_NOSYNC;
     }
+#endif
 
 #ifdef __hpux
     /*


libpromises/dbm_lmdb.c
```diff
@@ -282,10 +282,16 @@ DBPriv *DBPrivOpenDB(const char *const dbpath, const dbid id)
     }
 
     unsigned int open_flags = MDB_NOSUBDIR;
+#if defined(_AIX) || defined(__sun)
+    /* The locks and lastseen (on hubs) DBs are heavily used and using
+     * MDB_NOSYNC increases performance. However, AIX and Solaris often suffer
+     * from some serious issues with consistency (ENT-4002) so it's better to
+     * sacrifice some performance there in favor of stability. */
```
kacf: I'm not sure this will help actually (hard to say, it depends exactly on how the file cache works in AIX). The thing is that this only forces it to disk, but you still don't know what's in the cache. If the mmapped cache and the file descriptor cache indeed are not synchronized, then I would expect this not to help.

Wasn't your original question, @vpodzime, whether to use `WRITEMAP`? Because this *would* help, I'm pretty sure.

What this PR will help with is if they are experiencing corruption after a powerloss, but this is the *only* case where I would expect it to help.

vpodzime: > Wasn't your original question, @vpodzime, whether to use `WRITEMAP`? Because this _would_ help, I'm pretty sure.

Yes, that was my original suspicion. But then I found 
> Modifications to a file-system object are seen consistently, whether accessed from a mapped file region or from the read or write subroutine. 
(https://www.ibm.com/support/knowledgecenter/en/ssw_aix_72/m_bostechref/mmap.html)
so I don't think that's the problem on AIX.

My current suspicion is that something is force-killing some of our processes on AIX/Solaris and that results in corruptions. Which this could help with, maybe?

vpodzime: > Better said: But `NOSYNC` _has no effect_ on force killings

Unless force-killed processes are treated really badly on AIX. Which we cannot disprove.

libpromises/dbm_lmdb.c
```diff
@@ -282,10 +282,16 @@ DBPriv *DBPrivOpenDB(const char *const dbpath, const dbid id)
     }
 
     unsigned int open_flags = MDB_NOSUBDIR;
+#if defined(_AIX) || defined(__sun)
```
olehermanse: 🤔

vpodzime: :eyes:

olehermanse:
@cf-bottom jenkins, with exotics, please.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2981/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2981/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2981/

vpodzime:
Let's conservatively merge this to *master* and we will see if there's any noticeable negative impact on the performance.

PR #3857, opened 2019-09-24
Added user flags from configure to AM_CFLAGS

olehermanse:
Overriding CFLAGS in configure is not correct, according to:

https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html

However, we already do this, and it seems quite common:

https://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_check_enable_debug.m4

And we already do this substitution in other Makefile.am files.
So, for now, I think it's better to try to be consistent.

This patch allows you to do something like this:

```
$ ./autogen.sh --enable-debug
$ make -j2 CFLAGS='-Werror -Wall -Wno-pointer-sign'
```

And still get debugging information (line numbers in gdb) for
libpromises. Previously, the explicit `CFLAGS` in make command
would override the `-g` option from `--enable-debug`

Changelog: None
Ticket: None
Signed-off-by: Ole Herman Schumacher Elgesem <ole@northern.tech>


olehermanse:
@cf-bottom jenkins, please.

olehermanse:
Replaced by: https://github.com/cfengine/core/pull/3859

PR #3863, opened 2019-09-26
CFE-3168 Fix doc rendering

nickanderson:
The docs render with some empty boxes, this change tries to fix that.


cf-bottom:

PR #3864, opened 2019-09-27
CFE-3087: New functions Get and StatGet for protocol

karlhto:
Added functions `Get` and `StatGet`, where the latter just stats the file before calling the former function with the correct file size. `Get` uses `GET <block_size> <remote_path>` to retrieve a file.

Ticket: CFE-3087

### DIFF
--- a/libcfnet/protocol.c
+++ b/libcfnet/protocol.c
@@ -28,6 +28,7 @@
 #include <client_protocol.h>
 #include <definitions.h>
 #include <net.h>
+#include <stat_cache.h>
 #include <string_lib.h>
 #include <tls_generic.h>
 
@@ -87,4 +88,123 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
     }
 
     return seq;
+}
+
+bool ProtocolGet(AgentConnection *conn, const char *remote_path,
+                 const char *local_path, const uint32_t file_size, int perms)
+{
+    assert(conn != NULL);
+    assert(remote_path != NULL);
+    assert(local_path != NULL);
+    assert(file_size != 0);
+
+    perms = (perms == 0) ? CF_PERMS_DEFAULT : perms;
+
+    unlink(local_path);
+    FILE *file_ptr = safe_fopen_create_perms(local_path, "wx", perms);
+    if (file_ptr == NULL)
+    {
+        Log(LOG_LEVEL_WARNING, "Failed to open file %s (fopen: %s)",
+            local_path, GetErrorStr());
+        return false;
+    }
+
+    char buf[CF_MSGSIZE] = {0};
+    int to_send = snprintf(buf, CF_MSGSIZE, "GET %d %s",
+                           CF_MSGSIZE, remote_path);
+
+
+    int ret = SendTransaction(conn->conn_info, buf, to_send, CF_DONE);
+    if (ret == -1)
+    {
+        Log(LOG_LEVEL_WARNING, "Failed to send request for remote file %s:%s",
+            conn->this_server, remote_path);
+        unlink(local_path);
+        fclose(file_ptr);
+        return false;
+    }
+
+    char cfchangedstr[sizeof(CF_CHANGEDSTR1 CF_CHANGEDSTR2)];
+    snprintf(cfchangedstr, sizeof(cfchangedstr), "%s%s",
+             CF_CHANGEDSTR1, CF_CHANGEDSTR2);
+
+    bool success = true;
+    uint32_t received_bytes = 0;
+    while (received_bytes < file_size)
+    {
+        int len = TLSRecv(conn->conn_info->ssl, buf, CF_MSGSIZE);
+        if (len == -1)
+        {
+            Log(LOG_LEVEL_WARNING, "Failed to GET file %s:%s",
+                conn->this_server, remote_path);
+            success = false;
+            break;
+        }
+        else if (len > CF_MSGSIZE)
+        {
+            Log(LOG_LEVEL_WARNING,
+                "Incorrect length of incoming packet "
+                "while retrieving %s:%s, %d > %d",
+                conn->this_server, remote_path, len, CF_MSGSIZE);
+            success = false;
+            break;
+        }
+
+        if (BadProtoReply(buf))
+        {
+            Log(LOG_LEVEL_ERR,
+                "Error from server while retrieving file %s:%s: %s",
+                conn->this_server, remote_path, buf);
+            success = false;
+            break;
+        }
+
+        if (StringSafeEqualN(buf, cfchangedstr, sizeof(cfchangedstr) - 1))
+        {
+            Log(LOG_LEVEL_ERR,
+                "Remote file %s:%s changed during file transfer",
+                conn->this_server, remote_path);
+            success = false;
+            break;
+        }
+
+        ret = fwrite(buf, sizeof(char), len, file_ptr);
+        if (ret < 0)
+        {
+            Log(LOG_LEVEL_ERR,
+                "Failed to write during retrieval of file %s:%s (fwrite: %s)",
+                conn->this_server, remote_path, GetErrorStr());
+            success = false;
+            break;
+        }
+
+        received_bytes += len;
+    }
+
+    if (!success)
+    {
+        unlink(local_path);
+    }
+
+    fclose(file_ptr);
+    return success;
+}
+
+bool ProtocolStatGet(AgentConnection *conn, const char *remote_path,
+                     const char *local_path, int perms)
+{
+    assert(conn != NULL);
+    assert(remote_path != NULL);
+
+    struct stat sb;
+    int ret = cf_remote_stat(conn, false, remote_path, &sb, "file");
+    if (ret == -1)
+    {
+        Log(LOG_LEVEL_ERR,
+            "%s: Failed to stat remote file %s:%s",
+            __func__, conn->this_server, remote_path);
+        return false;
+    }
+
+    return ProtocolGet(conn, remote_path, local_path, sb.st_size, perms);
 }
--- a/libcfnet/protocol.h
+++ b/libcfnet/protocol.h
@@ -64,4 +64,55 @@
  */
 Seq *ProtocolOpenDir(AgentConnection *conn, const char *path);
 
+/**
+ * Receives a file from a remote host.
+ *
+ * The server will use "/var/cfengine" as working directory if absolute path
+ * is not specified. The client will send a request that looks like this:
+ *      `GET <buf_size> <remote_path>`
+ *
+ * `buf_size` is the local buffer size: how much of the file to receive in
+ * each transaction. It should be aligned to block size (which is usually
+ * 4096), but currently the protocol only allows _exactly_ 4095 bytes per
+ * transaction.
+ *
+ * The function shall fail if connection is not established, or if the server
+ * gives a bad response (denoted by a message preceded by "BAD").
+ *
+ * @param [in] conn         The connection to use
+ * @param [in] remote_path  Path on remote host
+ * @param [in] local_path   Path of received file
+ * @param [in] file_size    Size of file to get
+ * @param [in] perms        Permissions of local file
+ * @return True if file was successfully transferred, false otherwise
+ *
+ * Example (for printing each directory entry):
+ * @code
+ *     AgentConnection *conn = ServerConnection("127.0.0.1", "666", ...);
+ *     bool got_file = ProtocolGet(conn, "masterfiles/update.cf",
+ *                                 "update.cf", CF_MSGSIZE, 0644);
+ *     if (got_file)
+ *     {
+ *        struct stat sb;
+ *        stat("update.cf", &sb);
+ *        printf("This file is %ld big!\n", sb.st_size);
+ *     }
+ * @endcode
+ *
+ * In the protocol, this will look like this on the server side:
+ *     Received: GET masterfiles/update.cf
+ *     Translated to: GET /var/cfengine/masterfiles/update.cf
+ */
+bool ProtocolGet(AgentConnection *conn, const char *remote_path,
+                 const char *local_path, const uint32_t file_size, int perms);
+
+
+/**
+ * Receives a file from a remote host, see documentation for #ProtocolGet
+ *
+ * This funtion will first stat the remote path before attempting to receive
+ * it.
+ */
+bool ProtocolStatGet(AgentConnection *conn, const char *remote_path,
+                     const char *local_path, int perms);
 #endif
--- a/cf-net/cf-net.c
+++ b/cf-net/cf-net.c
@@ -695,17 +695,9 @@ static void *CFNetGetFile(void *arg)
         return NULL;
     }
 
-    struct stat sb;
-    data->ret = cf_remote_stat(conn, true, data->remote_file, &sb, "file");
-    if (data->ret != 0)
-    {
-        printf("Could not stat: '%s'\n", data->remote_file);
-    }
-    else
-    {
-        bool ok = CopyRegularFileNet(data->remote_file, data->local_file, sb.st_size, true, conn);
-        data->ret = ok ? 0 : -1;
-    }
+    bool ok = ProtocolStatGet(conn, data->remote_file,
+                               data->local_file, 0644);
+    data->ret = ok ? 0 : -1;
     CFNetDisconnect(conn);
     return NULL;
 }


karlhto:
@cf-bottom Jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4023/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4023/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4023/

libcfnet/protocol.c
```diff
@@ -88,3 +89,88 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
...
+    while (to_recv < file_size)
+    {
+        int len = TLSRecv(conn->conn_info->ssl, buf, CF_MSGSIZE);
+        if (len == -1)
+        {
+            break;
+        }
+        else if (len > CF_MSGSIZE)
+        {
+            fprintf(stderr, "What?\n");
```
vpodzime: Use `Log()` here.

cf-serverd/server_common.c
```diff
@@ -430,123 +443,113 @@ void CfGetFile(ServerFileGetState *args)
...
             TLSSend(ConnectionInfoSSL(conn_info), sendbuffer, args->buf_size);
         }
+
         return;
     }
 
 /* File transfer */
 
-    if ((fd = safe_open(filename, O_RDONLY)) == -1)
+    int div = 3;
```
craigcomstock: I know this was existing code but seems like "div" should be named something like "frequency_of_source_file_change_checks" ;) Or something a little shorter. ;)

libcfnet/protocol.c
```diff
@@ -88,3 +89,88 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
...
+{
+    assert(conn != NULL);
+    assert(remote_path != NULL);
+    assert(local_path != NULL);
+    assert(file_size != 0);
+
+    perms = (perms == 0) ? CF_PERMS_DEFAULT : perms;
+
+    unlink(local_path);
+    FILE *file_ptr = safe_fopen_create_perms(local_path, "w", perms);
```
craigcomstock: check for return value from unlink()?

libcfnet/protocol.c
```diff
@@ -88,3 +89,88 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
...
+    if (file_ptr == NULL)
+    {
+        return false;
```
craigcomstock: Does safe_fopen_create_perms() log what went wrong? Seems like it does in most cases in safe_open_create_perms() so probably OK to just return false with no logging here.

libcfnet/protocol.c
craigcomstock: Definitely change "What?" to some descriptive text right? Like "Received more bytes than CF_MSGSIZE(%d)"?

craigcomstock: Maybe prefix with "Protocol error:" as well?

libcfnet/protocol.c
```diff
@@ -88,3 +89,88 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
...
+            unlink(local_path);
+            fclose(file_ptr);
+            return false;
+        }
+
+        ret = fwrite(buf, sizeof(char), len, file_ptr);
+        if (ret < 0)
+        {
+            fprintf(stderr, "bad\n");
+            unlink(local_path);
```
craigcomstock: Use Log() here as well and be more descriptive like "local file write failed, fwrite: %s", GetErrorStr()?

libcfnet/protocol.c
```diff
@@ -88,3 +89,88 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
...
+            fclose(file_ptr);
+            return false;
```
craigcomstock: generally, do you want to pay attention to return value/errors on unlink(), fclose(), etc? There are many system calls here and I know it would get messy if each is handled. I guess the question is what can you do if one errors and is it significant to the code that follows?

libpromises/cf3.defs.h
```diff
@@ -132,12 +132,6 @@ typedef enum
 
 /*****************************************************************************/
 
-#define CF_FAILEDSTR "BAD: Unspecified server refusal (see verbose server output)"
-#define CF_CHANGEDSTR1 "BAD: File changed "     /* Split this so it cannot be recognized */
-#define CF_CHANGEDSTR2 "while copying"
```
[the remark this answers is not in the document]
karlhto: We figured it out, it's likely to avoid the string being recognised when transferring a CFEngine binary. It's terrible and hacky, but the way the protocol is designed it can't really be helped.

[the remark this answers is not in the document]
karlhto: ...with other words I need to revert this. It fits better where I moved it anyway, so won't be a big change.

libcfnet/protocol.c
karlhto: No, I need to add better log messages here. : )

libcfnet/protocol.c
karlhto: If it fails to unlink, there isn't really much we can do apart from trying again, but stalling the program until the file is successfully deleted (for instance if the file is busy, like a mount point or something) is a bad solution. We could notify the user that the file could not be deleted successfully?

The same goes for `fclose`. I don't know the implications of if a file descriptor remains open in a program until it quits, but `fclose` failing will mean memory leaks since the `FILE` struct won't be deallocated.

cf-serverd/server_common.c
```diff
@@ -373,8 +373,7 @@ static void AbortTransfer(ConnectionInfo *connection, char *filename)
     Log(LOG_LEVEL_VERBOSE, "Aborting transfer of file due to source changes");
 
     char sendbuffer[CF_BUFSIZE];
-    snprintf(sendbuffer, CF_BUFSIZE, "%s%s: %s",
-             CF_CHANGEDSTR1, CF_CHANGEDSTR2, filename);
+    snprintf(sendbuffer, CF_BUFSIZE, "%s: %s", CF_CHANGEDSTR, filename);
```
olehermanse: Why are you changing / cleaning up cf-serverd code? I thought we agreed to implement the new protocol functions, independently of the old functionality. That way we can ensure that we don't fix / change any parts of the implementation, and test cf-net with an "unaltered" cf-serverd. Also, why is there a diff in the commit message?

karlhto: We did agree to that, but at the time of making these changes, I was not aware of `CF_CHANGEDSTR` actually having an effect on functionality. The other changes I made are just code cleanup, and I want to split those into a new PR now.

The diff is there by accident, as you might have guessed. Something probably went wrong when I wanted to reuse the commit message in `COMMIT_EDITMSG` when a commit failed.

libcfnet/protocol.c
karlhto: I'm going to add an `O_EXCL` option for `fopen`. Checking result of `unlink` should be unnecessary then.

karlhto:
@cf-bottom Jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4059/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4059/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4059/

karlhto:
There's a mistake in docs, sec

karlhto:
@cf-bottom Jenkins

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4060/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4060/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4060/

libcfnet/protocol.h
```diff
@@ -64,4 +64,52 @@
...
+ *        printf("This file is %ld big!\n", sb.st_size);
+ *     }
+ * @endcode
+ *
+ * In the protocol, this will look like this on the server side:
+ *     Received: GET masterfiles/update.cf
+ *     Translated to: GET /var/cfengine/masterfiles/update.cf
+ */
+bool ProtocolGet(AgentConnection *conn, const char *remote_path,
+                 const char *local_path, const uint32_t file_size, int mode);
```
karlhto: Should be named `perms` instead of `mode` to correspond with how we name things in `file_lib`, maybe

karlhto:
No changes, only rebased to current master and removed `libntech` update.

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4074/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4074/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4074/

libcfnet/protocol.c
```diff
@@ -88,3 +89,123 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
...
+    char buf[CF_MSGSIZE] = {0};
+    int to_send = snprintf(buf, CF_MSGSIZE, "GET %d %s",
+                           CF_MSGSIZE, remote_path);
+
+    int ret = SendTransaction(conn->conn_info, buf, to_send, CF_DONE);
+    if (ret == -1)
+    {
+        Log(LOG_LEVEL_ERR,
+            "%s: Failed to send request for file %s on host %s",
+            __func__, remote_path, conn->this_server);
```
olehermanse: We want our log messages (verbose and above) to be user oriented. The function name doesn't belong here.

libcfnet/protocol.c
```diff
@@ -88,3 +89,123 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
...
+        fclose(file_ptr);
+        return false;
+    }
+
+    char cfchangedstr[sizeof(CF_CHANGEDSTR1 CF_CHANGEDSTR2)];
+    snprintf(cfchangedstr, sizeof(cfchangedstr), "%s%s",
+             CF_CHANGEDSTR1, CF_CHANGEDSTR2);
+
+    bool res = true;
+    uint32_t to_recv = 0;
```
olehermanse: `to_recv` should be called `received` or `received_bytes`. `to_recv` sounds like the target (file size), not how many bytes you have already received / read.

libcfnet/protocol.c
```diff
@@ -88,3 +89,123 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
...
+    while (to_recv < file_size)
+    {
+        int len = TLSRecv(conn->conn_info->ssl, buf, CF_MSGSIZE);
+        if (len == -1)
+        {
+            Log(LOG_LEVEL_ERR,
+                "%s: Failed to retrieve file %s:%s",
```
olehermanse: Remove function name, here and all errors below. (You could also rewrite slightly to present the same information, for example `Failed to GET file` or similar).

libcfnet/protocol.c
```diff
@@ -88,3 +89,123 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
...
+        if (BadProtoReply(buf))
+        {
+            Log(LOG_LEVEL_ERR,
+                "%s: Error from server while retrieving file %s:%s: %s",
+                __func__, conn->this_server, remote_path, buf);
+            res = false;
+            break;
+        }
+
+        if (strncmp(buf, cfchangedstr, sizeof(cfchangedstr) - 1) == 0)
```
olehermanse: `StringSafeEqualN()` is more readable.

libcfnet/protocol.c
```diff
@@ -88,3 +89,123 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
...
+        Log(LOG_LEVEL_ERR, "%s: Failed to open file %s (fopen: %s)",
+            __func__, local_path, GetErrorStr());
+        return false;
+    }
+
+    char buf[CF_MSGSIZE] = {0};
+    int to_send = snprintf(buf, CF_MSGSIZE, "GET %d %s",
+                           CF_MSGSIZE, remote_path);
+
+    int ret = SendTransaction(conn->conn_info, buf, to_send, CF_DONE);
```
olehermanse: I would add a comment about the fact that `SendTransaction()` NUL-terminates buf (always).

libcfnet/protocol.c
```diff
@@ -88,3 +89,123 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
...
+}
+
+bool ProtocolStatGet(AgentConnection *conn, const char *remote_path,
+                     const char *local_path, int perms)
+{
+    assert(conn != NULL);
+    assert(remote_path != NULL);
+
+    struct stat sb;
+    int ret = cf_remote_stat(conn, false, remote_path, &sb, "file");
```
olehermanse: This will be replaced by `ProtocolStat()` right?

libcfnet/protocol.h
```diff
@@ -64,4 +64,55 @@
...
+ *        printf("This file is %ld big!\n", sb.st_size);
+ *     }
+ * @endcode
+ *
+ * In the protocol, this will look like this on the server side:
+ *     Received: GET masterfiles/update.cf
+ *     Translated to: GET /var/cfengine/masterfiles/update.cf
+ */
+bool ProtocolGet(AgentConnection *conn, const char *remote_path,
+                 const char *local_path, const uint32_t file_size, int perms);
```
olehermanse: No big deal, but I don't think the `uint32_t` should be declared `const` in the declaration. IMO, it only makes sense to include it in the function definition (implementation).

libcfnet/protocol.c
karlhto: Yep.

libcfnet/protocol.h
karlhto: Considering the documentation is in the header file, I think it provides useful information that the function will not use another file size than the one provided by the caller.

libcfnet/protocol.c
karlhto: Not sure why this would be relevant here, since `snprintf` also `NUL`-terminates the buffer

karlhto:
@cf-bottom Jenkins pls

libcfnet/protocol.h
olehermanse: No. `const` in the declaration declares that you won't change the parameter. For `const char *`, this is useful because you can reasonably assume that a function with this signature won't change your string (the data the pointer points to). It in no way guarantees that the parameter `file_size` will be used in the protocol command, nothing stops the function from falling back to something else, for example:

```C
snprintf(buf, sizeof(buf), "GET %d /my/file", file_size ? file_size : 32);
```

`const int` or similar in the declaration is confusing, it does nothing, and implies that the function would be able to alter your int, if the `const` was not there. 

In the definiton (implementation), it does have a function:

```C
bool ProtocolGet(const char *const path, const int file_size)
{
  [... big function with lots of logic ...]
  // Since path and file_size are const, we know that the spaghetti above 
  // didn't change them. We've also prevented someone from altering them
  // perhaps accidentally.
  Log(LOG_LEVEL_INFO, "Got file '%s' of size %d", path, file_size);
  return true;
}
```

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4108/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4108/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4108/

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4109/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4109/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4109/

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4112/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4112/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4112/

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4113/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4113/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4113/

karlhto:
I should get better at testing this before saying it's ready.

libcfnet/protocol.h
karlhto: > It in no way guarantees that the parameter file_size will be used in the protocol command, nothing stops the function from falling back to something else

It's hard to not agree with this point, but I would still argue that it still adds understanding. I like to match parameters in declarations and definitions.

> In the definiton (implementation), it does have a function:

Yes, I am aware of that. But regarding the point that it does not do anything in the declaration: you might as well say that adding parameter names in the declaration is pointless because they don't have a function there.

olehermanse: @karlhto parameter names clearly have a purpose, they explain what the parameter means (a human readable name). I don't think anyone would argue this. However your interpretation of what `const` means is arbitrary, it's not based on how the language works, or what the const keyword usually means.

> It's hard to not agree with this point, but I would still argue that it still adds understanding. 

And I think it adds confusion. 

> I like to match parameters in declarations and definitions.

We've discussed this in another PR before, with craig I believe. Yes, you can sync them, that's kind of a different point than what you were suggesting above. Above, you stated that `const` adds understanding. If you really believe that, you could start adding more `const` to declarations where it's not even in the definitons.

That's fine for this case. When you get a lot of consts in the function declaration, I think it makes it a lot more easy to read if you drop the unnecessary ones. There are even some quite rare cases where I would say it's appropriate for the function definition to use different parameter names from the declaration.

vpodzime: +1 to @olehermanse from me

olehermanse: > it still provides a hint that another value won't be used (in my opinion).

@karlhto It doesn't. This is where we disagree. That's just not what `const` means. Within the function body, it means that the local variable `file_size` is constant, cannot be assigned to. Again, in the declaration it is just confusing. (If I saw that I would wonder why it was there, not assume that it was what you are hinting at).

> It's not a big deal and we probably shouldn't spend this much energy on this.

Yes. I wouldn't block your PR on this, but since this is the second time it comes up during PR review, we should probably "settle" it and add a short guideline to our style guide.

karlhto: > we should probably "settle" it and add a short guideline to our style guide.

@olehermanse Sounds good. I am fine with settling on your style since @vpodzime seems to agree.

karlhto:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4116/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4116/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4116/

libcfnet/protocol.c
```diff
@@ -88,3 +89,122 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
 
     return seq;
 }
+
+bool ProtocolGet(AgentConnection *conn, const char *remote_path,
+                 const char *local_path, const uint32_t file_size, int perms)
```
olehermanse: Just curious, why did you choose `uint32_t`?

vpodzime: I was curious about the same. I don't recall these types being used anywhere else in the codebase. And I think `ssize_t` would make more sense here?

olehermanse:
@karlhto I merged the PR, because I think the comments were not severe enough to block this any longer. Please reply to the comments and submit a follow-up PR if appropriate.

PR #3875, opened 2019-10-07
Use file.io instead of transfer.sh

Lex-2008:
They seem to be a more reliable storage facility

### DIFF
--- a/travis-scripts/after_script.sh
+++ b/travis-scripts/after_script.sh
@@ -16,7 +16,9 @@ test "$TRAVIS_PULL_REQUEST" = "false" && BRANCH_OR_PULL_REQUEST=$TRAVIS_BRANCH |
 filename=cfengine-$VERSION-$BRANCH_OR_PULL_REQUEST-$JOB_TYPE.artifacts.zip
 zip -q -r $filename artifacts/
 
-echo ===== uploading $filename to transfer.sh =====
-curl --upload-file $filename https://transfer.sh/$filename
+echo ===== uploading $filename to file.io =====
+curl -F "file=@$filename" https://file.io
+echo 'Note that file.io DELETES file from their servers after first download,'
+echo "so don't delete it from your machine if you still need it!"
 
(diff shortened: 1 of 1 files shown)


travis-scripts/after_script.sh
```diff
@@ -16,7 +16,9 @@ test "$TRAVIS_PULL_REQUEST" = "false" && BRANCH_OR_PULL_REQUEST=$TRAVIS_BRANCH |
 filename=cfengine-$VERSION-$BRANCH_OR_PULL_REQUEST-$JOB_TYPE.artifacts.zip
 zip -q -r $filename artifacts/
 
-echo ===== uploading $filename to transfer.sh =====
-curl --upload-file $filename https://transfer.sh/$filename
+echo ===== uploading $filename to file.io =====
+curl -F "file=@$filename" https://file.io
+echo 'Note that file.io DELETES file from their servers after first download,'
+echo "so don't delete it from your machine if you still need it!"
```
vpodzime: I suppose we only need to download the file once.? Or are multiple travis jobs using the same file?

Lex-2008: No, we never download these files. They are there only for the case if someone wants to check a travis failure - they can always download artifacts.

vpodzime: OK, cool. I wasn't sure if this wasn't used by the other jobs downloading the packages produced by travis builds.

olehermanse: +1. A comment about this would be nice.

olehermanse:
3.12 Cherry-pick: https://github.com/cfengine/core/pull/3891

olehermanse:
@Lex-2008 can you backport to 3.10.x ? There were conflicts.

Lex-2008:
we don't upload build artifacts to transfer.sh from 3.10.x branch, that's why they're not failing. We can add such feature to 3.10.x branch, it can't hurt

PR #3878, opened 2019-10-07
CFE-3172/master Separate examples for readstringlist(), readreallist(), and readintlist()

nickanderson:
Merge together:
https://github.com/cfengine/documentation/pull/2208

### DIFF
--- a/examples/readintlist.cf
+++ b/examples/readintlist.cf
@@ -1,0 +1,34 @@
+#+begin_src prep
+#@ ```
+#@ printf "one\ntwo\nthree\n" > /tmp/list.txt
+#@ printf "1\n2\n3\n"         >> /tmp/list.txt
+#@ printf "1.0\n2.0\n3.0"     >> /tmp/list.txt
+#@ ```
+#+end_src
+#+begin_src cfengine3
+bundle agent example_readintlist
+# @brief Example showing function return types
+{
+  vars:
+      "my_list_of_integers"
+        ilist => readintlist( "/tmp/list.txt",     # File to read
+                              "^(\D+)|(\d+[^\n]+)", # Ignore any lines that are not integers
+                              "\n",                 # Split on newlines
+                              inf,                  # Maximum number of entries
+                              inf);                 # Maximum number of bytes to read
+
+  reports:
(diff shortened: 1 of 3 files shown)


examples/readintlist.cf
```diff
@@ -0,0 +1,34 @@
...
+#@ ```
+#+end_src
+bundle agent example_readintlist
+# @brief Example showing function return types
+{
+  vars:
+      "my_list_of_integers"
+        ilist => readintlist( "/tmp/list.txt",      # File to read
+                              "^(\D+)|(\d+[^\n]+)", # Ignore any non integers
+                              "\n",                 # Split on newlines
```
craigcomstock: Why do you have to provide this regex for non-ints when the function name is readintlist()? Seems like that would be a default.

craigcomstock: Then again, you are just documenting here, not implementing how it works. So all good, just struck me as curious.

examples/readstringlist.cf
```diff
@@ -0,0 +1,39 @@
...
+#@ ```
+#+end_src
+bundle agent example_readstringlist
+# @brief Example showing readstringlist()
+{
+  vars:
+      "my_list_of_strings"
+        slist => readstringlist( "/tmp/list.txt", # File to read
+                              "",                 # Ignore any non-digits
+                              "\n",               # Split on newlines
```
craigcomstock: How is an empty regex `""` ignoring any non-digits where previously a different regex was used with the same comment? `"^(\D+)",         # Ignore any non-digits`

examples/readintlist.cf
nickanderson: I had to supply it because if non-numbers were parsed into the list, the agent would error. So yes, I was just trying to document how to use it with a file that might contain non-numbers. Previously, the examples showed just ignoring hash prefixed lines (which really made sense mostly for slist).

examples/readstringlist.cf
nickanderson: bogus comment, thanks.

nickanderson:
OK @craigcomstock i fixed it up a bit, and decided that i would include hash comments in the example to exclude (it's so common, we should show it).

PR #3879, opened 2019-10-08
CFE-3165/3.12.x: Added example illustrating function return types

nickanderson:


nickanderson:
Well, i guess something is different here from master. It's not important enough to chase down. Good and done for 3.15.0

PR #3892, opened 2019-10-11
CFE-2489/master: Added test for copy_from resolving correct value for this.promiser

nickanderson:


tests/acceptance/10_files/copy_from_this_promiser.cf
```diff
@@ -0,0 +1,85 @@
...
+bundle agent init {
+  vars:
+      "ROOT" string => "/tmp/CFE-2489";
+      "SOURCE" string => "$(ROOT)/SOURCE";
+      "use_bundles"
+        slist => { "bundle_uses_copy_from_without_param",
+                   "bundle_uses_copy_from_with_param" };
+
+  files:
+      "$(init.ROOT)/$(use_bundles)/." create => "true";
```
craigcomstock: why init.ROOT aren't you still in the same bundle so just ROOT would work?

tests/acceptance/10_files/copy_from_this_promiser.cf
```diff
@@ -0,0 +1,85 @@
...
+
+      # Setup source files to copy_from
+      "$(init.ROOT)/SOURCE/$(init.ROOT)/$(use_bundles)/file.txt"
+        create => "true";
```
craigcomstock: this path looks real weird. Shouldn't it just be `"$(SOURCE)/$(use_bundles)/file.txt"`?

tests/acceptance/10_files/copy_from_this_promiser.cf
```diff
@@ -0,0 +1,85 @@
...
+
+  methods:
+      "XXX" usebundle => $(init.use_bundles);
+}
+
+bundle agent bundle_uses_copy_from_without_param {
+
+  files:
+      # As expected, when a copy_from promise uses a non-parameterized body the this.promiser context is as expected (points to the promised file).
+      "$(init.ROOT)/$(this.bundle)/file.txt"
```
craigcomstock: chop down comment to near 80 chars?

tests/acceptance/10_files/copy_from_this_promiser.cf
```diff
@@ -0,0 +1,85 @@
...
+body copy_from my_copy_without_param() {
+        source  => "$(init.ROOT)/SOURCE/$(this.promiser)";
+        compare => "digest";
+}
+
+bundle agent bundle_uses_copy_from_with_param {
+
+  files:
+      # Oddly, when a copy_from promise uses a parameterized body, the this.promiser context does not contain the promised file, instead, it seems to contain the parents promiser.
+      # So, in this case, it wrongly gets XXX from the 'test' bundle.
```
craigcomstock: chop the comment to around 80 chars?

tests/acceptance/10_files/copy_from_this_promiser.cf
nickanderson: The only reason was so that it was the same in other places making it easier for me to understand what was going on.

nickanderson: yes, ROOT would work.

tests/acceptance/10_files/copy_from_this_promiser.cf
nickanderson: ACK

tests/acceptance/10_files/copy_from_this_promiser.cf
nickanderson: ACK

tests/acceptance/10_files/copy_from_this_promiser.cf
nickanderson: No, the case is using $(this.promiser) inside the copy_from body.
Basically, the use case is like: Promise file (`/etc/myfile`) is a copy from the same path in a remote location. So under SOURCE, I needed to seed the full path to the file.

That way you could do something like this:
```
/etc/profile
  when in prod, copy from /SOURCE/PROD/etc/profile
  when in dev, copy from /SOURCE/DEV/etc/profile
```
Or perhaps:
```
/path/$(application)/config
  when talking about app1, copy from /SOURCE/app1/config
  when talking about app2, copy from /SOURCE/app2/config.
```

craigcomstock: Right. I see it now. `$(init.ROOT)/SOURCE` is like  the root of a tarball/package that would be unpacked/copied to `/`. Ack.

PR #3897, opened 2019-10-15
Upload artifacts to file.io (3.10)

Lex-2008:


olehermanse:
Since it seems this is not necessary, I will close this.

PR #3902, opened 2019-10-20
Retry SSL_accept() if WANT_READ or WANT_WRITE is indicated

vpodzime:
SSL_accept() may fail with SSL_get_error() returning
SSL_ERROR_WANT_READ or WANT_WRITE in some cases. That doesn't
indicate a fatal error in SSL connection establishing, but rather
a potentially temporary problem which could be fixed by a
retry. And that's exactly what we should do.

See man:SSL_accept(3) for details.

This is a follow-up for ca691fe91ecea6fe7f091ce5680ad7f0a1a020ce
which did the same for SSL_connect, SSL_read and SSL_write.

Ticket: CFE-3066
Changelog: cf-serverd now tries to accept connection multiple times

### DIFF
--- a/cf-serverd/server_tls.c
+++ b/cf-serverd/server_tls.c
@@ -46,6 +46,7 @@
 
 static SSL_CTX *SSLSERVERCONTEXT = NULL;
 
+#define MAX_ACCEPT_RETRIES 5
 
 /**
  * @param[in]  priv_key private key to use (or %NULL to use the global PRIVKEY)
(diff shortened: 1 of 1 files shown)


vpodzime:
jenkins: [Build Status](https://ci.cfengine.com/job/pr-pipeline/3029/)

vpodzime:
green *bootstap-tests* build: [Build Status](https://ci.cfengine.com/job/bootstrap-tests/208/)

PR #3905, opened 2019-10-22
CFE-1655: process-related promises via '/proc'

tdlee:
Process-related promises have hitherto been implemented by invoking the
'ps' command and attempting to parse its output.  But 'ps' output
is highly human-oriented, and often not easy to machine-parse reliably.
It is also highly OS-variable.

CFE-1655 had initially proposed the idea of using "/proc" where available.
My own encounter with the flaws in the existing mechanism is at CFE-3020.

Ideally, this code would be chosen whenever "/proc" is available, and an
autotools "HAVE_PS_VIA_PROC" variable is used to manage this opt-in.

In this first implementation (commit), I have limited it to Linux
(my development platforms have been RHEL6 and RHEL7) and this extra
check, which would ideally be removed in due course, is governed
by a clause in "configure.ac".

This first commit passes most "tests/acceptance/05_processes/01_matching"
tests, including two recently added ones.  It currently fails two
existing ones, so needs a little further work.

There may also be a little more work needed to cover the case of
a process exiting during the small but finite period while the
information about it is being gathered.


tdlee:
I think this is now ready for serious human review.

Quoting from the original commit and PR text (which are identical modulo one trivial typo correction):

> This first commit passes most "tests/acceptance/05_processes/01_matching"
> tests, including two recently added ones.  It currently fails two
> existing ones, so needs a little further work.

It now passes all the "05_processes/01_matching/"  tests except one. That one ("stime_range.cf") is inappropriate in this '/proc'-based context.  It uniquely constructs its own "ps" command and tries to parse the output.  We need to arrange either for this test to be skipped in this '/proc' context or for a replacement (presumably designed by the maintainer of the existing test).

There are two "travis" reports.

The valgrind test mentions a possibly uninitialised value in "LoadProcStat ()".  This is the 'uid' variable.  This code is not reached unless 'found' is true, and this is itself dependent on having initialised 'uid' in the preceding loop.  (If there was a known never-valid value for uid, then this could be used instead for controlling the flow.)

> Ideally, this code would be chosen whenever "/proc" is available, and an
> autotools "HAVE_PS_VIA_PROC" variable is used to manage this opt-in.
> In this first implementation (commit), I have limited it to Linux
> (my development platforms have been RHEL6 and RHEL7) and this extra
> check, which would ideally be removed in due course, is governed
> by a clause in "configure.ac".

That "HAVE_PS_VIA_PROC" gives rise to the second "travis" report.  I'm happy to take advice on this if there is a fix I need to implement.

Finally, because we now have two '.c'  files, a case can be made for renaming the original file 'processes_select.c' as 'processes_select_ps.c' or 'processes_select_pscmd.c' to balance the new 'processes_select_proc.c' (and, of course, related adjustments to 'Makefile' entries, etc.  Again, if you think this is a good thing, let me know and I'll do it.

libpromises/processes_select_proc.c
```diff
@@ -0,0 +1,809 @@
...
+     *    pid (some text here) S n1 n2 n3 ...
+     * We already know the pid. And the text is a variant of command (known).
+     * We are interested in the state char 'S' and numbers n1, n2, n3, ...
+     */
+
+    /* Find closing ')' and place pointer on the char 'S'. */
+    char *pt;
+    pt = strrchr(statbuf, ')');
+    pt++;
+    pt++;
```
olehermanse: Here, you need to consider the worst case scenario(s):

- What if `strrchr()` didn't find a `)`? It will return `NULL`, you increment it twice, you now have a pointer to `0x000002`, which is (most likely) in kernel memory and will cause a segmentation fault (crash) when you try to use it.
- What if `)` is the last character `read()` from the file? You increment twice and `pt` is now pointing to the uninitialized part of `statbuf`.
- What if you `read()` exactly `CF_MAXVARSIZE - 1` characters from file, and the last one was a `)`? `pt` points to the last byte in your buffer, you increment twice, and it now points to something else on the stack.

You need to check that `strrchr` doesn't return NULL, and you need to make sure that your increments don't exceed `statbuf + len`, something like:

```
char *pt = strrchr(statbuf, ')');
if (pt == NULL)
{
  // Could not find closing parenthesis
  // Log some error?
  // return NULL; ?
}
size_t parenthesis_offset = pt - statbuf; // Length from statbuf start to closing parenthesis
if ((parenthesis_offset + 2) >= len)
{
  // After adding 2 to pt, it will point to uninitialized memory, 
  // either a part of statbuf that wasn't read from file or outside of statbuf
  // log error?
  // return NULL?
}
```

libpromises/processes_select_proc.c
```diff
@@ -0,0 +1,809 @@
...
+      "%d %d %*d %d %*d %*u "
+      "%*u %*u %*u %*u "
+      "%lu %lu %*d %*d "
+      "%ld %ld %ld %*d "
+      "%llu %lu %ld %*d ",
+      &pstate[0],
+      &ppid, &pgid, &ttyn,
+      &utime, &stime,
+      &priority, &nice, &threads,
+      &starttime, &vsize, &rss);
```
olehermanse: Taken from travis valgrind output:

```
==2109== Use of uninitialised value of size 8
==2109==    at 0x66D4550: rawmemchr (rawmemchr.S:25)
==2109==    by 0x66BAFF1: _IO_str_init_static_internal (strops.c:41)
==2109==    by 0x66AE266: vsscanf (iovsscanf.c:40)
==2109==    by 0x66A8916: sscanf (sscanf.c:32)
==2109==    by 0x4ED388C: LoadProcStat (processes_select_proc.c:679)
==2109==    by 0x4ED4335: LoadProcessTable (processes_select_proc.c:788)
==2109==    by 0x40D6AA: KeepAgentPromise (cf-agent.c:1667)
==2109==    by 0x4EA9293: ExpandPromiseAndDo (expand.c:215)
==2109==    by 0x4EA9293: ExpandPromise (expand.c:283)
==2109==    by 0x40DF6A: ScheduleAgentOperations (cf-agent.c:1414)
==2109==    by 0x41AD8D: VerifyMethod (verify_methods.c:172)
==2109==    by 0x41B13C: VerifyMethodsPromise (verify_methods.c:74)
==2109==    by 0x40D965: KeepAgentPromise (cf-agent.c:1729)
==2109==  Uninitialised value was created by a stack allocation
==2109==    at 0x4ED3560: LoadProcStat (processes_select_proc.c:616)
```

At this point, it's a little confusing to see what's happening, but I think this is because `pt` points to an unitialized part of statbuf (which is stack allocated in this funcion, LoadProcStat()), so should be fixed when you address the comments above.

libpromises/processes_select_proc.c
```diff
@@ -0,0 +1,809 @@
...
+    // a "shouldn't happen" error
+    if (!found)
+    {
+        Log(LOG_LEVEL_ERR, "error searching for uid %d in %s",
+          uid, statusfile);
+
+        return false;
+    }
+
+    JsonObjectAppendInteger(pdata, JPROC_KEY_UID, uid);
```
olehermanse: LGTM (code analysis) has an alert on this line, saying:

```
The variable uid may not be initialized here.
```

You say that it should be, because of the relationship with the `found` boolean variable, however there are problems. (I'll comment on relevant lines above).

libpromises/processes_select_proc.c
```diff
@@ -0,0 +1,809 @@
...
+    char key[CF_MAXVARSIZE];
+    while ((nread = getline(&line, &len, stream)) != -1)
+    {
+        sscanf(line, "%s %d ", key, &uid);
+        if (strcasecmp("Uid:", key) == 0)
+        {
+            found = true;
+            break;
+        }
+    }
```
olehermanse: `uid` starts off uninitialized, and you want to assure that this loop:

- assigns to `uid` and sets `found` to `true`.
- doesn't assign to `uid` and leaves `found` as false.

However, you haven't guaranteed these 2. Here is a relevant part of `man sscanf`:

```
These functions return the number of input items assigned. 
This can be fewer than provided for, or even zero, in the event of a matching failure.
```

So, if the `line` is something weird, like `"Uid: notanumber"`, or simply `"Uid:"` `key` will be assigned to, but `uid` will not be altered. The comparison with `key` will be true, and the `found` variable will be set to `true`, while `uid` remains uninitialized.

To fix this, you should check that `sscanf()` returns `2`, which means it matched / assigned to both variables.

**Please note** that it's good that you didn't initialize `uid`. If you had set it to `0`, `-1` or something else, the tools couldn't help us find problems like this. (We currently use a combination of tools to catch these mistakes; gcc / clang compiler warnings, ASAN (address sanitizer instrumentation), valgrind runtime checking, LGTM (code analysis).

libpromises/processes_select_proc.c
```diff
@@ -0,0 +1,809 @@
...
+    /*
+     * Big picture: "stat" is:
+     *    pid (some text here) S n1 n2 n3 ...
+     * We already know the pid. And the text is a variant of command (known).
+     * We are interested in the state char 'S' and numbers n1, n2, n3, ...
+     */
+
+    /* Find closing ')' and place pointer on the char 'S'. */
+    char *pt;
+    pt = strrchr(statbuf, ')');
```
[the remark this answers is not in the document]
tdlee: Thanks.  Part of 'eb874bd' below.

libpromises/processes_select_proc.c
```diff
@@ -0,0 +1,809 @@
...
+        return pdata;
+    }
+
+    /* read the 'stat' file into a buffer */
+    char statbuf[CF_MAXVARSIZE];
+    len = read(fd, statbuf, CF_MAXVARSIZE -1);
+    close(fd);
+    if (len <=0) {
+        return pdata;
+    }
```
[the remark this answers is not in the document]
tdlee: Thanks.  Part of 'eb874bd' below.  (Also tidies up the untidy "<=0" with a space.)

libpromises/processes_select_proc.c
tdlee: Thanks!  This caused me to look beyond the se details to the bigger picture.  The real parsing of this line is in the subsequent 'sscanf()'.  The '/proc' man page indicates that that the whole line is intended to be scanf-parseable.  This means that my 'strrchr()' to skip the early part can be subsumed into that 'sscanf()' which eliminates this part of my code altogether.  See 'eb874bd' below.

libpromises/processes_select_proc.c
tdlee: Thanks.  See next point.

libpromises/processes_select_proc.c
tdlee: Good points.  The code now checks that sscanf return count.

tdlee:
> The general idea of parsing proc when available, is good. In fact having code that parses proc to JSON, could be useful for other projects, so I think we should add / move some of this to libntech. (However, if you prefer to work in core repository only, this is something we can do later.)

When I began drafting this, it quickly became apparent that the existing JSON library would be a very useful way to collect and manage the data from the various sources and to allow queries on it from the higher-level CFEngine code.  So, yes, relocating 'LoadProcessTable()' etc. to a more general purpose library could make eminent sense.  But that refactoring would be of much wider scope than this PR and would require many other considerations about the overall API of such a library.  Whereas I am happy to be the owner of PR 3905, shouldn't this subsequent potential refactoring be primarily owned by Northern.tech?

If you open a 'mender' issue (or similar) for that refactoring I would be most interested to be an observer and, if appropriate, a minor participant or contributor.

> I will get back to review the whole PR, when I have time. Writing parsers is difficult and error prone, it will require a thorough review...

You'll see that my later commits eliminate a significant part of my earlier parsing.

> ... and I also think you should add some unit tests for parts of the code. (Unit tests in CFEngine test indiviual C functions, instead of evaluating a complete policy file). Here is an example of an existsing unit test: https://github.com/cfengine/core/blob/master/tests/unit/conversion_test.c

OK. I'll take a look.  I'll see what unit tests already exist for the ps-command version and see whether I can do something to an equivalent level.

olehermanse:
> When I began drafting this, it quickly became apparent that the existing JSON library would be a very useful way to collect and manage the data from the various sources and to allow queries on it from the higher-level CFEngine code. So, yes, relocating 'LoadProcessTable()' etc. to a more general purpose library could make eminent sense. But that refactoring would be of much wider scope than this PR and would require many other considerations about the overall API of such a library. Whereas I am happy to be the owner of PR 3905, shouldn't this subsequent potential refactoring be primarily owned by Northern.tech?

Another approach, which I would use, is to develop the new feature (/proc parsing only), in a library approach. (i.e. add the library functions to libntech, add unit tests, and make sure it works well, then introduce the new feature in cfengine afterwards, utilizing the functionality you built in libntech. This is how I would do it, but as I said before, since you're the one driving this, it's up to you. If you want to just develop the functionality here, in core, you are free to do so. We might refactor / move your code to libntech after it is merged.

> If you open a 'mender' issue (or similar) for that refactoring I would be most interested to be an observer and, if appropriate, a minor participant or contributor.

I will make sure to ping you if / when we do this.

configure.ac
```diff
@@ -1310,6 +1310,18 @@ case "$target_os" in
...
 esac
 
+dnl ######################################################################
+dnl Should "processes_select" use "/proc/<nnn>" rather than "ps" command?
+dnl New at March/2019.  Developing on RHEL6.  Eventually aim to use widely.
+dnl ######################################################################
+
+case "$target_os" in
+   linux*)
+        AM_CONDITIONAL([HAVE_PS_VIA_PROC], [test -f "/proc/1/stat"])
```
olehermanse: This fails the OS X build in travis:

```
checking that generated files are newer than configure... done
configure: error: conditional "HAVE_PS_VIA_PROC" was never defined.
Usually this means the macro was only invoked conditionally.
The command "./travis-scripts/script.sh" exited with 1.
```

There's many ways to address this, but I think the easiest / most appropriate is to just remove the linux restriction. your `test` command should already check whether `/proc` exists or not.

tdlee: My reason for that temporary working was conservatism while this work is in preparation.  I fully and enthusiastically agree that it should eventually go.

Various OSes also offer "/proc".  But a glance at their man pages (and existing other corners of CFEngine code) suggest that the new code may require some "#ifdef" work before it can go live on such OSes.

If I were to remove that Linux restriction right now, there is a significant chance that new builds on such "/proc"-based non-Linux systems would fail.

So... I'm happy to go with your call.  Do we temporarily keep the Linux restriction and keep the code safe for untested /proc OSes?  Or do we remove the restriction and risk failures on such OSes?

tdlee: Oh!... Could I simply add a default '`*)`' case for this interim period?  Like:

> case "$target_os" in
>    linux*)
>         AM_CONDITIONAL([HAVE_PS_VIA_PROC], [test -f "/proc/1/stat"])
>         ;;
>    *)
>         AM_CONDITIONAL([HAVE_PS_VIA_PROC], false)
>         ;;
> esac

with a comment that the 'case' scaffolding is temporary, to be removed in due course, leaving only the one line "`AM_CONDITIONAL(..., [test...])`"?

tdlee:
(Working note: I need to address that new "LoadBootTime()" thing flagged up by LGTM.)

configure.ac
olehermanse: > Oh!... Could I simply add a default '*)' case for this interim period? Like:

@tdlee Yes, I believe that is a way to do it. It essentially needs to be defined for  all branches (all systems) as either true or false. There might be better / cleaner ways to to do it, but my knowledge of m4/autotools/sh is limited.

tdlee: That is now place at commit '0e49619a' below, including a comment that the 'case' scaffolding is a transitional arrangement.

tdlee:
That's looking better.  Thanks for your help!

- Unit tests: The most recent significant commit (`f4bb523`) now applies all existing unit tests to the new code, so the new code gets just as much coverage as the established code.  Because the finally addressed test here, "`linux_process_test`", is now included, it is thus generic "/proc", so I have renamed it "`proc_process_test`".  This was also the final vestigial user of "`process_linux.c`" (whose functionality is now in the new code), so that has been removed.

- API into code: A nice side-effect of making the unit tests work was that it encouraged the splitting of the new code into two files either side of a JSON-returning API onto "/proc".  This should give you a good starting point for the proposed migration of some of it into "`libntech`".

- I suspect there may be an issue with acceptance test "`05_processes/01_matching/stime_range.cf`".  During development, I think it has always passed here on Githuib/travis, but has varied between failing and succeeding on my system.  Alas, that test is almost impenetrably hard to decipher, so I am unable to tell whether it is a weakness in my code or a flaw in the test itself.  Might you persuade the assigned owner of that test to rework it into something clearer?

- The related more general "`05_processes/01_matching/`" topic is report CFE-3025 that I had opened back in March; indeed, I contributed two features to that which were recently accepted and merged.

tdlee:
Is there access to a non-Linux machine for me to try?

It would be good to demonstrate this "/proc" and JSON-returning API for a second OS.

And I also see that some other OSes seem to have programmatic interfaces onto the reading of process information (AIX:` getprocs()`; HPUX: `pstat_getproc()`; etc.).  So even where such a system offers "/proc" it might be preferable to use such functions.  It would be good also to demonstrate this via that API, including the necessary `configure.ac` to determine the presence of those functions.

tdlee:
I've just seen "`cf-monitord/mon_processes.c`".  Although not essential, we might want to include that in our plans.

tdlee:
> I've just seen "cf-monitord/mon_processes.c".  Although not essential, we might want to include that in our plans.

That is now in place.

Its first results were slightly inconsistent with the non-"/proc" results.  This was because the "ps" command outputs the username based on the process's effective-uid, whereas this new "/proc" code had been drafted using the real-uid. So I have adjusted the new code to be euid.

This also adds a new function "`FetchProcessTable()`" to the embryonic API.

tdlee:
From my perspective I think this is reasonably stable.

It provides the neat split between the OS-dependent "gather JSON representation" and the OS-independent "use JSON representation" that might be useful for your proposed `libntech` component, making a JSON-API framework.

The one major thing I would really like to try is to code up a second OS: one that is non-Linux.  Unfortunately I don't have ready access to anything other than RHEL6/7.  I see that Github's invocation of travis includes a 'darwin'.  Does my GitHub account give me the ability to code it up there? Alternatively is there a willing volunteer on a non-Linux system who might try this?  (Might we ask on the 'dev' email list?)

One detail I have just spotted that I need to address is that the fail-return (and logging) from my `GetProcessStartTime()` and `GetProcessState()` isn't quite the same as in the ps-cmd implementation. I'll try to resolve that detail soon.  Feel free to set this as a "change requested" item so that we can keep track of it.

A thought.  Might it be possible to make some minimal additions to the existing ps-cmd code to include a JSON returning interface?  Then that, too, could align with the JSON-API framework.

configure.ac
tdlee: I'm still turning over ideas to make this unsatifactory scaffolding into something cleaner!

At the heart of the matter are:

- the "/proc" implementations seems to vary significantly across systems

- we don't yet have even embryonic implementations for any of the non-Linux systems

- there may be OSes which don't use "/proc" but do offer a neat programming interface, and which it would be good for our autoconf-etc. structure to encompass

tdlee: See commit 'aedbcda' below.  This gives a cleaner framework, capable of detecting whether or not the platform has a coded handler for its "/proc".

tdlee:
Hitherto, the "big picture" for the alternative to legacy `ps`-output parsing has been to use "/proc" if available.

This recent set of commits recognises that the "`/proc`" detail needs adjusting to be "the OS's interface to interrogate process information".  This allows inclusion OSes, such as FreeBSD (and other BSDs?)  which don't offer "/proc" but do offer something equivalent such as "`kvm_getprocs()`".

As a side-effect, the `configure.ac` scaffolding has actually become cleaner.

A BSD-driver file is in preparation (although not yet included here) that demonstrates this.

olehermanse:
Hello, @tdlee.

We'd very much like to get this functionality merged, but there is still quite a bit of work left here, so let's try to get the ball rolling again;

- [ ] The Pull Request has merge conflicts now, please rebase on master and resolve the conflicts.
- [ ] It looks like there are still some failures in acceptance tests, all acceptance tests need to be green on all our tested platforms before we can merge it.
- [ ] Some commits are fixing things introduced in one of your earlier commits, it would make it a lot easier to review if you could _squash_ these commits together, using an interactive rebase.
- [ ] The Pull Request is quite large, it would be helpful for us if you try to split it up into multiple PRs, if possible.

All of these points are explained in the contribution guidelines:

https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#checklist

I know splitting 1 feature into multiple PRs might seem unintuitive, but here is a suggestion:

1. Introduce 1 PR with just the "skeleton", i.e. the necessary autotools changes, and empty functions which do nothing and print errors that they are not implemented yet. Ensure proc parsing is not default anywhere for now, this is tested for "free" in this case, because we already have tests for processes, which would fail with the not yet implemented functions.
2. Add the core parts, the parsing functions, with detailed unit tests.
3. Start using the core parts for procs parsing in processess promises, tying together 1. and 2.
4. Start enabling proc parsing by default on some platforms where it makes sense.

I understand that this is some extra work, especially if you are not accustomed to using the more advanced git functionality to move changes around, but we are really trying to optimize for making things as easy to review as possible, lowering the chance of us missing severe problems. The 4 steps above are just a suggestion, and not a requirement in any way, but it would help us in reviewing and merging this in a safer way.

Sorry for the delay, I've started reviewing this at least a couple of times, but there is just so much going on, and it would take me a long time to review, understand the code, write meaningful feedback, try to understand the test failures, add my own fixes for problems I see, etc. We can't merge code without reviewing and understanding it, and being fairly certain it doesn't introduce a lot of problems for other users.

tdlee:
Apologies for the delay in replying.  (It's a pity we couldn't have kept the momentum going back in Oct/Nov.  But we are where we are.)

Thanks for the reply.

Could you clarify a fundamental principle in your request for multiple PRs?  Is the request effectively saying to abandon this PR (3905)?  And then to start the PR process again from scratch in the four step process you mention?  And that those multiple PRs should be against current master?  (Or putting it another way: rebase my work against current master, then submit multiple fresh PRs?)

Thanks.

olehermanse:
> Could you clarify a fundamental principle in your request for multiple PRs? Is the request effectively saying to abandon this PR (3905)? And then to start the PR process again from scratch in the four step process you mention? And that those multiple PRs should be against current master? (Or putting it another way: rebase my work against current master, then submit multiple fresh PRs?)

@tdlee Splitting it up into 4 PRs was not a requirement, just a suggestion to help move this along as quickly as possible.

The size of a Pull Requst (in terms of number of commits, lines of code changed and complexity) affects:

* How time consuming and challenging it is for us to review
  * If a PR takes hours or days to review, it is very hard for me to set aside that much time, and I'll have to delay reviewing it.
  * Increases the likelihood of us missing mistakes / bugs during a long review
* Number of errors introduced that you need to fix before we can merge
* Number of iterations we have to go back and forth (us asking questions or requesting changes, you addressing our requests)

All of these points mean it will take, days, weeks, or months before we can get a PR ready to merge. Then it will be out of date, and might require resolving conflicts, re-testing, re-reviewing, etc. Breaking the feature up into multiple PRs, so we can focus on fewer commits/changes at a time, getting them properly reviewed and merged, and then we can "forget" about them and move on to the next step.

Small, incremental, non-breaking changes are almost always better.

olehermanse:
Closing due to inactivity. Feel free to open a new PR if you get back to this.

PR #3909, opened 2019-10-22
address a 'dereferenced without an explicit null-check' issue

tdlee:
While working on PR 3905 (CFE-1655) a potential code fault was highlighted: multiple occurrences of a pointer argument not being null-checked in "SelectProcess()".  But this same fault was also present in the original code.  This fix addresses that original fault in the same way that it is already addressed in the nearby "SelectProcesses()".

### DIFF
--- a/libpromises/processes_select.c
+++ b/libpromises/processes_select.c
@@ -148,6 +148,7 @@ static bool SelectProcess(const char *procentry,
     Rlist *rp;
 
     assert(process_regex);
+    assert(a != NULL);
 
     StringSet *process_select_attributes = StringSetNew();
 


olehermanse:
@tdlee Thank you for addressing this alert. This looks good, but the two commits should be squashed into one. Are you able to squash them locally and force push, so the GitHub PR just shows 1 commit? (We generally recommend users do this themself, so they can control the commit message, and keep their GPG signature etc.) Otherwise, I am able to do it via the GitHub UI, so just tell me.

tdlee:
My preference was for the first commit: the 'assert' very early in the function.  During testing of something else, I got a suspicion that it might have been too early: that a NULL pointer might be legitimate, under some circumstances that cause an early return before it is used.  (I now suspect that earlier suspicion might have been false!) So, for safety I moved the assert down to near the first use of the pointer.

Part of me still would prefer that 'assert' to be early.

Could you advise?  If you are happy with the first commit ('assert' early in function), cherry-pick it as the complete PR. If you prefer the second place ('assert' lower down), could you give me the set of git commands to squash the commits, force-push the result, etc.?

olehermanse:
> My preference was for the first commit: the 'assert' very early in the function. During testing of something else, I got a suspicion that it might have been too early: that a NULL pointer might be legitimate, under some circumstances that cause an early return before it is used. (I now suspect that earlier suspicion might have been false!) So, for safety I moved the assert down to near the first use of the pointer.

> Part of me still would prefer that 'assert' to be early.

@tdlee I agree. Your line of thinking is sound, and I would think the same. On the one hand you don't want to break anything by adding an assert too early, on the other hand, it seems like it makes more sense to assert early, to try to catch any programming mistakes.

So. The current state is easy to merge, it's a conservative change. If you want this merged today, we should keep it as is, just squashing the commits. The necessary commands are (roughly):

```
$ git checkout ps-tidy-up
$ git rebase -i master
```

(Assuming master is a somewhat recent clone of our master, with no extra commits).

This will open an editor, for example vim or emacs, with something like this:

```
pick ac07e6d27 Moved Protocol version parsing to protocol.c
pick 001bd420b Network protocol version 2 is now called "tls"

# Rebase 36379afcb..001bd420b onto 36379afcb (2 commands)
#
# Commands:
# p, pick <commit> = use commit
# r, reword <commit> = use commit, but edit the commit message
# e, edit <commit> = use commit, but stop for amending
# s, squash <commit> = use commit, but meld into previous commit
# f, fixup <commit> = like "squash", but discard this commit's log message
# x, exec <command> = run command (the rest of the line) using shell
# b, break = stop here (continue rebase later with 'git rebase --continue')
# d, drop <commit> = remove commit
# l, label <label> = label current HEAD with a name
# t, reset <label> = reset HEAD to a label
# m, merge [-C <commit> | -c <commit>] <label> [# <oneline>]
# .       create a merge commit using the original merge commit's
# .       message (or the oneline, if no original merge commit was
# .       specified). Use -c <commit> to reword the commit message.
#
# These lines can be re-ordered; they are executed from top to bottom.
#
# If you remove a line here THAT COMMIT WILL BE LOST.
#
# However, if you remove everything, the rebase will be aborted.
#
# Note that empty commits are commented out
```

( I used my branch so the commit names and SHAs are different ).

On the second line, change `pick` to `squash` or `s`. (This will squash/combine the commit into the one above it). Save and quit.

An editor will open allowing you to change the commit message. Write one that makes sense for the final commit.

**However:** If you want to have the assert early instead, you can instead add `drop`, `d`, or simply remove the line while rebasing above. Then, I can run it through our test infrastructure, and see if it looks good. (We are currently upgrading it, so it will take a couple of days, then).

Once done rebasing, if you have your remotes and upstream set up correctly, it should be enough to do:

```
$ git push
# Review that the remote / branch and information look correct
$ git push --force
```

If that doesn't work (no upstream branch), you can do:

```
$ git push --set-upstream origin ps-tidy-up
# Review that the remote / branch and information look correct
$ git push --set-upstream origin ps-tidy-up --force
```

(Here, `origin` is the name of your remote, your fork, so you might need to change that).

tdlee:
I've just added another commit, intended to revert to the original commit (the 'assert' early in the function):

- first commit: assert early
- second commit: assert later
- third commit: revert second, to become first

As a git beginner, I've never done any rebase, or squash operations.  But I seem to recall git documentation warning against them, especially when already pushed to other repos where they might already have been used onwards (as, for example, I presume would apply in this "pull request" here).

I really want to spend my limited time concentrating on PR 3905 (_CFE-1655: process-related promises via '/proc'_).  This detail on a single line in the existing 'ps'-method code is a distraction from that; its importance to 3905 is that my work there should functionally mirror the outcome here.

Could I leave this one-line 'ps'-method adjustment to you, to decide the best way?

Thanks.

olehermanse:
> As a git beginner, I've never done any rebase, or squash operations. But I seem to recall git documentation warning against them, especially when already pushed to other repos where they might already have been used onwards (as, for example, I presume would apply in this "pull request" here).

@tdlee Rebasing is normal, and sometimes necessary to keep the commit history tidy. We never rebase / amend / do anything destructive to the main branches ( like `master`, `3.12.x`, `3.10.x` and so on ), but for a PR branch in a personal fork, it is completely okay to rebase and force push. 

> I really want to spend my limited time concentrating on PR 3905 (CFE-1655: process-related promises via '/proc'). This detail on a single line in the existing 'ps'-method code is a distraction from that; its importance to 3905 is that my work there should functionally mirror the outcome here.

> Could I leave this one-line 'ps'-method adjustment to you, to decide the best way?

Yes, that's okay. I'll run some tests, and squash and merge this myself, no worries.

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4004/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4004/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4004/

tdlee:
Many thanks!

PR #3908, opened 2019-10-22
CFE-3177: CONTRIBUTING.md should include a suggested git workflow

tdlee:

PR #3910, opened 2019-10-23
CFE-3177 - CONTRIBUTING.md: More detailed git workflow explanation

olehermanse:
Follow-up from: https://github.com/cfengine/core/pull/3908

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,64 +22,8 @@ Merged features and larger changes will be released in the first minor release
 [the scheduled release date](https://cfengine.com/product/supported-versions/)
 in order to make it to the first minor release.
 
-
-## A possible workflow
-
-If you are already comfortable with Github, you may skip this section.
-
-This is a suggested outline for those unfamiliar with working with Github.
-It does, however, assume a familiarity with the basics of using git
-so it does not attempt to provide every detail of that.
-
-You will be working with three git repositories: the central one on Github,
-your fork on Github and your local clone.
-
-Within Github, make your fork of the central CFEngine repository
-using the fork icon on the
-[Github CFEngine main page](https://github.com/cfengine/core).
-Your fork lives at `https://github.com/my-githud-id/cfengine-core`.
-
-Make your own local clone from the main page.
-Since the middle of 2019 the source code uses a git 'submodule'
-which requires the use of the `--recursive` flag.
-
-```
-git clone --recursive https://github.com/cfengine/core
-```
-
-In your local clone, add a reference to your fork on Github:
-
-```
-git remote add name-for-my-fork https://github.com/my-github-id/cfengine-core
-
-```
-
-Never work in the `master` branch; always use a separate branch for your topic.
-Prepare your topic on this topic-focussed branch.
-When it is ready, push it to your Github fork.
-
-```
-git checkout master # ensure you start from a good reference point
-git checkout mytopic
-./autogen.sh
-edit; check; git add; git commit; etc.
-git push name-for-my-fork mytopic
-```
-
-On your fork at Github you should now see an additional branch `mytopic`.
-This is what you use for the pull request there.
-
-### Occasional housekeeping
-
-From time to time, update your local clone from the central repository
-as a reasonably current baseline,
-and likewise push those changes to your fork:
-
-```
-git checkout master
-git pull origin master
-git push name-for-my-fork master
-```
+If you are not very familiar with Git or GitHub, you can look at our
+[suggested workflow](#suggested-git--github-workflow), later in this document.
 
 
 ## Pull Requests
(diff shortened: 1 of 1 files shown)


olehermanse:
@tdlee Feel free to review this.

CONTRIBUTING.md
```diff
@@ -800,45 +744,280 @@ Try restricting ifdefs in the header files, or in the beginning of
...
+commits:
+
+```
+pick ac07e6d27 Added new policy function isrealdir()
+pick 001bd420b Fixed mistake in isrealdir()
+
+# Commented out instructions ....
+```
+
+The commits / commands will be applied to `upstream/master` from top to bottom
```
vpodzime: I don't think `git rebase -i upstream/master` is really a good example. It mixes too many things. `git rebase -i HEAD~N` or something similar is much easier to understand, AFAICT.

For example the above *"commits/commands will be applied to `upstream/master`"* is not quite true, right? They are applied to the local branch which is being rebased to `upstream/master`.

CONTRIBUTING.md
```diff
@@ -800,45 +744,280 @@ Try restricting ifdefs in the header files, or in the beginning of
...
+`git` has some commands for working with submodules, for example:
+
+```
+$ git submodule update
+```
+
+This will update the submodule, checking out the correct commit for your
+current `HEAD`.
+
+Please note that a submodule is just a normal git repo, with the `SHA`
```
vpodzime: I wish this was true. Working in the submodule has proven to be impossible for me. I had to create a separate clone of the repo and work there. In particular, the submodule doesn't have its own `.git/config` and many things work in different ways.

CONTRIBUTING.md
olehermanse: > I don't think `git rebase -i upstream/master` is really a good example. It mixes too many things. `git rebase -i HEAD~N` or something similar is much easier to understand, AFAICT.

I disagree. I only mentioned HEAD in one other place, and I was considering to not do it. `HEAD~N` is even more confusing to me, then you have to explain what HEAD means, what `~` means, etc. `upstream/master` is discussed in other places, and it also has another advantage: you can use the same command in (more than) 3 different situations:

1) Your PR is out of date and you need to bring it up to date with master (upstream)
2) Bringing local master branch up to date (with upstream)
3) Adding `-i` option to interactively rebase (for squashing commits etc.)

Yes, teaching and explaining these things is difficult.

> For example the above "commits/commands will be applied to upstream/master" is not quite true, right? They are applied to the local branch which is being rebased to upstream/master.

This depends on how you interpret `apply`. I think it's right, as it says there, but I also understand that it can be confusing. Wording these things correctly and in an understandable way is hard. Something like `upstream/master is used as a base (starting point), then the commits in the list are applied (without altering upstream/master)` is how I read it.

CONTRIBUTING.md
olehermanse: I have no problems working inside the submodule. I never interact with `.git/config` directly, but the config options I set using `git config` work just fine. Everything I've described in this guide, I've been able to do inside the submodule, in both this project and other projects which use submodules.

CONTRIBUTING.md
olehermanse: @vpodzime Also, we should assume people to some extent understand git. I think of this as some sort of reference / command lookup. If you don't understand how rebase works or exactly what `upstream/master` you can google that.

olehermanse: I updated the wording to try to be a little more clear.

CONTRIBUTING.md
olehermanse: Updated the wording here as well.

olehermanse: Updated the wording here as well.

CONTRIBUTING.md
```diff
@@ -800,45 +744,280 @@ Try restricting ifdefs in the header files, or in the beginning of
...
+URLs instead, which might be more convenient).
+
+Replace `your-username` with your GitHub username, here and later.
+`--recursive` is needed to clone submodules, without it, you would have to run
+these commands separately:
+
+```
+$ git submodule init
+$ git submodule update
+```
```
nickanderson: very nice explination

CONTRIBUTING.md
```diff
@@ -800,45 +744,280 @@ Try restricting ifdefs in the header files, or in the beginning of
...
+
+### Branching
+
+Never work in the `master` branch; always use a separate branch for your topic.
+Prepare your topic on this topic-focussed branch.
+When it is ready, push it to your Github fork.
+
+```
+$ git checkout master
+$ git checkout -b mytopic
```
nickanderson: Perhaps `git checkout -b mytopic upstream/master`

CONTRIBUTING.md
```diff
@@ -800,45 +744,280 @@ Try restricting ifdefs in the header files, or in the beginning of
...
+Locally, it is usually sufficient to run unit tests, and select acceptance
+tests. Then, commit and push normally:
+
 ```
-$ git pull --rebase upstream master
-From https://github.com/cfengine/masterfiles
- * branch                master     -> FETCH_HEAD
-First, rewinding head to replay your work on top of it...
-Applying: ENT-3329: Allow hubs to collect from themselves over loopback
+$ git add -p
```
nickanderson: oh fancy suggesting partial adds, but one of my favorite git features

CONTRIBUTING.md
```diff
@@ -800,45 +744,280 @@ Try restricting ifdefs in the header files, or in the beginning of
...
+$ git fetch --all
+```
+
+Apply changes from the `upstream` `master` branch to our local `master` branch:
+
+```
+$ git rebase upstream/master
+```
+
+(You can also use `git pull`, but I find `rebase` a little easier when there
```
nickanderson: maybe `git pull --rebase`

CONTRIBUTING.md
```diff
@@ -800,45 +744,280 @@ Try restricting ifdefs in the header files, or in the beginning of
...
+### Forking
 
+You will be working with three git repositories: the central one on Github,
+your fork on Github and your local clone.
 
-#### How to rebase a branch
+Within Github, make your fork of the central CFEngine repository
+using the fork icon on the
+[Github CFEngine repository page](https://github.com/cfengine/core).
+Your fork lives at `https://github.com/your-username/core`.
```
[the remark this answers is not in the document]
olehermanse: Added: https://github.com/cfengine/core/pull/3911

CONTRIBUTING.md
olehermanse: I don't think this makes it easier to understand or read, and this isn't really a comprehensive git tutorial, so I'll leave it as is.

CONTRIBUTING.md
olehermanse: Sure. But I still think it's easier to just learn `fetch` + `rebase`, and after @jimis taught me long ago, I almost never use `pull` or `merge` commands. (This is of course very subjective, there's no right answer, I just tried to show my preferred approach, both now and when I learned git).

tdlee:
Thanks for taking this on!  Much appreciated.

Has this become too complex?

My original intention is that this top-level "git workflow" documentation should give a brief, simple outline to those of us who, at best, paddle in the gentle shallows of git; who have not pushed out into the riptides, whirlpools and wild currents of advanced git.  This is what I attempted to achieve in my original, consciously avoiding anything that was unnecessary for us beginners.

I can see why you might wish to include this more advanced material, including some of the advanced material that escaped from the advanced section into the implicitly non-advanced text.

So are we heading to the point where this really needs to be a separate document?  And even within that document to have a clear, maintained distinction between "basics", "intermediate" and "advanced"?

Just a thought...

olehermanse:
@tdlee You are of course welcome to submit pull requests to suggest changes. I think the current state of it is okay. There is an advanced section immediately following it (which we already had).

For the basic section, I decided to include things which most contributors will need to know. It has the parts needed to download, build/test make changes, and submit a PR. Then clean up commit history and keep PR up to date before it's merged. Yes, it's longer, but I believe for those who know git, it's pretty easy to skim through, look at the necessary commands and make sense of it. It shows the things you were missing (GitHub, forking, remotes, and submodules), and it also covers things we tell contributors to do frequently.

PR #3933, opened 2019-11-07
Fixed failure to build when source and build locations differ

tdlee:
The CFEngine code used to get this right, but a couple of issues have crept in
at some point.

Although most people build in the same directory-tree location as the source,
GNU autotools offers the ability for these locations to differ.
This is particularly useful when wanting to develop and test parallel builds
from a shared source location, for instance using different platforms and/or
different "--with-foo" options.

A typical use would be the following parallel builds:

    mkdir -p build/platform_1; cd build/platform_1; ../../autogen.sh
    mkdir -p build/platform_2; cd build/platform_2; ../../autogen.sh
    mkdir -p build/platform_3; cd build/platform_3; ../../autogen.sh

Subsequent development in the source can then be immediately tested
on those three platforms, in parallel, and with the previous cached 'make'
results within each still valid.  This is much, much faster!

This fix restores this capability.

There is one residual exception, outside CFEngine, but still easily fixable.
In the git submodule "libntech", file "libntech/tests/unit/csv_parser_test.c"
needs to use the "TESTDATADIR" macro in two places in a way very similar
to the nearby "json_test.c".  (I'm not quite sure how to submit that
particular fix as a pull-request, but can easily provide a 'diff' on request.)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -140,9 +140,9 @@ coverage: clean-coverage run-coverage collect-coverage
 BUILT_SOURCES = \
 	$(srcdir)/libpromises/enterprise_extension.c \
 	$(srcdir)/libpromises/enterprise_extension.h
-libpromises/enterprise_extension.c: libpromises/extensions_template.c.pre libpromises/enterprise_extension.sed
+$(srcdir)/libpromises/enterprise_extension.c: libpromises/extensions_template.c.pre libpromises/enterprise_extension.sed
 	$(V_SED) $(SED) -f $(srcdir)/libpromises/enterprise_extension.sed $< > $@
-libpromises/enterprise_extension.h: libpromises/extensions_template.h.pre libpromises/enterprise_extension.sed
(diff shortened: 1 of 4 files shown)


tdlee:
Note that the file changes can be done independently; this doesn't need to be an "atomic" commit.  If you wish to split them,"autogen.sh" should be done first, then "Makefile.am" then the others.  Finally the additional "libntech" item mentioned in the final paragraph above (which comes into play only at the "make check" stage).

olehermanse:
@cf-bottom please test this in Jenkins.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4739/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4739/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4739/

olehermanse:
Debain 9 failure unrelated.

PR #3934, opened 2019-11-08
Only use AI_NUMERICHOST for what looks like an IPv4 address

vpodzime:
Using AI_NUMERICHOST for getaddrinfo() is an optimization that
allows us to avoid potential long DNS queries when doing bind()
in cf-server. However, it cannot be used if the bind address is a
host name and not an IP address.

This change makes it being used only in cases where the bind
address looks like an IPv4 address. If it doesn't, getaddrinfo()
will take care of resolving things. Please note that this flag is
just a hint for getaddrinfo(), not using it never changes the
results (using it potentiall does).

Ticket: CFE-3190
Changelog: Allow using hostname in bindtoaddress

### DIFF
--- a/libcfnet/server_code.c
+++ b/libcfnet/server_code.c
@@ -7,6 +7,7 @@
 #include <systype.h>                    // CLASSTEXT
 #include <signals.h>                    // GetSignalPipe
 #include <cleanup.h>                    // DoCleanupAndExit
+#include <ctype.h>                      // isdigit
 
 /* Wait up to a minute for an in-coming connection.
  *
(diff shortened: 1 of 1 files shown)


vpodzime:

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4076/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4076/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4076/

libcfnet/server_code.c
```diff
@@ -7,6 +7,7 @@
 #include <systype.h>                    // CLASSTEXT
 #include <signals.h>                    // GetSignalPipe
 #include <cleanup.h>                    // DoCleanupAndExit
+#include <ctype.h>
```
olehermanse: Nitpick, but I think it's useful to add the comments for why you added the include, similar to lines above.

PR #3958, opened 2019-11-28
Mitigate race condition in the promiser_match_is_correct.cf test

vpodzime:
We spawn an agent process in a shell and then immediately check
that it is running. If it takes a bit to start (easy to happen on
a heavy-loaded machine running tests) the process is not
found. Let's give it a second to start before we try extracting
information about it.


vpodzime:
This requires a number of runs in Jenkins to see if it actually helps with the randomly failing test.

PR #3971, opened 2019-12-10
Debugging basename() SEGFAULT on Solaris 11

vpodzime:


olehermanse:
Thanks for adding the prints, helped me fix the issue;

https://github.com/cfengine/core/pull/3978

PR #3589, opened 2019-04-06
Add users support for FreeBSD

xaxo05:
Add users support for FreeBSD.

Use pw insread of useradd and usermod, user chpass instead of chpasswd, use rmuser instead of userdel. 

Implement fgetpwent() and fgetgrent() for FreeBSD.

Needs to be tested as I have only tested some cases myself for my personal needs.

PR #4002, opened 2020-01-03
FreeBSD: Changed ifdef guard to be OS-specific

olehermanse:
Based on review comment:

https://github.com/cfengine/core/pull/3589#discussion_r344640531


olehermanse:
Tested on FreeBSD 12.1 in AWS. Works.

PR #4016, opened 2020-01-16
ENT-5090: Removed (USE AT YOUR OWN RISK) from cf-key help menu for -x

olehermanse:
It is kind of implied that --force-removal can be destructive.
We don't have any specific reasoning for this phrasing, so it
is not so useful, just confusing / concerning for a user.


olehermanse:
Cherry-picks:

https://github.com/cfengine/core/pull/4020
https://github.com/cfengine/core/pull/4021

PR #4021, opened 2020-01-16
ENT-5090: Removed (USE AT YOUR OWN RISK) from cf-key help menu for -x (3.12)

olehermanse:
It is kind of implied that --force-removal can be destructive.
We don't have any specific reasoning for this phrasing, so it
is not so useful, just confusing / concerning for a user.


olehermanse:
> ACK. I don't think backports require re-review unless there are some significant conflicts.

I usually think of it as an "agree to backport" rather than a code review.

PR #4025, opened 2020-01-17
ENT-5304:Replaced determine-version.py with a manually updated file (3.15)

olehermanse:
Added determine-version.sh to replace it. A much simpler
solution which trusts the version number from the commited
file, but just adds a commit SHA if there is no tag.

This should get rid of all our problems with version mismatch.


olehermanse:
@cf-bottom try this in jenkins, please.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4430/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4430/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4430/

travis-scripts/script.sh
```diff
@@ -19,9 +19,7 @@ INSTDIR=$HOME/cf_install
 # fi
 
 # Unshallow the clone. Fetch the tags from upstream even if we are on a
```
Lex-2008: I wonder if we still need this. determine-version.py needed all tags to work, but new determine-version.sh needs only to check if the tip of the current branch points to a tag.

olehermanse: Yes, I was thinking the same. I can test later.

PR #4062, opened 2020-02-10
CFE-3198: Added macros before_version, at_version and after_version

olehermanse:
A final PR for the version detection macros. In some cases you want to specify just one version, or versions before or after (exclusive);

```
@if maximum_version(3.15.999) # Ugly
@if before_version(3.16) # More readable
```

```
# Something which was buggy in 3.15.4:
@if minimum_version(3.15.5)
@if after_version(3.15.4)
```

```
@if between_versions(3.15.0, 3.15.0) # Ugly
@if at_version(3.15.0) # More readable
```

I guess the last one missing is `not_version` (inverse of `at`), but I don't really see a case where it's useful, and you can acheive it using `at_version` + `else`.

### DIFF
--- a/libpromises/cf3lex.l
+++ b/libpromises/cf3lex.l
@@ -90,6 +90,9 @@ comment    #[^\n]*
 
 macro_if_minimum_version  @if\ minimum_version\([0-9]{1,5}(\.[0-9]{1,5}){0,2}\)
 macro_if_maximum_version  @if\ maximum_version\([0-9]{1,5}(\.[0-9]{1,5}){0,2}\)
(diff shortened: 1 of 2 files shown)


olehermanse:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4569/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4569/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4569/

vpodzime:
> Just to be conservative, I would not backport these without explicit reason that these provide a workaround for some specific issue.

If we don't backport them, they will be unusable until people stop using 3.15.x and 3.12.x

PR #4063, opened 2020-02-10
cf-remote: Fixed package selection for new package names

olehermanse:
Tested on Ubuntu 18, Ubuntu 16, CentOS 7, CentOS 6

### DIFF
--- a/contrib/cf-remote/cf_remote/remote.py
+++ b/contrib/cf-remote/cf_remote/remote.py
@@ -198,6 +198,23 @@ def install_host(
         release = releases.default
         if version:
             release = releases.pick_version(version)
+
+        if "os_release" in data and data["os_release"]:
+            distro = data["os_release"]["ID"]
+            major = data["os_release"]["VERSION_ID"].split(".")[0]
+            platform_tag = distro + major
+
+            # Add tags with version number first, to filter by them first:
+            tags.append(platform_tag) # Example: ubuntu16
+            if distro == "centos":
+                tags.append("el" + major)
+
+            # Then add more generic tags (lower priority):
+            tags.append(distro) # Example: ubuntu
+            if distro == "centos":
+                tags.append("rhel")
+                tags.append("el")
+
         artifacts = release.find(tags, extension)
         if not artifacts:
             user_error(
--- a/contrib/cf-remote/cf_remote/remote.py
+++ b/contrib/cf-remote/cf_remote/remote.py
@@ -250,8 +250,7 @@ def uninstall_host(host, *, connection=None):
     print_info(data)
 
     if not data["agent_version"]:
-        print("CFEngine is not installed on '{}' - moving on".format(host))
-        return data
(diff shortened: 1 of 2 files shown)


contrib/cf-remote/cf_remote/remote.py
```diff
@@ -97,6 +97,48 @@ def get_info(host, *, users=None, connection=None):
...
+                # Examples:
+                # CentOS release 6.10 (Final)
+                # Red Hat Enterprise Linux release 8.0 (Ootpa)
+                before, after = redhat_release.split(" release ")
+                distro = "rhel"
+                if before.lower().startswith("centos"):
+                    distro = "centos"
+                major = after.split(".")[0]
+                tags.append(distro + major)
+                if distro != "el":
```
vpodzime: `distro` cannot be equal to `"el"` here, right?

contrib/cf-remote/cf_remote/remote.py
```diff
@@ -97,6 +97,48 @@ def get_info(host, *, users=None, connection=None):
...
+                    tags.append("el" + major)
+                if distro != "rhel":
```
vpodzime: And here it can only be `"centos"` if it is not `"rhel"`.

contrib/cf-remote/cf_remote/remote.py
```diff
@@ -97,6 +97,48 @@ def get_info(host, *, users=None, connection=None):
...
+                    tags.append("rhel" + major)
+
+                tags.append(distro)
+                if distro != "rhel":
+                    tags.append("rhel")
+                if distro != "el":
+                    tags.append("el")
```
vpodzime: Same here.

contrib/cf-remote/cf_remote/remote.py
```diff
@@ -190,6 +232,7 @@ def install_host(
         extension = None
         if "package_tags" in data and "msi" in data["package_tags"]:
             extension = ".msi"
+            del data["package_tags"]["msi"]
```
vpodzime: What does this have to do with CentOS 6?

olehermanse: `data["package_tags"]` is now appended (extending) tags below, to get the tags from os-release stuff. As a side effect, it means that the msi tag would be added "twice" both as an extension and as a tag. This seemed unnecessary.

contrib/cf-remote/cf_remote/remote.py
olehermanse: Right, I was just following a pattern. Fixed.

contrib/cf-remote/cf_remote/remote.py
olehermanse: Yes, I guess my thought here was more like: don't add `"rhel"` twice to the list. I tried looking for a unique list / ordered set in python, but it seems it's not part of the standard library.

olehermanse: Fixed / tried to make code more clear.

PR #4070, opened 2020-02-17
Do not log an error in case of no match for a measurements promise

vpodzime:
If a measurements promise regex doesn't match any line of a given
file, it should not produce an error logged to syslog. It's just
noise and the default value 0.0 of the measurement indicates that
nothing was measured. Running with '--verbose' then shows the log
message and helps with debugging.

Ticket: ENT-5171
Changelog: Measurements promises with no match no longer produce errors

### DIFF
--- a/cf-monitord/history.c
+++ b/cf-monitord/history.c
@@ -551,7 +551,7 @@ static PromiseResult NovaExtractValueFromStream(EvalContext *ctx, const char *ha
 
     if (!found)
     {
-        cfPS(ctx, LOG_LEVEL_ERR, PROMISE_RESULT_FAIL, pp, a, "Could not locate the line for promise '%s'", handle);
+        cfPS(ctx, LOG_LEVEL_VERBOSE, PROMISE_RESULT_FAIL, pp, a, "Could not locate the line for promise '%s'", handle);
         *value_out = 0.0;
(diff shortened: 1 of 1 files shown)


olehermanse:
@cf-bottom jenkins

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4614/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4614/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4614/

vpodzime:
> why in this case the log message doesn't match the level of the promise result?

What and where says that a failed promise has to cause an error message?

nickanderson:
> What and where says that a failed promise has to cause an error message?

I am not aware of a strict rule, but it's definitely a-typical behavior. I added this to the ticket:

```
It is a-typical for a promise to have a not-kept outcome and not emit an error level log message, however, in this case, it does not make sense to emit a log message.

    Unaware of any documented rule that promises not kept should always result in an ERROR level log message.
    It is legitimate that there may be 0 occurrences of a log message to measure
    It makes some sense that the promise is recorded as a failure since it was unable to find the pattern it intends to count
```

PR #4098, opened 2020-03-04
CFE-3041: Added .error.cf possibility to acceptance tests (3.15)

olehermanse:
Purpose is explained in testall script;
* `.x.cf` is for tests which are supposed exit with error,
  for example, invalid policy syntax.
* `.error.cf` is for tests which are supposed to log errors,
  but exit with 0. For example failed promises.

PR #4099, opened 2020-03-04
CFE-3041 Travis: Run errorlog acceptance tests

olehermanse:
It is debatable whether errorexit tests should be considered safe
or not, but errorlog should definitely be considered safe.


olehermanse:
Cherry-pick: https://github.com/cfengine/core/pull/4098

PR #4109, opened 2020-03-12
ENT-5329: Don't go back further than field start when parsing CMD field.

kacf:
This can happen if the CMD field is empty and filled with space, for
example in the case of zombie processes. Presumably the reason this
has passed before is because it would later result in a strndup call
with a negative size, which on most platforms would just mean the same
as zero. However, on AIX 7 this crashes instead, which is how the
issue was detected (split_process_line_test:test_platform_extra_table
test).

Changelog: Fix rare crashing bug when parsing zombie entries in ps
output. The problem was only ever observed on AIX, but could
theoretically happen on any platform depending on exact libc behavior.

Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>

note to cf-bottom:
Merge together:
https://github.com/cfengine/buildscripts/pull/705

### DIFF
--- a/libpromises/processes_select.c
+++ b/libpromises/processes_select.c
@@ -882,7 +882,7 @@ Solaris 9:
         {
             // Last field, slurp up the rest, but discard trailing whitespace.
             last = linelen;
-            while (isspace(line[last - 1]))
+            while (last > pos && isspace(line[last - 1]))
             {
                 last--;
             }


olehermanse:
AIX only build:

[Build Status](https://ci.cfengine.com/job/pr-pipeline/4829/)

olehermanse:
@cf-bottom full jenkins run please, with exotics :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4830/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4830/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4830/

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4833/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4833/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4833/

kacf:
@olehermanse: @Lex-2008 confirmed that this was fixed in the latest run, and the problems we see now are due to https://tracker.mender.io/browse/ENT-5418. So I think we can ignore the AIX result here. The question is redhat, where this is also a failure. Does that look related to you (I can't see it)?

olehermanse:
No, redhat failure looks unrelated.

kacf:
Thanks!

olehermanse:
Cherry-picks:
https://github.com/cfengine/core/pull/4121
https://github.com/cfengine/core/pull/4120

PR #4129, opened 2020-04-03
Add multi-key/multi-host support to cf-keycrypt

vpodzime:
Encrypting secrets for hosts individually is error-prone and
results in many files. A better approach is to encrypt the
payload for multiple hosts (keys) at once and store the result in
a file the format of which allows the individual hosts to decrypt
the payload back using their keys.

This commit introduces a new 'Encrypted-for' header for the
encrypted file format which tells which keys the payload was
encrypted for and also what is the ordering of the segments
containing the encrypted symmetric key. Thus the format now looks
as follows:
```
  ------------
  | Headers  |
  ------------
  | AES IV   |
  ------------
  | AES key  |
  ------------
  | AES key  |
  ------------
  | ...      |
  ------------
  | AES key  |
  ------------
  | data     |
  ------------
```
Where each 'AES key' block is the same randomly generated AES key
encrypted by different RSA (host) key. The host then needs to
identify which 'AES key' block to decrypt and use as the key to
decrypt the payload.

Ticket: CFE-3271
Changelog: None

TODO:
- [x] test on real hub and hosts deployment
- [x] valgrind-check memory leaks
- [x] adapt tests


vpodzime:
Jenkins pipeline run: [Build Status](https://ci.cfengine.com/job/pr-pipeline/4948/)

olehermanse:
Jenkins failures look unrelated.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -403,73 +433,153 @@ static bool CheckHeader(const char *key, const char *value)
...
+ *
+ * @warning #buffer has to be at least #max_chars+1 big to accomodate for the
+ *          terminating NUL byte.
+ */
+static inline bool ReadUntilDelim(FILE *input_file, char *buffer, size_t max_chars, char delimiter, bool stop_on_space)
 {
+    bool done = false;
+    size_t i = 0;
+    char c = fgetc(input_file);
+    while (!done && (i <= max_chars))
```
olehermanse, line 461: Each iteration of the loop increments i by 1 and puts 1 byte into buffer. Then after the loop, 1 more byte is added to buffer (NUL). The less than or equal means this loop runs up to `max_chars + 1`times. The condition needs to be `i < max_chars`, to prevent 1 byte overflow.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -403,73 +433,153 @@ static bool CheckHeader(const char *key, const char *value)
...
+        else if (stop_on_space && (c == ' '))
+        {
+            done = true;
+            ungetc(c, input_file);
+        }
+        else
+        {
+            buffer[i] = c;
+            i++;
+            c = fgetc(input_file);
```
olehermanse, line 481: What happens to this character if `max_chars` is reached?

cf-keycrypt/cf-keycrypt.c
```diff
@@ -403,73 +433,153 @@ static bool CheckHeader(const char *key, const char *value)
...
+        }
+    }
+    buffer[i] = '\0';
```
olehermanse, line 484: Potential overflow here, see comment above about loop condition.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -403,73 +433,153 @@ static bool CheckHeader(const char *key, const char *value)
...
+
+    bool ran_out = ((i > max_chars) || (c == EOF));
+    return !ran_out;
+}
+
+static inline void SkipSpaces(FILE *input_file)
+{
+    char c = ' ';
+    while (!feof(input_file) && (c == ' '))
```
olehermanse, line 493: This function uses `feof()` while the one above uses `EOF`, should decide on one of them.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -403,73 +433,153 @@ static bool CheckHeader(const char *key, const char *value)
...
+    {
+        c = fgetc(input_file);
+    }
+    if (c != ' ')
+    {
+        ungetc(c, input_file);
```
olehermanse, line 499: I wonder what happens here if you `ungetc((char) EOF, input_file);` on a file that has reached EOF.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -403,73 +433,153 @@ static bool CheckHeader(const char *key, const char *value)
...
             }
+            *n_enc_keys = n_enc_for_headers;
+            free(key_digest);
+            return version_specified && found_matching_digest;
         }
         else
         {
-            return false;
+            /* keep trying */
+            ungetc(next, input_file);
```
olehermanse, line 570: Again, I believe `next` here can be `(char) EOF`.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -403,73 +433,153 @@ static bool CheckHeader(const char *key, const char *value)
...
+            if (StringSafeEqual(value, key_digest))
+            {
+                found_matching_digest = true;
+                *enc_key_pos = n_enc_for_headers;
+            }
+            n_enc_for_headers++;
+        }
+
+        /* headers are supposed to be terminated by a blank line */
         char next = fgetc(input_file);
```
olehermanse, line 552: May return `EOF`(int) in case of error or EOF.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -403,73 +433,153 @@ static bool CheckHeader(const char *key, const char *value)
...
-            else
+            if (!found_matching_digest)
             {
-                /* keep trying */
-                ungetc(next, input_file);
+                Log(LOG_LEVEL_ERR, "File not encrypted for host '%s'", key_digest);
             }
+            *n_enc_keys = n_enc_for_headers;
+            free(key_digest);
+            return version_specified && found_matching_digest;
```
olehermanse, line 565: Parenthesize, please.

cf-keycrypt/cf-keycrypt.c
```diff
@@ -674,16 +825,16 @@ int main(int argc, char *argv[])
         exit(EXIT_FAILURE);
     }
 
-    if (decrypt && (host == NULL) && (key_path == NULL))
+    if (decrypt && (host_arg == NULL) && (key_path_arg == NULL))
     {
         /* Decryption requires a private key which is usually only available for
-         * the local host. Let's just default to localhost if no other specific
+         * the local host. Let's just default to localhost_arg if no other specific
```
olehermanse, line 831: Seems like a find and replace error?

cf-keycrypt/cf-keycrypt.c
```diff
@@ -674,16 +825,16 @@ int main(int argc, char *argv[])
...
          * host/key is given for decryption. */
-        Log(LOG_LEVEL_VERBOSE, "Using the localhost private key for decryption");
-        host = "localhost";
+        Log(LOG_LEVEL_VERBOSE, "Using the localhost_arg private key for decryption");
```
olehermanse, line 833: Same here.

olehermanse:
@cf-bottom test this in jenkins again, just in case :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4951/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4951/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4951/

cf-keycrypt/cf-keycrypt.c
```diff
@@ -403,73 +433,153 @@ static bool CheckHeader(const char *key, const char *value)
...
+ *          terminating NUL byte.
+ */
+static inline bool ReadUntilDelim(FILE *input_file, char *buffer, size_t max_chars, char delimiter, bool stop_on_space)
 {
+    bool done = false;
+    size_t i = 0;
+    char c = fgetc(input_file);
+    while (!done && (i <= max_chars))
+    {
+        if (c == EOF)
```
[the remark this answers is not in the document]
vpodzime: Well, it just works with `char c` as well. `EOF` is `-1`, but `0xFF` could have been confused for `EOF` if it showed up. At least on some systems. Good catch!

cf-keycrypt/cf-keycrypt.c
```diff
@@ -403,73 +433,153 @@ static bool CheckHeader(const char *key, const char *value)
...
-static bool ParseHeaders(FILE *input_file)
+/**
+ * Read from #input_file into #buffer at most #max_chars or until #delimiter is
+ * encountered. If #stop_on_space, also stop when ' ' is read. #buffer is
+ * NUL-terminated when this function returns.
+ *
+ * @warning #buffer has to be at least #max_chars+1 big to accomodate for the
+ *          terminating NUL byte.
+ */
+static inline bool ReadUntilDelim(FILE *input_file, char *buffer, size_t max_chars, char delimiter, bool stop_on_space)
```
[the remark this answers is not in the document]
vpodzime: Yeah, I'm not 100% in favour of either of the two. But for this particular case I find it better to work with `max_chars` as it is now. And this is a `static` function for this particular use case. Was it a generic function in *libntech/libutils*, I'd prefer buffer size too.

cf-keycrypt/cf-keycrypt.c
vpodzime: Yup, good catch! `max_chars` is the highest index in `char buffer[max_chars + 1]`.

cf-keycrypt/cf-keycrypt.c
vpodzime: Needs `if (i < max_chars)`. :+1:

cf-keycrypt/cf-keycrypt.c
vpodzime: Why? Both ways exist and work because they can be a better fit for different cases. Here it's nice to use `feof()` because `c` doesn't have to be initialized by `fgetc()` and the loop condition makes more sense to me like this.

cf-keycrypt/cf-keycrypt.c
vpodzime: Nothing special, you just get `EOF` in the next `fgetc()`. `ungetc()` only works with the internal stream API (`FILE *`) buffers, I think.

vpodzime:
Comments addressed.

PR #4133, opened 2020-04-06
Update BSD.md

craigcomstock:
Changed to more FreeBSD style `doas` commands.
Removed emacs-nox, FreeBSD comes with ed(1) the standard UNIX editor. No need to be editor opinionated. ;)

### DIFF
--- a/docs/BSD.md
+++ b/docs/BSD.md
@@ -2,12 +2,19 @@
 
 BSD is not officially supported.
 These are some basic instructions for getting CFEngine Community to build on FreeBSD.
-Tested on FreeBSD 12.1 VM in AWS EC2.
+Tested on FreeBSD 12.1 VM in AWS EC2 and on a Lenovo X220 Tablet.
 
 ## Dependencies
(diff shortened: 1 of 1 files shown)


craigcomstock:
Ok. I'm no FreeBSD expert so maybe `su` is there by default. And we can assume a user knows how to `/usr/sbin/pw usermod <user> -G wheel` to allow them to use `su`. `doas` requires editing `/usr/local/etc/doas.conf` and adding `permit nopass <user> as root` or similar.

craigcomstock:
But at least I've confirmed that the instructions work even now.

olehermanse:
> Removed emacs-nox, FreeBSD comes with ed(1) the standard UNIX editor. No need to be editor opinionated. ;)

And yet you just were, preferring ed over emacs. With my instructions you'd have the choice ;)

craigcomstock:
sudo vs doas is about as useful as emacs-nox vs ed so closing. :)

PR #4135, opened 2020-04-07
Added dependency one-liners for various systems

craigcomstock:
Add some examples of how to install needed dependencies for FreeBSD and SUSE. TODO: debian, redhat, hp-ux, ???

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -35,6 +35,8 @@ following additional tools:
 
 Latest stable versions of the tools and libraries are generally advised.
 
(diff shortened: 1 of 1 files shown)


craigcomstock:
Granted we might need an autogen.sh command line for these one-liners as well. As is already present in https://github.com/cfengine/core/blob/master/docs/BSD.md so not sure if we should create SUSE.md, etc for various systems or not.

craigcomstock:
@olehermanse commits are sqaushed, will refine in further PRs when I find differences.

PR #4139, opened 2020-04-14
ENT-4653 Fixed ifelse() to return fallback in case of undefined variables

craigcomstock:
Root cause is that if there is only one iteration wheel remaining and
that wheel has an unresolved value PromiseIteratorNext() was returning
false which cause ExpandPromiseAndDo() to not evaluate the ifelse
function and not assign the default value in the ifelse() expression.

Changelog: title
Ticket: ENT-4653


craigcomstock:
not even close to a fix I think. will do more work and run more local tests first.

PR #4140, opened 2020-04-14
CFE-270/master: Activated test making sure classic templates do not add extra newlines

nickanderson:
----

#


nickanderson:
Soft failing in travis as expected
```
./10_files/templating/cftemplate.cf Soft fail (CFE-270)
```

PR #4148, opened 2020-04-20
Added where changelog entries are discovered

craigcomstock:
Only in commits. Not in PR descriptions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -209,6 +209,10 @@ This means that references to implementation details are not appropriate, leave 
 It is the behavior change which is important.
 Changes which don't affect users, like changes in tests or in tools which are not packaged, should *not* have a changelog entry.
 
+All changelog entries should also include a reference to a ticket:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -209,6 +209,10 @@ This means that references to implementation details are not appropriate, leave
 It is the behavior change which is important.
 Changes which don't affect users, like changes in tests or in tools which are not packaged, should *not* have a changelog entry.
 
+All changelog entries should also include a reference to a ticket:
```
Lex-2008, line 212: Note that it's not obligatory to type letters `Tickets`. You can just include `CFE-1234` somewhere in commit message text, like this:
```
Changelog: Fix issue mentioned in CFE-1234 and other places.
```

craigcomstock: I didn't know that! Hmm.

nickanderson: Me neither!

PR #4157, opened 2020-04-22
CFE-628/master: Added test for expected output

nickanderson:
Ticket: CFE-628
Changelog: None

### DIFF
--- a/tests/acceptance/01_vars/01_basic/CFE-628.cf
+++ b/tests/acceptance/01_vars/01_basic/CFE-628.cf
@@ -1,0 +1,44 @@
+###########################################################
+#
+# Test evaluation order expectations
+#
+###########################################################
+
+body common control
+{
+    inputs => { "../../default.cf.sub" };
+    bundlesequence => { default($(this.promise_filename)) };
+    version => "1.0";
+}
+
+###########################################################
+
+bundle agent test
+{
+  meta:
+      "description" -> { "CFE-628" }
+        string => "Test that the output is as expected";
+
+      "test_skip_needs_work" -> { "CFE-628" }
+        string => "windows",
+        meta => { "CFE-628" };
+
+}
+
+###########################################################
+
+bundle agent check
+{
+  vars:
+      "cmd" string => "$(sys.cf_agent) -Kf $(this.promise_filename).sub";
+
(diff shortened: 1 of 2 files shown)


tests/acceptance/01_vars/01_basic/CFE-628.cf
```diff
@@ -0,0 +1,39 @@
...
+    inputs => { "../../default.cf.sub" };
+    bundlesequence => { default($(this.promise_filename)) };
+    version => "1.0";
+}
+
+###########################################################
+
+bundle agent test
+{
+  meta:
```
craigcomstock, line 18: Mark as !windows only test. Since you are using `/bin/echo`.

tests/acceptance/01_vars/01_basic/CFE-628.cf.sub
```diff
@@ -0,0 +1,14 @@
+bundle agent main 
+{
+  classes:
+    "found_the_file" expression => fileexists( $(this.promise_filename) );
+
+  vars:
+    found_the_file::
```
craigcomstock, line 7: It's not clear to me how this tests a variable reference with nested parenthesis. I only ever see one set of parenthesis.

tests/acceptance/01_vars/01_basic/CFE-628.cf
nickanderson: ACK,

nickanderson:
For explination, i think that there is the jira ticket for reference. I just pulled from the report to synthesise the test.

tests/acceptance/01_vars/01_basic/CFE-628.cf.sub
craigcomstock: Even reading the bug I don't completely understand the emphasis on "nested parenthesis". It seems the more important distinction is whether the reports promises should print out twice or not, due to multiple passes and the found_the_file class being both true and false.

Does this test currently fail? Should it be marked as soft fail then?

nickanderson: No, the test passes.

Ah, the test header was from a copied file. un-related.

PR #4163, opened 2020-04-24
Evaluate 'if' even in case of class context or 'ifvarclass'

vpodzime:
If a class context is specified in the policy or if 'ifvarclass'
constraint is used, the 'if' constraint should be evaluated as
well.

Ticket: CFE-2615
Changelog: 'if' constraint now works in combination with class contexts

### DIFF
--- a/libpromises/promises.c
+++ b/libpromises/promises.c
@@ -625,25 +625,40 @@ Promise *ExpandDeRefPromise(EvalContext *ctx, const Promise *pp, bool *excluded)
 
     /* Look for 'if'/'ifvarclass' exclusion. */
     {
+        /* We need to make sure to check both 'if' and 'ifvarclass' constraints. */
+        bool checked_if = false;
         const Constraint *ifvarclass = PromiseGetConstraint(pp, "ifvarclass");
         if (!ifvarclass)
         {
             ifvarclass = PromiseGetConstraint(pp, "if");
+            checked_if = true;
         }
 
         // if - Skip if false or error:
-        if (ifvarclass && (CheckVarClassExpression(ctx, ifvarclass, pcopy) != EXPRESSION_VALUE_TRUE))
+        while (ifvarclass != NULL)
         {
-            if (LogGetGlobalLevel() >= LOG_LEVEL_VERBOSE)
+            if (CheckVarClassExpression(ctx, ifvarclass, pcopy) != EXPRESSION_VALUE_TRUE)
             {
-                char *ifvarclass_string =  RvalToString(ifvarclass->rval);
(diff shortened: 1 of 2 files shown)


vpodzime:
@cf-bottom, jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5049/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5049/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5049/

libpromises/promises.c
```diff
@@ -625,25 +625,40 @@ Promise *ExpandDeRefPromise(EvalContext *ctx, const Promise *pp, bool *excluded)
...
             ifvarclass = PromiseGetConstraint(pp, "if");
+            checked_if = true;
         }
 
         // if - Skip if false or error:
-        if (ifvarclass && (CheckVarClassExpression(ctx, ifvarclass, pcopy) != EXPRESSION_VALUE_TRUE))
+        while (ifvarclass != NULL)
         {
-            if (LogGetGlobalLevel() >= LOG_LEVEL_VERBOSE)
+            if (ifvarclass && (CheckVarClassExpression(ctx, ifvarclass, pcopy) != EXPRESSION_VALUE_TRUE))
```
olehermanse, line 640: `ifvarclass` checked in both `while` and `if` condition.

libpromises/promises.c
```diff
@@ -625,25 +625,40 @@ Promise *ExpandDeRefPromise(EvalContext *ctx, const Promise *pp, bool *excluded)
 
     /* Look for 'if'/'ifvarclass' exclusion. */
     {
+        /* We need to make sure to check both 'if' and 'ifvarclass' constraints. */
```
nickanderson, line 628: since if is supposed to be an alias to ifvarclass, I don't think that we should apply both if they are on the same promise. Rather, I think we should disallow using both on a single promise.

olehermanse: @nickanderson It's a little more subtle, since class guards with variables in them are implemented by putting them in an `if` contstraint. The easiest right now, would be to add a check, to ensure that you are only using 1 if/ifvarclass (in addition to the one added by class guard).

vpodzime: > implemented by putting them in an if contstraint
Actually, they are put in `ifvarclass` constraint.

olehermanse: @vpodzime but `ifvarclass` is an alias for `if` ;)

PR #4169, opened 2020-04-28
CFE-3319/master: Test that if can check a class expression from a data container

nickanderson:
----

#


cf-bottom:

PR #4175, opened 2020-05-08
ENT-5966/master: Added example illustrating template_method => "inline_mustache"

nickanderson:


examples/template_method-inline_mustache.cf
```diff
@@ -0,0 +1,72 @@
+#  Copyright 2019 Northern.tech AS
```
craigcomstock, line 1: 2020

nickanderson: ACK. DONE

PR #4193, opened 2020-05-22
ENT-6002 Added association between host info and release package names (3.12.x)

craigcomstock:
3.12.x package filenames do not include any "distro" information so installing
with --version 3.12.x doesn't work well. e.g. debian_7 gets ubuntu package.

Added "debian_7" tag in addition to current "debian7" tag which matches
package attribute values from releases.json such as
PACKAGES_HUB_x86_64_linux_debian_7

Changelog: none
Ticket: ENT-6002

### DIFF
--- a/contrib/cf-remote/cf_remote/packages.py
+++ b/contrib/cf-remote/cf_remote/packages.py
@@ -7,6 +7,7 @@ import re
 
 class Artifact:
     def __init__(self, data, filename=None):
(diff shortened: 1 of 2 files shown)


craigcomstock:
This fixes an issue I had installing 3.12.x nightlies on debian_7. Not sure if it breaks something else, thus the "WIP".

craigcomstock:
Not an issue anymore. Apparently we fixed the filenames. :+1:

PR #4195, opened 2020-05-23
CFE-1655: process-related promises via '/proc' Pt. 1

tdlee:
Process-related promises have hitherto been implemented by invoking the
'ps' command and attempting to parse its output. But 'ps' output
is highly human-oriented, and often not easy to machine-parse reliably.
It is also highly OS-variable.

CFE-1655 had initially proposed the idea of using "/proc" where available.
My own encounter with the flaws in the existing mechanism is at CFE-3020.

This will be a gradual, evolving transition, on a per-OS "opt-in" basis,
from old ps-command working to new system-probe (os "sysprobe") working.

This first commit sets up the autotools framework to permit this functionality
to be developed but does not include the code for any OS.

To allow gradual introduction, files "libpromises/process_solaris.c" and similar
are renamed to "libpromises/process_cmd_solaris.c" to reflect their old
command-based nature.  When an OS is implemented in the new way (e.g. "/proc")
a corresponding "libpromises/process_sysprobe_solaris.c" (etc.) is introduced.

Where conditional compilation is needed, "HAVE_PS_VIA_SYSPROBE" indicates
the new way.

Although not in this intial commit, it is intended that the new system-probing
will use JSON-based data structures to separate out the OS-dependent and
OS-independent aspects; the latter could later be moved into "libntech".


cf-bottom:

configure.ac
```diff
@@ -1347,6 +1347,80 @@ case "$target_os" in
...
+dnl
+dnl Most UNIX/Linux systems offer more machine-oriented mechanisms, for instance:
+dnl   o  Linux: /proc
+dnl   o  Solaris: /proc (different details)
+dnl   o  FreeBSD: kvm_getprocs()
+dnl
+dnl We have begun (2019) to investigate using these on a per-platform basis.
+dnl Because there is variety, we refer to this by the general term "system probe".
+dnl The previous ps-cmd mechanism is retained as a fallback for platforms.
+dnl ######################################################################
```
olehermanse, line 1362: Thanks for the comment, great explanation.

olehermanse:
@cf-bottom Please test this in Jenkins, including exotic platforms.

configure.ac
```diff
@@ -1347,6 +1347,80 @@ case "$target_os" in
...
+dnl For BSD ensure we have "kvm_getprocs()"
+case "$target_os" in
+  *bsd*|dragonfly*)
+    AC_CHECK_LIB([kvm], [kvm_getprocs],
+        [AC_DEFINE([HAVE_LIBKVM], 1, [Define if -lkvm is available])
+        LIBS="-lkvm $LIBS"],
+        AC_MSG_ERROR(Cannot find kvm library required for BSD)
+    )
+    ;;
+esac
```
olehermanse, line 1377: Not a big deal, but this could have been left out. (And introduced in a separate PR for BSD support later). Try to keep things as small and safe as possible. We don't have automated testing on BSD, so I'm not 100% sure this doesn't break the build for someone.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5221/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5221/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5221/

libpromises/Makefile.am
```diff
@@ -165,29 +165,58 @@ libpromises_la_SOURCES += \
...
+
+### Not yet implemented
+# libpromises_la_SOURCES += \
+#        process_sysprobe_@PROCESS_OS_TAG@.c \
+#        processes_load_sysprobe.c \
+#        processes_select_sysprobe.c
+
+else # HAVE_PS_VIA_SYSPROBE
+
+libpromises_la_SOURCES += processes_select_cmd.c
```
olehermanse, line 195: All of this is inside a `if !NT` above, so this source will not be added on Windows builds.

libpromises/Makefile.am
```diff
@@ -129,7 +129,7 @@ libpromises_la_SOURCES = \
 	parser_state.h \
 	patches.c \
 	pipes.h pipes.c \
-	processes_select.c processes_select.h \
+	processes_select.h \
```
olehermanse, line 132: As explained below, this removes the source from the windows build, you will need to add it back. Something like:

```
if NT
libpromises_la_SOURCES += processes_select_cmd.c
endif
```

olehermanse:
Closing due to inactivity.

PR #4201, opened 2020-05-29
Added usage of os-release ID_LIKE/VERSION_ID_LIKE for cf-remote install

craigcomstock:
Added parsing and use of ID_LIKE and VERSION_ID_LIKE to cf-remote to support installing on PureOS. Should work for many other derivative distributions.

Changelog: Title
Ticket: none

### DIFF
--- a/contrib/cf-remote/README.md
+++ b/contrib/cf-remote/README.md
@@ -85,6 +85,17 @@ Note that this demo setup (`--demo`) is notoriously insecure.
 It has default passwords and open access controls.
 Don't use it in a production environment.
 
+Note also that for distributions that we don't have packages for you might add entries to
+`/etc/os-release` on the target system to give cf-remote an alternative. For example, PureOS
+is based on debian testing, so currently the appropriate package we create would be debian 10.
+
+```
+ID_LIKE="debian"
+VERSION_ID_LIKE="10.0"
+```
+
+The non-standard `VERSION_ID_LIKE` is used. See [os-release documentation](https://www.linux.org/docs/man5/os-release.html) for more info.
+
 ### Specify an SSH key
 
 If you have more than one key in `~/.ssh` you may need to specify which key `cf-remote` is to use.
--- a/contrib/cf-remote/cf_remote/remote.py
+++ b/contrib/cf-remote/cf_remote/remote.py
@@ -106,8 +106,9 @@ def get_info(host, *, users=None, connection=None):
 
         tags = []
         if data["os_release"]:
-            distro = data["os_release"]["ID"]
-            major = data["os_release"]["VERSION_ID"].split(".")[0]
+            distro = data["os_release"]["ID_LIKE"] or data["os_release"]["ID"]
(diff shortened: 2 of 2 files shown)


contrib/cf-remote/README.md
```diff
@@ -85,6 +85,17 @@ Note that this demo setup (`--demo`) is notoriously insecure.
 It has default passwords and open access controls.
 Don't use it in a production environment.
 
+Note also that for distributions that we don't have packages for you might add entries to
+`/etc/os-release` on the target system to give cf-remote an alternative. For example, PureOS
+is based on debian testing, so currently the appropriate package we create would be debian 10.
+
+```
+ID_LIKE="debian"
+VERSION_ID_LIKE="10.0"
```
nickanderson, line 94: I dunno about adding this custom one to os-release. Feels fiddly, and something that may not be retained through OS upgrades, so they would also need to deploy policy to ensure the custom entries are always re-added.

More complex, but a better user experience IMHO wold be to prompt the user to pick the most similar platform then cache that info to re-use for any future same case.

olehermanse: @nickanderson I don't think cf-remote should turn interactive by default. We could add a better error message, better documentation, or `--interactive` option, but IMO, this is pretty low priority.

@craigcomstock FYI, I would just do:

```
olehermanse@OH-WIN ~ $ cf-remote packages debian 64 client
Available releases: 3.15.x, 3.15.2, 3.15.1, 3.15.0, 3.15.0b1, 3.12.x, 3.12.5, 3.12.4, 3.12.3, 3.12.2, 3.12.1, 3.12.0
Using 3.15.2 LTS (default):
Package already downloaded: '/home/olehermanse/.cfengine/cf-remote/packages/cfengine-nova_3.15.2-1.debian7_amd64.deb'
Package already downloaded: '/home/olehermanse/.cfengine/cf-remote/packages/cfengine-nova_3.15.2-1.debian8_amd64.deb'
Package already downloaded: '/home/olehermanse/.cfengine/cf-remote/packages/cfengine-nova_3.15.2-1.debian9_amd64.deb'
Package already downloaded: '/home/olehermanse/.cfengine/cf-remote/packages/cfengine-nova_3.15.2-1.debian10_amd64.deb'
olehermanse@OH-WIN ~ $ cf-remote install --client-package /home/olehermanse/.cfengine/cf-remote/packages/cfengine-nova_3.15.2-1.debian9_amd64.deb ...
```

contrib/cf-remote/cf_remote/remote.py
```diff
@@ -106,8 +106,9 @@ def get_info(host, *, users=None, connection=None):
 
         tags = []
         if data["os_release"]:
-            distro = data["os_release"]["ID"]
-            major = data["os_release"]["VERSION_ID"].split(".")[0]
+            distro = data["os_release"]["ID_LIKE"] or data["os_release"]["ID"]
```
olehermanse: I think this is problematic. Since you prefer `ID_LIKE`, I would expect this to cause cf-remote to download debian packages for ubuntu.

craigcomstock:
I don't see any need to pursue this one so closing.

PR #4206, opened 2020-06-04
sysinfo.c: Limit string sizes in sscanf calls

lluiscampos:
In practice, most of these calls were safe as the strings being scanned
are read from buffers in of the same size.

Still LGTM was alerting on "Unbounded write" so it is nice to fix.

Changelog: None
Ticket: ENT-6019
Signed-off-by: Lluis Campos <lluis.campos@northern.tech>


libenv/sysinfo.c
```diff
@@ -972,7 +972,7 @@ static void Get3Environment(EvalContext *ctx)
 
         if (*context == '@')
         {
-            if (sscanf(context + 1, "%[^=]=%[^\n]", name, value) == 2)
+            if (sscanf(context + 1, "%" STRING(CF_MAXVARSIZE - 1) "[^=]=%" STRING(CF_BUFSIZE - 1) "[^\n]", name, value) == 2)
```
lluiscampos, line 975: Other part of this file, for instance a bit below on line 997, use hardcoded numbers instead. Should I unify everywhere?

olehermanse: I wouldn't worry about that, yet.

lluiscampos:
> Please add a unit test to ensure `STRING(CF_BUFSIZE - 1)` works :)

It doesn't work; my bad.

I pushed new work using a hardcoded max length instead. Also don't think that adding `assert` before each of these to check that `sizeof(buffer) > 255` makes sense, as discussed, because the buffers are all declared locally. If any, I suggest we modify these buffers to be `version[256]` instead of `version[CF_BUFSIZE]` so that we have local control on their sizes.

libenv/sysinfo.c
```diff
@@ -972,7 +972,7 @@ static void Get3Environment(EvalContext *ctx)
 
         if (*context == '@')
         {
-            if (sscanf(context + 1, "%[^=]=%[^\n]", name, value) == 2)
+            if (sscanf(context + 1, "%255[^=]=%255[^\n]", name, value) == 2)
```
olehermanse, line 975: As discussed, add these lines above:

```
assert((sizeof(name) - 1) == 255);  // TODO: static_assert()
assert((sizeof(value) - 1) == 255); // TODO: static_assert()
```

It's just an attempt to prevent developers from shooting themselves in the foot.
What if someone copy pastes this code to another function with different size `name`?
What if someone changes `CF_MAXVARSIZE`?
What if someone changes `name` to be a pointer?

libenv/sysinfo.c
```diff
@@ -2061,7 +2061,7 @@ static int Linux_Suse_Version(EvalContext *ctx)
     {
         Item *list, *ip;
 
-        sscanf(relstring, "%[-_a-zA-Z0-9]", vbuf);
+        sscanf(relstring, "%255[-_a-zA-Z0-9]", vbuf);
```
olehermanse, line 2064: Similar here.

libenv/sysinfo.c
```diff
@@ -2158,8 +2158,8 @@ static int Linux_Suse_Version(EvalContext *ctx)
         }
         else
         {
-            sscanf(strversion, "VERSION = %s", strmajor);
-            sscanf(strpatch, "PATCHLEVEL = %s", strminor);
+            sscanf(strversion, "VERSION = %255s", strmajor);
+            sscanf(strpatch, "PATCHLEVEL = %255s", strminor);
```
olehermanse, line 2162: And here. And all places below.

I know it's ugly, but in C, safety is more important :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5219/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5219/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5219/

libenv/sysinfo.c
lluiscampos: I added asserts now :+1: Please look at the rework :smiley: 

However I sill think that it is more sensible to set these buffers to the required size (in this example, 256) instead of the macros. That will allow us to remove the `assert` calls and will also optimize the stack memory usage (at the moment we allocate arrays of 4096 bytes to use only 255 max)

libenv/sysinfo.c
```diff
@@ -2158,8 +2163,10 @@ static int Linux_Suse_Version(EvalContext *ctx)
         }
         else
         {
-            sscanf(strversion, "VERSION = %s", strmajor);
-            sscanf(strpatch, "PATCHLEVEL = %s", strminor);
+            assert((sizeof(strmajor)) > 255); // TODO: static_assert()
+            assert((sizeof(strminor)) > 255); // TODO: static_assert()
+            sscanf(strversion, "VERSION = %255s", strmajor);
+            sscanf(strpatch, "PATCHLEVEL = %255s", strminor);
```
lluiscampos, line 2169: There is an issue in these, though. They are defined as:

```
char strmajor[PRINTSIZE(major)], strminor[PRINTSIZE(minor)];
```

which means that the size is defined at runtime.

I suggest we change the declarations to be "big enough", following the way it is done everywhere else in the file. I just don't see why we want to optimize here to the byte while we are wasting so much memory in the others.

libenv/sysinfo.c
olehermanse: I agree, except for this part:

> That will allow us to remove the assert calls

We don't want to remove them, see reasons above.

libenv/sysinfo.c
olehermanse: We try to avoid hardcoded sizes, but I agree it should be constant, and a local macro for it might make sense.

olehermanse:
@cf-bottom jenkins, please.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5231/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5231/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5231/

libenv/sysinfo.c
lluiscampos: Fixed in my last push: I made these of size `CF_MAXVARSIZE`.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5232/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5232/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5232/

libenv/sysinfo.c
lluiscampos: > I agree, except for this part:
> 
> > That will allow us to remove the assert calls
> 
> We don't want to remove them, see reasons above.

The final code has buffers of `CF_MAXVARSIZE` size and the `assert` calls, so this discussions is not relevant for the PR any more.

However, what I was suggesting was something like:
```
char strmajor[256], strminor[256];
...
sscanf(strversion, "VERSION = %255s", strmajor);
sscanf(strpatch, "PATCHLEVEL = %255s", strminor);
```

And for that case, I argue that we don't need asserts: you are hardcoding the size of an array locally, just before using it on a `sscanf` also with a hardcoded max size.

olehermanse: @lluiscampos I understand, and I argued that you should still have the asserts there, like so:

```
char strmajor[256];
...
assert(sizeof(strmajor) == 256);
sscanf(strversion, "VERSION = %255s", strmajor);
```

The reason is that someone might try to change the 256 in the top of the function, and if the function is big and hard to read, they might miss the 255 in the format string (you can't even search for 256, you would miss it). Similarly, someone could copy the sscanf line, or just the format string, to another function, where the buffer size is different.

C is an unsafe language, code becomes more complicated, less maintainable, and more "copy-pasted" over time. I am just trying to be as safe and explicit as possible. When you write that format string, you are assuming that the buffer is 256 bytes, or more. So, make that assumption explicit, add an assert as close as possible to the code that warrants it.

The `...` in your example could be hundreds of lines of code in the future. You can't really control developers working on the project in the future, you can't guarantee that they don't make mistakes when they try to change it, the best we can do is try to prevent them from making those mistakes, by writing clear/readable code, useful comments and asserts / errors that show when something is wrong.

libenv/sysinfo.c
```diff
@@ -2391,12 +2400,14 @@ static int Linux_Debian_Version(EvalContext *ctx)
     }
 
     os[0] = '\0';
+    assert((sizeof(version)) > 250); // TODO: static_assert()
     sscanf(buffer, "%250s", os);
```
olehermanse, line 2404: should be `sizeof(os)`.

lluiscampos: Good catch. Fixed now.

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5252/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5252/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5252/

lluiscampos:
@olehermanse The failing jobs in Jenkins are `php-tests` and `php-selenium-tests`. I think this PR is good to go.

PR #4215, opened 2020-06-11
A few LMDB related fixes and improvements

vpodzime:
Goals:
- [x] fix potentially unaligned access to LMDB data in *cf-check*
- [x] make locks DB purging dynamic
- [x] handle full LMDBs
- [x] add `--test-write` to `cf-check repair`

### DIFF
--- a/cf-check/diagnose.c
+++ b/cf-check/diagnose.c
@@ -614,18 +614,26 @@ size_t diagnose_files(
 
         if (symlink_target != NULL)
         {
+            int usage;
+            bool needs_rotation = lmdb_file_needs_rotation(symlink_target, &usage);
             Log(LOG_LEVEL_INFO,
-                "Status of '%s' -> '%s': %s\n",
+                "Status of '%s' -> '%s': %s [%d%% usage%s]\n",
                 symlink,
                 symlink_target,
-                CF_CHECK_STRING(r));
+                CF_CHECK_STRING(r),
+                usage,
+                needs_rotation ? ", needs rotation" : "");
         }
         else
         {
+            int usage;
+            bool needs_rotation = lmdb_file_needs_rotation(filename, &usage);
             Log(LOG_LEVEL_INFO,
-                "Status of '%s': %s\n",
+                "Status of '%s': %s [%d%% usage%s]\n",
                 filename,
-                CF_CHECK_STRING(r));
+                CF_CHECK_STRING(r),
+                usage,
+                needs_rotation ? ", needs rotation" : "");
         }
 
 
--- a/cf-check/diagnose.c
+++ b/cf-check/diagnose.c
@@ -39,6 +39,11 @@ size_t diagnose_files(
 #include <validate.h>
 #include <openssl/rand.h>
 
+/* NOTE: Must be in sync with LMDB_MAXSIZE in libpromises/dbm_lmdb.c. */
+#ifndef LMDB_MAXSIZE
+#define LMDB_MAXSIZE    104857600
+#endif
+
 #define CF_CHECK_CREATE_STRING(name) \
   #name,
 
@@ -522,6 +527,25 @@ static char *follow_symlink(const char *path)
     }
     target_buf[r] = '\0';
     return xstrdup(target_buf);
+}
+
+bool lmdb_file_needs_rotation(const char *file, int *usage)
+{
+    struct stat sb;
+    if (stat(file, &sb) == 0)
+    {
+        int usage_pct = (((float) sb.st_size) / LMDB_MAXSIZE) * 100;
+        if (usage != NULL)
+        {
+            *usage = usage_pct;
+        }
+        return (usage_pct >= 95);
+    }
+    else
+    {
+        Log(LOG_LEVEL_ERR, "Failed to stat() '%s' when checking usage: %s", file, GetErrorStr());
+        return false;
+    }
 }
 
 /**
--- a/libpromises/dbm_lmdb.c
+++ b/libpromises/dbm_lmdb.c
@@ -700,6 +700,18 @@ bool DBPrivClean(DBPriv *db)
     return (mdb_drop(txn->txn, db->dbi, EMPTY_DB) != 0);
 }
 
+int DBPrivGetDBUsagePercentage(const char *db_path)
+{
+    struct stat sb;
+    int ret = stat(db_path, &sb);
+    if (ret == -1)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to get size of '%s': %s", db_path, GetErrorStr());
+        return -1;
+    }
+    return (int) ((((float) sb.st_size) / LMDB_MAXSIZE) * 100);
+}
+
 void DBPrivCommit(DBPriv *db)
 {
     assert(db != NULL);
--- a/libpromises/dbm_lmdb.c
+++ b/libpromises/dbm_lmdb.c
@@ -79,12 +79,17 @@ static int DB_MAX_READERS = -1;
 /******************************************************************************/
 
 static void HandleLMDBCorruption(MDB_env *env, const char *msg);
+static void HandleFullLMDB(MDB_env *env);
 
-static inline void CheckLMDBCorrupted(int rc, MDB_env *env)
+static inline void CheckLMDBUsable(int rc, MDB_env *env)
 {
     if (rc == MDB_CORRUPTED)
     {
         HandleLMDBCorruption(env, "");
+    }
+    else if (rc == MDB_MAP_FULL)
+    {
+        HandleFullLMDB(env);
     }
 }
 
@@ -134,7 +139,7 @@ static int GetWriteTransaction(DBPriv *const db, DBTxn **const txn)
     if (db_txn->txn != NULL && !db_txn->rw_txn)
     {
         rc = mdb_txn_commit(db_txn->txn);
-        CheckLMDBCorrupted(rc, db->env);
+        CheckLMDBUsable(rc, db->env);
         if (rc != MDB_SUCCESS)
         {
             Log(LOG_LEVEL_ERR, "Unable to close read-only transaction in '%s': %s",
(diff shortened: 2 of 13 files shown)


vpodzime:
> [@vpodzime](/vpodzime) [vpodzime](/vpodzime) added the [onhold](/cfengine/core/labels/onhold) label [now](#event-3434397134)

I needed this PR and the first commit in it to get a build for HP-UX that would help me extract some valuable data from an LMDB file that I needed to analyze. The further progress with this ticket is not expected to happen in the upcoming week.

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/10046/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/10046/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-10046/

cf-check/diagnose.c
```diff
@@ -39,6 +39,11 @@ size_t diagnose_files(
 #include <validate.h>
 #include <openssl/rand.h>
 
+/* NOTE: Must be in sync with LMDB_MAXSIZE in libpromises/dbm_lmdb.c. */
+#ifndef LMDB_MAXSIZE
+#define LMDB_MAXSIZE    104857600
```
nickanderson, line 44: If this is not aligned, what happens?

vpodzime: `cf-check diagnose/repair` will have bad information about DB usage and `cf-check repair` may rotate DBs that don't need it or miss some DBs that do need it, depending on the mismatch.

olehermanse: Could consider adding a test for this. Can either grep both files and compare, or create a unit tests which compiles to binaries just to print the 2 constants and compare them.

libpromises/dbm_lmdb.c
```diff
@@ -79,13 +79,18 @@ static int DB_MAX_READERS = -1;
 /******************************************************************************/
 
 static void HandleLMDBCorruption(MDB_env *env, const char *msg);
+static void HandleFullLMDB(MDB_env *env);
 
-static inline void CheckLMDBCorrupted(int rc, MDB_env *env)
+static inline void CheckLMDBUsable(int rc, MDB_env *env)
```
olehermanse, line 84: I know you didn't introduce this, but IMO in cases like this a const is very nice, to prevent something like an accidental `if (rc = 1)` (always true):

```suggestion
static inline void CheckLMDBUsable(const int rc, MDB_env *env)
```

libpromises/dbm_lmdb.c
```diff
@@ -280,6 +286,43 @@ static bool RepairedAfterOpen(const char *lmdb_file, int fd_tstamp)
     return false;
 }
 
+/**
+ * @warning Expects @fd_stamp to be locked.
+ */
+static bool RotatedAfterOpen(const char *lmdb_file, int fd_tstamp)
+{
+    time_t rotated_tstamp = -1;
+    ssize_t n_read = read(fd_tstamp, &rotated_tstamp, sizeof(time_t));
```
olehermanse, line 295:
```suggestion
    const ssize_t n_read = read(fd_tstamp, &rotated_tstamp, sizeof(time_t));
```

libpromises/dbm_lmdb.c
```diff
@@ -490,6 +533,220 @@ static void HandleLMDBCorruption(MDB_env *env, const char *msg)
...
+
+    /* 1. Acquire the lock for the DB to prevent more processes trying to use
+     *    it while it is corrupted (wait till the lock is available). */
+    while (ExclusiveFileLock(&db_lock, false) == -1)
+    {
+        /* busy wait to do the logging */
+        Log(LOG_LEVEL_INFO, "Waiting for the lock on the '%s' DB",
+            lmdb_file);
+        sleep(1);
+    }
```
olehermanse, line 688: Forever?

libpromises/dbm_lmdb.c
```diff
@@ -490,6 +533,220 @@ static void HandleLMDBCorruption(MDB_env *env, const char *msg)
...
+        close(fd_tstamp);       /* releases locks */
+        free(db_lock_file);
+        free(tstamp_file);
+
+        exit(EC_CORRUPTION_REPAIRED);
+    }
+
+    /* 3. Rotate the DB or at least move it out of the way. */
+    ret = rotate_lmdb_file(lmdb_file, fd_tstamp);
+    bool rotation_successful = (ret == 0);
```
olehermanse, line 707:
```suggestion
    const bool rotation_successful = (ret == 0);
```

libpromises/dbm_quick.c
```diff
@@ -206,6 +206,12 @@ bool DBPrivClean(DBPriv *db)
     return true;
 }
 
+int DBPrivGetDBUsagePercentage(const char *db_path)
```
olehermanse, line 209:
```suggestion
int DBPrivGetDBUsagePercentage(ARG_UNUSED const char *db_path)
```

libpromises/dbm_tokyocab.c
```diff
@@ -245,6 +245,12 @@ bool DBPrivClean(DBPriv *db)
     return true;
 }
 
+int DBPrivGetDBUsagePercentage(const char *db_path)
```
olehermanse, line 248:
```suggestion
int DBPrivGetDBUsagePercentage(ARG_UNUSED const char *db_path)
```

cf-check/diagnose.c
```diff
@@ -524,6 +529,25 @@ static char *follow_symlink(const char *path)
     return xstrdup(target_buf);
 }
 
+bool lmdb_file_needs_rotation(const char *file, int *usage)
```
olehermanse, line 532:
```suggestion
bool lmdb_file_needs_rotation(const char *const file, int *const usage)
```

olehermanse (dismissed):
Looks great in general, left some smaller comments, but nothing criical.

cf-check/diagnose.c
vpodzime: Yeah a test would be nice. But we have multiple instances of the same thing so I'll leave to a separate PR.

libpromises/dbm_lmdb.c
vpodzime: Yup. If this gives up, the next instance of the same program will again end up here anyway. I guess I could make the log level increment over time to make this easier to detect, but that's as much as we can do, AFAICT.

libpromises/dbm_lmdb.c
olehermanse: This is not inline with our code style / guidelines:

https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#safety--correctness

I don't agree that it loses the point - it effectively communicates and enforces that this variable doesn't change through the function. I do agree that it's more important for larger chunks of code. I find it very helpful to see a variable and know that it's written to exactly once, and where / what value that is. Also, using language feature to prevent accidents like the one mentioned is good. Having checks is also nice, of course.

> We (should) have checks for this.

Maybe we do, maybe we don't. There are probably other examples, that might be harder to check for. The point is that it's safer and more readable / maintainable, IMO.

Not a hard requirement from my side, especially since you didn't introduce it, but I don't agree with the reasoning.

larsewi: ^^

libpromises/dbm_lmdb.c
```diff
@@ -490,6 +533,220 @@ static void HandleLMDBCorruption(MDB_env *env, const char *msg)
...
+     * ExclusiveFileUnlock above. */
+
+    if (rotation_successful)
+    {
+        exit(EC_CORRUPTION_REPAIRED);
+    }
+    else
+    {
+        exit(EC_CORRUPTION_REPAIR_FAILED);
+    }
```
larsewi, line 746:
```suggestion
    exit((rotation_successful) ? EC_CORRUPTION_REPAIRED
                               : EC_CORRUPTION_REPAIR_FAILED);
```

cf-check/diagnose.c
```diff
@@ -524,6 +529,25 @@ static char *follow_symlink(const char *path)
...
+        {
+            *usage = usage_pct;
+        }
+        return (usage_pct >= 95);
+    }
+    else
+    {
+        Log(LOG_LEVEL_ERR, "Failed to stat() '%s' when checking usage: %s", file, GetErrorStr());
+        return false;
+    }
```
larsewi, line 548:
```suggestion
    Log(LOG_LEVEL_ERR, "Failed to stat() '%s' when checking usage: %s", file, GetErrorStr());
    return false;
```

larsewi (approved):

libpromises/dbm_lmdb.c
vpodzime: Then we don't agree. And you are free to improve my code if you think it needs it.

vpodzime: > This is not inline with our code style / guidelines: https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#safety--correctness

IMHO, the guidelines actually refer to the case where you have `const *` parameters (you cannot alter a non-pointer parameter, of course, only the local copy of it).

libpromises/dbm_lmdb.c
```diff
@@ -491,6 +533,220 @@ static void HandleLMDBCorruption(MDB_env *env, const char *msg)
...
+{
+    const char *lmdb_file = mdb_env_get_userctx(env);
+    Log(LOG_LEVEL_CRIT, "'%s' DB full!", lmdb_file);
+
+    /* Freeze the DB ASAP. This also makes the call to exit() safe regarding
+     * this particular DB because exit handlers will ignore it. */
+    DBHandle *handle = GetDBHandleFromFilename(lmdb_file);
+    FreezeDB(handle);
+
+#ifdef __MINGW32__
```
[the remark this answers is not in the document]
vpodzime: Yup, we are very inconsistent with this. I'd say right now it doesn't really matter because the moment we want to do non-mingw-based builds, we will have to go through all `__MINGW32__` appearances and check them. But might as well use the proper one in new code. :+1:

libpromises/dbm_lmdb.c
vpodzime: Nope, thanks. :grin:

cf-check/diagnose.c
vpodzime: Also *nope, thanks*.

larsewi (approved):

PR #4224, opened 2020-06-17
verify_vars.c: Removed -1 returns from bool function

olehermanse:
Note that this may change behavior, I changed them to false,
because that's what makes sense to me, but previously they would
return `(bool) -1`, which is `1` / `true`.

### DIFF
--- a/libpromises/verify_vars.c
+++ b/libpromises/verify_vars.c
@@ -521,7 +521,7 @@ static bool CompareRval(const void *rval1_item, RvalType rval1_type,
 {
     if (rval1_type != rval2_type)
     {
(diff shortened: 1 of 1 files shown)


olehermanse:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5276/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5276/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5276/

PR #4234, opened 2020-06-19
Added tag for 32 bit systems to enable cf-remote install

craigcomstock:


cf-bottom:

contrib/cf-remote/cf_remote/remote.py
```diff
@@ -237,6 +237,8 @@ def install_host(
         if edition == "enterprise":
             tags.append("hub" if hub else "agent")
         tags.append("64" if data["arch"] in ["x86_64", "amd64"] else data["arch"])
+        if data["arch"] in ["i386", "i486", "i586", "i686"]:
```
vpodzime, line 240: What about `x86`? I think we use just that in some place too.

craigcomstock: You mean, if `uname -m` in get_info() got `x86`, set the 32 bit tag? Is it a possible value?

https://gcc.gnu.org/backends.html
https://gcc.gnu.org/onlinedocs/gcc-4.5.3/gcc/i386-and-x86_002d64-Options.html

Don't mention plain x86 or even specifically x86_64, but do mention i[3-6]86.

vpodzime: I didn't know where `data["arch"]` was coming from, I thought it was from our data somewhere.

PR #4236, opened 2020-06-19
CFE-3377 Fixed ps command on Alpine OS

craigcomstock:


craigcomstock:
I'm now considering just suggesting a change to alpine packaging: run configure with `CFLAGS="-D__BUSYBOX__"`

alpine-configure.sh
```diff
@@ -0,0 +1,9 @@
+./configure \
+  CFLAGS="-D__ALPINE__" \
```
craigcomstock, line 2: @olehermanse @vpodzime what do you think about just submitting a PR to the alpine packaging repo which adds ```CFLAGS="$CFLAGS -D__BUSYBOX__"```

I have tested this locally on an alpine laptop and seems to fix ps parsing nicely.

I will run tests next to check, there may be a lot of various commands changed to busybox versions in the C code because of this change.

libenv/sysinfo.c
```diff
@@ -458,7 +458,7 @@ static void GetNameInfo3(EvalContext *ctx)
      */
     strcpy(VSYSNAME.sysname, "android");
 #endif
-#ifdef __BUSYBOX__
+#if defined(__BUSYBOX__) || defined(__ALPINE__)
```
vpodzime, line 461: Would it make sense to do
```
#ifdef __ALPINE__
#define __BUSYBOX__
#endif
```
instead?

alpine-configure.sh
vpodzime: > a PR to the alpine packaging repo which adds CFLAGS="$CFLAGS -D__BUSYBOX__"

Where would it be added? And what about people compiling on Alpine Linux on their own? I'd prefer this would be detected in `configure.ac` auto-magically. And in there, it should add `-D__BUSYBOX__` to `CPPFLAGS` not `CFLAGS`.

craigcomstock: right, make ./autogen.sh work :+1: agreed.

craigcomstock:
Closing. Will work on making autogen.sh/configure.ac do the right thing and post a different PR.

PR #4242, opened 2020-06-24
Added example to show how to include this. vars in body file or common

craigcomstock:


craigcomstock:
example was included directly in the docs, so closing.

PR #4249, opened 2020-07-15
Wait for background processes before generating reports

vpodzime:
Reports generation only happens if there is exactly one
'cf-agent' process running (to avoid corruptions in the reporting
data). If some promise in the policy uses an action policy with
'background => true', there might still be a forked child process
evaluating the promise when the main process reaches the point
when it wants to generate reports. In such cases, it is important
that the main process waits for the forked child processes.

Ticket: ENT-6042
Changelog: Promises with 'action => bg()' no longer break reporting data


vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/5365/)

nickanderson:
So, I tested this and it seemed to work. Can I get a build for 3.15.x?

nickanderson:
Found small issue testing this, please see ticket for details.

vpodzime:
> Found small issue testing this, please see ticket for details.

That's a different story out of scope of this PR.

PR #4257, opened 2020-07-24
CFE-3379: Backgrounded commands are now correctly executed in the child process (3.15)

olehermanse:

PR #4259, opened 2020-07-24
CFE-3379: Stopped printing background log message on windows

olehermanse:
Just a small mistake from previous PR.


olehermanse:
Cherr-picked with other commits in: https://github.com/cfengine/core/pull/4257

PR #4254, opened 2020-07-22
ENT-5335 add workarounds for tests failing on SUSE.

Lex-2008:
Issue is that `usermod` on suse doesn't like editing seconday groups together
with any other attributes. In real life, this isn't a big issue - the promise
will be completely repaired on second run. However, this test failed.

This commit changes this behsvior to expect secondary group to *not* be set
after first promise run. When this get fixed, this commit should be reverted :)

merge together: https://github.com/cfengine/buildscripts/pull/760

PR #4277, opened 2020-08-13
ENT-5335 add workarounds for tests failing on SUSE - 3.12.x

Lex-2008:
Issue is that `usermod` on suse doesn't like editing secondary groups together
with any other attributes. In real life, this isn't a big issue - the promise
will be completely repaired on second run. However, this test failed.

This commit changes this behavior to expect secondary group to *not* be set
after first promise run. When this get fixed, this commit should be reverted :)

(cherry picked from commit 3d36de3949c0e4a5a1e7e7d3b3fabc1d6b822801)

----

#


Lex-2008:
backport of https://github.com/cfengine/core/pull/4254

PR #4298, opened 2020-09-04
CFE-3389 CFE-3409: Fixed compiler warnings for format truncation and pointer sign

olehermanse:
All of the pointer sign ones are fixed. Some format truncation remaining, but they are not trivial.

### DIFF
--- a/cf-serverd/server_classic.c
+++ b/cf-serverd/server_classic.c
@@ -728,8 +728,12 @@ char iscrypt, enterprise_field;
                 recvlen - CF_RSA_PROTO_OFFSET, crypt_len);
         }
 
-        int ret = RSA_private_decrypt(crypt_len, recvbuffer + CF_RSA_PROTO_OFFSET,
-                                      decrypted_challenge, PRIVKEY, RSA_PKCS1_PADDING);
+        int ret = RSA_private_decrypt(
+            crypt_len,
+            (const unsigned char *) recvbuffer + CF_RSA_PROTO_OFFSET,
+            (unsigned char *) decrypted_challenge,
+            PRIVKEY,
+            RSA_PKCS1_PADDING);
         if (ret < 0)
         {
             Log(LOG_LEVEL_ERR, "Authentication failure: "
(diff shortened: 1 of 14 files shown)


vpodzime (commented):
A couple things:

- I'd squash the commits changing the sizes of buffers in *sys_info.c* to 256, they are all the same

- I don't think the commit just type-casting the `char *` pointers to `unsigned char *` is a good idea -- it doesn't convert the values on platforms where `char` and `unsigned char` differ. Maybe we can just disable the warning in the specific file(s)?

- Please adapt the `CFLAGS` in *static-checks*.

Thanks for looking at this!

olehermanse:
> I don't think the commit just type-casting the char * pointers to unsigned char * is a good idea -- it doesn't convert the values on platforms where char and unsigned char differ. Maybe we can just disable the warning in the specific file(s)?

@vpodzime The typecasts I added are already there (implicitly) - there is no change in behavior. This is the behavior we want. Not a conversion, just a type cast. `char` and `unsigned char` differ on most systems, the norm is that `char` means `signed char`. `-127` becomes `128`, `-1` becomes `255` and so on. In these cases, for hashing and sending / receiving bytes, this is what we want.

As I noted  in the commit message, I'd like to get back to the signatures of these 3 functions:

* `HashString()`
* `SendTransaction()`
* `ReceiveTransaction()`

With those fixed, less type casts will be necessary.

olehermanse:
> I'd squash the commits changing the sizes of buffers in sys_info.c to 256, they are all the same

@vpodzime Squashed.

olehermanse:
> Please adapt the `CFLAGS` in *static-checks*.

@vpodzime Added.

olehermanse:
@cf-bottom jenkins, please.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5574/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5574/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5574/

vpodzime:
So we store the bytes with a wrong pointer type, but then send those bytes to another place where it again treats it as the original type. Amazing. But OK then.

vpodzime (commented):
Do we still need `-Wno-format-truncation` for GCC in *static-checks*?

vpodzime:
> Do we still need `-Wno-format-truncation` for GCC in _static-checks_?

Answered on Slack, yes, there are still some.

vpodzime (approved):
ACK

PR #4312, opened 2020-09-11
Only respect 'ifelapsed' if there is a previous promise evaluation

vpodzime:
If the promise is being evaluated for the first time (no records
in the locks LMDB), it doesn't make sense to check 'ifelapsed'
for it.


olehermanse (approved):
Makes sense, but I would like it if you added some more detail to the commit message. Does this fix an issue you've encountered? Is there a relevant ticket number?

vpodzime:
> Makes sense, but I would like it if you added some more detail to the commit message. Does this fix an issue you've encountered? Is there a relevant ticket number?

It fixed an issue I was seeing in the logs which got me to the next issue which I didn't want to spend time on. The ticket number would be ENT-6200, I guess, that's the test I was debugging.

PR #4315, opened 2020-09-14
Fixed unbounded write in policy_server.c

olehermanse:
I used StringCopy because it guarantees destination will be
terminated. strncpy does not.

Reported by LGTM:
https://lgtm.com/rules/2158250570/


vpodzime (approved):
ACK

PR #4318, opened 2020-09-18
ENT-6165 Updated contrib/masterfiles-stage scripts and instructions to be accurate

craigcomstock:
Hub works as installed but there are a few missing pieces in this core/contrib dir.


contrib/masterfiles-stage/README.org
```diff
@@ -60,7 +60,7 @@ earlier versions of git may or may not work.
 :ID:       2aeaaa9b-1229-4c14-b130-6d86e370de42
 :END:
 Installation simply consists of saving =masterfiles-stage.sh= and its supporting
-=common.sh= to the same directory on your policy server, making sure that
+=common.sh= and =ssh-wrapper.sh= to the same directory on your policy server, making sure that
```
nickanderson, line 63: Did you pull ssh-wrapper.sh from nova or mission-portal?

craigcomstock: there is only one in nova AFAICT

https://github.com/cfengine/nova/tree/master/api/http/dc-scripts

Can't find one in mission-portal.

craigcomstock: may be more files we need to pull/sync. Will re-test using these current docs and adjust. :+1:

craigcomstock: yes, in order to satisfy the examples we also need `git-askpass.sh`

nickanderson (approved):

PR #4321, opened 2020-09-23
Use a newer version of RHEL 8 in 'cf-remote spawn'

vpodzime:
It's RHEL 8.2 and it already has the updated SELinux policy that
we require.

Ticket: ENT-6103
Changelog: None


vpodzime:
We have been using the same AMI in jenkins since last week.

vpodzime:
oh, different region :disappointed:

vpodzime:
> oh, different region disappointed

Fixed.

Lex-2008 (approved):

PR #4322, opened 2020-09-23
CFE-3390: Small fixes to CFEngine promise module library

olehermanse:
Not in use yet, discovered while testing out promise modules.

Added necessary logging functions, and log levels to assertion.

Fixed result for repaired promises.

### DIFF
--- a/docs/custom_promise_types/cfengine.py
+++ b/docs/custom_promise_types/cfengine.py
@@ -90,7 +90,8 @@ class PromiseModule:
         log_level = request.get("log_level", "info")
         self._response["operation"] = operation
 
-        assert log_level == "info"
+        # Agent will never request log level critical
+        assert log_level in ["error", "warning", "notice", "info", "verbose", "debug"]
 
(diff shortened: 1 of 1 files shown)


docs/custom_promise_types/cfengine.py
```diff
@@ -141,14 +141,32 @@ def _log(self, level, message):
         self._out.write(f"log_{level}={message}\n")
         self._out.flush()
 
+    def log_critical(self, message):
```
craigcomstock, line 144: What are warning and critical for? Unused right?

olehermanse: https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#log-levels

craigcomstock (dismissed):

craigcomstock (approved):
excellent. Yes, the lack of warning and critical in the assert was my clue, and also no use in the library directly. Looks good!

PR #4330, opened 2020-10-02
Added demo script for testing CFEngine on a group of hosts

olehermanse:
I use this quite often when I want to test or demo something on a real setup.

### DIFF
--- a/docs/cf_remote_demo.sh
+++ b/docs/cf_remote_demo.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
+
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+
+cf-remote spawn --platform ubuntu-18-04-x64 --count 1 --name hub --role hub
+cf-remote spawn --platform ubuntu-18-04-x64 --count 1 --name ubuntu18 --role client
+cf-remote spawn --platform ubuntu-16-04-x64 --count 1 --name ubuntu16 --role client
+cf-remote spawn --platform ubuntu-14-04-x64 --count 1 --name ubuntu14 --role client
+cf-remote spawn --platform centos-6-x64 --count 1 --name centos6 --role client
+cf-remote spawn --platform centos-7-x64 --count 1 --name centos7 --role client
+cf-remote spawn --platform debian-7-x64 --count 1 --name debian7 --role client
+cf-remote spawn --platform debian-8-x64 --count 1 --name debian8 --role client
+cf-remote spawn --platform debian-9-x64 --count 1 --name debian9 --role client
+cf-remote spawn --platform rhel-5-x64 --count 1 --name rhel5 --role client
+cf-remote spawn --platform rhel-6-x64 --count 1 --name rhel6 --role client
+cf-remote spawn --platform rhel-7-x64 --count 1 --name rhel7 --role client
+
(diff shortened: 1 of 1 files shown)


docs/cf_remote_demo.sh
```diff
@@ -0,0 +1,37 @@
...
+cf-remote spawn --platform debian-7-x64 --count 1 --name debian7 --role client
+cf-remote spawn --platform debian-8-x64 --count 1 --name debian8 --role client
+cf-remote spawn --platform debian-9-x64 --count 1 --name debian9 --role client
+cf-remote spawn --platform rhel-5-x64 --count 1 --name rhel5 --role client
+cf-remote spawn --platform rhel-6-x64 --count 1 --name rhel6 --role client
+cf-remote spawn --platform rhel-7-x64 --count 1 --name rhel7 --role client
+
+cf-remote --version master install --demo --bootstrap hub --hub hub
+
+cf-remote deploy --hub hub /northern.tech/cfengine/masterfiles
```
vpodzime, line 20: Would be better to use a relative path here.

docs/cf_remote_demo.sh
```diff
@@ -0,0 +1,37 @@
...
+cf-remote --version master install --demo --bootstrap hub --clients ubuntu16
+cf-remote --version master install --demo --bootstrap hub --clients ubuntu14
+cf-remote --version master install --demo --bootstrap hub --clients centos6
+cf-remote --version master install --demo --bootstrap hub --clients centos7
+cf-remote --version master install --demo --bootstrap hub --clients debian7
+cf-remote --version master install --demo --bootstrap hub --clients debian8
+cf-remote --version master install --demo --bootstrap hub --clients debian9
+cf-remote --version master install --demo --bootstrap hub --clients rhel5
+cf-remote --version master install --demo --bootstrap hub --clients rhel6
+cf-remote --version master install --demo --bootstrap hub --clients rhel7
```
vpodzime, line 32: Hmm, would `ubuntu18,ubuntu16,ubuntu14,...` work?

docs/cf_remote_demo.sh
```diff
@@ -0,0 +1,37 @@
...
+
+cf-remote sudo -H centos6 "service iptables stop"
```
vpodzime, line 34: I think *centos7* needs `systemctl stop firewalld`.

vpodzime: And so does *rhel7* and *rhel6* (iptables).

vpodzime (commented):
Looks good to me otherwise.

docs/cf_remote_demo.sh
olehermanse: Done.

docs/cf_remote_demo.sh
olehermanse: Yes, but I find it handy that it's easy to comment out lines.

olehermanse:
@vpodzime I created this bug, because of the new suse rpms, this setup no longer works, cf-remote needs some adjustments before we merge this:

https://tracker.mender.io/browse/CFE-3445

olehermanse:
Fix for above mentioned bug: https://github.com/cfengine/core/pull/4335

docs/cf_remote_demo.sh
olehermanse: @vpodzime With the AMIs we use, those are not needed. This works (just tested it, all hosts reporting). So I don't think I need to add more commands.

vpodzime (approved):
ACK

PR #4333, opened 2020-10-06
CFE-3447: cf-remote: Now lists master as an available release

olehermanse:


contrib/cf-remote/cf_remote/packages.py
```diff
@@ -165,7 +165,11 @@ def __init__(self, edition="enterprise"):
             rel = Release(release)
             if "status" in release and release["status"] == "unsupported":
```
vpodzime, line 166: So *master* doesn't have `status`? Isn't that just a matter of time before it does and things break again? I'd just do `if release["version"] != "master":...`.

vpodzime (commented):
Looks good to me, but let's make this more @Lex-2008-proof :wink: (no offense, @Lex-2008 :grin:)

contrib/cf-remote/cf_remote/packages.py
Lex-2008: there is no "status" in https://cfengine.com/release-data/enterprise/releases.json, not anymore.

A brief of history: it was there, indeed, initially - when we thought that it would be great if `release.py` would specify which releases are supported, and which are not. However, back then `release.py` was running only during releases - and soon we realised that some versions can expire in-between releases. Hence, it was decided that `releases.json` file will include expiration dates for each branch, and it will be responsibility of the file reader to compare expiration dates with current dates to figure out which branches are supported. But "status" entries were left in the file for historical reasons, where @olehermanse saw them.

Lex-2008 (approved):
Another option would be to include `"lts_branch"` entry to the "master" release. Compare these two releases:
```
    "releases": [
        {
            "URL": "https://cfengine.com/release-data/enterprise/master.json", 
            "debug": true, 
            "release_date": "2020-10-07", 
            "version": "master", 
            "alpha": true
        }, 
        {
            "URL": "https://cfengine.com/release-data/enterprise/3.15.x.json", 
            "debug": true, 
            "release_date": "2020-10-07", 
            "version": "3.15.x", 
            "lts_branch": "3.15.x"
        }, 
```

Why `"lts_branch"` is in second, but not in first?
…
p.s. I can tell you why master has `"alpha": true` - because its "version" field has letter "a"! :smile:

olehermanse:
@Lex-2008 I don't think master should have an `lts_branch` because master does not have an LTS branch.

PR #4335, opened 2020-10-06
CFE-3445: cf-remote: Fixed package selection on RHEL

olehermanse:
cf-remote was picking SUSE rpms on RHEL machines, because
it was missing the extra tags we added for CentOS.


vpodzime (approved):
ACK

PR #4338, opened 2020-10-08
--audit=manifest mode

vpodzime:
TODO:
- [x] test manually
- [x] make things build on Windows
- [x] get a :green_heart: jenkins

PR #4353, opened 2020-10-19
ENT-5294: --audit=diff mode

vpodzime:
TODO:
- [x] green jenkins build
- [x] manual testing
- [x] merge #4338 first
- [x] make sure we have *diff* available everywhere

### DIFF
--- a/cf-agent/audit_mode.c
+++ b/cf-agent/audit_mode.c
@@ -408,7 +408,6 @@ static bool RunDiff(const char *path1, const char *path2)
 
 static bool DiffFile(const char *path)
 {
-    PrintDelimiter();
     const char *chrooted_path = ToChangesChroot(path);
 
     struct stat st_orig;
@@ -421,6 +420,7 @@ static bool DiffFile(const char *path)
         return true;
     }
 
+    PrintDelimiter();
     if (lstat(chrooted_path, &st_chrooted) == -1)
     {
         /* TODO: should this do print info about the original file? */
--- a/cf-agent/audit_mode.c
+++ b/cf-agent/audit_mode.c
@@ -30,6 +30,8 @@
 #include <file_lib.h>           /* IsAbsoluteFileName() */
 #include <dir.h>                /* DirOpen(),...*/
 #include <string_lib.h>         /* StringEqual() */
+#include <string_sequence.h>    /* ReadLenPrefixedString() */
+#include <changes_chroot.h>     /* CHROOT_CHANGES_LIST_FILE */
 
 #include <audit_mode.h>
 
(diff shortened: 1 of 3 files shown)


vpodzime:
> This pull request **introduces 4 alerts** when merging [900d23d](https://github.com/cfengine/core/commit/900d23d1bb0c9fe5f3faa44498f13215f76542e4) into [e8db8f4](https://github.com/cfengine/core/commit/e8db8f4a18a199962494bcab0ab5647058cc836c) - [view on LGTM.com](https://lgtm.com/projects/g/cfengine/core/rev/pr-0e5f81c34de3a57c5c9ab54c433f559f9e5664b3)

No new alerts here, it's just code moved around.

vpodzime:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/5802/)

cf-agent/audit_mode.c
```diff
@@ -342,7 +345,104 @@ bool ManifestRename(const char *orig_name, const char *new_name)
...
 
-bool ManifestChangedFiles()
+static bool RunDiff(const char *path1, const char *path2)
+{
+    char diff_path[PATH_MAX];
+    strncpy(diff_path, GetBinDir(), sizeof(diff_path));
+    JoinPaths(diff_path, sizeof(diff_path), "diff");
+    char *command;
+    int ret = xasprintf(&command, "%s -u --label 'original %s' --label 'changed  %s' '%s' '%s'",
+                        diff_path, path1, path1, path1, path2);
```
olehermanse, line 355: Should it be `1, 1, 1, 2` or `1, 2, 1, 2`? If it is correct how you wrote it, that both labels should have `path1`, then I think this deserves a comment.

cf-agent/audit_mode.c
```diff
@@ -342,7 +345,104 @@ bool ManifestRename(const char *orig_name, const char *new_name)
...
+                to_write -= n_written;
+            }
+            else
+            {
+                Log(LOG_LEVEL_ERR, "Failed to print results from 'diff' for '%s' and '%s'",
+                    path1, path2);
+                failure = true;
+                break;
+            }
+        }
```
olehermanse, line 381: I think there are problems with this while loop.

Here is the man page for fwrite:

```
     The functions fread() and fwrite() advance the file position indicator
     for the stream by the number of bytes read or written.  They return the
     number of objects read or written.  If an error occurs, or the end-of-
     file is reached, the return value is a short object count (or zero).

     The function fread() does not distinguish between end-of-file and error;
     callers must use feof(3) and ferror(3) to determine which occurred.  The
     function fwrite() returns a value less than nitems only if a write error
     has occurred.
```

At the very least, I think you need to advance `buf`, by creating and updating a temporary pointer / index, so that the retry tries to write from the right offset, in case of a short write.

Also, it seems fwrite should only return a short count in case of errors, is it better to log error then or retry?

cf-agent/audit_mode.c
```diff
@@ -342,7 +345,104 @@ bool ManifestRename(const char *orig_name, const char *new_name)
...
+    {
+        switch (st_chrooted.st_mode & S_IFMT) {
+        case S_IFREG:
+        case S_IFDIR:
+            return RunDiff(path, chrooted_path);
+        default:
+            printf("'%s' is a %s\n", path, GetFileTypeDescription(st_chrooted.st_mode));
+            ManifestStatInfo(&st_chrooted);
+            ManifestFileDetails(path, &st_chrooted, true);
+        return true;
```
olehermanse, line 440: Weird indentation.

cf-agent/audit_mode.c
```diff
@@ -342,7 +345,104 @@ bool ManifestRename(const char *orig_name, const char *new_name)
...
+        ManifestStatInfo(&st_chrooted);
+        ManifestFileDetails(path, &st_chrooted, true);
+        return true;
+    }
+    else
+    {
+        switch (st_chrooted.st_mode & S_IFMT) {
+        case S_IFREG:
+        case S_IFDIR:
+            return RunDiff(path, chrooted_path);
```
olehermanse, line 435: TIL about `diff` of two directories.

olehermanse (dismissed):
LGTM. Please look at my comments and wait for diff utility to be ready before merging.

cf-agent/audit_mode.c
vpodzime: It is correct because the output then looks like this:
```
--- original /etc/motd.d/cfengine
+++ changed  /etc/motd.d/cfengine
```
I'll add an example output. :+1:

cf-agent/audit_mode.c
vpodzime: Yup, good catches! It's a code moved around so I'll add a new commit with a fix. I was actually reading the man page of `write()` and that's not so specific about the reasons for a failure.

vpodzime: > Also, it seems fwrite should only return a short count in case of errors, is it better to log error then or retry?

Actually, I think it's better to retry and log an error if the retry doesn't write anything more.

vpodzime:
@cf-bottom one more jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5841/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5841/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5841/

cf-agent/audit_mode.c
olehermanse: Makes sense 👍

olehermanse (dismissed):

olehermanse (approved):

PR #4364, opened 2020-10-22
CFE-3391: if for custom promises

olehermanse:
Didn't have to implement this, it already worked. I added a test,
removed it from the list of attributes which are not yet
implemented, and made sure the if attribute is not sent to the
module.


tests/acceptance/30_custom_promise_types/02_if.cf
```diff
@@ -0,0 +1,129 @@
...
+        not(fileexists("$(G.testfile)-8")),
+        not(fileexists("$(G.testfile)-9")),
+        not(fileexists("$(G.testfile)-10")),
+      };
+
+  reports:
+    ok::
+      "$(this.promise_filename) Pass";
+    !ok::
+      "$(this.promise_filename) FAIL";
```
nickanderson, line 128: instead of these explicit pass/fail reports, consider: `methods: "Pass/FAIL" usebundle => dcs_passif( "ok", $(this.promise_filename) ), inherit => "true";`

nickanderson (dismissed):
One suggestion for leveraging dcs bundles, but looks good to me.

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5847/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5847/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5847/

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5848/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5848/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5848/

nickanderson (commented):
You haven't convinced me that ifvarclass should behave differently than if. ifvarclass was the name for the attribute originally, later we added if, and now it's the preffered name, but I don't think that ifvarclass should explicitly not work, nor should it work differently.

olehermanse:
> You haven't convinced me that ifvarclass should behave differently than if. ifvarclass was the name for the attribute originally, later we added if, and now it's the preffered name, but I don't think that ifvarclass should explicitly not work, nor should it work differently.

@nickanderson With custom promise types, it's much better to be restrictive with what we support from the beginning. ifvarclass is a terrible name, it makes the policy harder to read. I believe we should deprecate or at least strongly encourage users to move away from it. It also has and has had many problems in the way it's implemented. I thought it was much more sensible to just disallow it in custom promises, than to try to validate things correctly. For example, I don't think you should be able to use if and ifvarclass together. Also, because of how it's implemented, I think it doesn't work correctly if you use `ifvarclass` and a class guard with variable expansion. To me, that is a mess, and I think it's much better to just error if you try to use ifvarclass, with a helpful error message telling you to use if instead.

nickanderson:
> @nickanderson With custom promise types, it's much better to be restrictive with what we support from the beginning. ifvarclass is a terrible name, it makes the policy harder to read. I believe we should deprecate or at least strongly encourage users to move away from it. It also has and has had many problems in the way it's implemented. I thought it was much more sensible to just disallow it in custom promises, than to try to validate things correctly. For example, I don't think you should be able to use if and ifvarclass together. Also, because of how it's implemented, I think it doesn't work correctly if you use `ifvarclass` and a class guard with variable expansion. To me, that is a mess, and I think it's much better to just error if you try to use ifvarclass, with a helpful error message telling you to use if instead.

I agree it's a bad name.
I agree that I don't think that you should be able to use if and ifvarclass on the same promise.
I agree that we should encourage people to use if instead, but via documentation, not implementation.
I think that ifvarclass /should/ work with variable class expressions, we should fix it if it's not working, I guess we fixed it only in the case of if.
I don't think that ifvarclass and if should have different behaviour.
I don't think that we should explicitly prevent ifvarclass from working.

vpodzime (approved):
Looks good to me.

PR #4368, opened 2020-10-23
CFE-3464: Added `--remote-download` argument to cf-remote

aleksandrychev:
Added `--remote-download` argument to the cf-remote that enables to download packages directly to the target machine with curl


aleksandrychev:
@olehermanse please review

olehermanse (approved):
Great, thanks @aleksandrychev! I haven't tested it yet, but I'll make sure I do next week.

PR #4362, opened 2020-10-21
CFE-3470: Changed return type of and(), not(), or() to boolean (context)

olehermanse:
Let's see where this blows up.

PR #4371, opened 2020-10-23
CFE-3470: Added changelog entry for unexpected consequence of change in and(), or(), not()

olehermanse:
This commit changed the return type of and(), or(), not():
https://github.com/cfengine/core/pull/4362/commits/c63152c0876d537c71fea52ad8e43df963e0215d

They now match all the other boolean functions.
This is good, but it means in some edge cases the type checking is affected.
This was not (ab)used in any of our policy or tests, so it should be a fairly rare edge case.

```
    Changelog: and(), or(), not() now return boolean and cannot be used directly in slist vars.
               They can now be used in other places where a boolean is expected.
               (Most notably and / or promise attributes).
               The return values can be converted to strings using concat(), if necessary.
```


vpodzime (approved):
ACK

PR #4385, opened 2020-11-02
Use Fedora 33 as the base image for static checks

vpodzime:
Freshly released with newer versions of compilers and static
checkers.


vpodzime:
Manually checked that things pass with the F33 image.

olehermanse (approved):
👍

PR #4390, opened 2020-11-05
ENT-6205 add ifelapsed locks test without abort

craigcomstock:
Goal is to merge this test and let it run in CI for a while (a week?) and hope that the new test fails on solaris sparc at some point, or that the test passes in a run where the original redmine test fails.

### DIFF
--- a/tests/acceptance/00_basics/ifelapsed_and_expireafter/locks.cf.sub
+++ b/tests/acceptance/00_basics/ifelapsed_and_expireafter/locks.cf.sub
@@ -1,0 +1,52 @@
+# Test that locking works
+# Called twice from main.cf
+
+bundle agent example
+{
+  methods:
+      "test" usebundle => test, action => now;
+}
+
+bundle agent test
+{
+  vars:
+    first::                                                      
+       "count" string => "first";
+    second::
+       "count" string => "second";
+
+  methods:
+# should print "Called from test" for both runs
+      "Report"
+      usebundle => report_now($(this.bundle)),
+      action => now;
+
+# should print "Called from call_report_now" for both runs
+      "Nested Report"
+      usebundle => call_report_now,
+      action => now;
+
+# should print "Called from only_first" only for the first run
+      "Only first"
+      usebundle => report_now("only_first");
+}
+
+bundle agent call_report_now
+{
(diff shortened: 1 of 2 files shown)


craigcomstock:
Looks like both tests are currently passing on solaris sparc64 10 and 11:

```
09:58:47 ./00_basics/ifelapsed_and_expireafter/locks.cf Pass
09:59:03 ./00_basics/ifelapsed_and_expireafter/redmine_6334.cf Pass
```

If travis is green, I'd like to merge this PR @nickanderson

tests/acceptance/00_basics/ifelapsed_and_expireafter/locks.cf.sub
```diff
@@ -0,0 +1,52 @@
+# Test that locking works
+# Called twice from main.cf
```
nickanderson, line 2: i wuld jiust clarify this to say that cf-agent is run with this policy as the entry twice. so that there isn't confusion that a bundle in here is called twice within one agent run.

nickanderson (commented):
You are trying to test ignoring locks for specific promises using action bodies?

e.g. that a methods promise will be executed in the event of cf-agent running the same policy over separate executions within quick succession (1m by default)?

tests/acceptance/00_basics/ifelapsed_and_expireafter/locks.cf.sub
craigcomstock: Right.

craigcomstock:
> You are trying to test ignoring locks for specific promises using action bodies?
> 
> e.g. that a methods promise will be executed in the event of cf-agent running the same policy over separate executions within quick succession (1m by default)?

Well. I am specifically trying to prove-out whether the inclusion of the abort in the `00_basics/ifelapsed_and_expireafter/redmine_6334.cf` test is causing the failures on solaris sparc64 systems periodically.

So if you don't see value in this test intrinsically we could certainly schedule for reverting this commit after we see if this removal of abort can shed light on the intermittent failures in `00_basics/ifelapsed_and_expireafter/redmine_6334.cf`.

nickanderson (approved):

PR #4401, opened 2020-11-19
CFE-3499: Fixed cf-remote destroy command for 0 instances

olehermanse:


contrib/cf-remote/cf_remote/spawn.py
```diff
@@ -367,7 +367,10 @@ def spawn_vms(vm_requests, creds, region, key_pair=None, security_groups=None,
 
 
 def destroy_vms(vms):
-    with Pool(len(vms)) as pool:
+    n = len(vms)
+    if (n == 0):
+        return
```
vpodzime, line 372: I'd do `if not vms: return`. It's more generic.

olehermanse: Very well. Done.

vpodzime (approved):
ACK

PR #4408, opened 2020-11-24
Add the '--append' option to 'cf-remote spawn'

vpodzime:
So that new VMs can be added to an existing group of VMs which
allows creation of big groups of VMs in batches (e.g. to avoid
rate limits on cloud APIs).

Ticket: CFE-3502
Changelog: 'cf-remote spawn' now supports adding new VMs to an existing group

### DIFF
--- a/contrib/cf-remote/cf_remote/commands.py
+++ b/contrib/cf-remote/cf_remote/commands.py
@@ -257,7 +257,7 @@ def download(tags=None, version=None, edition=None):
     return _iterate_over_packages(tags, version, edition, True)
 
 def spawn(platform, count, role, group_name, provider=Providers.AWS, region=None,
-          size=None, network=None, public_ip=True):
+          size=None, network=None, public_ip=True, extend_group=False):
     if os.path.exists(CLOUD_CONFIG_FPATH):
         creds_data = read_json(CLOUD_CONFIG_FPATH)
     else:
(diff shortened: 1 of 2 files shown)


Lex-2008 (approved):

contrib/cf-remote/cf_remote/commands.py
```diff
@@ -298,8 +299,16 @@ def spawn(platform, count, role, group_name, provider=Providers.AWS, region=None
 
         region = region or creds_data["gcp"].get("region", "europe-west1-b")
 
+    # TODO: Do we have to complicate this instead of just assuming existing VMs
+    # were created by this code and thus follow the naming pattern from this
+    # code?
+    if group_exists:
+        range_start = len(vms_info[group_key]) - 1
```
olehermanse, line 306: The `- 1` seems incorrect? if there are 1 vms, the starting index should be 1, not 0, right?

olehermanse (commented):
Looks good overall, just 1 question.

contrib/cf-remote/cf_remote/commands.py
vpodzime: No, because we use `range(N)` if the group doesn't exist so the VM indices start with 0.

olehermanse: @vpodzime First run spawns 3 VMs:
```
range(3) -> (0, 1, 2)
```

VMs have indices 0, 1 and 2.

Next run does:

```
len(vms) -> 3
```

That's the correct starting index for the next batch, we want:

```
range(len(vms), len(vms) + count) -> (3, 4, 5, ...)
```

What am I missing?

vpodzime: You are right, of course. I wasn't thinking about this enough. :grin: The reason for the `-1` is this extra object that we add to the group in the JSON file:
```
    "meta": {
      "provider": "gcp",
      "region": "europe-west1-b"
    }
```
I'll make sure the code doesn't count this key in.

Lex-2008 (approved):

PR #4420, opened 2020-12-08
CFE-3519 Added gpg_keys promise type

craigcomstock:
Use a directory of .asc (public key) and .fp (fingerprint) files
to ensure gpg keys are up to date.

```
gpg_keys:
    "/home/user/.gnupg"
      keydir => "/home/user/.password-store/.pub-keys";
```

Ticket: CFE-3519
Changelog: title

- [x] refactor to keep high-level logic first, hide details in functions
- [x] cleanup code
- [x] re-test after refactor
- [x] rework the example policy

### DIFF
--- a/docs/custom_promise_types/password-store-sync.cf
+++ b/docs/custom_promise_types/password-store-sync.cf
@@ -1,0 +1,33 @@
+promise agent git
+{
+  path => "/var/cfengine/modules/promises/git_using_lib.py";
+  interpreter => "/usr/bin/python3";
+}
+
+promise agent gpg_keys
+{
+  path => "/var/cfengine/modules/promises/gpg.py";
+  interpreter => "/usr/bin/python3";
+}
+
+bundle agent main
+{
+  vars:
+    # for gpg_keys promise, pass a keylist data similar to this proposed standard with the addition of the `ascii` property
+    # https://datatracker.ietf.org/doc/draft-mccain-keylist/ (from https://github.com/firstlookmedia/gpgsync project)
+  #  "keylist" data => readjson("/home/user/keylist.json");
+
+    # Or assembly the information using some policy data manipulation and jq (https://stedolan.github.io/jq/) magic
+    "fingerprint_files" slist => findfiles("/home/user/.password-store/.pub-keys/*.fp");
+    "keys[$(fingerprint_files)][fingerprint]" string => readfile( $(fingerprint_files) );
+    "keys[$(fingerprint_files)][ascii]" string => readfile( regex_replace( $(fingerprint_files), ".fp$", ".asc", "") );
+    "keylist" data => mapdata("json_pipe", '$(def.jq) {"keys":[.[]]}', keys);
+
+  git:
+    "/home/user/.password-store"
+      repo => "https://github.com/user/my-passwords";
+
+  gpg_keys:
+    "/home/user/.gnupg"
+      keylist => storejson( @(keylist[0]) );
+}
--- a/docs/custom_promise_types/gpg.py
+++ b/docs/custom_promise_types/gpg.py
@@ -1,0 +1,154 @@
+"""Custom Promise Type module for adding missing gpg keys
+
+Given some GPG user id values (see man page for all possiblities, there are many)
+and the ascii value of a key, ensure that the key is present in the given GPG homedir.
+Typically the user id will be a key fingerprint or an email address.
+
+Installation:
+  cp gpg.py $(sys.masterdir)/modules/promises
+
+typically:
+  cp gpg.py /var/cfengine/masterfiles/modules/promises
+
+Usage:
+
+Provide the user id and ascii values to this promise with the `keylist` attribute in
+a data container with the format:
+
+similar to this standard: https://datatracker.ietf.org/doc/draft-mccain-keylist/
+
+```json
+{
+  "keys": [
+    {
+      "user_id": "86EB84C96B2E62676B47C4919BB29FF9FD3ED09F",
+      "ascii": "<large chunk of ascii public key content for importing"
+    },
+  ]
+}
+```
+
+The promiser in this case is the GPG homedir. Typically `$HOME/.gnupg`
+
+```cfe3
+bundle agent main
+{
+  gpg:
+    "$(sys.user_data[home_dir])$(const.dirsep).gnupg"
+      keylist => readjson("$(this.promise_dirname)/keylist.json");
+}
+```
+
+"""
+
+import json
+from subprocess import Popen, PIPE
+import sys
+from cfengine import PromiseModule, ValidationError, Result
+
+class GpgKeysPromiseTypeModule(PromiseModule):
+    def __init__(self):
+        super().__init__("gpg_keys_promise_module", "0.0.1")
+
+    def gpg_import_ascii(self, homedir, ascii):
+        with Popen(
+            ["gpg", "--homedir", f"{homedir}", "--import"],
+            stdout=PIPE,
+            stdin=PIPE,
+            stderr=PIPE,
+        ) as proc:
+            try:
+                sdtout, stderr = proc.communicate(input=ascii.encode())
+                if proc.returncode == 0:
+                    return True
+                else:
+                    self.log_error(
+                        f"Error importing gpg key return code '{proc.returncode}'"
+                    )
+                    self.log_verbose(
+                        f"Import gpg key failed, stderr was '{stderr.decode()}'"
+                    )
+                    return False
+            except TimeoutExpired:
+                proc.kill()
+                proc.communicate()
+                self.log_error("Timed out importing gpg key")
+                return False
+
+    def clean_storejson_output(self, storejson_output):
+        # workaround custom promise types not supporting data container or slist attribute values
+        # this function cleans up the result of sending a data container or slist with
+        # attribute => storejson( @(data) )
+
+        return storejson_output.replace('\\"', '"').replace("\n", "")
+
+    def gpg_key_present(self, homedir, user_id):
+        with Popen(
+            ["gpg", "--homedir", f"{homedir}", "-k", f"{user_id}"],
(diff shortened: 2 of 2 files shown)


docs/custom_promise_types/gpg_keys_using_lib.py
```diff
@@ -0,0 +1,45 @@
...
+          if entry.name.endswith('.fp') and entry.is_file():
+            fp_filename = keydir + os.sep + entry.name
+            fp_file = open(fp_filename, "r")
+            fp = fp_file.read()
+            fp = fp.replace('\n','')
+            fp_file.close()
+            self.log_debug(f"file {fp_filename} has fingerprint {fp}")
+            rc = os.system(f"gpg --homedir {promiser} -k {fp} 2>/dev/null >/dev/null")
+            if rc != 0:
+              asc_file = keydir + os.sep + entry.name.replace(".fp",".asc")
```
olehermanse, line 30: Might not be a big deal, but this fails if ".fp" appears anywhere else in the name. You could write a helper to do exactly what you want:

```python
replace_end(entry.name, ".fp", ".asc")
```

```python
def replace_end(string, remove, append):
    if not string.endswith(remove):
        raise ValueError
    return string[0:-len(remove)] + append
```

docs/custom_promise_types/gpg_keys_using_lib.py
```diff
@@ -0,0 +1,45 @@
...
+      with os.scandir(keydir) as it:
+        for entry in it:
+          if entry.name.endswith('.fp') and entry.is_file():
+            fp_filename = keydir + os.sep + entry.name
+            fp_file = open(fp_filename, "r")
+            fp = fp_file.read()
+            fp = fp.replace('\n','')
+            fp_file.close()
+            self.log_debug(f"file {fp_filename} has fingerprint {fp}")
+            rc = os.system(f"gpg --homedir {promiser} -k {fp} 2>/dev/null >/dev/null")
```
olehermanse, line 28: I think you should add a comment to explain what this command actually does.

docs/custom_promise_types/gpg_keys_using_lib.py
```diff
@@ -0,0 +1,45 @@
...
+            if rc != 0:
```
olehermanse, line 29: And a comment here to explain what the return code means.

docs/custom_promise_types/gpg_keys_using_lib.py
```diff
@@ -0,0 +1,45 @@
...
+              asc_file = keydir + os.sep + entry.name.replace(".fp",".asc")
+              self.log_info(f"Importing {asc_file} public key file with fingerprint {fp} into GNUPGHOME {promiser}")
```
olehermanse, line 31: Again, I think this log message would be less confusing if you deleted the `GNUPGHOME`.

docs/custom_promise_types/gpg_keys_using_lib.py
```diff
@@ -0,0 +1,45 @@
...
+              rc = os.system(f"gpg --homedir {promiser} --import {asc_file} 2>/dev/null >/dev/null")
```
olehermanse, line 32: I guess this won't work for folders with spaces, and you should add quotes to address this.

docs/custom_promise_types/gpg_keys_using_lib.py
```diff
@@ -0,0 +1,45 @@
...
+# TODO handle errors with os calls and such
+      with os.scandir(keydir) as it:
+        for entry in it:
+          if entry.name.endswith('.fp') and entry.is_file():
+            fp_filename = keydir + os.sep + entry.name
+            fp_file = open(fp_filename, "r")
+            fp = fp_file.read()
+            fp = fp.replace('\n','')
+            fp_file.close()
+            self.log_debug(f"file {fp_filename} has fingerprint {fp}")
```
olehermanse, line 27: F in file should be capitalized.

docs/custom_promise_types/gpg_keys_using_lib.py
```diff
@@ -0,0 +1,45 @@
...
+            rc = os.system(f"gpg --homedir {promiser} -k {fp} 2>/dev/null >/dev/null")
+            if rc != 0:
+              asc_file = keydir + os.sep + entry.name.replace(".fp",".asc")
+              self.log_info(f"Importing {asc_file} public key file with fingerprint {fp} into GNUPGHOME {promiser}")
+              rc = os.system(f"gpg --homedir {promiser} --import {asc_file} 2>/dev/null >/dev/null")
+              if rc != 0:
+                self.log_error(f"Unable to import key file {asc_file} for fingerprint {fp}, exit code {rc}")
```
olehermanse, line 34: I like to put single quotes around strings of data in log messages. It is especially useful if that variable might contain a newline, space, tab or something similar. Makes it much easier to see where it starts and ends.

olehermanse (changes requested):
Looks good overall @craigcomstock , many smaller comments and suggestions.

docs/custom_promise_types/gpg_keys_using_lib.py
craigcomstock: I log an error right after this explaining right?

craigcomstock: ah, I don't, I was looking at the OTHER `rc != 0`. Agreed, will comment.

docs/custom_promise_types/gpg_keys_using_lib.py
```diff
@@ -0,0 +1,61 @@
...
+                        rc = os.system(
+                            f"gpg --homedir '{promiser}' --import '{asc_file}' 2>/dev/null >/dev/null"
+                        )
+                        if rc != 0:
+                            self.log_error(
+                                f"Unable to import key file '{asc_file}' for fingerprint '{fp}', exit code '{rc}'"
+                            )
+                            result = Result.NOTKEPT
+                        else:
+                            if result != Result.NOTKEPT:
```
olehermanse, line 53: Should still be `Result.NOT_KEPT`.

olehermanse (changes requested):
Looks good overall, please fix the last comment and also squash commits.

docs/custom_promise_types/gpg_keys_using_lib.py
craigcomstock: oh no! ;) Yep, will fix and squash.

olehermanse (changes requested):
@craigcomstock Please re-request review once you've addressed the last few comments from my first review 🐈

craigcomstock:
refactored quite a bit. still WIP now though as I need to test/cleanup. :)

docs/custom_promise_types/password-store-sync.cf
```diff
@@ -0,0 +1,22 @@
+promise agent git
```
craigcomstock, line 1: @olehermanse not sure what to do with an example. How to manage putting a few examples in here and getting them to "work easy" for new users? Or just leave out the example .cf now that I have added a complete enough example to the python module?

olehermanse: This example looks fine to me.

craigcomstock: fixed up based on my refactoring.

docs/custom_promise_types/gpg_keys_using_lib.py
```diff
@@ -0,0 +1,154 @@
...
+                    return True
+                else:
+                    self.log_verbose(
+                        f"Querying gpg key failed, stderr was '{stderr.decode()}'"
+                    )
+                    return False
+            except TimeoutExpired:
+                proc.kill()
+                proc.communicate()
+                self.log_error("Timed out querying for gpg key '{user_id}'")
```
olehermanse, line 104: Missing `f` in front of `"` to make this an f-string (formatted string).

docs/custom_promise_types/gpg_keys_using_lib.py
```diff
@@ -0,0 +1,154 @@
+"""Custom Promise Type module for adding missing gpg keys
```
olehermanse, line 1: The only reason why I called my modules `from_scratch` and `using_lib` was because there were 2 different ones. I think this one should be just `gpg.py`, and the promise type should be `gpg`.

docs/custom_promise_types/gpg_keys_using_lib.py
```diff
@@ -0,0 +1,154 @@
...
+
+Given some GPG user id values (see man page for all possiblities, there are many)
+and the ascii value of a key, ensure that the key is present in the given GPG homedir.
+Typically the user id will be a key fingerprint or an email address.
+
+Installation:
+  cp gpg_keys_using_lib.py $(sys.workdir)/modules/promises
```
olehermanse, line 8: OK, but if you are using masterfiles, the preferred location is `masterfiles/modules/promises/`, which will be copied there automatically.

olehermanse (approved):
Looks good overall, just some smaller comments.

docs/custom_promise_types/password-store-sync.cf
```diff
@@ -0,0 +1,33 @@
...
+{
+  path => "/var/cfengine/modules/promises/git_using_lib.py";
+  interpreter => "/usr/bin/python3";
+}
+
+promise agent gpg_keys
+{
+  path => "/var/cfengine/modules/promises/gpg_keys_using_lib.py";
```
olehermanse, line 9: Outdated now, no longer `using_lib`.

olehermanse (changes requested):

docs/custom_promise_types/password-store-sync.cf
craigcomstock: yep, fixed. thanks! :)

olehermanse (approved):
LGTM, thanks :)

PR #4423, opened 2020-12-10
CFE-2507: Policy function storejson() no longer truncates strings lager than 4096 bytes (3.12)

larsewi:
Policy function storejson() no longer truncates strings lager than 4096 bytes


vpodzime (approved):
ACK

PR #4431, opened 2020-12-17
CFE-2686: Policy function format() no longer truncates strings lager than 4KiB (3.15.x)

larsewi:
- Added edge case test files for acceptance test.
 - Policy function format() no longer truncates strings lager than 4KiB.
 - Removed commented-out code.
 - Added explicit comparisons.
 - Removed `snprintf` and assigned a string literal instead.


vpodzime (approved):
ACK

PR #4438, opened 2020-12-21
ENT-6633 cf-remote: Fixed bug with package selection

olehermanse:
Introduced in:
8b3f2dd1f31832980e2fa66a3cdcf94ba2e63549


craigcomstock (approved):
Yes. Makes a LOT of sense given what I had investigated so far. :+1:

PR #4443, opened 2021-01-07
Fixed vm-check libxml dependency package name

craigcomstock:
Ticket: none
Changelog: none

### DIFF
--- a/contrib/vagrant-ci/centos-7-x64/Vagrantfile
+++ b/contrib/vagrant-ci/centos-7-x64/Vagrantfile
@@ -43,10 +43,12 @@ SHELL
     nd.vm.hostname = no_deps_vm_name.to_sym
 
     nd.vm.provision "shell", inline: <<-SHELL
-yum -q -y install lmdb-devel pcre-devel openssl-devel libyaml-devel libxml-devel
+set -e
+yum -q -y install lmdb-devel pcre-devel openssl-devel libyaml-devel libxml2-devel
 SHELL
 
     nd.vm.provision "shell", inline: <<-SHELL
+set -e
 shopt -s expand_aliases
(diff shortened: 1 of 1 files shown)


craigcomstock:
@vpodzime I also added `-e` to these scripts so that next time something like this package name changes and breaks things later we catch it sooner.

contrib/vagrant-ci/centos-7-x64/Vagrantfile
```diff
@@ -43,10 +43,12 @@ SHELL
     nd.vm.hostname = no_deps_vm_name.to_sym
 
     nd.vm.provision "shell", inline: <<-SHELL
-yum -q -y install lmdb-devel pcre-devel openssl-devel libyaml-devel libxml-devel
+set -e
+yum -q -y install lmdb-devel pcre-devel openssl-devel libyaml-devel libxml2-devel
```
olehermanse, line 47: Seems like the commit message should have some explanation. Was this working before? Was there a change from CentOS 6 to 7? Has this always been wrong?

craigcomstock: @vpodzime had mentioned that before with centos-6 when he finished this it was passing all tests.

We don't run this on a regular basis so not sure when it failed.

Checking online repos of centos6 it seems like libxml2-devel is present and not libxml-devel so maybe a change in centos6 recently as well.

vpodzime: I think maybe I was just lucky and something pulled in `libxml2-devel` too or it was already in the CentOS 6 box I used. No idea.

vpodzime (approved):
ACK

PR #4444, opened 2021-01-08
GitHub Actions: Added ASAN

olehermanse:


olehermanse:
Regarding increasing job count from 1 to 8: No science behind it, but it did decrease the time of that step from 1m35s to 43s, so I think that's good enough :)

craigcomstock (approved):

larsewi (approved):
Nice!

PR #4452, opened 2021-01-13
Added custom promise type to check if a site is up

aleksandrychev:


aleksandrychev:
@olehermanse please review

docs/custom_promise_types/site_up.py
```diff
@@ -0,0 +1,45 @@
...
+        except urllib.error.HTTPError as e:
+             error = f"Site '{url}' is DOWN! Response code: '{e.code}'"
+        except urllib.error.URLError as e:
+             error = f"Site '{url}' is DOWN! Reason: '{e.reason}'"
+        except Exception as e:
+            error = str(e)
+
+        if error is not None:
+            self.log_error(error)
+            return Result.NOT_KEPT
```
olehermanse, line 36: Since this is always true, you set error in all the branches, it should be an assert:

```suggestion
        assert(error is not None)
        self.log_error(error)
        return Result.NOT_KEPT
```

docs/custom_promise_types/site_up.py
```diff
@@ -0,0 +1,45 @@
...
+class SiteUpPromiseTypeModule(PromiseModule):
+    def __init__(self):
+        super().__init__("site_up_promise_module", "0.0.1")
+
+    def validate_promise(self, promiser, attributes):
+        if self.is_url_valid(promiser) == False :
+            raise ValidationError(f"URL '{promiser}' is invalid")
+
+
+    def evaluate_promise(self, url, attributes):
```
olehermanse, line 17:
```suggestion
            raise ValidationError(f"URL '{promiser}' is invalid")

    def evaluate_promise(self, url, attributes):
```

Nitpicking, but shouldn't be 2 lines between here. I think [black](https://github.com/psf/black) would remove this.

docs/custom_promise_types/site_up.py
```diff
@@ -0,0 +1,45 @@
...
+import urllib.error
+from cfengine import PromiseModule, ValidationError, Result
+
+
+class SiteUpPromiseTypeModule(PromiseModule):
+    def __init__(self):
+        super().__init__("site_up_promise_module", "0.0.1")
+
+    def validate_promise(self, promiser, attributes):
+        if self.is_url_valid(promiser) == False :
```
olehermanse, line 13:
```suggestion
        if not self.is_url_valid(promiser):
```

More clean / pythonic.

docs/custom_promise_types/site_up.py
```diff
@@ -0,0 +1,45 @@
...
+
+        error = None
+        try:
+            code = urllib.request.urlopen(url, context=ssl_ctx).getcode()
+            self.log_info(f"Site '{url}' is UP!")
+            return Result.KEPT
+        except urllib.error.HTTPError as e:
+             error = f"Site '{url}' is DOWN! Response code: '{e.code}'"
+        except urllib.error.URLError as e:
+             error = f"Site '{url}' is DOWN! Reason: '{e.reason}'"
```
olehermanse, line 30: What is the difference between an HTTPError and a URLError? Could explain with a comment.

olehermanse (approved):
Looks good to me, just some style suggestions.

docs/custom_promise_types/site_up.py
aleksandrychev: added comments

olehermanse (approved):

PR #4453, opened 2021-01-13
Support for hailing agents with cf-runagent via cf-execd 

vpodzime:
Merge together:
https://github.com/cfengine/core/pull/4453
https://github.com/cfengine/masterfiles/pull/1891

### DIFF
--- a/cf-execd/cf-execd-runagent.c
+++ b/cf-execd/cf-execd-runagent.c
@@ -1,0 +1,139 @@
+/*
+  Copyright 2021 Northern.tech AS
+
+  This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+  This program is free software; you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by the
+  Free Software Foundation; version 3.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#include <stdio.h>
+
+#include <logging.h>
+#include <buffer.h>
+#include <json.h>
+#include <pipes.h>
+#include <alloc.h>
+#include <file_lib.h>
+#include <known_dirs.h>
+
+#include <cf-execd-runagent.h>
+
+
+/**
+ * Handle request to run cf-runagent.
+ *
+ * @note Expected to be called from a forked child process (blocks and forks).
+ */
+void HandleRunagentRequest(int conn_fd)
+{
+    FILE *conn_file = fdopen(conn_fd, "r+");
+    if (conn_file == NULL)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to create file stream from the connection file descriptor: %s",
+            GetErrorStr());
+
+        /* SIGH: There's no dprintf() on exotics. */
+        char error_buf[256];
+        int length = snprintf(error_buf, sizeof(error_buf),
+                              "{\"error\": \"Failed to create file stream from the connection file descriptor: %s\"}",
+                              GetErrorStr());
+        assert(length < sizeof(error_buf));
+        write(conn_fd, error_buf, MIN(sizeof(error_buf), (size_t) length));
+        close(conn_fd);
+        return;
+    }
+
+    char *request = NULL;
+    size_t line_size = 0;
+    ssize_t n_read = getline(&request, &line_size, conn_file);
+    if (n_read <= 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to read the runagent request: %s",
+            (n_read < 0) ? GetErrorStr() : "no data");
+        fprintf(conn_file, "{\"error\": \"Failed to read the runagent request: %s\"}",
+                (n_read < 0) ? GetErrorStr() : "no data");
+        free(request);
+        fclose(conn_file);
+        return;
+    }
+
+    /* Strip the trailing newline (if any). */
+    if (request[n_read - 1] == '\n')
+    {
+        request[n_read - 1] = '\0';
+    }
+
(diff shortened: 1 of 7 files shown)


vpodzime:
The only remaining bit here I'm aware of is SELinux policy adaptations for *cf-execd* to be allowed to work with the `runagent.socket`.

cf-execd/cf-execd-runagent.c
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /* TODO: '--raw' for just getting output from cf-runagent directly? */
+    for (char *c = (request + n_read - 1); (c > request) && (*c != '\n'); c--)
+    {
+        if (*c == '\n')
+        {
+            *c = '\0';
+        }
+    }
```
larsewi, line 83: Is this really necessary?
The `getline()` function reads a line from ***stream***, delimited by the newline character. It then returns the number of characters written, excluding the terminating NUL character. Does this not mean that the newline character is always located at `request[n_read - 1]`. Thus we could do `request[n_read - 1] = '\0';` ?

cf-execd/cf-execd-runagent.c
```diff
@@ -0,0 +1,150 @@
...
+
+    char *command;
+    int ret = xasprintf(&command, "%s%ccf-runagent -b -H %s", GetBinDir(), FILE_SEPARATOR, request);
+    assert(ret != -1); /* should never happen */
+    if (ret == -1)
```
larsewi, line 88: I don't understand why we assert the return value from `xasprintf` with a comment saying "this should never happen", and then handle it as if it could happen? Does not `xasprintf` call `DoCleanupAndExit`, thus the error handling will never happen?

cf-execd/cf-execd-runagent.c
```diff
@@ -0,0 +1,150 @@
...
+    {
+        Log(LOG_LEVEL_ERR, "Failed to start cf-runagent and connect to its output");
+        fprintf(conn_file, "{\"error\": \"Failed to start cf-runagent and connect to its output\"}");
+        fclose(conn_file);
+        return;
+    }
+
+    Buffer *collected_output = BufferNewWithCapacity(1024);
+    char output_buffer[CF_BUFSIZE];
+    size_t read_bytes = fread(output_buffer, 1, sizeof(output_buffer), runagent_output);
```
larsewi, line 111: Personally I would use `sizeof(char)` to avoid using magic numbers although it's pretty obvios what the `1` means in this case.

cf-execd/cf-execd-runagent.c
vpodzime: *The buffer is null-terminated and includes the newline character, **if one was found.***

The request could be sent into the socket without a newline.

cf-execd/cf-execd-runagent.c
vpodzime: Good point! I'll gladly drop this. :+1: :grin:

cf-execd/cf-execd-runagent.c
vpodzime: I'd have to call the variable `read_chars` then. :wink:

vpodzime:
jenkins (with exotics): [Build Status](https://ci.cfengine.com/job/pr-pipeline/6060/)

vpodzime:
> jenkins (with exotics): [Build Status](https://ci.cfengine.com/job/pr-pipeline/6060/)

The HP-UX failure is: *14:07:47 ssh: connect to host cloud.siteox.com port 35622: No route to host* But there was a previous green build on this platform so it can be ignored.

larsewi (dismissed):
Looks good !

vpodzime:
>  [@vpodzime](/vpodzime) [vpodzime](/vpodzime) [force-pushed](/cfengine/core/compare/5d4e3ddcfb2b8e45328730d75670b726ad92664a..cabe28a981fd304e7e51a0598325188d261a423c) the vpodzime:master-cf_execd_runagent branch from [`5d4e3dd`](/cfengine/core/commit/5d4e3ddcfb2b8e45328730d75670b726ad92664a) to [`cabe28a`](/cfengine/core/commit/cabe28a981fd304e7e51a0598325188d261a423c) [now](#event-4234402807)

Fixed the ticket number in commit messages.

olehermanse (approved):
LGTM

cf-execd/cf-execd.c
```diff
@@ -431,12 +455,29 @@ static void CFExecdMainLoop(EvalContext *ctx, Policy **policy, GenericAgentConfi
...
+        {
+            break;
+        }
     }
 }
 
 #else  /* ! __MINGW32__ */
 
+// msg should include exactly one reference to unsigned int.
+static inline unsigned int MaybeSleepLog(LogLevel level, const char *msg, unsigned int seconds)
```
olehermanse: I didn't get this at first read. I'd prefer `msg_format` as the parameter name and:

```
// msg_format should include exactly one '%u'
```

cf-execd/cf-execd-runagent.c
```diff
@@ -0,0 +1,150 @@
...
+        Log(LOG_LEVEL_ERR, "Failed to create file stream from the connection file descriptor: %s",
+            GetErrorStr());
+
+        /* SIGH: There's no dprintf() on exotics. */
+        char error_buf[256];
+        int length = snprintf(error_buf, sizeof(error_buf),
+                              "{\"error\": \"Failed to create file stream from the connection file descriptor: %s\"}",
+                              GetErrorStr());
+        assert(length < sizeof(error_buf));
+        write(conn_fd, error_buf, MIN(sizeof(error_buf), (size_t) length));
```
olehermanse, line 57: Could use an allocating printf function to avoid the hard coded buffer size limit.

larsewi (approved):

cf-execd/cf-execd-runagent.c
vpodzime: Yes, but this is a special code block because `fdopen()` is very unlikely to fail, one of the reasons could be problems with memory allocation. And the buffer is big enough for all `GetErrorStr()` strings, AFAICT.

cf-execd/cf-execd-runagent.c
vpodzime: That's true. I can simplify the code then. :+1:

cf-execd/cf-execd.c
vpodzime: It's not a function I added, though. I'll add a commit improving it.

cf-execd/cf-execd-runagent.c
olehermanse: Makes sense 👍

larsewi (approved):
Nice !

vpodzime:
@olehermanse I've addressed your comments and @larsewi re-reviewed things. I'm merging this,  but feel free to re-review too and post comments.

PR #4460, opened 2021-01-18
CFE-3516 Custom promise type library, cp and backup

Lex-2008:


docs/custom_promise_types/cfengine.sh
```diff
@@ -0,0 +1,153 @@
...
+    response_result="valid"
+    if [ "$saw_unknown_attribute" != "no" -a "$all_attributes_are_valid" != "yes" ] ; then
+        echo "log_error=Unknown attribute/s: ${unknown_attribute_names#, }"
+        response_result="invalid"
+    fi
+
+    if [ ! -z "$obligatory_arrtibutes" ]; then
+        for attribute_name in $obligatory_attributes; do
+            varname="attribute_$obligatory_arrtibutes"
+            if [ -z "${!varname}" ]; then
```
vpodzime, line 88: What does the `!` do here?

docs/custom_promise_types/cfengine.sh
```diff
@@ -0,0 +1,153 @@
...
+    # Skip until (and including) the first empty line
+    while IFS='$\n' read -r line; do
+        if [ "x$line" = "x" ] ; then
+            return;
+        fi
+    done
+}
+
+# Skip the protocol header given by agent:
+skip_header
```
vpodzime, line 140: Should this really just always skip the header?

docs/custom_promise_types/backup.sh
```diff
@@ -0,0 +1,58 @@
...
+    # call the script
+    . "$(dirname "$0")/backup3/backup.sh"
+
+    # Analyse its results
+    if [ ! -f "$BACKUP_TMP".files ]; then
+        response_result="not_kept"
+        return 0
+    fi
+    local filelist_size="$(stat -c %s "$BACKUP_TMP".files)"
+    if [ "$filelist_size" -gt 21 ]; then
```
vpodzime, line 48: ?

vpodzime (commented):
Looks good to me otherwise.

docs/custom_promise_types/cfengine.sh
Lex-2008: It was the most sensible way of dynamic variable names in Bash I found:
https://stackoverflow.com/questions/16553089/dynamic-variable-names-in-bash

docs/custom_promise_types/backup.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/sh
+#
+# Backup custom promise type, uses cfengine.sh library located in same dir,
+# and https://github.com/Lex-2008/backup3
```
tzz, line 4: Should the example depend on external resources like that?

docs/custom_promise_types/backup.sh
```diff
@@ -0,0 +1,58 @@
...
+    BACKUP_KEEP_TMP="1"
+    run_this() {
+        run_rsync always "$request_promiser" "$request_attribute_from"
+    }
+
+    # call the script
+    . "$(dirname "$0")/backup3/backup.sh"
+
+    # Analyse its results
+    if [ ! -f "$BACKUP_TMP".files ]; then
```
tzz, line 43: Shouldn't you check the return code `$?` also?

docs/custom_promise_types/cfengine.sh
```diff
@@ -0,0 +1,153 @@
...
+}
+
+operation_validate() {
+    response_result="valid"
+    if [ "$saw_unknown_attribute" != "no" -a "$all_attributes_are_valid" != "yes" ] ; then
+        echo "log_error=Unknown attribute/s: ${unknown_attribute_names#, }"
+        response_result="invalid"
+    fi
+
+    if [ ! -z "$obligatory_arrtibutes" ]; then
```
tzz, line 85: misspelled `attributes` here and below?

docs/custom_promise_types/cfengine.sh
olehermanse: Adding library support for getting the information from the header would be nice, but it's also quite safe to skip it now.

The header becomes important when we add more features / protocol versions, and we want to write modules which are compatible with both old and new agents.

docs/custom_promise_types/cfengine.sh
```diff
@@ -0,0 +1,153 @@
...
+            return;
+        fi
+    done
+}
+
+# Skip the protocol header given by agent:
+skip_header
+
+# Write our header to request line based protocol:
+echo "example_promises 0.0.1 v1 line_based"
```
olehermanse, line 143: library needs a way to set this from module (module name and version). Protocol version and flags can be controlled by library.

docs/custom_promise_types/cfengine.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/false
```
olehermanse, line 1: Commit title (first line), and PR title, should not end in punctuation (period).

https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#commit-message-example

docs/custom_promise_types/backup.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/sh
```
olehermanse, line 1: Commit title (first line), and PR title, should not end in punctuation (period).

https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#commit-message-example

olehermanse (changes requested):

docs/custom_promise_types/backup.sh
olehermanse: The other example, `cp.sh` in previous commit, does not rely on anything else, so this is okay. We will move all of these to a more appropriate place soon.

docs/custom_promise_types/cfengine.sh
craigcomstock: I'm not sure I understand how this part works. It seems like you would want to `varname="attribute_$attribute_name"` and not what you have right now? If `$obligatory_attributes` (spelling fixed) is what I think you expect, a space delimited set of attribute names to use with `for`, then wouldn't `varname` end up being pretty odd? e.g. if `obligatory_attributes` is `from attribute1 attribute2` then varname would be `attribute_from attribute1 attribute2` and then the `-z` doesn't seem like it would work properly.

craigcomstock: FWIW, this link says think before you use this indirection feature.

http://mywiki.wooledge.org/BashFAQ/006#Indirection

So maybe an array would be a better choice?

craigcomstock (commented):
a few questions, looks good!

docs/custom_promise_types/backup.sh
Lex-2008: Sounds reasonable! Although backup.sh doesn't have code to return proper error codes, checking for non-0 return code here makes sense.

docs/custom_promise_types/cfengine.sh
Lex-2008: :facepalm: That's what I get when I just want to "cleanup few variable names" in already-tested code before creating PR: not working code. I should feel awhamed, and I honestly do.

Lex-2008: Regarding arrays - that would be a better choice, indeed, but I'd rather keep this library POSIX-compatible. However, my opinion is not very strong, so I can be convinced :)

vpodzime: I'm OK with `${!varname}` if you add a comment explaining it. Because searching for `{!` in `man bash` only refers to arrays.

docs/custom_promise_types/backup.sh
Lex-2008: indeed, it was a bit of magic. Replaced with explanation

docs/custom_promise_types/cfengine.sh
Lex-2008: explaining comment added

docs/custom_promise_types/cfengine.sh
Lex-2008: Corrected, thanks!

docs/custom_promise_types/cfengine.sh
Lex-2008: done!

docs/custom_promise_types/cfengine.sh
Lex-2008: fixed!

docs/custom_promise_types/backup.sh
Lex-2008: fixed!

docs/custom_promise_types/cfengine.sh
```diff
@@ -0,0 +1,153 @@
...
+reset_state() {
+    # Set global variables before we begin another request
+
+    # Variables parsed directly from request:
+    request_operation=""
+    request_log_level=""
+    request_promise_type=""
+    request_promiser=""
+
+    for var in $(env | cut -d '=' -f 1 | grep 'request_attribute_'); do
```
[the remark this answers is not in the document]
Lex-2008: makes sense!

docs/custom_promise_types/cfengine.sh
```diff
@@ -0,0 +1,164 @@
...
+operation_validate() {
+    response_result="valid"
+    if [ "$saw_unknown_attribute" != "no" -a "$all_attributes_are_valid" != "yes" ] ; then
+        echo "log_error=Unknown attribute/s: ${unknown_attribute_names#, }"
+        response_result="invalid"
+    fi
+
+    if [ ! -z "$obligatory_attributes" ]; then
+        for attribute_name in $obligatory_attributes; do
+            # Note: ${!varname} synthax expands to value of valiable, which name
```
basvandervlies, line 87: valiable --> variable

Lex-2008: thanks, indeed!

olehermanse (approved):
LGTM, thanks!

docs/custom_promise_types/cfengine.sh
```diff
@@ -0,0 +1,164 @@
...
+operation_validate() {
+    response_result="valid"
+    if [ "$saw_unknown_attribute" != "no" -a "$all_attributes_are_valid" != "yes" ] ; then
+        echo "log_error=Unknown attribute/s: ${unknown_attribute_names#, }"
+        response_result="invalid"
+    fi
+
+    if [ ! -z "$obligatory_attributes" ]; then
+        for attribute_name in $obligatory_attributes; do
+            # Note: ${!varname} synthax expands to value of variable, which name
```
craigcomstock, line 87: s/synthax/syntax/ :)

craigcomstock: Or were you referring to the Roland AX-Synth? ;) JK.

Lex-2008: I make this mistake so often I should add it to OS-wide auto-correct :)

Lex-2008:
Dear reviewers, please have a look at the last commit ("fixup! Move main loop to function"). Does it look like a good improvement to be squashed?

olehermanse (approved):

PR #4464, opened 2021-01-21
CFE-3551/master: Added example illustrating inline yaml

nickanderson:
Ticket: CFE-3551
Changelog: None

### DIFF
--- a/examples/inline-yaml.cf
+++ b/examples/inline-yaml.cf
@@ -1,0 +1,58 @@
+#+begin_src cfengine3
+bundle agent example_inline_yaml
+# @brief Example illustrating inline yaml
+{
+  vars:
+      # YAML requires "---" header (followed by newline)
+      # NOTE \n is not interpreted as a newline, instead use $(const.n)
+
+      "yaml_multi_line" data => '---
+- "CFEngine Champions":
+   - Name: "Aleksey Tsalolikhin"
+     Year: 2011
+   - Name: "Ted Zlatanov"
+     Year : 2013';
+
+
+      "yaml_single_line" data => '---$(const.n)- key1: value1$(const.n)- key2: value2';
+
+  reports:
+      "Data container defined from yaml_multi_line: $(with)"
+        with => storejson( @(yaml_multi_line) );
+
+      "Data container defined from yaml_single_line: $(with)"
+        with => storejson( @(yaml_single_line) );
+}
+bundle agent __main__
+{
+  methods:
+      "example_inline_yaml";
+}
+#+end_src
+###############################################################################
+#+begin_src example_output
+#@ ```
+#@ R: Data container defined from yaml_multi_line: [
+#@   {
+#@     "CFEngine Champions": [
+#@       {
+#@         "Name": "Aleksey Tsalolikhin",
+#@         "Year": 2011
+#@       },
+#@       {
+#@         "Name": "Ted Zlatanov",
+#@         "Year": 2013
+#@       }
+#@     ]
+#@   }
+#@ ]
+#@ R: Data container defined from yaml_single_line: [
+#@   {
+#@     "key1": "value1"
+#@   },
+#@   {
+#@     "key2": "value2"
+#@   }
+#@ ]
+#@ ```
+#+end_src


nickanderson:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6065/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6065/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6065/

examples/inline-yaml.cf
```diff
@@ -0,0 +1,49 @@
...
+      "yaml_multi_line" data => '---
+- key1: value1
+- key2: value2';
+
+
+      "yaml_single_line" data => '---$(const.n)- key1: value1$(const.n)- key2: value2';
+
+  reports:
+      "Data container defined from yaml_multi_line: $(with)"
+        with => string_mustache( "{{%-top-}}", @(yaml_multi_line) );
```
craigcomstock, line 18: I think the example might be clearer by using `with => storejson(@(yaml_multi_line));` I find the mustache stuff a bit noisy and confusing (and just a touch longer to type ;).

examples/inline-yaml.cf
```diff
@@ -0,0 +1,49 @@
...
+
+      "Data container defined from yaml_single_line: $(with)"
+        with => string_mustache( "{{%-top-}}", @(yaml_single_line));
+}
+bundle agent __main__
```
craigcomstock, line 23: Do you have to use `__main__` for some reason? Otherwise I would just change the `example_inline_yaml` bundle name to `main`. For the sake of brevity I guess.

nickanderson: I like to use `__main__` because it highlights the feature, and it makes it much easier to scoop up a bunch of examples into a policy and then run the bundles you want. If they are all `bundle agent main` then you have to go fix duplicate bundle definition.

examples/inline-yaml.cf
tzz: My favorite is `with => format("%S", variable)` because it works for any variable :)

examples/inline-yaml.cf
```diff
@@ -0,0 +1,49 @@
...
+  vars:
+      # YAML requires "---" header (followed by newline)
+      # NOTE \n is not interpreted as a newline, instead use $(const.n)
+
+      "yaml_multi_line" data => '---
+- key1: value1
+- key2: value2';
+
+
+      "yaml_single_line" data => '---$(const.n)- key1: value1$(const.n)- key2: value2';
```
tzz, line 14: Totally looking for something to criticize in this beautiful example: maybe use a different key and value pattern in the second container?

examples/inline-yaml.cf
craigcomstock: Makes sense. I typically only put an example stand-alone and run it with `cf-agent -KIf ./%`. :+1: on your method.

tzz: I've always been a fan of self-contained examples I can run directly so I'm with Craig.

Although it would be neat if `cf-agent -f myfile.cf` ran whatever agent bundle it found if it was the only one.

examples/inline-yaml.cf
craigcomstock: @tzz not bad actually! That sort of improves the output in a way.

Gives me
```
R: yml [{"key1":"value1"},{"key2":"value2"}]
```

(on one line)

instead of the multi-line and structured output of `string_mustache` or `storejson`.

Feature request would be supporting this though right? ;)

```
  "Data container defined from yaml_multi_line: $(yaml_multi_line)";
```

That would be the "easiest" and most obvious I think. Kind of like python `f"` string formatting with variable references inside. :)

nickanderson: I think that multiline output for data is much easier to read, so I tend to avoid format() unless I am on some older version because as far as I know, it won't do multiline output.

I tend to use mustache because then if I want to render parts of the data it's easier and already there, but also I didn't realize that storejson() would produce multiline output. I can switch to that unless @tzz knows magic format() to make it multiline.


I prefert

examples/inline-yaml.cf
nickanderson: Aaaaaaaaaas yoooooooooooooooooooooooooou wiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiish

examples/inline-yaml.cf
tzz: > Feature request would be supporting this though right? ;)
> 
> ```
>   "Data container defined from yaml_multi_line: $(yaml_multi_line)";
> ```
> 
> That would be the "easiest" and most obvious I think. Kind of like python `f"` string formatting with variable references inside. :)

Absolutely. I proposed something like that years ago and it was rejected by the team, so we ended up converging (ha!) on the `with` attribute. Maybe the current team is willing to reconsider.

tzz: > unless @tzz knows magic format() to make it multiline.

Nope, sorry... Craig's way is best.

nickanderson: > > Feature request would be supporting this though right? ;)
> > ```
> >   "Data container defined from yaml_multi_line: $(yaml_multi_line)";
> > ```
> > 
> > 
> > That would be the "easiest" and most obvious I think. Kind of like python `f"` string formatting with variable references inside. :)
> 
> Absolutely. I proposed something like that years ago and it was rejected by the team, so we ended up converging (ha!) on the `with` attribute. Maybe the current team is willing to reconsider.

https://tracker.mender.io/browse/CFE-3552

nickanderson:
ping @atsaloli as well.

vpodzime (commented):
Please squash the commits.

nickanderson:
> Please squash the commits.

https://www.youtube.com/watch?v=b8jrhtPz2Ac

craigcomstock (approved):

PR #4465, opened 2021-01-22
Make bundle promise outcomes INFO instead of VERBOSE level

tzz:
Bundle promise outcomes are useful, but require running with `--verbose` which generates a LOT of output. This patch reports promise outcomes with `--inform`.

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -2103,13 +2103,13 @@ static int NoteBundleCompliance(const Bundle *bundle, int save_pr_kept, int save
     }
     else
     {
-        Log(LOG_LEVEL_VERBOSE, "A: Promises kept in '%s' = %.0lf", bundle->name, delta_pr_kept);
-        Log(LOG_LEVEL_VERBOSE, "A: Promises not kept in '%s' = %.0lf", bundle->name, delta_pr_notkept);
-        Log(LOG_LEVEL_VERBOSE, "A: Promises repaired in '%s' = %.0lf", bundle->name, delta_pr_repaired);
+        Log(LOG_LEVEL_INFO, "A: Promises kept in '%s' = %.0lf", bundle->name, delta_pr_kept);
+        Log(LOG_LEVEL_INFO, "A: Promises not kept in '%s' = %.0lf", bundle->name, delta_pr_notkept);
+        Log(LOG_LEVEL_INFO, "A: Promises repaired in '%s' = %.0lf", bundle->name, delta_pr_repaired);
 
         bundle_compliance = (delta_pr_kept + delta_pr_repaired) / (delta_pr_kept + delta_pr_notkept + delta_pr_repaired);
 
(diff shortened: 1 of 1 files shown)


cf-bottom:

olehermanse (changes requested):
It certainly is useful, but should not be in INFO. We've made a lot of changes to ensure INFO output is clean by default, and only has output when changes are made to the system. This would reverse that and create unnecessary noise in that case.

https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#log-levels

I agree that the information is useful, but it hase to be in a separate log level, a new command line flag, or something similar. Please create a ticket on the bug tracker to discuss further:

https://tracker.mender.io/

cf-agent/cf-agent.c
```diff
@@ -2103,13 +2103,13 @@ static int NoteBundleCompliance(const Bundle *bundle, int save_pr_kept, int save
     }
     else
     {
-        Log(LOG_LEVEL_VERBOSE, "A: Promises kept in '%s' = %.0lf", bundle->name, delta_pr_kept);
-        Log(LOG_LEVEL_VERBOSE, "A: Promises not kept in '%s' = %.0lf", bundle->name, delta_pr_notkept);
-        Log(LOG_LEVEL_VERBOSE, "A: Promises repaired in '%s' = %.0lf", bundle->name, delta_pr_repaired);
+        Log(LOG_LEVEL_INFO, "A: Promises kept in '%s' = %.0lf", bundle->name, delta_pr_kept);
+        Log(LOG_LEVEL_INFO, "A: Promises not kept in '%s' = %.0lf", bundle->name, delta_pr_notkept);
```
nickanderson, line 2107: Promises not kept should already be elevated as errors right?

cf-agent/cf-agent.c
```diff
@@ -2103,13 +2103,13 @@ static int NoteBundleCompliance(const Bundle *bundle, int save_pr_kept, int save
...
+        Log(LOG_LEVEL_INFO, "A: Promises repaired in '%s' = %.0lf", bundle->name, delta_pr_repaired);
```
nickanderson, line 2108: Promises repaired makes /some/ sense in Inform, but the specific changes should already be in the inform outut.

cf-agent/cf-agent.c
tzz: Currently there's no way to know which bundles are failing their promises, you have to dig into the source code and infer the promise that failed. This is really useful to avoid hunting across dozens or hundreds of bundles.

cf-agent/cf-agent.c
tzz: I don't think it's consistently reported, and per bundle is very useful.

tzz:
> I agree that the information is useful, but it hase to be in a separate log level, a new command line flag, or something similar. Please create a ticket on the bug tracker to discuss further:

Sure: https://tracker.mender.io/browse/CFE-3553

PR #4466, opened 2021-01-22
Report normal file editing mismatches as INFO not ERROR

tzz:
Mismatches while editing are common and should not be treated as errors.


nickanderson:
@tzz do you have a stand-alone policy file that we can run to see the specifics of what you are talking about?

tzz:
sorry @nickanderson @olehermanse 

I think https://github.com/cfengine/core/pull/4783 will DTRT for this one

PR #4467, opened 2021-01-22
CFE-3266: package cf-remote for pypi to install with pip

craigcomstock:


cf-bottom (changes requested):
Please use a company e-mail instead of c***@unreasonablefarm.org

craigcomstock:
other bits of information are still pending, like pypi and testpypi credentials and build process.

currently working on getting a console_script working. works a little bit locally with `python3 setup.py --user develop` but not quite.

Gives:
```
craig@other:/northern.tech/cfengine/core/contrib/cf-remote$ cf-remote --version
Traceback (most recent call last):
  File "/home/craig/.local/bin/cf-remote", line 33, in <module>
    sys.exit(load_entry_point('cf-remote', 'console_scripts', 'cf-remote')())
TypeError: 'module' object is not callable
```

olehermanse:
Please go to here:

PR #4468, opened 2021-01-25
cf-remote: Moved to separate repo, added deprecation warning and links

olehermanse:
`cf-remote` was moved to it's own repo, so it can have it's own versioning, CI, etc.

https://github.com/cfengine/cf-remote


craigcomstock (approved):

vpodzime (approved):
Looks good to me. Just out of curiosity -- why not another submodule?

olehermanse:
> Looks good to me. Just out of curiosity -- why not another submodule?

@vpodzime That's a good idea actually, I didn't think of it.

I guess when it's available via pip, you won't want/need the extra clone in cfengine/core, but I guess it doesn't hurt either?

I don't have anything against reintroducing it as a submodule here.

PR #4469, opened 2021-01-26
Added a new --simulate=manifest-full mode

vpodzime:
New simulation mode that manifests all changed files as well as
all other files evaluated by the agent run which were not skipped
(by file selection rules).

Ticket: CFE-3506
Changelog: Commit

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -236,7 +236,7 @@ static const char *const HINTS[] =
     "Show *final* evaluated variables, including those defined without dependency to user-defined classes in policy. Optionally can take a regular expression.",
     "Do not run policy as the last step of the bootstrap process",
     "Do not run database integrity checks and repairs at startup",
-    "Run in simulate mode, either 'manifest' or 'diff'",
+    "Run in simulate mode, either 'manifest', 'manifest-full' or 'diff'",
     NULL
 };
 
@@ -338,22 +338,34 @@ int main(int argc, char *argv[])
 
     GenericAgentFinalize(ctx, config);
 
-    if (EVAL_MODE == EVAL_MODE_SIMULATE_MANIFEST)
+    StringSet *audited_files = NULL;
+    if ((EVAL_MODE == EVAL_MODE_SIMULATE_MANIFEST) ||
+        (EVAL_MODE == EVAL_MODE_SIMULATE_MANIFEST_FULL))
     {
-        bool success = ManifestChangedFiles();
+        bool success = ManifestChangedFiles(&audited_files);
         if (!success)
         {
             Log(LOG_LEVEL_ERR, "Failed to manifest changed files");
         }
+        if (EVAL_MODE == EVAL_MODE_SIMULATE_MANIFEST_FULL)
+        {
+            /* Skips the files already manifested above. */
+            success = ManifestAllFiles(&audited_files);
+            if (!success)
+            {
+                Log(LOG_LEVEL_ERR, "Failed to manifest unmodified files");
+            }
+        }
     }
     else if (EVAL_MODE == EVAL_MODE_SIMULATE_DIFF)
     {
-        bool success = DiffChangedFiles();
+        bool success = DiffChangedFiles(&audited_files);
         if (!success)
         {
             Log(LOG_LEVEL_ERR, "Failed to show differences for changed files");
         }
     }
+    StringSetDestroy(audited_files);
 
 #ifdef HAVE_LIBXML2
         xmlCleanupParser();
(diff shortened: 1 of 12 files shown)


olehermanse:
@vpodzime Not a big deal, but if the manifest includes kept, not kept and repaired, I'd prefer another name, like: `manifest-full` or simply `full`.

@nickanderson what do you think?

vpodzime:
> @vpodzime Not a big deal, but if the manifest includes kept, not kept and repaired, I'd prefer another name, like: `manifest-full` or simply `full`.

Makes sense. And I guess the code should also not refer to `KEPT` when it's all non-skipped files (incl. failed)

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6074/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6074/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6074/

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6075/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6075/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6075/

nickanderson:
> @vpodzime Not a big deal, but if the manifest includes kept, not kept and repaired, I'd prefer another name, like: `manifest-full` or simply `full`.
> 
> @nickanderson what do you think?

I don't think so.

To me, simulate diff is the short report showing just the delta. This one should I think show repaired and I think even not-kept as they are changes that couldn't be completed. They would be attempted during a non-simulation.

Manifest is the full monty. This is what your system should look like at the end. I think it makes sense this includes all the files that cfengine expects to consider. Kept, repaired, and not kept.

Yes, the output could be volumous and include many things that the user doesn't care about.
First thought is that maybe we need to have tags that will prevent promises from getting into this report. Then if we tagged all the stock promises  those could be excluded, but then you might actually want to see some of that and getting them into a report might be difficult. Or perhaps an option to exclude promises  based on a path,e.g. exclude promises for /var/cfenigne/ but that won't help with commands or processes or packages. Not sure exactly.

I don't love adding another manifest-full option. diff-show-full-files sounds more accurate if your only showing promises repaired.

vpodzime:
> Manifest is the full monty.

This is what happens with the stock policy on a hub:
```
[root@hub vagrant]# cf-agent -KI --simulate=manifest-full > /tmp/manifest_full.out
[root@hub vagrant]# ll -h /tmp/manifest_full.out
-rw-r--r--. 1 root root 251M Jan 27 09:12 /tmp/manifest_full.out
```
I don't think this is useful in any way.

olehermanse:
> > @vpodzime Not a big deal, but if the manifest includes kept, not kept and repaired, I'd prefer another name, like: `manifest-full` or simply `full`.
> > @nickanderson what do you think?
> 
> I don't think so.
> 
> To me, simulate diff is the short report showing just the delta. This one should I think show repaired and I think even not-kept as they are changes that couldn't be completed. They would be attempted during a non-simulation.
> 
> Manifest is the full monty. This is what your system should look like at the end. I think it makes sense this includes all the files that cfengine expects to consider. Kept, repaired, and not kept.
> 
> Yes, the output could be volumous and include many things that the user doesn't care about.
> First thought is that maybe we need to have tags that will prevent promises from getting into this report. Then if we tagged all the stock promises those could be excluded, but then you might actually want to see some of that and getting them into a report might be difficult. Or perhaps an option to exclude promises based on a path,e.g. exclude promises for /var/cfenigne/ but that won't help with commands or processes or packages. Not sure exactly.
> 
> I don't love adding another manifest-full option. diff-show-full-files sounds more accurate if your only showing promises repaired.

@nickanderson there are many words and interpretations here. In my opinion, diiff mode and manifest mode are not the same as repaired (changed) only vs everything mode. Diff means to show the changes as diffs. Manifest is to show a manifest of the final state (it can also be only repaired promises, it's just that we want it presented as the full final file contents/state, not diffs). I know the ticket stated that manifest should include the kept promises, but that was because of the particular user/use-case, which I believe does not use masterfiles.

As @vpodzime said, showing kept promises if you are using masterfiles is just not useful currently. (250+MB of logs). Having a separate manifest-full mode, which is essentially only useful for small policy sets / standalone policy files, and keeping the existing manifest mode useful makes perfect sense to me.

larsewi:
[ cf-agent/cf-agent.c:694](https://github.com/cfengine/core/pull/4469/files#diff-a728d21fe93822246ac006cb79a16cf57a2fc501eb713d322dafb42726e042bbR694)

Should it be ...
```C
Log(LOG_LEVEL_ERR, "Missing argument for --simulate, 'manifest', 'manifest-full' or 'diff' required");
```
instead of ...
```C
Log(LOG_LEVEL_ERR, "Missing argument for --simulate, 'manifest' or 'diff' required");
```
?

larsewi (approved):
Awesome !

olehermanse (approved):

PR #4478, opened 2021-01-29
libntech: Removed unnecessary assert comparison

larsewi:
libntech updated


vpodzime (approved):
ACK

olehermanse (approved):

PR #4486, opened 2021-02-01
ENT-6750/master: Stopped emitting warning and recording result when observing new SETGID files

nickanderson:
Ticket: ENT-6750
Changelog: Title


vpodzime (approved):
ACK

PR #4495, opened 2021-02-08
Travis: Disabled MacOS

olehermanse:
Failing Mac builds have caused master to be red for some days.
This seems to be because Travis changed the installed packages
on their default image. Since we are already testing MacOS
on GitHub Actions, and they are much faster and more reliable,
I don't see a point in fixing the travis build, so I disabled it.


vpodzime (approved):
ACK

Lex-2008 (approved):

olehermanse:
Backports:
https://github.com/cfengine/core/pull/4496
https://github.com/cfengine/core/pull/4497

PR #4498, opened 2021-02-09
Check if result from evaluate_promise() is a string in custom_promise…

vpodzime:
…_types/cfengine.py

'class Result' only defines attributes that are strings, it's not
an Enum. So things like 'Result.KEPT' are actually just strings
with no special type.

Changelog: None


larsewi (approved):
Yes !

olehermanse (approved):

PR #4501, opened 2021-02-10
CFE-3573: Assertion error no longer logs empty error msg

larsewi:
Promise module python library now handles all exception classes and additionally logs the exception type followed by the message. This way we at least get a hint in case of an assertion error without any message.

### DIFF
--- a/docs/custom_promise_types/cfengine.py
+++ b/docs/custom_promise_types/cfengine.py
@@ -1,5 +1,6 @@
 import sys
 import json
+import traceback
 
 
 def _skip_until_empty_line(file):
@@ -140,6 +141,7 @@ class PromiseModule:
         except Exception as e:
             self.log_critical(f"{type(e).__name__}: {e}")
             self._result = Result.ERROR
+            self._log_traceback()
         self._add_result()
         _put_response(self._response, self._out)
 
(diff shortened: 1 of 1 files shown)


vpodzime (approved):
Looks good to me.

docs/custom_promise_types/cfengine.py
```diff
@@ -137,6 +137,9 @@ def _handle_validate(self, promiser, attributes):
         except ValidationError as e:
             self.log_error(e)
             self._result = Result.INVALID
+        except Exception as e:
+            self.log_error(f"{type(e).__name__}: {e}")
+            self._result = Result.INVALID
```
olehermanse: According to the spec, I think this should be log level critical, and the result should be error:

https://github.com/cfengine/core/blob/master/docs/custom_promise_types/modules.md#results

olehermanse (approved):
Makes sense, one comment about following the spec, but not super critical.

docs/custom_promise_types/cfengine.py
larsewi: > Makes sense, one comment about following the spec, but not super critical.
>> According to the spec, I think this should be log level critical, and the result should be error:

It's not super critical, but it should be critical :laughing: \
I will fix !

olehermanse (approved):
LGTM, but should not have a Changelog entry - doesn't change anything in the released packages.

olehermanse (approved):

PR #4519, opened 2021-02-25
Assert results from write() in cf-execd/cf-execd-runagent.c with NDEBUG_UNUSED

vpodzime:
Otherwise LGTM complains that write() return value is ignored and
if it's only asserted, it complains it's unused in non-debug
builds.


olehermanse (dismissed):
Great, thanks. Please merge the libntech PR first and update this PR with the libntech merge commit, so hashes are synced.

larsewi (approved):
ACK

PR #4521, opened 2021-02-27
CFE-3570: Added groups promises prototype

larsewi:
Ticket: CFE-3570
Changelog: None

### DIFF
--- a/docs/custom_promise_types/groups_experimental.cf
+++ b/docs/custom_promise_types/groups_experimental.cf
@@ -1,0 +1,61 @@
+
+promise agent groups_experimental
+{
+    path => "$(this.promise_dirname)/../modules/promises/groups_experimental.py";
+    interpreter => "/usr/bin/python3";
+}
+
+bundle common foo {
+    vars:
+        # alice and bob should be in foo
+        # malcom should not be in foo
+        "members" string => '{"include": ["alice", "bob"], "exclude": ["malcom"]}';
+}
+
+bundle common bar {
+    vars:
+        # only alice should be in bar 
+        # only cannot be used with include or exclude
+        "members" string => '{"only": ["alice"]}';
+}
+
+bundle agent groups_experimental_prototype
+{
+    groups_experimental:
+        "foo"
+            policy => "present", # foo should exist
+            members => "$(foo.members)";
+
+    groups_experimental:
+        "bar"
+            # policy defaults to present, thus bar should exist
+            gid => "1337", # bar should have gid = 1337
+            members => "$(bar.members)";
+
+    groups_experimental:
+        "baz"
+            policy => "absent"; # should not exist
+}
+
+bundle agent example_users
+{
+    # make sure the users in the groups promise actually exists
+    users:
+        "alice"
+            policy => "present";
+    
+    users:
+        "bob"
+            policy => "present";
+    
+    users:
+        "malcom"
+            policy => "present";
+}
+
+bundle agent __main__
+{
+    methods:
+        "example_users";
+        "groups_experimental_prototype";
+}
--- a/docs/custom_promise_types/groups_experimental.py
+++ b/docs/custom_promise_types/groups_experimental.py
@@ -1,0 +1,350 @@
+import re
+import json
+from subprocess import Popen, PIPE
+from cfengine import PromiseModule, ValidationError, Result
+
+
+class GroupsExperimentalPromiseTypeModule(PromiseModule):
+    def __init__(self):
+        super().__init__("groups_experimental_promise_module", "0.0.1")
+        self._name_regex = re.compile(r"^[a-z_][a-z0-9_-]*[$]?$")
+        self._name_maxlen = 32
+
+
+    def validate_promise(self, promiser, attributes):
+        # check promiser type
+        if type(promiser) is not str:
+            raise ValidationError("Invalid type for promiser: expected string")
+
+        # check promiser value
+        if self._name_regex.match(promiser) is None:
+            self.log_warning(f"Promiser groupname '{promiser}' should match regular expression '[a-z_][a-z0-9_-]*[$]?'")
+
+        # check promiser length not too long
+        if len(promiser) > self._name_maxlen:
+            raise ValidationError(f"Promiser '{promiser}' is too long: ({len(promiser)} > {self._name_maxlen})")
+
+        # check attribute policy if present
+        if "policy" in attributes:
+            policy = attributes["policy"]
+
+            # check attribute policy type
+            if type(policy) is not str:
+                raise ValidationError("Invalid type for attribute policy: expected string")
+
+            # check attribute policy value
+            if policy not in ("present", "absent"):
+                raise ValidationError(f"Invalid value '{policy}' for attribute policy: must be 'present' or 'absent'")
+
+            # check attributes gid and members are not used with policy absent
+            if policy == "absent":
+                if "gid" in attributes:
+                    self.log_warning(f"Cannot assign gid to absent group '{promiser}'")
+                if "members" in attributes:
+                    self.log_warning(f"Cannot assign members to absent group '{promiser}'")
+
+        # check attribute gid if present
+        if "gid" in attributes:
+            gid = attributes["gid"]
+
+            # check attribute gid type
+            if type(gid) not in (str, int):
+                raise ValidationError("Invalid type for attribute gid: expected string or int")
+
+            # check attribute gid value
+            if type(gid) == str:
+                try: 
+                    int(gid)
+                except ValueError:
+                    raise ValidationError(f"Invalid value '{gid}' for attribute gid: expected integer literal")
+
+        # check attribute members if present
+        if "members" in attributes:
+            # remove escape chars followed by parsing json
+            members = json.loads(json.loads('"' + attributes["members"] + '"'))
+            attributes["members"] = members
+            
+            # check attribute only not used with attributes include or exclude
+            if "only" in members and ("include" in members or "exclude" in members):
+                raise ValidationError("Attribute 'only' may not be used with attributes 'include' or 'exclude'")
+            
+            # check attributes of attibutes in members
+            for attr in members:
+                if attr not in ("only", "include", "exclude"):
+                    raise ValidationError(f"Invalid value '{attr}' in attribute members: must be 'only', 'exclude' or 'include'")
+            
+            # make sure users aren't both included and excluded
+            if "include" in members and "exclude" in members:
+                duplicates = set(members["include"]).intersection(set(members["exclude"]))
+                if duplicates != set():
+                    raise ValidationError(f"Users {duplicates} both included and excluded from group '{promiser}'")
+
+
+    def evaluate_promise(self, promiser, attributes):
+        # keep track of any repairs or failed repairs
+        failed_repairs = 0
+        repairs = 0
+
+        # get group from '/etc/group' if present
+        group = None
+        try:
+            group = Group.lookup(promiser)
+        except GroupException as e:
+            self.log_error(f"Failed to lookup group '{promiser}': {e}")
+            failed_repairs += 1
+        
+        # get promised gid if present
+        promised_gid = attributes.get("gid")
+
+        # parse json in attribute members
+        if "members" in attributes:
+            # remove escape chars followed by parsing json
+            members = json.loads(json.loads('"' + attributes["members"] + '"'))
+            attributes["members"] = members
+
+        # set policy to present by default, if not specified
+        if "policy" not in attributes:
+            self.log_verbose(f"Policy not specified, defaults to present")
+            attributes["policy"] = "present"
+
+        # create group if policy present and group absent
+        if attributes["policy"] == "present" and group is None:
+            self.log_debug(f"Group '{promiser}' should be present, but does not exist")
+            try:
+                group = Group.create(promiser, promised_gid)
+            except GroupException as e:
+                self.log_error(f"Failed to create group '{promiser}': {e}")
+                failed_repairs += 1
+            else:
+                self.log_info(f"Created group '{promiser}'")
+                repairs += 1
+
+        # delete group if policy absent and group present
+        elif attributes["policy"] == "absent" and group is not None:
+            self.log_debug(f"Group '{promiser}' should be absent, but does exist")
+            try:
+                # group is set to None here
+                group = group.delete()
+            except GroupException as e:
+                self.log_error(f"Failed to delete group '{promiser}': {e}")
+                failed_repairs += 1
+            else:
+                self.log_info(f"Deleted group '{promiser}'")
+                repairs += 1
+
+        # if group is now present, check attributes 'gid' and 'members'
+        if group is not None:
+            # check gid if present
+            if promised_gid is not None and promised_gid != group.gid:
+                self.log_error(f"There is an existing group '{promiser}' with a different GID ({group.gid}) than promised ({promised_gid})")
+                # We will not try to repair this, as this might grant permissions to group
+                failed_repairs += 1
+            
+            # check members if present
+            if "members" in attributes:
+                members = attributes["members"]
+                set_members_repairs, set_members_failed_repairs = self._set_members(group, members)
+                repairs += set_members_repairs
+                failed_repairs += set_members_failed_repairs
+
+        self.log_debug(f"'{repairs}' repairs and '{failed_repairs}' failed repairs to promiser '{promiser}'")
+        if failed_repairs > 0:
+            self.log_error(f"Promise '{promiser}' not kept")
+            return Result.NOT_KEPT
+
+        if repairs > 0:
+            self.log_info(f"Promise '{promiser}' repaired")
+            return Result.REPAIRED
+
+        self.log_verbose(f"Promise '{promiser}' kept")
+        return Result.KEPT
+
+
+    def _set_members(self, group, members):
+        repairs = 0
+        failed_repairs = 0
+
+        for attribute in members:
+            if attribute == "include":
+                users = members["include"]
+                include_users_repairs, include_users_failed_repairs = self._include_users(group, users)
+                repairs += include_users_repairs
+                failed_repairs += include_users_failed_repairs
+
+            elif attribute == "exclude":
+                users = members["exclude"]
+                exclude_users_repairs, exclude_users_failed_repairs = self._exclude_users(group, users)
+                repairs += exclude_users_repairs
+                failed_repairs += exclude_users_failed_repairs
+
+            elif attribute == "only":
+                users = members["only"]
+                only_users_repairs, only_users_failed_repairs = self._only_users(group, users)
+                repairs += only_users_repairs
+                failed_repairs += only_users_failed_repairs
+
+        return repairs, failed_repairs
+
+    
+    def _include_users(self, group, users):
+        repairs = 0
+        failed_repairs = 0
+
+        for user in users:
+            self.log_debug(f"User '{user}' should be included in group '{group.name}'")
+            if user in group.members:
+                self.log_debug(f"User '{user}' already included in group '{group.name}'")
+            else:
+                self.log_debug(f"User '{user}' not included in group '{group.name}'")
+                try:
+                    group.add_member(user)
+                except GroupException as e:
+                    self.log_error(f"Failed to add user '{user}' to group '{group.name}': {e}")
+                    failed_repairs += 1
+                else:
+                    self.log_info(f"Added user '{user}' to group '{group.name}'")
+                    repairs += 1
+        
+        return repairs, failed_repairs
+
+    
+    def _exclude_users(self, group, users):
+        repairs = 0
+        failed_repairs = 0
+
+        for user in users:
+            self.log_debug(f"User '{user}' should be excluded from group '{group.name}'")
+            if user in group.members:
+                self.log_debug(f"User '{user}' not excluded from group '{group.name}'")
+                try:
+                    group.remove_member(user)
+                except GroupException as e:
+                    self.log_error(f"Failed to remove user '{user}' from group '{group.name}': {e}")
+                    failed_repairs += 1
+                else:
+                    self.log_info(f"Removed user '{user}' from group '{group.name}'")
+                    repairs += 1
+            else:
+                self.log_debug(f"User '{user}' already excluded from group '{group.name}'")
+
+        return repairs, failed_repairs
+
+    
+    def _only_users(self, group, users):
+        repairs = 0
+        failed_repairs = 0
+
+        self.log_debug(f"Group '{group.name}' should only contain members {users}")
+        if (set(users) != set(group.members)):
+            self.log_debug(f"Group '{group.name}' does not only contain members {users}")
+            try:
+                group.set_members(users)
+            except GroupException as e:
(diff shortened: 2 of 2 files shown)


docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
...
+        # if attribute gid is present
+        if "gid" in attributes:
+            gid = attributes["gid"]
+
+            # check attribute gid type
+            if type(gid) is not str:
+                raise ValidationError("invalid type for attribute gid: expected string")
+
+            # check attribute gid value
+            if type(gid) == str:
```
vpodzime, line 56: This condition is not necessary. If it was `False`, an exception would be raised above. Also, did you try specifying the GID as an integer in the policy? Does CFEngine send a string to the module in such case? I think we should be able to send integers in the JSON.

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
...
+
+        # if attribute members is present
+        if "members" in attributes:
+            # some black magic which removes back slashes parsing before parsing json
+            members = json.loads(json.loads('"' + attributes["members"] + '"'))
+            attributes["members"] = members
+            
+            # check attribute only not used with attributes include or exclude
+            if "only" in members and ("include" in members or "exclude" in members):
+                raise ValidationError("attribute only may not be used with attributes include or exclude")
```
vpodzime, line 70: I'd add single quotes around the attribute names to make the message easier to read.

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
...
+            
+            # check attributes of attibutes in members
+            for attr in members:
+                if (attr != "only" and attr != "include" and attr != "exclude"):
```
vpodzime, line 74: `if any(attr not in ("only", "include", "exclude") for attr in members)`

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
...
+                    raise ValidationError(f"invalid value '{attr}' in attribute members: must be only, exclude or include")
+            
+            # check values of attributes in members
+            for users in members.values():
+                for user in users:
+                    match = re.match(self._name_regex, user)
+                    if match == None:
+                        self.log_warning(f"username '{user}' not according to recomendations in system reference manual")
+                    if len(user) > self._name_maxlen:
+                        raise ValidationError(f"username too long: ({len(user)} > {self._name_maxlen})")
```
vpodzime, line 84: I don't think this should care about the validity of the usernames, they are or aren't users existing in the system, that's all we this needs to care about, right?

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
...
+
+            group = Group.lookup(promiser)
+            if "gid" in attributes:
+                gid = attributes["gid"]
+
+                if gid != group.gid:
+                    try:
+                        self.log_info(f"setting GID of group '{group.name}' to '{gid}'")
+                        Group.gid_set(group.name, gid)
+                        repairs += 1
```
vpodzime, line 143: I'm not quite sure this should be done. This is effectively renaming the existing group not setting GID for a new group. There might be files on the file system with that `gid` and so this could grant permissions to those files to this new group. I think it would be safer to make this an error.

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
...
+                    except GroupException as e:
+                        self.log_error(f"failed to set GID of group '{group.name}' to '{gid}'")
+                        failed_repairs += 1
+            
+            if "members" in attributes:
+                members = attributes["members"]
+                ret = self._set_members(group, members)
```
vpodzime, line 150: `(set_members_failed_repairs, set_members_repairs) = self._set_members(group, members)`

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
...
+                    repairs += 1
+                except GroupException as e:
+                    self.log_error(f"failed to delete group '{promiser}': {e}")
+                    failed_repairs += 1
+            else:
+                self.log_debug(f"group {promiser} does not exist")
+
+        self.log_debug(f"{repairs} repairs, {failed_repairs} failed repairs")
+        if failed_repairs > 0:
+            self.log_verbose("promise not kept")
```
vpodzime, line 170: These log messages should at least mention the promiser.

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
...
+        self.log_verbose("promise kept")
+        return Result.KEPT
+
+
+    def _set_members(self, group, members):
+        failed_repairs = 0
+        repairs = 0
+
+        for attribute in members:
+            if attribute == "include":
```
vpodzime, line 186: Please create separate functions for the branches in this `if` and then call the functions based on `if "include" in members"` etc.

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
...
+                        except GroupException as e:
+                            self.log_error(f"failed to remove user '{user}' from group '{group.name}': {e}")
+                            failed_repairs += 1
+                    else:
+                        self.log_debug(f"user '{user}' already excluded from group '{group.name}'")
+
+            elif attribute == "only":
+                users = members["only"]
+                self.log_debug(f"group '{group.name}' should only contain members {users}")
+                if (sorted(users) != sorted(group.members)):
```
vpodzime, line 221: `if set(users) != set(group.members)`

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
...
+        self.name = name
+        self.gid = gid
+        self.members = members
+
+    
+    @staticmethod
+    def lookup(group):
+        command = ["grep", f"^{group}", "/etc/group"]
+        process = Popen(command, stdout=PIPE, stderr=PIPE)
+        stdout, stderr = process.communicate()
```
vpodzime, line 252: You can just do:
```
with open("/etc/group") as f:
  for line in f:
    if line.startswith(group):
      ...
```
No need to fork-exec *grep*.

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
...
+
+    @staticmethod
+    def members_set(group, users):
+        command = ["gpasswd", "--members", ",".join(users), group]
+        process = Popen(command, stdout=PIPE, stderr=PIPE)
+        stdout, stderr = process.communicate()
+
+        if process.returncode != 0:
+            msg = stderr.decode("utf-8").strip().split()
+            raise GroupException(msg)
```
vpodzime, line 331: Please make these actual methods of `Group` objects -- `lookup()` and `create()` might be static methods returning new instances, or move these into a separate module as functions.

vpodzime (changes requested):
Looks good to me overall.

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,335 @@
+import re
+import json
+from subprocess import Popen, PIPE
+from cfengine import PromiseModule, ValidationError, Result
+
+
+class GroupsPromiseTypeModule(PromiseModule):
+    def __init__(self):
+        super().__init__("git_promise_module", "0.0.2")
+        self._name_regex = r"[a-z_][a-z0-9_-]*[$]?\Z"
```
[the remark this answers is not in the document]
larsewi: Sure. The `\Z` anchor specifies that a match must occur at the end of the input string, or before `\n` at the end of the input string.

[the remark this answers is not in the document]
vpodzime: Why not `$` then? And there should also be `^` in the beginning, I guess.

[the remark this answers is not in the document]
larsewi: Regular expression is used by `re.match` function, which matches from the beginning of the string. But I can add `$` to the beginning and call `re.search` to make this more explicit. Also as far as I know `\Z` is equivalent to `$`. But I don't mind changing it to `$` if you'd rather prefer that.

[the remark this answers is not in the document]
olehermanse: @larsewi I think `^` and `$` are more widely used and understood than `\A` and `\Z`.

I have been surprised by the match function before, making the beginning of line part explicit might be good.

docs/custom_promise_types/groups.py
olehermanse: Worth trying, but I'm afraid the integers are translated to string early and there's no difference between `512` and `"512"` during evaluation.

docs/custom_promise_types/groups.py
olehermanse: I think the halfway point is more readable:

```python
for attr in members:
    if attr not in ("only", "include", "exclude"):
```

vpodzime: I like `any()` and `all()` in conditions like these a lot. But YMMV, of course.

larsewi: Both suggestions greatly improve the code. But I lean towards Oles suggestion, as it was easier for me to read and understand.

docs/custom_promise_types/groups.py
larsewi: Good point!

docs/custom_promise_types/groups.py
larsewi: Good point ! :+1:

docs/custom_promise_types/groups.py
larsewi: :+1:

docs/custom_promise_types/groups.py
larsewi: It should only create group with GID, if it does not exist? E.g. `groupadd --gid 123  foo`

docs/custom_promise_types/groups.py
larsewi: It was not necessary, but now it is :wink:

docs/custom_promise_types/groups.py
larsewi: Makes more sense, yes

docs/custom_promise_types/groups.py
larsewi: Agreed !

docs/custom_promise_types/groups.py
larsewi: I'm not quite sure if the changes I've now made, corresponds to what you meant. But at least it's more OOP now, i think.

larsewi:
Thanks @vpodzime and @olehermanse for great reviews !

docs/custom_promise_types/groups.py
vpodzime: It's exactly what I mean. Very nice!

vpodzime (commented):
Many nitpicks, but nothing serious. Starting to look very nice, @larsewi. You are starting to be a good candidate for this nice badge: badge

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,338 @@
...
+        # get group from '/etc/group' if present
+        group = None
+        try:
+            group = Group.lookup(promiser)
+        except GroupException as e:
+            self.log_error(f"failed to lookup group '{promiser}': {e}")
+            failed_repairs += 1
+        
+        # get promised gid if present
+        promised_gid = attributes["gid"] if "gid" in attributes else None
```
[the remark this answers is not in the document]
larsewi: Nice to know! :clap:

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,338 @@
...
+        # set policy to present by default, if not specified
+        if "policy" not in attributes:
+            self.log_verbose(f"policy not specified, defaults to present")
+            attributes["policy"] = "present"
+
+        # create group if policy present and group absent
+        if attributes["policy"] == "present" and group is None:
+            self.log_debug(f"group '{promiser}' should be present, but does not exist")
+            try:
+                self.log_info(f"creating group '{promiser}'")
```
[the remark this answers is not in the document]
larsewi: Confusing indeed. I thought it would branch of to the except block immediately if an exception was caught. Did not think it would finish the try block before branching to except.

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,338 @@
...
+                group = group.delete()
+                repairs += 1
+            except GroupException as e:
+                self.log_error(f"failed to delete group '{promiser}': {e}")
+                failed_repairs += 1
+
+        # if group is now present, check attributes 'gid' and 'members'
+        if group:
+            # check gid if present
+            if promised_gid and promised_gid != group.gid:
```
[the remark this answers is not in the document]
larsewi: Good catch !

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,338 @@
...
+                self.log_error(f"GID '{group.gid}' for promiser '{promiser}' does not match promised GID '{promised_gid}'")
```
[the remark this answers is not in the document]
larsewi: - Yes, if we failed to create group with GID, its probably do to the fact that an existing group already has that GID. 
 - If group already exist, but with a different GID, it does not imply that an existing group has that GID. Thus  "There is an existing group ..."  may be a false statement.

docs/custom_promise_types/groups.py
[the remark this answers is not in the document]
larsewi: Should not `repairs += 1` be in else-block as well ? :thinking:

docs/custom_promise_types/groups.py
larsewi: :clap:

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,340 @@
...
+        self._name_maxlen = 32
+
+
+    def validate_promise(self, promiser, attributes):
+        # check promiser type
+        if type(promiser) is not str:
+            raise ValidationError("invalid type for promiser: expected string")
+
+        # check promiser value
+        match = re.match(self._name_regex, promiser)
```
vpodzime, line 20: Why not `self._name_regex.match(promiser)`?

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,340 @@
...
+        # create group if policy present and group absent
+        if attributes["policy"] == "present" and group is None:
+            self.log_debug(f"group '{promiser}' should be present, but does not exist")
+            try:
+                group = Group.create(promiser, promised_gid)
+            except GroupException as e:
+                self.log_error(f"failed to create group '{promiser}': {e}")
+                failed_repairs += 1
+            else:
+                self.log_info(f"creating group '{promiser}'")
```
vpodzime, line 120: *created*

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,340 @@
...
+        # delete group if policy absent and group present
+        elif attributes["policy"] == "absent" and group is not None:
+            self.log_debug(f"group '{promiser}' should be absent, but does exist")
+            try:
+                group = group.delete()
+            except GroupException as e:
+                self.log_error(f"failed to delete group '{promiser}': {e}")
+                failed_repairs += 1
+            else:
+                self.log_info(f"deleting group '{promiser}'")
```
vpodzime, line 132: *deleted*

docs/custom_promise_types/groups.py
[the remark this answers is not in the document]
vpodzime: >     * Yes, if we failed to create group with GID, its probably do to the fact that an existing group already has that GID.

But that should be handled above where `Group.create()` is called.


>     * If group already exist, but with a different GID, it does not imply that an existing group has that GID.

`Group.lookup()` searches by name, right? So `group.gid` would be GID of an existing group with the same name here.

Am I missing something?

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,340 @@
...
+                self.log_debug(f"user '{user}' already included in group '{group.name}'")
+            else:
+                self.log_debug(f"user '{user}' not included in group '{group.name}'")
+                try:
+                    group.add_member(user)
+                except GroupException as e:
+                    self.log_error(f"failed to add user '{user}' to group '{group.name}': {e}")
+                    failed_repairs += 1
+                else:
+                    self.log_info(f"adding user '{user}' to group '{group.name}'")
```
vpodzime, line 205: *added*

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,340 @@
...
+            self.log_debug(f"user '{user}' should be excluded from group '{group.name}'")
+            if user in group.members:
+                self.log_debug(f"user '{user}' not excluded from group '{group.name}'")
+                try:
+                    group.remove_member(user)
+                except GroupException as e:
+                    self.log_error(f"failed to remove user '{user}' from group '{group.name}': {e}")
+                    failed_repairs += 1
+                else:
+                    self.log_info(f"removing user '{user}' from group '{group.name}'")
```
vpodzime, line 225: *removed*

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,340 @@
...
+
+    @staticmethod
+    def create(name, gid=None):
+        command = ["groupadd", name]
+        if (gid):
+            command += ["--gid", gid]
+        process = Popen(command, stdout=PIPE, stderr=PIPE)
+        stdout, stderr = process.communicate()
+
+        if (process.returncode != 0):
```
vpodzime, line 290: Extra parentheses here.

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,340 @@
...
+            msg = stderr.decode("utf-8").strip().split('\n')[0]
```
vpodzime, line 291: Why only the first line from the error output?

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,340 @@
...
+            raise GroupException(msg)
+
+
+    def set_members(self, users):
+        command = ["gpasswd", "--members", ",".join(users), self.name]
+        process = Popen(command, stdout=PIPE, stderr=PIPE)
+        stdout, stderr = process.communicate()
+
+        if process.returncode != 0:
+            msg = stderr.decode("utf-8").strip().split()
```
vpodzime, line 335: `split()` on what?

docs/custom_promise_types/groups.py
larsewi: My thoughts were to try to follow the logging guide lines in [CONTRIBUTING.md](https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#logging-guidelines), stating "Normally, try to keep each message to one line of output, produced by one call to Log()."

Also the first line of the command line tools used here, clearly explains what's wrong. The remaining lines produced usually dwell into details on how to use that specific tool. This can be quite confusing to the user.

docs/custom_promise_types/groups.py
larsewi: :sweat_smile:

docs/custom_promise_types/groups.py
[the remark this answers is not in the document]
larsewi: It compares the promiser's GID to the promised GID. Thus if the promiser's GID does not match the GID it was promised, it does not imply that there is another group with that GID. It only means the promiser's GID did not match the promised GID.

This can happen if the promiser already exists, but with another GID. The error message will be: `"GID '123' for promiser 'foo' does not match promised GID '321'"`

If another group has the promised GID and the promiser does not exist. Creating the promiser will fail. And the error message will be: `"failed to create group 'foo': groupadd: group 'foo' already exists"`

The `lookup` function is used to look up the promiser group.

docs/custom_promise_types/groups.py
larsewi: Yes, that is much better :+1:

docs/custom_promise_types/groups.py
vpodzime: Makes sense, but then it should be explained in a comment in the code.

larsewi: Good point :+1:

docs/custom_promise_types/groups.py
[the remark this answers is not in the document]
vpodzime: I do understand that. It's just that `"GID '123' for promiser 'foo' does not match promised GID '321'"` is, for me, a very cryptic way of saying *There's an existing group foo which has a different GID ('123') than what was promised ('321')*.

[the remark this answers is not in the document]
larsewi: OK, then I understand. My bad. I thought you meant "there is another group than `foo`".
I agreee `"There's an existing group foo which has a different GID ('123') than what was promised ('321')"` is a lot clearer.

docs/custom_promise_types/groups.py
```diff
@@ -0,0 +1,345 @@
...
+    def __init__(self):
+        super().__init__("git_promise_module", "0.0.2")
+        self._name_regex = re.compile(r"^[a-z_][a-z0-9_-]*[$]?$")
+        self._name_maxlen = 32
+
+
+    def validate_promise(self, promiser, attributes):
+        # check promiser type
+        if type(promiser) is not str:
+            raise ValidationError("invalid type for promiser: expected string")
```
olehermanse, line 17: I think all your error messages and log messages should be changed to have Stor forbokstav ™

larsewi: Stor forbokstav ™ it is

vpodzime (dismissed):
Nice and clean! :+1:

docs/custom_promise_types/groups.cf
```diff
@@ -0,0 +1,61 @@
+
+promise agent groups
```
olehermanse, line 2:
```suggestion
promise agent groups_experimental
```

docs/custom_promise_types/groups.cf
```diff
@@ -0,0 +1,61 @@
...
+bundle common bar {
+    vars:
+        # only alice should be in bar 
+        # only cannot be used with include or exclude
+        "members" string => '{"only": ["alice"]}';
+}
+
+bundle agent groups_prototype
+{
+    groups:
```
olehermanse, line 24:
```suggestion
    groups_experimental:
```

docs/custom_promise_types/groups.cf
```diff
@@ -0,0 +1,61 @@
...
+        "foo"
+            policy => "present", # foo should exist
+            members => "$(foo.members)";
+
+    groups:
```
olehermanse, line 29:
```suggestion
    groups_experimental:
```

docs/custom_promise_types/groups.cf
```diff
@@ -0,0 +1,61 @@
...
+        "bar"
+            # policy defaults to present, thus bar should exist
+            gid => "1337", # bar should have gid = 1337
+            members => "$(bar.members)";
+
+    groups:
```
olehermanse, line 35:
```suggestion
    groups_experimental:
```

larsewi:
Funny that no one has commented on this. The danger of copy paste.
```python
super().__init__("git_promise_module", "0.0.2")
```

Change it to this
```python
super().__init__("groups_experimental_promise_module", "0.0.1")
```

Also changed promise type name to `groups_experimental`

vpodzime (approved):
ACK

PR #4523, opened 2021-03-01
Do not print error if there was no error in AllowAccessForUsers()

vpodzime:
Otherwise this confusing error message is logged:
  Failed to get UID for user 'cfapache': Success


olehermanse (approved):
👍 Thanks!

PR #4527, opened 2021-03-02
CFE-3599/master: Clarified error log message about untrusted state directory not being private

nickanderson:
Ticket: CFE-3599
Changelog: Title


vpodzime (approved):
ACK

PR #4530, opened 2021-03-04
Fix iteration over the 'extra_hashes' map in PolicyMerge

vpodzime:
'extra_hashes' is a StringMap which means the actual hash map is
in its 'impl' field.

### DIFF
--- a/libpromises/policy.c
+++ b/libpromises/policy.c
@@ -451,7 +451,7 @@ Policy *PolicyMerge(Policy *a, Policy *b)
 
     if (extra_hashes != NULL)
     {
-        MapIterator it = MapIteratorInit((Map*) extra_hashes);
+        MapIterator it = MapIteratorInit(extra_hashes->impl);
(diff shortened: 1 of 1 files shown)


libpromises/policy.c
```diff
@@ -451,7 +451,7 @@ Policy *PolicyMerge(Policy *a, Policy *b)
 
     if (extra_hashes != NULL)
     {
-        MapIterator it = MapIteratorInit((Map*) extra_hashes);
+        MapIterator it = MapIteratorInit(extra_hashes->impl);
```
olehermanse, line 454: Using the impl member outside like this looks weird, I would expect `StringMapIteratorInit()` or something like that, but okay.

olehermanse (approved):

libpromises/policy.c
vpodzime: Yeah, that's what I expected too. But that's out of scope of this change.

PR #4535, opened 2021-03-09
CFE-3569: Added policy variable sys.os_name_human

larsewi:
Ticket: CFE-3569
Changelog: Title

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -3335,6 +3335,99 @@ void GetDefVars(EvalContext *ctx)
         "jq --compact-output --monochrome-output --ascii-output --unbuffered --sort-keys",
         CF_DATA_TYPE_STRING, "invocation,source=agent,command_name=jq");
 }
+
+/*****************************************************************************/
+
+static void SysOSNameHuman(EvalContext *ctx)
+{
+    // This function sets the $(sys.os_name_human) variable
+    assert(ctx != NULL);
+    const char *const lval = "os_name_human";
+
+    /**
+     * Order is important. The first if-statement is more specific and wins.
+     * E.g. prefer Ubuntu over Debian.
+     */
+    if (EvalContextClassGet(ctx, NULL, "ubuntu") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "Ubuntu", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=ubuntu");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "debian") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "Debian", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=debian");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "centos") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "CentOS", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=centos");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "fedora") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "Fedora", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=fedora");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "redhat") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "RHEL", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=redhat");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "aix") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "AIX", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=aix");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "hpux") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "HP-UX", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=hpux");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "suse") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "SUSE", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=suse");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "macos") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "macOS", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=macos");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "windows") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "Windows", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=windows");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "freebsd") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "FreeBSD", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=freebsd");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "solaris") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "Solaris", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=solaris");
+    }
+    else
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "Unknown", CF_DATA_TYPE_STRING,
+                                      "source=agent");
+    }
+}
+
 /*****************************************************************************/
 
 void DetectEnvironment(EvalContext *ctx)
@@ -3347,4 +3440,5 @@ void DetectEnvironment(EvalContext *ctx)
     OSClasses(ctx);
     GetSysVars(ctx);
     GetDefVars(ctx);
+    SysOSNameHuman(ctx);
 }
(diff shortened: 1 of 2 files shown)


larsewi:
@cf-bottom Jenkins please

libenv/sysinfo.c
```diff
@@ -3335,6 +3335,80 @@ void GetDefVars(EvalContext *ctx)
...
+    const char *const lval = "os_name_human";
+    const char *const tags = "source=agent";
+
+    /**
+     * Order is important.
+     * E.g.
+     * |    A    |    B    | A -> B |
+     * |---------|---------|--------|
+     * | ubuntu  | debian  | true   |
+     * | debian  | ubuntu  | false  |
```
olehermanse, line 3353:
```suggestion
     * Order is important. The first if is more specific and wins,
     * for example, prefer Ubuntu over Debian.
```

This comment makes me more confused, I would go for a simpler sentence and no table.

larsewi: Yes, your totally right. Guess I got a little inspired by the illustration you showed me in the docs on files promise-type.

libenv/sysinfo.c
```diff
@@ -3335,6 +3335,76 @@ void GetDefVars(EvalContext *ctx)
...
+    const char *const tags = "source=agent";
+
+    /**
+     * Order is important. The first if-statement is more specific and wins.
+     * E.g. prefer Ubuntu over Debian.
+     */
+    if (EvalContextClassGet(ctx, NULL, "ubuntu") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "Ubuntu", CF_DATA_TYPE_STRING, tags);
```
olehermanse, line 3354: I like to add tags for exactly where the variable came from, so in this case:

```suggestion
        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
                                      "Ubuntu", CF_DATA_TYPE_STRING, "source=agent,derived-from=ubuntu");
```

Like I've done here: https://github.com/cfengine/masterfiles/blob/8748fbd14323e18a326ae7a2ddc0a7ca2fda1bfd/inventory/os.cf#L33-L35

(And same for all cases below).

@nickanderson do you agree?

tests/acceptance/01_vars/01_basic/os_name_human.cf
```diff
@@ -0,0 +1,54 @@
...
+    hpux::
+      "expected" string => "HPUX";
+    suse::
+      "expected" string => "SUSE";
+    macos::
+      "expected" string => "macOS";
+    windows::
+      "expected" string => "Windows";
+    freebsd::
+      "expected" string => "FreeBSD";
```
olehermanse, line 38: Since macOS is based on FreeBSD I would put it after, but I'm pretty sure `freebsd` is not defined on mac so it shouldn't make a difference.

olehermanse (approved):
Looks good, just some minor comments.

libenv/sysinfo.c
olehermanse: It also makes it very easy to grep for where the variable was defined, just grep for the tags.

tests/acceptance/01_vars/01_basic/os_name_human.cf
larsewi, line 38: Should solaris be in here as well ?

tests/acceptance/01_vars/01_basic/os_name_human.cf
```diff
@@ -0,0 +1,54 @@
...
+    redhat::
+      "expected" string => "RHEL";
+    centos::
+      "expected" string => "CentOS";
+    fedora::
+      "expected" string => "Fedora";
+    aix::
+      "expected" string => "AIX";
+    hpux::
+      "expected" string => "HPUX";
```
larsewi, line 30: Guess HPUX should be HP_UX ?

tests/acceptance/01_vars/01_basic/os_name_human.cf
larsewi: And maybe NetBSD, OpenBSD as well ?

tests/acceptance/01_vars/01_basic/os_name_human.cf
olehermanse: Yep!

tests/acceptance/01_vars/01_basic/os_name_human.cf
olehermanse: @larsewi Yes, solaris should be added. We don't have to add platforms outside of what we officially support and cover in CI / release testing, but I usually include FreeBSD in my PRs, because I have tested on FreeBSD (manually) before and can do it again. Same goes for macOS. I would probably skip OpenBSD and NetBSD, unless you or someone else can do a quick test and confirm that it works.

libenv/sysinfo.c
```diff
@@ -3335,6 +3335,93 @@ void GetDefVars(EvalContext *ctx)
...
+    else if (EvalContextClassGet(ctx, NULL, "suse") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "SUSE", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived=suse");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "macos") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "macOS", CF_DATA_TYPE_STRING,
```
vpodzime, line 3402: *MacOS*

libenv/sysinfo.c
```diff
@@ -3335,6 +3335,93 @@ void GetDefVars(EvalContext *ctx)
...
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "FreeBSD", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived=freebsd");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "solaris") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "Solaris", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived=windows");
+    }
```
vpodzime, line 3422: I think we should also have a default like *Unknown*. It's better than not expanding the variable when it's attempted to be used and doesn't exist.

vpodzime (commented):
Looks good to me otherwise.

libenv/sysinfo.c
larsewi: Are you sure? They tend to use lower case 'm' in the OS name. E.g. "macOS Big Sur"

libenv/sysinfo.c
larsewi: Yes, that makes sense

libenv/sysinfo.c
olehermanse: @larsewi @vpodzime lowercase is correct

libenv/sysinfo.c
```diff
@@ -3335,6 +3335,93 @@ void GetDefVars(EvalContext *ctx)
...
+
+    /**
+     * Order is important. The first if-statement is more specific and wins.
+     * E.g. prefer Ubuntu over Debian.
+     */
+    if (EvalContextClassGet(ctx, NULL, "ubuntu") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "Ubuntu", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived=ubuntu");
```
olehermanse, line 3355:
```suggestion
                                      "source=agent,derived-from=ubuntu");
```

Here and below.

olehermanse (changes requested):

olehermanse (approved):

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6172/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6172/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6172/

libenv/sysinfo.c
vpodzime: *macOS, previously Mac OS X...* :man_facepalming:

libenv/sysinfo.c
```diff
@@ -3335,6 +3335,99 @@ void GetDefVars(EvalContext *ctx)
...
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "FreeBSD", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=freebsd");
+    }
+    else if (EvalContextClassGet(ctx, NULL, "solaris") != NULL)
+    {
+        EvalContextVariablePutSpecial(ctx, SPECIAL_SCOPE_SYS, lval,
+                                      "Solaris", CF_DATA_TYPE_STRING,
+                                      "source=agent,derived-from=windows");
```
vpodzime, line 3421: *derived-from=solaris*

vpodzime (commented):
Also, fails on SUSE 15: https://ci.cfengine.com/job/testing-pr/label=PACKAGES_x86_64_linux_suse_15/13124/testReport/._01_vars_01_basic_os_name_human/cf/os_name_human_cf/
The failure there is because the class `suse` is not defined there which is not a problem in this PR. But the failure shows that `$(test.expected)` is not expanded in such a case while it should be `"Unknown"`.

vpodzime:
> But the failure shows that `$(test.expected)` is not expanded in such a case while it should be `"Unknown"`.

Hint: https://docs.cfengine.com/docs/3.15/reference-functions-isvariable.html

vpodzime (approved):
Looks good to me.

tests/acceptance/01_vars/01_basic/os_name_human.cf
```diff
@@ -0,0 +1,64 @@
...
+        string => "Test for expected values of variable sys.os_name_human";
+
+  #
+  # Note:
+  #
+  # Order is important, the last variable assignment wins.
+  # E.g. prefer Ubuntu over Debian.
+  #
+  # This test should also succeed if none of the os-classes are defined.
+  # This should be covered in another test.
```
olehermanse, line 19: @vpodzime @larsewi But it isn't covered in another test though, right? I would say that discovering suse is not defined on suse 15 was quite valuable, and we should cover this. Are you going to create a separate test for `sys.os_name_human != Unknown`? I think that's a valid test that we want to pass in all our tested platforms, always.

vpodzime: I think it should be covered in a different test which I thought Lars was working on.

larsewi: @olehermanse @vpodzime Yes I'm planning doing that next. But I believe that this test belongs to [ENT-6873](https://tracker.mender.io/browse/ENT-6873) along with the fix.

olehermanse (approved):
Great work!

larsewi:
Tested on Windows Server 2019 :heavy_check_mark:

tests/acceptance/01_vars/01_basic/os_name_human.cf
larsewi: @vpodzime @olehermanse  I have instead written a test that checks wether or not `sys.os_name_human` resolves to `"Unknown"` here [PR 4554](https://github.com/cfengine/core/pull/4554). My comment above is not longer valid.

PR #4552, opened 2021-03-25
Describe how 'classes' is handled with custom promise types

vpodzime:
Ticket: CFE-3437
Changelog: None

### DIFF
--- a/docs/custom_promise_types/modules.md
+++ b/docs/custom_promise_types/modules.md
@@ -37,12 +37,13 @@ bundle agent main
     linux::
       "/opt/cfengine/masterfiles"
         if => "redhat",
+        classes => some_classes_body,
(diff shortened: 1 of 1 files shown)


docs/custom_promise_types/modules.md
```diff
@@ -55,6 +56,7 @@ These attributes are handled by the agent, and cannot be used inside promise mod
 * `handle`
 * `meta`
 * `with`
+* `classes`
```
olehermanse, line 59: What happens if I try to make a custom promise with attribute `scope` or `promise_kept`, does it work? Based on your recent changes, I guess not, and so I think the full list of attributes which are not allowed should be listed here.

vpodzime: Good point!

olehermanse (approved):

PR #4576, opened 2021-04-02
Added switch-case fall through code style guideline

KotzaBoss:
Ticket: CFE-3416
Changelog: None

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -342,6 +342,24 @@ If you are a Guru, try to restrain yourself, only do magic when absolutely neces
           // Log message already printed above
       }
       ```
+* Switch-case deliberate fall through:
+  It should be explicitly indicated with a specific comment:
+  `// fall through` or `/* fall through */`
+  ```c
+  switch(x) {
+  case 1:
+  case 2:
+      // ...
+  // fall through
+  // Optional comment explaining why
+  case 3:
+      // ...
+      break;
+  default:
+      // ...
+      break;
+  }
+  ```
 * Variable names:
     * Describe what the variable is.
     * Are English words with underscores: `string_index`


CONTRIBUTING.md
```diff
@@ -342,6 +342,23 @@ If you are a Guru, try to restrain yourself, only do magic when absolutely neces
           // Log message already printed above
       }
       ```
+    * Deliberate switch-case fall through should be explicitly indicated with a specific comment:
```
vpodzime, line 345: I think this can be an item in the top-level list, not inside *Conditionals*.

vpodzime (commented):
Looks good to me otherwise.

CONTRIBUTING.md
KotzaBoss: My reasoning is that switch is technically a conditional statement so thats why i put it there. Do you think its more appropriate outside of contidionals as its own thing?

vpodzime: You are right, but the fall-through is specific to switch(), it has nothing to do with syntax and style for (other) conditionals.

KotzaBoss: alright, should i name the list item `* switch-case deliberate fallthrough`?

vpodzime: Yeah, that's fine.

CONTRIBUTING.md
```diff
@@ -342,6 +342,23 @@ If you are a Guru, try to restrain yourself, only do magic when absolutely neces
...
+      `// fall through` or `/* fall through */`
+      ```c
+      switch(x) {
+      case 1:
+      case 2:
+          // ...
+      // fall through
+      /* Optional comment explaining why */
```
olehermanse, line 353: I don't think you should mix the comment styles `/* */` and `//` unless it's necessary because you want a multiline comment. So in this case, I would:

```suggestion
      // Optional comment explaining why
```

KotzaBoss: I put the `/* */` to imply that it may be a multiline comment. It didnt come out well. Will fix along side Vratislavs change request

vpodzime (approved):
ACK

olehermanse (approved):

PR #4586, opened 2021-04-12
CFE-2262: Update old test for this.promiser in ifvarclass (master)

nickanderson:
No functional change, only updating metadata to refer to ticket in jira instead of redmine.

This test is still soft failing.


cf-bottom:

craigcomstock (approved):

PR #4594, opened 2021-04-15
Exceptions from Python-based custom promise modules

vpodzime:


misc/custom_promise_types/cfengine.py
```diff
@@ -2,6 +2,7 @@
 import json
 import traceback
 
+_LOG_LEVELS = {level: idx for idx, level in enumerate(("critical", "error", "warning", "notice", "info", "verbose", "debug"))}
```
larsewi, line 5: Wow, this is some high level python magic 👏

larsewi (approved):
Nice changes !

olehermanse (approved):

misc/custom_promise_types/cfengine.py
vpodzime: This is what people call Python these days. :wink:

PR #4604, opened 2021-04-18
Added autoconf message for whether debug option is present

olehermanse:
This message was useful to confirm that libntech is built
with debug options. I thought I might leave it in.


vpodzime (approved):
ACK

PR #4610, opened 2021-04-23
CFE-3476: Added string() policy function

KotzaBoss:
Ticket: CFE-3476
Changelog: None

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -8722,6 +8722,34 @@ static FnCallResult FnCallCFEngineCallers(EvalContext *ctx, ARG_UNUSED const Pol
     return FnReturnContainerNoCopy(callers);
 }
 
+static FnCallResult FnCallString(EvalContext *ctx,
+                                 ARG_UNUSED const Policy *policy,
+                                 const FnCall *fp,
+                                 const Rlist *finalargs)
+{
+    assert(finalargs != NULL);
+
+    char *ret = RvalToString(finalargs->val);
+    if (StringStartsWith(ret, "@("))
+    {
+        bool allocated = false;
+        JsonElement *json = VarNameOrInlineToJson(ctx, fp, finalargs, false, &allocated);
+
+        if (json != NULL)
+        {
+            Writer *w = StringWriter();
+            JsonWriteCompact(w, json);
+            ret = StringWriterClose(w);
+            if (allocated)
+            {
+                JsonDestroy(json);
+            }
+        }
+    }
+
+    return FnReturnNoCopy(ret);
+}
+
 /*********************************************************************/
 /* Level                                                             */
 /*********************************************************************/
@@ -9255,6 +9283,12 @@ static const FnCallArg HASH_TO_INT_ARGS[] =
     {CF_INTRANGE, CF_DATA_TYPE_INT, "Lower inclusive bound"},
     {CF_INTRANGE, CF_DATA_TYPE_INT, "Upper exclusive bound"},
     {CF_ANYSTRING, CF_DATA_TYPE_STRING, "Input string to hash"},
+    {NULL, CF_DATA_TYPE_NONE, NULL}
+};
+
+static const FnCallArg STRING_ARGS[] =
+{
+    {CF_ANYSTRING, CF_DATA_TYPE_STRING, "Convert argument to string"},
     {NULL, CF_DATA_TYPE_NONE, NULL}
 };
 
@@ -9948,6 +9982,8 @@ const FnCallType CF_FNCALL_TYPES[] =
     FnCallTypeNew("hash_to_int", CF_DATA_TYPE_INT, HASH_TO_INT_ARGS, &FnCallHashToInt, "Generate an integer in given range based on string hash",
               FNCALL_OPTION_NONE, FNCALL_CATEGORY_DATA, SYNTAX_STATUS_NORMAL),
 
+    FnCallTypeNew("string", CF_DATA_TYPE_STRING, STRING_ARGS, &FnCallString, "Convert argument to string",
+                  FNCALL_OPTION_NONE, FNCALL_CATEGORY_DATA, SYNTAX_STATUS_NORMAL),
     // read functions for reading from file
     FnCallTypeNew("readdata", CF_DATA_TYPE_CONTAINER, READDATA_ARGS, &FnCallReadData, "Parse a YAML, JSON, CSV, etc. file and return a JSON data container with the contents",
                   FNCALL_OPTION_NONE, FNCALL_CATEGORY_IO, SYNTAX_STATUS_NORMAL),
--- a/tests/acceptance/01_vars/02_functions/string.cf
+++ b/tests/acceptance/01_vars/02_functions/string.cf
@@ -1,0 +1,57 @@
+body common control
+{
+        inputs => { "../../default.cf.sub" };
+        bundlesequence => { default( "$(this.promise_filename)" ) };
+}
+
+bundle agent init
+{
+  classes:
+      "ExistingClass";
+
+  vars:
+      "a_int" int => "1k";
+      "a_real" real => "6.066";
+      "a_string" string => "Really interesting string";
+      "a_data_container_1" data => '{"foo": "bar"}';
+      "a_data_container_2" data => '[1, 2, 3]';
+      "a_data_container_3" data => '{"list": [1, 2, 3, 4], "obj": {"key": ["b", "a", "r"]}}';
+
+      "string_of_int" string => string("$(a_int)");
+      "string_of_real" string => string("$(a_real)");
+      "string_of_string" string => string("$(a_string)");
+      "string_of_bool_true" string => string(and("ExistingClass"));
+      "string_of_bool_false" string => string(and("NonExistingClass"));
+      "string_of_data_container_1" string => string(a_data_container_1);  # Strings not interpretted as references
+      "string_of_data_container_2" string => string("a_data_container_2");  # Strings not interpretted as references
+      "string_of_data_container_3" string => string(@(a_data_container_3));
+
+      #    "test_args" string => string();  # This is caught by the parser
+}
+
+bundle agent check
+{
+  meta:
+      "description" -> { "CFE-3476" }
+        string => "Test whether string() properly converts argument to string
+                   or string-serialises JSON *only* when explicitely passed as a '@()' reference";
+
+  classes:
+      "ok_int" expression => strcmp("1000", "$(init.string_of_int)");
+      "ok_real" expression => strcmp("6.066000", "$(init.string_of_real)");
+      "ok_string" expression => strcmp("Really interesting string", "$(init.string_of_string)");
+      "ok_bool_true" expression => strcmp("any", "$(init.string_of_bool_true)");
+      "ok_bool_false" expression => strcmp("!any", "$(init.string_of_bool_false)");
+      "ok_data_container_1" expression => strcmp("a_data_container_1", "$(init.string_of_data_container_1)");
+      "ok_data_container_2" expression => strcmp("a_data_container_2", "$(init.string_of_data_container_2)");
+      "ok_data_container_3" expression => strcmp('{"list":[1,2,3,4],"obj":{"key":["b","a","r"]}}',
+                                                 "$(init.string_of_data_container_3)");
+
+      "ok" expression => and("ok_int", "ok_real", "ok_string", "ok_bool_true",
+                             "ok_bool_false", "ok_data_container_1", "ok_data_container_2", "ok_data_container_3");
+
+  methods:
+      "Pass/Fail"
+        usebundle => dcs_passif( "ok", "$(this.promise_filename)" ),
+        inherit => "true";
+}
--- a/examples/string.cf
+++ b/examples/string.cf
@@ -1,0 +1,24 @@
+#+begin_src cfengine3
+bundle agent main
+{
+  classes:
+      "classA";
+      "classB";
+
+  vars:
+      "some_string" string => "cba";
+      "class_expressions" slist => {"classA.classB",
+                                    string(and("classA", strcmp("$(some_string)", "abc")))
+      };
+
+  reports:
+      "$(class_expressions)";
+}
+#+end_src
+
+#+begin_src example_output
+#@ ```
+#@ R: classA.classB
+#@ R: !any
+#@ ```
+#+end_src


KotzaBoss:
The test [file](tests/acceptance/TEST_FNCALLSTRING.cf) is obviously out of place. I couldnt make out how to properly make acceptance tests yet so i just run `cf-agent` on this to see my progress. Will be dealling with this soon

libpromises/evalfunction.c
```diff
@@ -8722,6 +8722,28 @@ static FnCallResult FnCallCFEngineCallers(EvalContext *ctx, ARG_UNUSED const Pol
...
+    }
+
+    char* ret;
+    switch(finalargs->val.type) {
+        case RVAL_TYPE_SCALAR:
+            ret = RvalToString(finalargs->val);
+            break;
+        default:
+            ProgrammingError("Shouldnt reach this TODO");
+    }
```
KotzaBoss, line 8742: Switch is overkill but i wrote it expecting to deal with different finalargs types. If the functions arguments are `CF_DATA_TYPE_STRING` can i expect that cfengine wont pass anything else?

tests/acceptance/01_vars/02_functions/string.cf
```diff
@@ -0,0 +1,56 @@
...
+	bundlesequence => { "init", "check" };
+}
+
+bundle agent init
+{
+	classes:
+		"ExistingClass";
+
+	vars:
+		"aInt" int => "1k";
```
olehermanse, line 12: please don't use camelcase, we use lowercase and underscores:

```suggestion
		"a_int" int => "1k";
```

tests/acceptance/01_vars/02_functions/string.cf
```diff
@@ -0,0 +1,56 @@
...
+		!okReal::
+			"$(thisfilename) real FAIL";
+		!okString::
+			"$(thisfilename) string FAIL";
+		!okBoolTrue::
+			'$(thisfilename) bool "any" FAIL';
+		!okBoolFalse::
+			'$(thisfilename) bool "!any" FAIL';
+		okALL::
+			"$(thisfilename) pass";
```
olehermanse: Change this to 2 messages, one for `ok` and one for `!ok`:

https://github.com/cfengine/core/blob/master/tests/acceptance/README#L159-L160

https://github.com/cfengine/core/blob/5d361487ce2da4d61c6fad81afbe22806ca9f5b2/tests/acceptance/30_custom_promise_types/11_unless.cf#L125-L128

tests/acceptance/01_vars/02_functions/string.cf
```diff
@@ -0,0 +1,56 @@
...
+		"convBoolFalse" string => string(and("NonExistingClass"));
+		"convRandInt" string => string(randomint(10, 1000));
+
+#		"testArgs" string => string();  # This is caught by the parser
+}
+
+bundle agent check
+{
+	vars:
+		"thisfilename" string => "$(this.promise_filename)";
```
olehermanse: Unnecessary

```suggestion
```

tests/acceptance/01_vars/02_functions/string.cf
```diff
@@ -0,0 +1,56 @@
...
+		"aInt" int => "1k";
+		"aReal" real => "6.066";
+		"aString" string => "Really interesting string";
+		"aIList" ilist => {1, 2, 3};
+		"aSList" slist => {
+			"ExistingClass",
+			string(and("ExistingClass", strcmp("bca", "abc")))
+		};
+
+		"convInt" string => string("$(aInt)");
```
olehermanse, line 21:
```suggestion
		"string_of_int" string => string("$(aInt)");
```

olehermanse (changes requested):
I'll do a more thorough review later, but here are some comments on the test.

tests/acceptance/01_vars/02_functions/string.cf
```diff
@@ -0,0 +1,41 @@
+body common control
+{
+	bundlesequence => { "init", "check" };
```
olehermanse, line 3: Please use (2) spaces for indentation, to be consistent with the other tests.

libpromises/evalfunction.c
```diff
@@ -8722,6 +8722,28 @@ static FnCallResult FnCallCFEngineCallers(EvalContext *ctx, ARG_UNUSED const Pol
     return FnReturnContainerNoCopy(callers);
 }
 
+static FnCallResult FnCallString(EvalContext *ctx,
```
olehermanse, line 8725:
```suggestion
static FnCallResult FnCallString(ARG_UNUSED EvalContext *ctx,
```

libpromises/evalfunction.c
```diff
@@ -8722,6 +8722,28 @@ static FnCallResult FnCallCFEngineCallers(EvalContext *ctx, ARG_UNUSED const Pol
...
+                                 const Policy *policy,
```
olehermanse, line 8726:
```suggestion
                                 ARG_UNUSED const Policy *policy,
```

libpromises/evalfunction.c
```diff
@@ -8722,6 +8722,28 @@ static FnCallResult FnCallCFEngineCallers(EvalContext *ctx, ARG_UNUSED const Pol
...
+                                 const FnCall *fp,
```
olehermanse, line 8727:
```suggestion
                                 ARG_UNUSED const FnCall *fp,
```

libpromises/evalfunction.c
```diff
@@ -8722,6 +8722,28 @@ static FnCallResult FnCallCFEngineCallers(EvalContext *ctx, ARG_UNUSED const Pol
...
+                                 const Rlist *finalargs)
+{
+    // Parser doesnt allow this to be true since STRING_ARGS has one argument
+    if(!finalargs) {
```
olehermanse, line 8731:
```suggestion
    if(finalargs == NULL) {
```

https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#safety--correctness

tests/acceptance/01_vars/02_functions/string.cf
```diff
@@ -0,0 +1,41 @@
...
+
+#		"test_args" string => string();  # This is caught by the parser
+}
+
+bundle agent check
+{
+	classes:
+		"ok_int" expression => strcmp("$(init.a_int)", "$(init.string_of_int)");
+		"ok_real" expression => strcmp("$(init.a_real)", "$(init.string_of_real)");
+		"ok_string" expression => strcmp("$(init.a_string)", "$(init.string_of_string)");
```
olehermanse, line 31: I think this is better:

```suggestion
		"ok_int" expression => strcmp("1000", "$(init.string_of_int)");
		"ok_real" expression => strcmp("6.066", "$(init.string_of_real)");
		"ok_string" expression => strcmp("Really interesting string", "$(init.string_of_string)");
```

I see 2 advantages:

- Test is more straight forward / easy to read
- What if both strings expand to `""` (empty string)? Your test would pass. (I guess this could happen if there is something wrong with variable expansion, or maybe you typed the wrong variable name).

In general, it is good to not repeat yourself, but a test is one of the exceptions where it makes sense.

tests/acceptance/01_vars/02_functions/string.cf
```diff
@@ -0,0 +1,41 @@
...
+		"ok_bool_true" expression => strcmp("any", "$(init.string_of_bool__true)");
+		"ok_bool_false" expression => strcmp("!any", "$(init.string_of_bool__false)");
```
olehermanse, line 33:
```suggestion
		"ok_bool_true" expression => strcmp("any", "$(init.string_of_bool_true)");
		"ok_bool_false" expression => strcmp("!any", "$(init.string_of_bool_false)");
```

I don't think the 2 underscores is necessary.

tests/acceptance/01_vars/02_functions/string.cf
```diff
@@ -0,0 +1,41 @@
...
+	vars:
+		"a_int" int => "1k";
+		"a_real" real => "6.066";
+		"a_string" string => "Really interesting string";
+
+		"string_of_int" string => string("$(a_int)");
+		"string_of_real" string => string("$(a_real)");
+		"string_of_string" string => string("$(a_string)");
+		"string_of_bool__true" string => string(and("ExistingClass"));
+		"string_of_bool__false" string => string(and("NonExistingClass"));
```
olehermanse, line 20:
```suggestion
		"string_of_bool_true" string => string(and("ExistingClass"));
		"string_of_bool_false" string => string(and("NonExistingClass"));
```

I don't think the 2 underscores is necessary.

tests/acceptance/01_vars/02_functions/string.cf
```diff
@@ -0,0 +1,41 @@
...
+bundle agent init
+{
+	classes:
+		"ExistingClass";
+
+	vars:
+		"a_int" int => "1k";
+		"a_real" real => "6.066";
+		"a_string" string => "Really interesting string";
+
```
olehermanse, line 15: `string()` should stringify a data container into a JSON serialized string, and the test should test this:

```suggestion
		"a_data_container" data => '{"foo": "bar"}';

```

libpromises/evalfunction.c
```diff
@@ -8722,6 +8722,28 @@ static FnCallResult FnCallCFEngineCallers(EvalContext *ctx, ARG_UNUSED const Pol
...
+        return FnFailure();
+    }
+
+    char* ret;
+    switch(finalargs->val.type) {
+        case RVAL_TYPE_SCALAR:
+            ret = RvalToString(finalargs->val);
+            break;
+        default:
+            ProgrammingError("Shouldnt reach this TODO");
```
olehermanse, line 8741: You should add a case for a data container, and convert it to a json string using:

https://github.com/cfengine/libntech/blob/6092009490d4219ec160a17a70ff083b7ef3a835/libutils/json.h#L581

olehermanse (changes requested):
A good start, I added a bunch of suggestions and things you need to fix. (I will do another review, and probably suggest more changes when you are done with these).

libpromises/evalfunction.c
tzz: Maybe this should be doing the same thing that `format("%s", anything)` does currently - no need to reimplement all that code, just factor out a stringification function.

tests/acceptance/01_vars/02_functions/string.cf
KotzaBoss: woops, that was careless vim substitution

libpromises/evalfunction.c
KotzaBoss: I got a hint from vratislav to checkout `VarNameOrInlineToJson` and i got it to work. With the current "signature" for the policy function only RVAL_TYPE_SCALAR is passed which means that a switch is unecessary.  But i still have doubts on the logics "integrity"

libpromises/evalfunction.c
```diff
@@ -8722,6 +8722,35 @@ static FnCallResult FnCallCFEngineCallers(EvalContext *ctx, ARG_UNUSED const Pol
...
+
+    char *ret;
+    bool allocated = false;  // Does it imply heap allocation is present?
+    JsonElement* json = VarNameOrInlineToJson(ctx, fp, finalargs, false, &allocated);
+
+    if(json != NULL)
+    {
+        Writer* w = StringWriter();
+        JsonWriteCompact(w, json);
+        ret = StringWriterClose(w);
```
vpodzime, line 8744: Looks like there should be `JsonDestroy()` somewhere here.

libpromises/evalfunction.c
```diff
@@ -8722,6 +8722,35 @@ static FnCallResult FnCallCFEngineCallers(EvalContext *ctx, ARG_UNUSED const Pol
...
+{
+    // Parser doesnt allow this to be true since STRING_ARGS has one argument
+    if(finalargs == NULL)
+    {
+        return FnFailure();
+    }
+
+    char *ret;
+    bool allocated = false;  // Does it imply heap allocation is present?
+    JsonElement* json = VarNameOrInlineToJson(ctx, fp, finalargs, false, &allocated);
```
vpodzime, line 8738: So if IIRC this would actually make `string("var_name")` return the string representation of the `$(var_name)` variable (if it exists). I'm not sure that's what we want here. I think `string("some_string")` should just be `"some_string"`.

vpodzime: IOW, I don't think this function should be a collecting function.

KotzaBoss: I tested it with collecting and it doesnt seem to do so, but true im not really sure COLLECTING is necessary. With NONE it works properly anyway

libpromises/evalfunction.c
KotzaBoss: Is the JsonDestroy necessary if [allocated](https://github.com/cfengine/core/blob/b248be4d04c9c508f6b76fe9b24fe40be9658aad/libpromises/evalfunction.c#L495) is set to true by the function? There are nested calls that push that pointer to bool further down but i cant find documentation on it.

vpodzime: Yes, `allocated == true` means the `json` was allocated and either `JsonDestroy()` is needed after it is used/converted or it needs to be returned to the caller (not this case).

libpromises/evalfunction.c
```diff
@@ -8722,6 +8722,35 @@ static FnCallResult FnCallCFEngineCallers(EvalContext *ctx, ARG_UNUSED const Pol
...
+                                 ARG_UNUSED const FnCall *fp,
+                                 const Rlist *finalargs)
+{
+    assert(finalargs != NULL);
+
+    char *ret;
+    bool allocated = false;
+    JsonElement *json = VarNameOrInlineToJson(ctx, fp, finalargs, false, &allocated);
+
+    if(json != NULL)
```
vpodzime, line 8736: Still missing spaces after `if`s.

tests/acceptance/01_vars/02_functions/string.cf
```diff
@@ -0,0 +1,49 @@
...
+    "a_string" string => "Really interesting string";
+    "a_data_container" data => '{"foo": "bar"}';
+    "a_data_container_2" data => '{"list": [1, 2, 3, 4], "obj": {"key": ["b", "a", "r"]}}';
+
+    "string_of_int" string => string("$(a_int)");
+    "string_of_real" string => string("$(a_real)");
+    "string_of_string" string => string("$(a_string)");
+    "string_of_bool_true" string => string(and("ExistingClass"));
+    "string_of_bool_false" string => string(and("NonExistingClass"));
+    "string_of_data_container" string => string(a_data_container);
```
vpodzime, line 23: How does `string("a_data_container")` behaves?

KotzaBoss:
```
vars:
  "a_data_container_3" data => '[1, 2, 3]';
  "string_of_data_container_3" string => string("a_data_container_3");
reports:
  "string of container 3 = $(string_of_data_container_3)";

### output
R: [1,2,3]
```
Is this desired behaviour?

vpodzime: Not what I would expect. @nickanderson @tzz @olehermanse thoughts?

olehermanse: > Not what I would expect. @nickanderson @tzz @olehermanse thoughts?

@vpodzime Why? Because there are no spaces after the commas?

KotzaBoss: >  Because there are no spaces after the commas?

I believe thats because i used the JsonWriteCompact

vpodzime: > @vpodzime Why? Because there are no spaces after the commas?

So you give the function a string. And because there's a variable with a name matching the string it gives you the string version of the variable? I don't like this. It may be OK for functions that clearly act on `slist/data` **variables**, but `string("some_string")` should just return `"some_string"`, I believe.

tzz: I see. Sorry if I misled @KotzaBoss with my earlier suggestion. I hope I can write a better explanation here.

Based on https://tracker.mender.io/browse/CFE-3476 I think @vpodzime and others are asking for converting **anything** to a string without awareness if it's a variable or not. That's very similar to `format("%s", ...)` except for how booleans are treated, so I would mention the equivalence in the docs for `string()`. Of course, you can choose to also make `format("%s", and(...))` work the same way.

In the docs for `string()` I would also mention `format("%S", ...)` to get the "other" behavior, serializing variables by name.

KotzaBoss: So we are interested in the InlineToJson part and not the VarName i assume?. For example a reference to json will serialize it but any "some_string" is not interpreted as anything other than a string

vpodzime: > For example a reference to json will serialize it but any "some_string" is not interpreted as anything other than a string

Yes, `json_var` and `@(json_var)` should use the variable, but `"json_var"` should just be a string.

olehermanse: > Yes, `json_var` and `@(json_var)` should use the variable, but `"json_var"` should just be a string.

@vpodzime AFAIK quotes are optional for strings in policy language. `json_var` and `"json_var"` are the same. `json_var` is not a reference, it is a string (that happens to be the name of a variable).

Maybe @nickanderson can clarify.

olehermanse: > So you give the function a string. And because there's a variable with a name matching the string it gives you the string version of the variable? I don't like this. It may be OK for functions that clearly act on `slist/data` **variables**, but `string("some_string")` should just return `"some_string"`, I believe.

@vpodzime @KotzaBoss So, based on this and my last comment, I think `string("json_var")` and `string(json_var)` should both return the string `"json_var"`.

My understanding is that the only things that would make sense to look up the variable and serialize the JSON would be: `string(@(json_var))` (or we could create another function which takes variable names, but let's not).

olehermanse (changes requested):

libpromises/evalfunction.c
```diff
@@ -8722,6 +8722,35 @@ static FnCallResult FnCallCFEngineCallers(EvalContext *ctx, ARG_UNUSED const Pol
...
+static FnCallResult FnCallString(ARG_UNUSED EvalContext *ctx,
+                                 ARG_UNUSED const Policy *policy,
+                                 ARG_UNUSED const FnCall *fp,
+                                 const Rlist *finalargs)
+{
+    assert(finalargs != NULL);
+
+    char *ret;
+    bool allocated = false;
+    JsonElement *json = VarNameOrInlineToJson(ctx, fp, finalargs, false, &allocated);
```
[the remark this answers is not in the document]
KotzaBoss: However that means that string(a_data_container) will not be considered a reference.

olehermanse (dismissed):

nickanderson (commented):
I prefer to use dcs for consistency but generally, this looks pretty good.

vpodzime (commented):
*Changelog: Title*

Looks good to me otherwise.

KotzaBoss:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6566/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6566/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6566/

tests/acceptance/01_vars/02_functions/string.cf
nickanderson: Definately, I would expect `string( @(json_var) )` to return the seralized refernece.

I don't know that I could point to anything and be able to consistently say if you should expect a bare reference to be the literal string or the expansion of a variable with the same name.

Look at all the ways `mergedata()` can work.

    bundle agent __main__
    {
      vars:
          "one" data => '{ "one": "red" }';
          "two" data => '{ "two": "blue" }';
          "merged_bare" data => mergedata( one, two);
          "merged_quote" data => mergedata( "one", "two");
          "merged_at" data => mergedata( @(one), @(two));
          "merged_quote_at" data => mergedata( "@(one)", "@(two)");
          "merged_bare_ref" data => mergedata( "@(one)", "@(two)", '{ "three": one }'); # Here, the unquoted /one/ will be the expansion of the /one/ data container, putting it under the three key.
    
      reports:
        "merged_bare: $(with)" with => string_mustache( "{{%-top-}}", merged_bare );
        "merged_quote: $(with)" with => string_mustache( "{{%-top-}}", "merged_quote" );
        "merged_at: $(with)" with => string_mustache( "{{%-top-}}", @(merged_at) );
        "merged_quote_at: $(with)" with => string_mustache( "{{%-top-}}", "@(merged_quote_at)" );
        "merged_bare_ref: $(with)" with => string_mustache( "{{%-top-}}", merged_bare_ref );
    }

    R: merged_bare: {
      "one": "red",
      "two": "blue"
    }
    R: merged_quote: {
      "one": "red",
      "two": "blue"
    }
    R: merged_at: {
      "one": "red",
      "two": "blue"
    }
    R: merged_quote_at: {
      "one": "red",
      "two": "blue"
    }
    R: merged_bare_ref: {
      "one": "red",
      "three": {
        "one": "red"
      },
      "two": "blue"
    }

olehermanse (approved):
Looks good now, thanks for working on this! :)

olehermanse:
Merging before CI completes, as the new acceptance test passed everywhere.

tests/acceptance/01_vars/02_functions/string.cf
olehermanse: @nickanderson `mergedata()` takes variable names (or data in {} or a `@` reference). You can see from your output that quoting does not matter, just like I said. (Yes there is a "special feature" with the unquoted `one` inside the data container, but that is a separate thing, IMO).

PR #4612, opened 2021-04-26
CFE-3603: Added example illustrating inherit_from (master)

nickanderson:
New example, intended for use in docs.

### DIFF
--- a/examples/inherit_from_classes.cf
+++ b/examples/inherit_from_classes.cf
@@ -1,0 +1,52 @@
+# This example illustrates the use of inherit_from to inherit body attribute
+# values from another body.
+###############################################################################
+#+begin_src cfengine3
+bundle agent __main__
+{
+  commands:
+      "/bin/true"
+        handle => "some meaningful string",
+        classes => my_set_some_extra_fancy_classes( "$(this.promiser)",
+                                                    "$(this.handle)",
+                                                    "some_class_to_handle_dependencies" );
+
+      "/bin/false"
+        handle => "some meaningless string",
+        classes => my_set_some_extra_fancy_classes( "$(this.promiser)",
+                                                    "$(this.handle)",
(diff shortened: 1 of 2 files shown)


craigcomstock (commented):
I wonder about the length of the body names, maybe foo and bar would be easier to read? I understand the meaning but it's a bit hard to "see" at a glance with such large names.

Maybe make the simplest example you can first, and then a more advanced example like what you have here second?

examples/inherit_from_classes.cf
```diff
@@ -0,0 +1,52 @@
...
+        classes => my_set_some_extra_fancy_classes( "$(this.promiser)",
+                                                    "$(this.handle)",
+                                                    "some_class_to_handle_dependencies" );
+
+  reports:
+      "Defined classes:$(const.n)$(with)"
+        with => join( "$(const.n)", sort( classesmatching( "some_.*"), "lex" ));
+}
+
+body classes my_set_some_extra_fancy_classes( x, y, z)
```
olehermanse, line 25:
```suggestion
body classes my_set_some_extra_fancy_classes(x, y, z)
```

examples/inherit_from_classes.cf
```diff
@@ -0,0 +1,52 @@
...
+# @brief In addition to the classes set by `set_some_fancy_classes` define `z` when the promise is repaired
+{
+        inherit_from => set_some_fancy_classes( $(x), $(y) );
+        promise_repaired => { "some_fancy_class_${x}_${y}_repaired", $(z) };
+}
+
+body classes set_some_fancy_classes ( x, y)
```
olehermanse, line 32:
```suggestion
body classes set_some_fancy_classes(x, y)
```

nickanderson:
Replaced by #4635 (i was targeting the wrong ticket number here).

PR #4618, opened 2021-04-27
ENT-6941 Increased range of allowed values of number of processes

Lex-2008:


olehermanse (changes requested):
The commit message should have an explanation of why.

olehermanse (approved):

PR #4622, opened 2021-04-29
Do not build 30_custom_promise_types/custom_promise_binary.c on Windows

vpodzime:
It fails to compile there and it's not needed because we don't
run acceptance tests on Windows.

Also use the platform-based AM conditional, building test
binaries have nothing to do with us building builtin extensions
even though it happens to be in the same cases.

Ticket: CFE-3562
Changelog: None


vpodzime:
RHEL 7 + mingw build: [Build Status](https://ci.cfengine.com/job/pr-pipeline/6490/)

larsewi (approved):
ACK

craigcomstock (commented):
ack. thanks! :)

PR #4642, opened 2021-05-10
ENT-6941: Increased range of allowed values of number of processes (3.15)

Lex-2008:
Issue was that even with current ±10% allowed spread of number of processes
enumerated by CFEngine vs `ps` this test was failing sometimes, with errors
like this:

> 27 is not within the range 21-26

Changelog: None

(cherry picked from commit 2c90edf9678d825ea23a37353aa904797cbb71b1)


cf-bottom:

olehermanse (approved):

PR #4649, opened 2021-05-14
ENT-6490: Add a new 'a_entry_required' parameter to DeleteDigestFromLastSeen()

vpodzime:
The lastseen local DB has entries of multiple types. In
particular, there are k-entries mapping hostkeys to IP addresses
and a-entries for the reverse mapping of IP addresses to
hostkeys. The ideal state is to have the entries of those two
types coherent (bijective), but it's not always possible. For
this reason, we need to be able to delete k-entries even if there
is no matching a-entry.

Do not require a-entries when deleting digests in
RemoveKeysFromLastSeen() if must_be_coherent is false.

Ticket: ENT-6490
Changelog: None

PR #4653, opened 2021-05-17
[3.15.x] Add a new 'a_entry_required' parameter to DeleteDigestFromLastSeen()

vpodzime:
The lastseen local DB has entries of multiple types. In
particular, there are k-entries mapping hostkeys to IP addresses
and a-entries for the reverse mapping of IP addresses to
hostkeys. The ideal state is to have the entries of those two
types coherent (bijective), but it's not always possible. For
this reason, we need to be able to delete k-entries even if there
is no matching a-entry.

Do not require a-entries when deleting digests in
RemoveKeysFromLastSeen() if must_be_coherent is false.

Ticket: ENT-6490
Changelog: None
(cherry picked from commit 93b778ca3aa60e130fc5d063dee9744d398a683d)

### DIFF
--- a/libpromises/lastseen.c
+++ b/libpromises/lastseen.c
@@ -510,9 +510,11 @@ clean:
  * @param[in,out] ip  : return the key corresponding host.
  *                      If NULL, return nothing
  * @param[in] ip_size : length of ip parameter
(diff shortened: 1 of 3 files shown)


vpodzime:
3.15.x version of https://github.com/cfengine/core/pull/4649

PR #4655, opened 2021-05-17
Make the ActionAttributeLogLevelFromString function be nicer on users

vpodzime:
Recognize both 'info' and 'inform' as LOG_LEVEL_INFO log level
and produce a warning if the log level is not recognized.

Also use StringEqual() instead of strcmp() == 0.

### DIFF
--- a/libpromises/attributes.c
+++ b/libpromises/attributes.c
@@ -649,16 +649,22 @@ LogLevel ActionAttributeLogLevelFromString(const char *log_level)
         return LOG_LEVEL_ERR;
     }
 
-    if (strcmp("inform", log_level) == 0)
+    if (StringEqual(log_level, "inform") || StringEqual(log_level, "info"))
     {
         return LOG_LEVEL_INFO;
     }
-    else if (strcmp("verbose", log_level) == 0)
+    else if (StringEqual(log_level, "verbose"))
     {
         return LOG_LEVEL_VERBOSE;
     }
+    else if (StringEqual(log_level, "error") || StringEqual(log_level, "log"))
+    {
+        /* XXX: what is 'log' and why is it the same as 'error'? */
+        return LOG_LEVEL_ERR;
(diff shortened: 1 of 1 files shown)


libpromises/attributes.c
```diff
@@ -649,16 +649,22 @@ LogLevel ActionAttributeLogLevelFromString(const char *log_level)
...
-    else if (strcmp("verbose", log_level) == 0)
+    else if (StringEqual(log_level, "verbose"))
     {
         return LOG_LEVEL_VERBOSE;
     }
+    else if (StringEqual(log_level, "error") || StringEqual(log_level, "log"))
+    {
+        /* XXX: what is 'log' and why is it the same as 'error'? */
+        return LOG_LEVEL_ERR;
+    }
```
olehermanse, line 664: I guess you added this because otherwise there were warnings? My wild guess would say that `log` is just the default if nothing is specified.

olehermanse (approved):

libpromises/attributes.c
vpodzime: Added what? There were no warnings before, `LOG_LEVEL_ERR` was just used as a fallback. But `log` is a valid value mentioned in the documentation so it shouldn't produce warnings. If nothing is specified, the global log level is used.

olehermanse: @vpodzime The `else if` block I am commenting on is something you added. Including the `XXX` comment inside of it. (I understand you didn't change the behavior).

I guess you added this **else if block** because otherwise there were warnings?

vpodzime: > I guess you added this else if block because otherwise there were warnings?
Yes, I wanted to add the warning for the unrecognized value because that was my initial motivation here (I was using `"info"` and it didn't work). So I needed all the valid values to be covered in `else if` branches with no warning produced.

PR #4656, opened 2021-05-17
Wrapped example_output with backticks

KotzaBoss:


nickanderson (approved):

olehermanse:
@nickanderson @KotzaBoss why?

KotzaBoss:
Compare the two documentations int _without_ backticks and string _with_ backticks
int: 
image
string:
image

You can see that because string is wrapped with \`\`\` it becomes a code block whereas int doesnt

olehermanse (approved):

PR #4662, opened 2021-05-19
Associate comments with variables and classes

vpodzime:
Missing bits:
- [x] trivial acceptance tests checking the `comment` attribute propagation from `vars` and `classes` promises

PR #4667, opened 2021-05-21
Verbose-log comments associated with 'vars' and 'classes' promises

vpodzime:
Ticket: CFE-2442
Ticket: CFE-2443
    
Changelog: Verbose log now contains comments associated with 'vars' and 'classes' promises


vpodzime:
Only the last commit is relevant for this PR, the rest is from #4662 which needs to be merged first.

larsewi (approved):
ACK

PR #4671, opened 2021-05-25
CFE-3041: Added check to ensure regular tests that pass with errors fail

KotzaBoss:
Ticket: CFE-3041
Changelog: None


KotzaBoss:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6616/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6616/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6616/

olehermanse (changes requested):
As mentioned in slack, you will need to rename the new failing tests due to this, to make CI green. Then we can work through fixing them one at a time. The extension for a test which produces an error should be `.error.cf`.

KotzaBoss:
Need some help on those 16_server acceptance tests.

vpodzime (commented):
Should the second and third commits be squashed? I don't see a point in renaming some tests there and back.

vpodzime:
@olehermanse shouldn't we have a new special extension and mechanism for these tests? With changes from here, if they fail to parse or cause a segfault or whatever, they will be considered as passed, right? While they should only print some `error:` messages but work otherwise.

olehermanse:
> @olehermanse shouldn't we have a new special extension and mechanism for these tests? With changes from here, if they fail to parse or cause a segfault or whatever, they will be considered as passed, right? While they should only print some error: messages but work otherwise.

@vpodzime no. `.error.cf` means test logs an error. Crash, segfault, syntax error, etc. would still cause test failure. Do you think it's beneficial to add another category for `test prints errors and prints pass message`? ( I think not ).

Explanation:

https://tracker.mender.io/browse/CFE-3041

olehermanse:
I think the problem here is that some tests print errors unintentionally, but not always. Only on some platforms / CIs.

vpodzime:
> Do you think it's beneficial to add another category for `test prints errors and prints pass message`? ( I think not ).

I can totally see such cases. But probably not worth the even greater complexity, it's quite crazy already.

KotzaBoss:
> I think the problem here is that some tests print errors unintentionally, but not always. Only on some platforms / CIs.

@olehermanse So there isnt much todo from the repos side of things?

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6851/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6851/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6851/

olehermanse:
@KotzaBoss As discussed, I think the change here is correct, however it does uncover some problems already existing in the codebase, making it red. We cannot merge it as is.

I am closing this for now, but as I said, please open a PR to do a few renames, and see if those are valid or not. For example you can look at the first test:

```
tests/acceptance/00_basics/01_compiler/string_contexts.cf
```

See if you can find the error, and see if it works to rename it to `.error.cf`.

PR #4676, opened 2021-05-26
[3.12.x] Do not assert namespace is not specified for 'def.' variables

vpodzime:
'def' is a pretty common name for a bundle and so using it
outside of the default namespace should be allowed. It doesn't
produce an error in PolicyCheckBundle() and referring to
variables from a 'def' bundle in a non-default namespace
should just work.

Ticket: CFE-3668
Changelog: None
(cherry picked from commit cfb55d6dd4f5f26ca6a9b89b9b5afe630aad7646)


vpodzime:
Not backporting this to *3.12.x* after all.

PR #4678, opened 2021-05-26
Actions: Fix apt failures by running update

olehermanse:
These actions started failing with 404 messages sometimes, hope this helps.


vpodzime (approved):
ACK

PR #4688, opened 2021-05-31
complicated_cpu_bound_benchmark.cf: Made default range lower to reduce runtime

olehermanse:
This makes the policy take around 1 second instead of 10 on my
machine. Users can always change this range, but having it
somewhat fast by default makes sense.


vpodzime (approved):
ACK

KotzaBoss:
Can confirm ~10 speed up on my machine too.

image

KotzaBoss (approved):

larsewi (commented):
ACK

PR #4690, opened 2021-06-01
Handle 'null' as valid data for augments and CMDB items

vpodzime:
Things like '"vars": null' should be valid and just noop.

Ticket: ENT-7201
Changelog: None

### DIFF
--- a/libpromises/cmdb.c
+++ b/libpromises/cmdb.c
@@ -503,18 +503,21 @@ bool LoadCMDBData(EvalContext *ctx)
 
     bool success = true;
     JsonElement *vars = JsonObjectGet(data, "vars");
-    if ((vars != NULL) && !ReadCMDBVars(ctx, vars))
+    if ((vars != NULL) && (JsonGetType(vars) != JSON_TYPE_NULL) &&
+        !ReadCMDBVars(ctx, vars))
     {
         success = false;
     }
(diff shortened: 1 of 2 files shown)


libpromises/cmdb.c
```diff
@@ -503,18 +503,21 @@ bool LoadCMDBData(EvalContext *ctx)
 
     bool success = true;
     JsonElement *vars = JsonObjectGet(data, "vars");
-    if ((vars != NULL) && !ReadCMDBVars(ctx, vars))
+    if ((vars != NULL) && (JsonGetType(vars) != JSON_TYPE_NULL) &&
+        !ReadCMDBVars(ctx, vars))
     {
         success = false;
```
olehermanse, line 509: What does `success` do here? At first glance, it seems to me that `null` is expected and should not cause `success = false`.

olehermanse (approved):
LGTM, just one question:

libpromises/cmdb.c
vpodzime: That's how it works, though, isn't it? `success` is only set to `false` if there is valid non-null data and processing fails.

olehermanse: Ah, yes you are correct. My thinking was inverted.

PR #4693, opened 2021-06-01
Make sure program name is used as identifier in syslog

vpodzime:
Merge together:
https://github.com/cfengine/core/pull/4693
https://github.com/cfengine/enterprise/pull/661
https://github.com/cfengine/nova/pull/1766

PR #4696, opened 2021-06-03
[3.15.x] Make sure program name is used as identifier in syslog

vpodzime:
As man:openlog(3) says:

  The string pointed to by 'ident' is prepended to every message,
  and is typically set to the program name.  If 'ident' is NULL,
  the program name is used.  (POSIX.1-2008 does not specify the
  behavior when ident is NULL.)

And of course not all systems we support are nice and use program
name if we pass NULL as the 'ident' argument.

Ticket: ENT-7100
Changelog: CFEngine processes are now properly identified in syslog on non-GNU/Linux systems
(cherry picked from commit 5147f4a93a87b4ba651c6f2c3fbe1853ad1b0a86)

### DIFF
--- a/cf-agent/cf-agent.c
+++ b/cf-agent/cf-agent.c
@@ -258,7 +258,10 @@ int main(int argc, char *argv[])
 
     GenericAgentConfigApply(ctx, config);
 
-    GenericAgentDiscoverContext(ctx, config);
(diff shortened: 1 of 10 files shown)


vpodzime:
Backport of #4693

PR #4702, opened 2021-06-04
cfengine.py: Added critical error message when validate_promise returns a value

olehermanse:


vpodzime (approved):
ACK

misc/custom_promise_types/cfengine.py
```diff
@@ -289,6 +289,7 @@ def _handle_validate(self, promiser, attributes, request):
                 self._result = Result.VALID
             else:
                 # Bad, validate method shouldn't return anything else
+                self.log_critical(f"Bug in promise module {self.name} - validate_promise() should not return anything")
```
vpodzime, line 292: Actually, this should have the nice info about the particular promise too.

vpodzime (approved):
(could be a follow-up)

PR #4703, opened 2021-06-04
cfengine.py: Add support for int attributes in promise types

olehermanse:
See:
https://github.com/cfengine/modules/pull/9

### DIFF
--- a/misc/custom_promise_types/cfengine.py
+++ b/misc/custom_promise_types/cfengine.py
@@ -149,12 +149,18 @@ class PromiseModule:
             if name not in self._validator_attributes:
                 # Unknown attribute, this will cause a validation error later
                 continue
-            # Only known conversion needed: "true"/"false" -> True/False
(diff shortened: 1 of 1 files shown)


tranchitella (approved):

misc/custom_promise_types/cfengine.py
```diff
@@ -149,12 +149,18 @@ def _convert_types(self, promiser, attributes):
...
                 if value == "true":
                     replacements[name] = True
                 elif value == "false":
                     replacements[name] = False
+            # "int" -> int()
+            elif self._validator_attributes[name]["typing"] is int:
+                try:
+                    replacements[name] = int(value)
+                except ValueError:
+                    pass
```
vpodzime, line 163: Shouldn't this result in a validation error?

tranchitella: It does, but it is handled by another part of the code.

PR #4705, opened 2021-06-04
CFE-3639: Refactored ilist/rlist parsing to expand numeric literals

KotzaBoss:
Ticket: CFE-3639
Changelog: Title


libpromises/cf3parse_logic.h
```diff
@@ -1028,4 +1030,58 @@ static inline void ParserBeginBundleBody()
...
+static inline void ParserHandleQuotedListItems()
+{
+    ParserDebug("\tP:%s:%s:%s:%s list append: "
+                "qstring = %s\n",
+                ParserBlockString(P.block), P.blocktype, P.blockid,
+                (P.currentclasses ?
+                    P.currentclasses : "any"),
+                P.currentstring);
+
+    if(StringContains(P.lval, "list") && *P.lval != 's')  // deal with numbers in ilist/rlist
```
olehermanse, line 1042: This is oddly specific and imprecise at the same time.

You wanted to capture `(ilist|rlist)`, instead you made something that matches `[^s].*list.*`

I think you should try to focus more on correctness and readability here, rather than performance and cleverness.

For example:

```C
if (StringEquals(P.lval, "ilist") || StringEquals(P.lval, "rlist"))
```

Or:

```C
if ((P.lval[0] == 'i' || P.lval[0] == 'r') && StringEquals(P.lval + 1, "list"))
```

libpromises/cf3parse_logic.h
```diff
@@ -1028,4 +1030,58 @@ static inline void ParserBeginBundleBody()
...
+
+        bool ok = DoubleFromString(P.currentstring, &value);
+        if(!ok)
+        {
+            // TODO: we dont really want DoubleFromStrings error output...
+            ParseError("Lval '%s' expects numbers", P.lval);
+            goto cleanup;
+        }
+
+        if(*P.lval == 'i')  // ilist
```
olehermanse, line 1055:
```suggestion
        if (P.lval[0] == 'i')  // ilist
```

libpromises/cf3parse_logic.h
```diff
@@ -1028,4 +1030,58 @@ static inline void ParserBeginBundleBody()
     P.useargs = NULL;
 }
 
+static inline void ParserHandleQuotedListItems()
+{
```
olehermanse, line 1034: I think it's a good idea to set up some of your assumptions here, as assertions. Especially the ones that are quite important for the reader to understand:

```C
assert(P.lval != NULL);
assert(StringEqual(P.lval, "slist") || StringEqual(P.lval, "ilist") || StringEqual(P.lval, "rlist"));
```

libpromises/cf3parse_logic.h
```diff
@@ -1028,4 +1030,58 @@ static inline void ParserBeginBundleBody()
...
+        }
+
+        RlistAppendScalar((Rlist **) &P.currentRlist,
+                          ret);
+    }
+    else
+    {
+        RlistAppendScalar((Rlist **) &P.currentRlist,
+                          (void *) P.currentstring);
+    }
```
olehermanse, line 1080: This is essentially to handle the `slist` case, right?

Handling this much earlier, with an early return, would be better IMO. Then all the code after it can assume it is not slist, and doesn't have to be indented inside an if.

Something like:

```
if (StringEqual(P.lval, "slist"))
```

libpromises/cf3parse_logic.h
```diff
@@ -1028,4 +1030,58 @@ static inline void ParserBeginBundleBody()
...
+
+cleanup:
+    free(P.currentstring);
+    P.currentstring = NULL;
```
olehermanse, line 1084: There is a macro for this, and I don't think such a simple cleanup warrants a goto:

```suggestion
    FREE_AND_NULL(P.currentstring);
```

libpromises/cf3parse_logic.h
```diff
@@ -1028,4 +1030,58 @@ static inline void ParserBeginBundleBody()
...
+    {
+        char ret[PRINTSIZE(intmax_t)] = {0};
+        double value;
+
+        bool ok = DoubleFromString(P.currentstring, &value);
+        if(!ok)
+        {
+            // TODO: we dont really want DoubleFromStrings error output...
+            ParseError("Lval '%s' expects numbers", P.lval);
+            goto cleanup;
```
olehermanse, line 1052:
```suggestion
            FREE_AND_NULL(P.currentstring);
            return;
```

libpromises/cf3parse_logic.h
```diff
@@ -1028,4 +1030,58 @@ static inline void ParserBeginBundleBody()
...
+                    P.currentclasses : "any"),
+                P.currentstring);
+
+    if(StringContains(P.lval, "list") && *P.lval != 's')  // deal with numbers in ilist/rlist
+    {
+        char ret[PRINTSIZE(intmax_t)] = {0};
+        double value;
+
+        bool ok = DoubleFromString(P.currentstring, &value);
+        if(!ok)
```
olehermanse, line 1048:
```suggestion
        if (!ok)
```

libpromises/cf3parse.y
```diff
@@ -749,16 +749,7 @@ litem:                 IDENTIFIER
...
-                                       "qstring = %s\n",
-                                       ParserBlockString(P.block), P.blocktype, P.blockid,
-                                       (P.currentclasses ?
-                                            P.currentclasses : "any"),
-                                       P.currentstring);
-                           RlistAppendScalar((Rlist **) &P.currentRlist,
-                                             (void *) P.currentstring);
-                           free(P.currentstring);
-                           P.currentstring = NULL;
+                           ParserHandleQuotedListItems();
```
olehermanse, line 752: Not a big deal, but something to consider for the next time:

If you copied this code into a function without changing anything, only whitespace, in a separate commit, it would be easier to review and see what changes you made to it.

libpromises/cf3parse_logic.h
```diff
@@ -1028,4 +1030,58 @@ static inline void ParserBeginBundleBody()
...
+            if(StringContainsChar(P.currentstring, '.'))
+            {
+                ParseWarning(PARSER_WARNING_TRUNCATED,
+                             "Real number '%s' will be truncated",
+                             P.currentid);
+            }
+
+            xsnprintf(ret, PRINTSIZE(intmax_t), "%jd", (intmax_t)value);
+        }
+        else  // rlist
```
olehermanse, line 1068: I like that you put comments like this.

You could also make it an assert, then you have more assurance that it is correct:

```
else
{
    assert(StringEqual(P.lval, "rlist"));
```

tests/acceptance/01_vars/01_basic/parse_list_quoted_number_literals.cf
```diff
@@ -0,0 +1,32 @@
...
+      "only_str_ints_neg"  ilist => { "-1", "-1k", "-1m", "-1g",
+                                      "-1K", "-1M", "-1G" };
+
+      "only_str_reals_pos"  rlist => { "1.0", "1.0k", "1.0m", "1.0g",
+                                       "1.0K", "1.0M", "1.0G" };
+
+      "only_str_reals_neg"  rlist => { "-1.0", "-1.0k", "-1.0m", "-1.0g",
+                                       "-1.0K", "-1.0M", "-1.0G" };
+
+      "truncation" ilist => { "1.0", "1.9" , "1", };  # Parser must show warning
```
olehermanse, line 23: Maybe add some numbers which don't start with 1

libpromises/cf3parse_logic.h
KotzaBoss: These exact assertions fail because the parser currently will call the litem rule for more than just ilist/rlist/slist but also inputs like in the test. So these assertions at the top dont work _unless_ i expect this handle to be called only for them. Implying an if check at the parser.
```
litem: QUOTED_STRING
          {
                 if (StringEqual(P.lval, "ilist") || StringEqual(P.lval, "rlist"))
                 {
                         ParserHandleIRlistQuotedStrings();
                 }
                 ...
          }
```
the logic is the same `if i/rlist expand otherwise just pass the string` its just how it is structured.
As i see it the two ways are handle it completely in the Handle or use the Handle only for the irlists.

I will transform the current handle to be more focused on ir/lists and see if it is cleaner.

libpromises/cf3parse.y
```diff
@@ -755,10 +755,21 @@ litem:                 IDENTIFIER
                                        (P.currentclasses ?
                                             P.currentclasses : "any"),
                                        P.currentstring);
-                           RlistAppendScalar((Rlist **) &P.currentRlist,
-                                             (void *) P.currentstring);
-                           free(P.currentstring);
-                           P.currentstring = NULL;
+
+                           if (!StringEqual(P.lval + 1, "list") || P.lval[0] == 's')  // slists and everything else
```
olehermanse, line 759: This code assumes `!NULL_OR_EMPTY(P.lval);`. Please add another if or assert for this.

libpromises/cf3parse.y
```diff
@@ -755,10 +755,23 @@ litem:                 IDENTIFIER
...
                                             P.currentclasses : "any"),
                                        P.currentstring);
-                           RlistAppendScalar((Rlist **) &P.currentRlist,
-                                             (void *) P.currentstring);
-                           free(P.currentstring);
-                           P.currentstring = NULL;
+
+                           assert(!NULL_OR_EMPTY(P.lval));
+
+                           if (!StringEqual(P.lval + 1, "list") || P.lval[0] == 's')  // slists and everything else
```
vpodzime, line 761: Please swap the two conditions, they will be easier to read and understand and also checking one character is faster/cheaper than checking a substring.

libpromises/cf3parse.y
```diff
@@ -755,10 +755,23 @@ litem:                 IDENTIFIER
...
+                               RlistAppendScalar((Rlist **) &P.currentRlist,
+                                                 (void *) P.currentstring);
+                               FREE_AND_NULL(P.currentstring);
+                           }
+                           else if (P.lval[0] == 'i')  // ilist
+                           {
+                               ParserHandleQuotedIlistItems();
+                           }
+                           else // rlist
+                           {
```
vpodzime, line 772: It's better to add asserts instead of these `ilist/rlist` comments. They bring more value than just the comments.

libpromises/cf3parse_logic.h
```diff
@@ -1028,4 +1030,57 @@ static inline void ParserBeginBundleBody()
     P.useargs = NULL;
 }
 
+static inline void ParserHandleQuotedIlistItems()
+{
+    assert(P.lval != NULL);
+    assert(StringEqual(P.lval, "ilist"));
+
+    char ret[PRINTSIZE(intmax_t)] = {0};
```
vpodzime, line 1038: Please only define `ret` right before it is needed.

vpodzime (commented):
Looks good to me otherwise. As mentioned in Slack, please add an explanation of what is being changed/enhanced and why to the commit message.

KotzaBoss:
The current implementation has a problem which propably stems from my lack of knowledge of the parser as a whole
```
"i" int => "100";
"lst" ilist => { "${i}" };

Parser output:
   error: Failed to parse real number: ${i}
./01_vars/01_basic/parse_list_quoted_number_literals.cf:13:26: error: Lval 'ilist' expects numbers
  "lst" ilist => { "${i}" };
                         ^
  ```
Shouldnt the evaluation of a variable happen before the handle is called?

**Comment by vpodzime** [comment]
> Shouldnt the evaluation of a variable happen before the handle is called?

No, the parser must preserve the variable reference. A variable reference is a different thing than a value. Like a pointer and a value in C, for example.

File: libpromises/parser.c [hunk 15]
```diff
@@ -168,6 +168,10 @@ int ParserWarningFromString(const char *warning_str)
     {
         return PARSER_WARNING_ALL;
     }
+    else if (strcmp("truncated", warning_str) == 0)
```
vpodzime, line 171: `StringEqual()`

vpodzime (commented):
Looks like this breaks some acceptance tests.

libpromises/parser.c
KotzaBoss: If i change only this line it will be inconsitent as there will be `strcmp`s above and just one `StringEqual` at the bottom. Should i change the above too?

vpodzime: Either change those in a separate commit or leave `strcmp()` here too.

libpromises/cf3parse_logic.h
```diff
@@ -1028,4 +1031,63 @@ static inline void ParserBeginBundleBody()
...
+    if (!ok)
+    {
+        // TODO: we dont really want DoubleFromStrings error output...
+        ParseError("Lval '%s' expects numbers", P.lval);
+        FREE_AND_NULL(P.currentstring);
+        return;
+    }
+
+    char ret[PRINTSIZE(intmax_t)] = {0};
+    xsnprintf(ret, PRINTSIZE(intmax_t), "%lf", value);
```
vpodzime, line 1087: It doesn't make much sense to use `intmax_t` here.

vpodzime (commented):
Thanks for the improvements!

libpromises/cf3parse_logic.h
```diff
@@ -1028,4 +1031,63 @@ static inline void ParserBeginBundleBody()
...
+    if (!ok)
+    {
+        // TODO: we dont really want DoubleFromStrings error output...
+        ParseError("Lval '%s' expects numbers", P.lval);
+        FREE_AND_NULL(P.currentstring);
+        return;
+    }
+
+    char ret[PRINTSIZE(double)] = {0};
+    xsnprintf(ret, PRINTSIZE(double), "%lf", value);
```
vpodzime, line 1087: I don't think `PRINTSIZE()` works for `double`.

KotzaBoss:
```c
#include <stdio.h>

#define PRINTSIZE(what) (3 + 53 * sizeof(what) / 22)

int main() {
    printf("%d", PRINTSIZE(double));
}
```
Locally the above program works and outputs `22`

KotzaBoss:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6799/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6799/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6799/

libpromises/cf3parse_logic.h
vpodzime: And how long is the biggest double value with the highest possible precision?

KotzaBoss: i looked more closely into printsize and indeed it only works for integrals and not IEEE floating point dark magic. However that leaves me in a predicament.

I gave a ludicrously large number as list item and i crossreferenced that with
`#define CF_REALRANGE "-9.99999E100,9.99999E100"` which gives 101 digits (excluding dot, sign, decimals and null terminator)
Decimals seem to not influence the parsing but im still investigating (on the other hand we seem to be interested in 6 decimal precision so that may not influence the problem too much).

With that in mind does it make sense to allocate a buffer of:
```
sign   digits   dot   "forced" precision   '\0'
1    + 101    + 1   + 6                  + 1    = 120 characters?
```
And use snprintf:
```c
#define DOUBLE_PRINTSIZE (1 + 101 +  1 + 6 + 1)
char ret[DOUBLE_PRINTSIZE] = {0};

snprintf(ret, DOUBLE_PRINTSIZE, "%.6lf", value);
```

vpodzime: What if people actually care about the precision? Floating point numbers are designed in a way to support storage of small numbers with high precision or large numbers with low precision. We should support both. The easiest case might be just using the string from the policy without the conversion there and back in cases where that is possible.

KotzaBoss:
After discussion with @olehermanse most of this work here will be discarded and a new pr with more specific changes will be introduced

PR #4718, opened 2021-06-18
Restore hostname back to 'localhost' in case it was changed

Lex-2008:
...by something else.

Changelog: none

### DIFF
--- a/tests/acceptance/04_examples/outputs/check_outputs.cf
+++ b/tests/acceptance/04_examples/outputs/check_outputs.cf
@@ -22,6 +22,12 @@ bundle agent init
   reports:
     DEBUG_LOTS::
       "$(this.bundle): will consider example $(all_examples)";
+
+  commands:
+      # restore hostname to be "localhost" - it's set to be like this in
+      # https://github.com/mendersoftware/mender-qa/blob/master/scripts/initialize-user-data.sh#L57
+      # but might be changed somewhere else
+      "hostname localhost";
 }
 
 #######################################################


tests/acceptance/04_examples/outputs/check_outputs.cf
```diff
@@ -22,6 +22,12 @@ bundle agent init
   reports:
     DEBUG_LOTS::
       "$(this.bundle): will consider example $(all_examples)";
+
+  commands:
+      # restore hostname to be "localhost" - it's set to be like this in
+      # https://github.com/mendersoftware/mender-qa/blob/master/scripts/initialize-user-data.sh#L57
+      # but might be changed somewhere else
+      "hostname localhost";
```
vpodzime, line 30: Is this test marked as unsafe or is it going to change hostname of my machine the next time I run `./testall` or `make check`?

Lex-2008: good point! Yes, it will change your hostname.

vpodzime: Then it's not a good way to go. (Unless the *check_outputs.cf* test is skipped by default)

olehermanse: Seems like it should not be a part of this test at all? Restore the hostname after the test which changes it finishes.

Lex-2008:
FU at https://github.com/cfengine/buildscripts/pull/843

PR #4721, opened 2021-06-23
Enabled non-object bare values inside augments variables key

olehermanse:
When discussing this feature, we mentioned that the new
format was only a problem/incompatibility for objects.
If the value is something like a string (or list)
the old structure "just works". This also means that
in the future, when 3.18.0+ is common, we can tell users
to just rename "vars" to "variables". As long as they
don't have any objects they don't have to make further
changes. This can be added as a warning, even.

Looking at our examples it seems quite common that
augments variables are strings or lists, and quite
rare that they are objects (data containers), so for
most users, they can just rename "vars" to "variables".
(With this change).

Without this change, the code crashes (abort assertion
error) if you try to do this.

### DIFF
--- a/libpromises/cmdb.c
+++ b/libpromises/cmdb.c
@@ -308,6 +308,7 @@ static bool ReadCMDBVariables(EvalContext *ctx, JsonElement *variables)
             if (data == NULL)
             {
                 Log(LOG_LEVEL_ERR, "Missing value in '%s' variable specification in CMDB data (value field is required)", key);
+                VarRefDestroy(ref);
                 continue;
             }
 
--- a/libpromises/cmdb.c
+++ b/libpromises/cmdb.c
@@ -288,24 +288,41 @@ static bool ReadCMDBVariables(EvalContext *ctx, JsonElement *variables)
     while (JsonIteratorHasMore(&iter))
     {
         const char *key = JsonIteratorNextKey(&iter);
-        JsonElement *var_info = JsonObjectGet(variables, key);
+        JsonElement *const var_info = JsonObjectGet(variables, key);
 
-        JsonElement *json_tags = JsonObjectGet(var_info, CMDB_VARIABLES_TAGS);
-        JsonElement *data = JsonObjectGet(var_info, CMDB_VARIABLES_DATA);
+        JsonElement *data;
+        StringSet *tags;
+        const char *comment = NULL;
 
-        if (data == NULL)
+        if (JsonGetType(var_info) == JSON_TYPE_OBJECT)
         {
-            Log(LOG_LEVEL_ERR, "Missing value in '%s' variable specification in CMDB data", key);
-            continue;
+            data = JsonObjectGet(var_info, CMDB_VARIABLES_DATA);
+
+            if (data == NULL)
+            {
+                Log(LOG_LEVEL_ERR, "Missing value in '%s' variable specification in CMDB data (value field is required)", key);
+                continue;
+            }
+
+            JsonElement *json_tags = JsonObjectGet(var_info, CMDB_VARIABLES_TAGS);
+            tags = GetTagsFromJsonTags("variable", key, json_tags, CMDB_SOURCE_TAG);
+            comment = GetCMDBComment("variable", key, var_info);
         }
+        else
+        {
+            // Just a bare value, like in "vars", no metadata
+            data = var_info;
+            tags = GetTagsFromJsonTags("variable", key, NULL, CMDB_SOURCE_TAG);
+        }
+
+        assert(tags != NULL);
+        assert(data != NULL);
 
         VarRef *ref = GetCMDBVariableRef(key);
         if (ref == NULL)
         {
             continue;
         }
-        StringSet *tags = GetTagsFromJsonTags("variable", key, json_tags, CMDB_SOURCE_TAG);
-        const char *comment = GetCMDBComment("variable", key, var_info);
 
         bool ret = AddCMDBVariable(ctx, key, ref, data, tags, comment);
         VarRefDestroy(ref);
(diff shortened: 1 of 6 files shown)


olehermanse:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6803/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6803/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6803/

vpodzime (commented):
Please add/modify an acceptance test for this. Also, what about CMDB data? Do we want to allow the same there?

olehermanse:
@cf-bottom jenkins, please.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6807/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6807/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6807/

libpromises/cmdb.c
```diff
@@ -288,24 +288,41 @@ static bool ReadCMDBVariables(EvalContext *ctx, JsonElement *variables)
...
+            tags = GetTagsFromJsonTags("variable", key, NULL, CMDB_SOURCE_TAG);
+        }
+
+        assert(tags != NULL);
+        assert(data != NULL);
 
         VarRef *ref = GetCMDBVariableRef(key);
         if (ref == NULL)
         {
             continue;
```
vpodzime, line 324: This can now leak `tags`, I think.

basvandervlies:
I am just looking at this pull request.  We are making a lot of usage of the json files and use list, strings and datacontainers. So we use the `vars` a lot. Must that be renamed in the json files and cfengine files?

vpodzime (commented):
Thanks for the test! Check the potential leak, please. Looks good to me otherwise.

olehermanse:
> I am just looking at this pull request. We are making a lot of usage of the json files and use list, strings and datacontainers. So we use the vars a lot. Must that be renamed in the json files and cfengine files?

@basvandervlies No, `vars` and `variables` are separate keys, both will work. To use the new functionality (adding tags and comments from augments) you will have to start using `variables` instead of vars, and do the following transformation:

```json 
{
  "vars": {
    "string": "just works",
    "data": {"my": "data"}
  }
}
```

```json
{
  "variables": {
    "string": "just works",
    "data": {"value": {"my": "data"}, "comment": "New comment feature!"}
  }
}
```

libpromises/generic_agent.c
```diff
@@ -518,19 +518,36 @@ static bool LoadAugmentsData(EvalContext *ctx, const char *filename, const JsonE
...
+                    // Just a bare value, like in "vars", no metadata
+                    data = var_info;
+                    tags = GetTagsFromAugmentsTags("variable", vkey, NULL, "source=augments_file", filename);
                 }
 
-                const JsonElement *json_tags = JsonObjectGet(var_info, AUGMENTS_VARIABLES_TAGS);
-                StringSet *tags = GetTagsFromAugmentsTags("variable", vkey, json_tags, "source=augments_file", filename);
-                const char *comment = GetAugmentsComment("variable", vkey, filename, var_info);
+                assert(tags != NULL);
+                assert(data != NULL);
```
craigcomstock, line 550: So leaking tags here as well and it is interesting that the code is duplicated between the two. Maybe a refactor TODO another time? `// TODO` ?

basvandervlies:
> > I am just looking at this pull request. We are making a lot of usage of the json files and use list, strings and datacontainers. So we use the vars a lot. Must that be renamed in the json files and cfengine files?
> 
> @basvandervlies No, `vars` and `variables` are separate keys, both will work. To use the new functionality (adding tags and comments from augments) you will have to start using `variables` instead of vars, and do the following transformation:
> 
> ```json
> {
>   "vars": {
>     "string": "just works",
>     "data": {"my": "data"}
>   }
> }
> ```
> 
> ```json
> {
>   "variables": {
>     "string": "just works",
>     "data": {"value": {"my": "data"}, "comment": "New comment feature!"}
>   }
> }
> ```

Thanks for the explanation ;-). Makes a lot of sense and I can just gradually upgrade to the new feature. Thanks for the explanation.

libpromises/generic_agent.c
vpodzime: We already have a ticket for that.

olehermanse: No leak here, since there is no continue here, like in the other file. Also there is already a ticket to refactor this.

vpodzime: But I don't think there's a leak possible here.

vpodzime (approved):
Looks good to me now.

vpodzime (changes requested):
Actually, there's a potential leak of `ref` now.

vpodzime (approved):
ACK!

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6808/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6808/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6808/

libpromises/cmdb.c
olehermanse: Fixed, thanks!

olehermanse:
New acceptance tests passed, merging.

PR #4734, opened 2021-07-05
Testing stricter DoubleFromString when bad quantifier is found

KotzaBoss:


KotzaBoss:
@olehermanse In the unit test it says ignoring bad quantified for backwards compatibility. What is the history behind this decision?

olehermanse:
I am only guessing, but I think the behavior of some of these string parsing functions has been like that before, intentionally, for convenience when parsing protocols, or similar.

Let's say you have a ascii, line based protocol that looks something like this:

```
QUERY 1024s REBASE
SET 512\n
UPDATE 512,1024
MEASURE 5.12,5.15,5.20
```

(Just an example, not exactly our protocol).

Having a function which just parses the number, stopping at whatever doesn't fit is certainly convenient. Not saying it's "correct", but that would be my guess. As for maintaining backwards compatibility, that should be your default approach, it is always much safer. If you want to change behavior you should try to analyze exactly what it might break, and how. (Often that's not worth it or desirable).

PR #4735, opened 2021-07-05
WIP: Changed inform example to match default value and logging level

KotzaBoss:
Ticket: None
Changelog: None

### DIFF
--- a/examples/inform.cf
+++ b/examples/inform.cf
@@ -1,7 +1,7 @@
 # The inform constraint can be used to select which commands promises
-# generate output in INFO log level. INFO log level, is intendeted for
+# generate output in NOTICE log level. NOTICE log level, is intendeted for
 # events which alter the state of the system, such as editing files, users,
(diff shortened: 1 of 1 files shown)


KotzaBoss:
Blindly changin the INFO to NOTICE in the example comment may be incorrect.

olehermanse (changes requested):
Please expand the commit message here with what you are doing, and _why_.

KotzaBoss:
This PR is wrong. At first i was playing with an example using grep but the output threw me off and i thought that the `notice: ` output was the one being suppressed. 
While looking at the output of the different output when playing with the example in the [documentation](https://docs.cfengine.com/docs/3.18/reference-promise-types-commands.html#inform) its clear that my changes are nonsensical.

However the default value is not false but true. I will be opening a [PR](https://github.com/cfengine/documentation/pull/2476) in the docs providing my test results.

PR #4744, opened 2021-07-26
CFE-3718: Added example illustrating rxdirs perms body attribute

nickanderson:


examples/rxdirs.cf
```diff
@@ -0,0 +1,59 @@
+###############################################################################
+#+begin_src cfengine3
+bundle agent __main__
+# @brief Illustrating the behavior of rxdirs in perms bodies. Note, by default, when read is requested for a directory, execute is given as well.
+{
```
craigcomstock, line 5: Also note that by default if `rxdirs` attribute is not specified then it defaults to true. This is clear after reading the entire policy or maybe looking at the docs, but not as this lone example. Does this example replace the one here? 

https://docs.cfengine.com/docs/3.18/reference-promise-types-files.html#rxdirs

nickanderson: Yes, I intend to replace that useless snippet with this example.

craigcomstock: This is in the docs already so example is good to go. :+1:

craigcomstock (approved):

PR #4771, opened 2021-08-13
CFE-3628 Move bash and python CPT libraries and tests to masterfiles

Lex-2008:


vpodzime (commented):
Please move the ticket reference into a `Ticket: ` metadata part of each commit.

vpodzime (dismissed):
ACK

vpodzime (approved):
ACK

PR #4773, opened 2021-08-16
CFE-3723: Changed log level of message about reaching maximum recursion from ERROR to VERBOSE

nickanderson:
If I promise to copy a directory recursively up to a depth of N and there are
levels deeper than N the promise should not error. Not promising enough depth to
reach the bottom isn't an error, at most it's a message more relevant for
verbose logs, if not debug logs.

Ticket: CFE-3723
Changelog: Title

### DIFF
--- a/cf-agent/verify_files_utils.c
+++ b/cf-agent/verify_files_utils.c
@@ -736,8 +736,7 @@ static PromiseResult SourceSearchAndCopy(EvalContext *ctx, const char *from, cha
 
     if (maxrecurse == 0)        /* reached depth limit */
     {
-        RecordFailure(ctx, pp, attr, "Maximum recursion level reached at '%s'", from);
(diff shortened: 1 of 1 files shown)


cf-agent/verify_files_utils.c
```diff
@@ -736,8 +736,7 @@ static PromiseResult SourceSearchAndCopy(EvalContext *ctx, const char *from, cha
 
     if (maxrecurse == 0)        /* reached depth limit */
     {
-        RecordFailure(ctx, pp, attr, "Maximum recursion level reached at '%s'", from);
-        return PROMISE_RESULT_FAIL;
+        Log(LOG_LEVEL_VERBOSE, "Maximum recursion level reached at '%s'", from);
```
nickanderson, line 739: maybe I should keep the early return? But what to return?

PR #4774, opened 2021-08-17
ENT-7434: Use JSON_NOT_NULL()/NULL_JSON() to better handle augments/CMDB data

vpodzime:
JSON has a special value 'null' which represents and empty
value. We should thus treat it as such -- in most cases this
means the same as if the child object was not present. For
example with this JSON::

  "obj": {
    "child" : null
  }

the 'child' sub-object should be treated as a potential 'child2'
sub-object which is not present at all. That is, when handling
JSON/CMDB data.

Ticket: ENT-7434
Changelog: 'null' JSON value is now handled as empty data in augments/host-specific data

### DIFF
--- a/libpromises/cmdb.c
+++ b/libpromises/cmdb.c
@@ -214,7 +214,7 @@ static StringSet *GetTagsFromJsonTags(const char *item_type,
                                       const char *default_tag)
 {
     StringSet *tags = NULL;
-    if (json_tags != NULL)
+    if (JSON_NOT_NULL(json_tags))
     {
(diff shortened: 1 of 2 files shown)


larsewi (approved):
ACK

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/7055/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/7055/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-7055/

PR #4775, opened 2021-08-17
CFE-3733: Added example illustrating that the last key wins with mergedata()

nickanderson:
Ticket: CFE-3733
Changelog: None


craigcomstock (approved):

vpodzime (approved):
ACK

PR #4828, opened 2021-10-21
CFE-3806: Added test for trailing newline with insert_tree not removing path end tag

nickanderson:
Ticket: CFE-3806
Changelog: None


nickanderson:
> I believe this is good to go.

Can i get a :heavy_check_mark: ?

vpodzime (approved):
Sure. :heavy_check_mark:

PR #4869, opened 2021-11-24
Use StringEqual() in DefaultTemplateData()

vpodzime:
Otherwise static checkers complain that 'wantbundle' may be NULL.


olehermanse (approved):

larsewi (commented):
ACK

PR #4872, opened 2021-11-25
Static checks using newer versions of Fedora

vpodzime:


vpodzime:
@olehermanse @larsewi just in case it's not clear from the *Makefile.am* -- you can test this locally by doing `make static-check` and then for example `make static-check-f35`. `make static-check` keeps doing the same as it used to so this doesn't break jenkins (*static-checks* job runs on *oslo-dc*).

larsewi (approved):
Tested locally

PR #4875, opened 2021-11-30
GH Actions: Add CIFuzz integration

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.
This is a service offered by OSS-Fuzz, on which cfengine already runs (although there is only a single fuzzer for now)

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'cfengine'
+        dry-run: false
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'cfengine'
+        fuzz-seconds: 600
+        dry-run: false
+    - name: Upload Crash
+      uses: actions/upload-artifact@v1
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts


vpodzime:
> ** CIFuzz / Fuzzing (pull_request) ** Successful in 7m

Well, it seems to be triggering. But I see this in the details:
```
 ==21==ERROR: AddressSanitizer: stack-overflow on address 0x7fffd6877ff8 (pc 0x00000054c989 bp 0x7fc18ee75000 sp 0x7fffd6878000 T0)
SCARINESS: 10 (stack-overflow)
    <empty stack>

SUMMARY: AddressSanitizer: stack-overflow
==21==ABORTING
```
which brings two questions:
1. Why the status says *success* when it actually failed?
2. And how do I actually find the exact case that triggered the failure so that I can reproduce it locally and fix it?

Thanks for this!

DavidKorczynski:
> > ** CIFuzz / Fuzzing (pull_request) ** Successful in 7m
> 
> Well, it seems to be triggering. But I see this in the details:
> 
> ```
>  ==21==ERROR: AddressSanitizer: stack-overflow on address 0x7fffd6877ff8 (pc 0x00000054c989 bp 0x7fc18ee75000 sp 0x7fffd6878000 T0)
> SCARINESS: 10 (stack-overflow)
>     <empty stack>
> 
> SUMMARY: AddressSanitizer: stack-overflow
> ==21==ABORTING
> ```
> 
> which brings two questions:
> 
> 1. Why the status says _success_ when it actually failed?
> 2. And how do I actually find the exact case that triggered the failure so that I can reproduce it locally and fix it?
> 
> Thanks for this!

I was also under the impression the check would fail when a crash happens - @jonathanmetzman can you advice here?

jonathanmetzman:
> > > ** CIFuzz / Fuzzing (pull_request) ** Successful in 7m
> > 
> > 
> > Well, it seems to be triggering. But I see this in the details:
> > ```
> >  ==21==ERROR: AddressSanitizer: stack-overflow on address 0x7fffd6877ff8 (pc 0x00000054c989 bp 0x7fc18ee75000 sp 0x7fffd6878000 T0)
> > SCARINESS: 10 (stack-overflow)
> >     <empty stack>
> > 
> > SUMMARY: AddressSanitizer: stack-overflow
> > ==21==ABORTING
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which brings two questions:
> > 
> > 1. Why the status says _success_ when it actually failed?
> > 2. And how do I actually find the exact case that triggered the failure so that I can reproduce it locally and fix it?
> > 
> > Thanks for this!
> 
> I was also under the impression the check would fail when a crash happens - @jonathanmetzman can you advice here?

CIFuzz only reports novel crashes. Crashes that reproduce on a build that is already running on OSS-Fuzz are not reported (because they were not introduced by the PR).
We might be able to make this behavior configurable if you need, but i think it makes sense for most cases even if it is a bit confusing some times.

I can tell this is what happened, because of the log line:
```
The crash is reproducible on previous build. Code change (pr/commit) did not introduce crash.

DavidKorczynski:
Thanks a lot for the clarification @jonathanmetzman !

vpodzime:
Makes sense, thanks!

vpodzime:
I'm still interested in the answer to this question, though:
> And how do I actually find the exact case that triggered the failure so that I can reproduce it locally and fix it?

TBH, I don't even know where to find the reproducer for the crash that's already present.

DavidKorczynski:
Your email need to be in the `project.yaml` of cfengine at oss-fuzz, which it is: https://github.com/google/oss-fuzz/blob/master/projects/cfengine/project.yaml

Then, you need to visit "https://oss-fuzz.com" where the interface will direct you. You need to log on with your email from the `project.yaml` by way of Google auth so the email must be linked to a Google account.

Alterntatively, you will have received an email at some point with the subject "Issue XXXXX  in oss-fuzz: cfengine:XXXXX" which will have information on the issue as well

vpodzime (approved):
ACK

PR #4876, opened 2021-11-30
Change build badges from travis to CI

craigcomstock:
Travis build badges reflect PR status and we only want
official status of branches.

Ticket: none
Changelog: none


craigcomstock:
Sadly at this point the most recent nightly builds are all failing except 3.12.x :( So I suppose we could "fix" that also.

vpodzime (dismissed):
ACK

craigcomstock:
will submit a different PR for removing.

PR #4877, opened 2021-12-01
SELinux: Allow access to /opt/cfengine for commands over SSH

vpodzime:
Needed for Federated Reporting.

Ticket: ENT-7493
Changelog: Federated Reporting is now supported on RHEL 8 hubs


nickanderson (approved):
yes, FR communicates over ssh and the users home dir as well as data to be transported is inside opt.

olehermanse (approved):
Thank you for working on this!

PR #4882, opened 2021-12-08
Added bootstrap/failsafe exclusion for directories named .no-distrib

nickanderson:
Merge together:
- https://github.com/cfengine/masterfiles/pull/2179
- https://github.com/cfengine/documentation/pull/2597

With build.cfengine.com and cfbs released, we would like to be able to
distribute compliance reports as cfbs modules. We don't need (or necessarily
want) to distribute compliance report definitions to every host since they are
only useful for an enterprise hub, so we introduced this directory '.no-distrib'
as an excluded directory for bootstrap and failsafe.

With this change bootstrap/failsafe will exclude any directory named .no-distrib.
Note, until the update policy is updated with the same
exclusion, any policy update will receive the files in those directories.

Ticket: ENT-8079
Changelog: Title

### DIFF
--- a/libpromises/failsafe.cf
+++ b/libpromises/failsafe.cf
@@ -121,7 +121,7 @@ bundle agent failsafe_cfe_internal_update
       "$(sys.inputdir)"
         handle => "failsafe_cfe_internal_bootstrap_update_files_sys_workdir_inputs_shortcut",
         copy_from => failsafe_scp("$(masterfiles_dir_remote)"),
-        depth_search => failsafe_recurse("inf"),
+        depth_search => failsafe_u_infinite_client_policy,
         file_select => failsafe_exclude_vcs_files,
         classes => failsafe_results("namespace", "inputdir_update");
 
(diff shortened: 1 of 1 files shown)


cf-bottom:

libpromises/failsafe.cf
```diff
@@ -121,7 +121,7 @@ bundle agent failsafe_cfe_internal_update
       "$(sys.inputdir)"
         handle => "failsafe_cfe_internal_bootstrap_update_files_sys_workdir_inputs_shortcut",
         copy_from => failsafe_scp("$(masterfiles_dir_remote)"),
-        depth_search => failsafe_recurse("inf"),
+        depth_search => failsafe_u_infinate_client_policy,
```
olehermanse, line 124:
```suggestion
        depth_search => failsafe_u_infinite_client_policy,
```

libpromises/failsafe.cf
```diff
@@ -420,6 +420,15 @@ body copy_from failsafe_scp(from)
       portnumber => "$(sys.policy_hub_port)";
 }
 ############################################
+body depth_search failsafe_u_infinate_client_policy
```
olehermanse, line 423:
```suggestion
body depth_search failsafe_u_infinite_client_policy
```

olehermanse (changes requested):

libpromises/failsafe.cf
nickanderson: ack

libpromises/failsafe.cf
nickanderson: ack

PR #4890, opened 2021-12-27
Replace backticks with single quotes in changelogs

Lex-2008:
Issue was that backticks in changelogs break docs builds, and Nick had
to fix it manually. Quote from cfengine#4863:

> The documentation pulls in changelogs and having markdown inside these files
> breaks when the doc build is unable to automatically resolve links.

Ticket: ENT-7979

Changelog: None

### DIFF
--- a/misc/changelog-generator/changelog-generator
+++ b/misc/changelog-generator/changelog-generator
@@ -127,7 +127,8 @@ for repo in repos:
         log_entry_local = False
         log_entry = ""
         for line in blob.stdout:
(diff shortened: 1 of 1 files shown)


vpodzime (dismissed):
Would be nice to add an explanation to the commit so that one doesn't have to go to the ticket and then to Nick's PR to find why backticks are problematic.

Lex-2008:
good idea! explanation added to the commit

vpodzime (approved):
ACK

PR #4895, opened 2022-01-07
Only replace the dot after the scope in MangleScopedVarNameIntoSpecia…

vpodzime:
…lScopeName()

In very rare cases where a scoped variable which has the same
string as "scope." inside square brackets (classic array index)
needs to be put into a special scope,
MangleScopedVarNameIntoSpecialScopeName() has to be more careful
and only replace the dot after the scope, not in the same string
inside the square brackets.

Ticket: ENT-8289
Changelog: None

PR #4896, opened 2022-01-10
[3.18.x] ENT-8289: Only replace the dot after the scope in MangleScopedVarNameIntoSpecia…

vpodzime:
…lScopeName()

In very rare cases where a scoped variable which has the same
string as "scope." inside square brackets (classic array index)
needs to be put into a special scope,
MangleScopedVarNameIntoSpecialScopeName() has to be more careful
and only replace the dot after the scope, not in the same string
inside the square brackets.

Ticket: ENT-8289
Changelog: None
(cherry picked from commit d4e7a40e2006714251b16648c2aba0e095a59057)

Conflicts:
  libntech


vpodzime:
3.18.x version of #4895

PR #4907, opened 2022-01-21
Fix the log message for remote copies

vpodzime:
This, for example, is clearly wrong:

  info: Copied file 'hub_cmdb/host_specific.json' from '/var/cfengine/data/./host_specific.json.cfnew' to '63.33.55.11'

should say "... from '63.33.55.11' to '/var/cfengine...'".


larsewi (approved):
ACK

olehermanse (approved):

PR #4912, opened 2022-01-26
Update libntech to current master

vpodzime:
So that we make the free_array_items() function available.

Ticket: ENT-8307
Changelog: None


larsewi (approved):
ACK

PR #4920, opened 2022-02-10
CFE-3906: Fix wrong use of log level in users promises log messages (3.18.x)

larsewi:
Log messages at log level info should state what the actual changes on
the system are, not the promise outcome. The promise outcome should be
logged at log level verbose.

Ticket: CFE-3906
Changelog: Title
Signed-off-by: larsewi <lars.erik.wik@northern.tech>
(cherry picked from commit 6b3b7f48f28e4fd49bf91c2fde5c86bba5b0f9a3)


olehermanse (approved):
This is not good enough, but I feel bad for you.

PR #4923, opened 2022-02-18
Fixed segfaults on Solaris 10 caused by NULL as argument in string format functions

larsewi:
Fixed segfaults on Solaris 10 caused by NULL as arguments in different format functions.

E.g. on Solaris 10 the following statement segfaults
```C
printf("%s\n", NULL);
```
while on other platforms it would output `(null)`

### DIFF
--- a/cf-agent/verify_files.c
+++ b/cf-agent/verify_files.c
@@ -873,7 +873,14 @@ static PromiseResult RenderTemplateMustache(EvalContext *ctx,
     }
     else
     {
-        RecordFailure(ctx, pp, attr, "Error rendering mustache template '%s'", attr->edit_template);
+        /* Use `edit_template` attribute when template method is "mustache".
+         * Use `edit_template_string` attribute when template method is
+         * "inline_mustache".
+         */
+        char *tmpl = StringEqual(attr->template_method, "mustache")
+                   ? attr->edit_template : attr->edit_template_string;
+        RecordFailure(ctx, pp, attr, "Error rendering mustache template '%s'",
+                      tmpl);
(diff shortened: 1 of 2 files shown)


vpodzime (approved):
ACK, thanks!

libpromises/mod_custom.c
```diff
@@ -1111,8 +1111,10 @@ PromiseResult EvaluateCustomPromise(EvalContext *ctx, const Promise *pp)
     /* Used below, constructed here while path and interpreter are definitely
      * valid pointers. */
     char custom_promise_id[CF_BUFSIZE];
-    NDEBUG_UNUSED size_t ret = snprintf(custom_promise_id, sizeof(custom_promise_id),
-                                        "%s-%s-%s", pp->promiser, path, interpreter);
+    NDEBUG_UNUSED size_t ret = snprintf(custom_promise_id,
```
olehermanse, line 1114: @larsewi `custrom` in commit message.

cf-agent/verify_files.c
```diff
@@ -873,7 +873,10 @@ static PromiseResult RenderTemplateMustache(EvalContext *ctx,
     }
     else
     {
-        RecordFailure(ctx, pp, attr, "Error rendering mustache template '%s'", attr->edit_template);
+        char *tmpl = StringEqual(attr->template_method, "mustache")
+                   ? attr->edit_template : attr->edit_template_string;
```
olehermanse, line 877: I would add a comment about what is for `inline_mustache`

olehermanse (approved):

larsewi:
@olehermanse, I added the comment and fixed the typo in commit message that you addressed.

olehermanse (approved):

PR #4937, opened 2022-03-23
Add more details to the warning in PromiseGetConstraintAsBooleanWithDefault()

vpodzime:
When a default value for an attribute that should be set
explicitly is used, a warning is emitted. It should be easier for
the user to identify which promise exactly it is coming from.

Ticket: CFE-951
Changelog: None


nickanderson (approved):
Thanks!

PR #4939, opened 2022-03-24
DEBUG: try to assert that sscanf("0", "%ju", &tmp) works on HP-UX

vpodzime:


vpodzime:
There we go: `Assertion failed: ret == 1, file conversion.c, line 1109`. Alright, HPUX, `int` it is for you. :stuck_out_tongue:

vpodzime:
Fix: https://github.com/cfengine/core/pull/4940

PR #4949, opened 2022-04-14
CFE-3962: Aligned format of verbose BEGIN and END promise logs

nickanderson:
A naive attempt ...

This change aligns the format of the BEGIN and END verbose promise log messages
making it easier to identify the logs that are related to each promise.

Ticket: CFE-3962
Changelog: Title


nickanderson:
bah, no surprise it didn't build

PR #4957, opened 2022-05-24
Fixes: CFE-3982 do not kill unrelated process

peckpeck:
Pull request to fix https://tracker.mender.io/browse/CFE-3982

### DIFF
--- a/libpromises/locks.c
+++ b/libpromises/locks.c
@@ -555,29 +555,30 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
 }
 
 // returns true if pid points to a cfengine process
-static bool IsCfengineProcess(const int pid)
+static bool IsCfengineProcess(pid_t pid)
 {
-    // There is no /proc on windows
-#ifndef __linux__
     char procfile[PATH_MAX];
     char cmd[PATH_MAX]; // we don't need the full ARG_MAX
     snprintf(procfile, PATH_MAX, "/proc/%d/cmdline", pid);
     FILE *f = fopen(procfile, "r");
+    // this will just return null on platform where /proc doesn't exist and return true as before
     if (f != NULL){
         size_t size;
-        size = fread(cmd, sizeof(char), PATH_MAX, f);
-        if (size > 0){
+        size = FullRead(f, cmd, PATH_MAX);
+        if (size > 0)
+        {
             if ('\n' == cmd[size-1]) {
                 cmd[size-1] = '\0';
             }
         }
         fclose(f);
-        // cmd contains the process command path
-        char *name = basename(cmd);
-        return (strncmp("cf-", name, 3) == 0);
+        if (size > 0) {
+            // cmd contains the process command path
+            char *name = basename(cmd);
+            return StringStartsWith(name, "cf-");
+        }
     }
-#endif
-    // assume true if we don't know
+    // assume true where we cannot check
     return true;
 }
 
@@ -617,6 +618,7 @@ static bool KillLockHolder(const char *lock)
     if (!IsCfengineProcess(lock_data.pid)) {
         Log(LOG_LEVEL_VERBOSE,
             "Lock holder with pid %d was replaced by a non cfengine process, do not try to kill it!, ", lock_data.pid);
+        // if the process doesn't exist anymore, then the original lock is not there, avoid producing an error
         return true;
     }
 
--- a/libpromises/locks.c
+++ b/libpromises/locks.c
@@ -554,6 +554,34 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
     }
 }
 
+// returns true if pid points to a cfengine process
+static bool IsCfengineProcess(const int pid)
+{
+    // There is no /proc on windows
+#ifndef _WIN32
+    char* procfile = (char*)xcalloc(1024,sizeof(char));
+    char* cmd = (char*)xcalloc(1024,sizeof(char));
+    if(cmd && procfile){
+        sprintf(procfile, "/proc/%d/cmdline",pid);
+        FILE* f = fopen(procfile,"r");
+        if(f){
+            size_t size;
+            size = fread(cmd, sizeof(char), 1024, f);
+            if(size>0){
+                if('\n'==cmd[size-1])
+                    cmd[size-1]='\0';
+            }
+            fclose(f);
+            // cmd contains the process command path
+            char *name = basename(cmd);
+            return (strncmp("cf-", name, 3) == 0);
+        }
+    }
+#endif
+    // assume true if we don't know
+    return true;
+}
+
 static bool KillLockHolder(const char *lock)
 {
     bool ret;
@@ -586,6 +614,12 @@ static bool KillLockHolder(const char *lock)
     }
 
     CloseLock(dbp);
+
+    if (!IsCfengineProcess(lock_data.pid)) {
+        Log(LOG_LEVEL_INFO,
+            "Lock holder disappeared");
+        return true;
+    }
 
     if (GracefulTerminate(lock_data.pid, lock_data.process_start_time))
     {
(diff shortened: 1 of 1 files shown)


peckpeck:
About the coding style, is there anything else than the function name that you want to point me to ?

olehermanse:
> About the coding style, is there anything else than the function name that you want to point me to ?

@peckpeck Some things I see at first glance:

1. Space after if, commas, and around operators (like `>`).
2. Opening curly brace on separate line (`{`).
3. Don't do `if`'s without curly braces.
4. Avoid unnecessary typecasts.
5. Pointer asterisk to the right, next to the name (`FILE *f`).
6. Compare pointers to `NULL` explicitly (`if (f != NULL)`).

libpromises/locks.c
```diff
@@ -554,6 +554,35 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
...
 
+// returns true if pid points to a cfengine process
+static bool IsCfengineProcess(const int pid)
+{
+    // There is no /proc on windows
+#ifndef _WIN32
+    char* procfile = xcalloc(1024, sizeof(char));
+    char* cmd = xcalloc(1024, sizeof(char));
+    if (cmd && procfile){
+        sprintf(procfile, "/proc/%d/cmdline", pid);
```
vpodzime, line 565: This emits an error during compilation:
```
locks.c:565:9: error: 'sprintf' is deprecated

      [-Werror,-Wdeprecated-declarations]

        sprintf(procfile, "/proc/%d/cmdline", pid);

        ^

./../libntech/libutils/deprecated.h:42:5: note: 'sprintf' has been explicitly

      marked deprecated here

    FUNC_DEPRECATED("Try snprintf() or xsnprintf() or xasprintf()");

    ^

./../libntech/libutils/compiler.h:44:49: note: expanded from macro

      'FUNC_DEPRECATED'

#    define FUNC_DEPRECATED(msg) __attribute__((deprecated))

                                                ^
```

libpromises/locks.c
```diff
@@ -554,6 +554,35 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
     }
 }
 
+// returns true if pid points to a cfengine process
+static bool IsCfengineProcess(const int pid)
+{
+    // There is no /proc on windows
+#ifndef _WIN32
+    char* procfile = xcalloc(1024, sizeof(char));
+    char* cmd = xcalloc(1024, sizeof(char));
```
vpodzime, line 563: Either `free()` these variables or, better, use arrays like `char procfile[PATH_MAX]` and similar.

libpromises/locks.c
```diff
@@ -554,6 +554,35 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
...
+    if (cmd && procfile){
```
vpodzime, line 564: No need to check these, `xcalloc()` kills the process in case of `ENOMEM`.

libpromises/locks.c
```diff
@@ -587,6 +616,12 @@ static bool KillLockHolder(const char *lock)
 
     CloseLock(dbp);
 
+    if (!IsCfengineProcess(lock_data.pid)) {
+        Log(LOG_LEVEL_INFO,
+            "Lock holder disappeared");
```
vpodzime, line 621: This log message should be a bit more verbose and logged as `VERBOSE`.

vpodzime (commented):
A bit more work needed.

peckpeck:
sorry for the delay

larsewi:
Hi @peckpeck ! as far as i know, Windows is not the only platform without the `/proc` file. Please correct me if I'm wrong. In `libpromises/process_select.h` we have some functions for querying processes. Maybe you can use those? E.g.
```
ClearProcessTable(); /* Make sure old process table is not cached */

Item *procs = NULL;
if (LoadProcessTable())
{
    ProcessSelect ps = PROCESS_SELECT_INIT;
    ps.max_pid = ps.min_pid = lock_data.pid;
    procs = SelectProcesses("cf-.*", &ps, true);
}
if (procs == NULL)
{
    /* Process select returned nothing, it must be an unrelated process */
    return false;
}
DeleteItemList(procs);
return true;
```

Not sure if this would work, but feel free to give it a try. Otherwise you could maybe guard it with the `#ifdef __linux__` macro.

libpromises/locks.c
```diff
@@ -554,6 +554,33 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
     }
 }
 
+// returns true if pid points to a cfengine process
+static bool IsCfengineProcess(const int pid)
```
vpodzime, line 558: `const` has no meaning here, and the type should be `pid_t` not `int.

libpromises/locks.c
```diff
@@ -554,6 +554,33 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
...
+{
+    // There is no /proc on windows
+#ifndef __linux__
+    char procfile[PATH_MAX];
+    char cmd[PATH_MAX]; // we don't need the full ARG_MAX
+    snprintf(procfile, PATH_MAX, "/proc/%d/cmdline", pid);
+    FILE *f = fopen(procfile, "r");
+    if (f != NULL){
+        size_t size;
+        size = fread(cmd, sizeof(char), PATH_MAX, f);
```
vpodzime, line 568: Please check `FullRead()` in *libntech/libutils*

libpromises/locks.c
```diff
@@ -554,6 +554,33 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
...
+        if (size > 0){
+            if ('\n' == cmd[size-1]) {
+                cmd[size-1] = '\0';
+            }
+        }
+        fclose(f);
+        // cmd contains the process command path
+        char *name = basename(cmd);
+        return (strncmp("cf-", name, 3) == 0);
```
vpodzime, line 577: Please check `StringStartsWith()` in *libntech/libutils*

libpromises/locks.c
```diff
@@ -554,6 +554,33 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
...
+        size_t size;
+        size = fread(cmd, sizeof(char), PATH_MAX, f);
+        if (size > 0){
+            if ('\n' == cmd[size-1]) {
+                cmd[size-1] = '\0';
+            }
+        }
+        fclose(f);
+        // cmd contains the process command path
+        char *name = basename(cmd);
```
vpodzime, line 576: If `size` is `0` then `cmd` is uninitialized here.

libpromises/locks.c
```diff
@@ -554,6 +554,33 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
...
+        return (strncmp("cf-", name, 3) == 0);
+    }
+#endif
+    // assume true if we don't know
+    return true;
```
vpodzime, line 581: Wouldn't it be safer to assume the opposite?

libpromises/locks.c
```diff
@@ -587,6 +614,12 @@ static bool KillLockHolder(const char *lock)
 
     CloseLock(dbp);
 
+    if (!IsCfengineProcess(lock_data.pid)) {
+        Log(LOG_LEVEL_VERBOSE,
+            "Lock holder with pid %d was replaced by a non cfengine process, do not try to kill it!, ", lock_data.pid);
+        return true;
```
vpodzime, line 620: So the function returns `true` here in this case, but it doesn't really kill the lock holder. That can be a source of issues.

libpromises/locks.c
peckpeck: we don't know means use the old behaviour on other platforms

libpromises/locks.c
```diff
@@ -554,6 +554,34 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
     }
 }
 
+// returns true if pid points to a cfengine process
+static bool IsCfengineProcess(pid_t pid)
+{
+    char procfile[PATH_MAX];
+    char cmd[PATH_MAX]; // we don't need the full ARG_MAX
+    snprintf(procfile, PATH_MAX, "/proc/%d/cmdline", pid);
```
vpodzime, line 562: `pid_t` doesn't have to be `int` and this can easily fail to compile in such a case because of the format (`%d`) and type (`pid_t`) mismatch. We usually use a trick with `uintmax_t` for things like this.

libpromises/locks.c
```diff
@@ -554,6 +554,34 @@ static void PromiseTypeString(char *dst, size_t dst_size, const Promise *pp)
...
+    FILE *f = fopen(procfile, "r");
+    // this will just return null on platform where /proc doesn't exist and return true as before
+    if (f != NULL){
+        size_t size;
+        size = FullRead(f, cmd, PATH_MAX);
```
vpodzime, line 567: `size_t size = FullRead(f, cmd, PATH_MAX);`

libpromises/locks.c
vpodzime: This is still applicable.

vpodzime:
@peckpeck I took your code as a base for [my work on the related ticket](https://github.com/cfengine/core/pull/5362). Thank you for your contribution!

PR #4962, opened 2022-06-02
CFE-3987 Added tests related to inserting text based on a location within select_region

nickanderson:
This change adds two tests, the first (acceptance/31_tickets/CFE-3987/1/test.cf)
tests the behavior described by the related issue, inserting text before a
selected region should insert immediately before the selected region even if
select_line_matching is not explicitly specified. This test is currently set to soft fail.

The second (acceptance/31_tickets/CFE-3987/2/test.cf) tests the workaround where
after select_line_matching is explicitly defined, the text is inserted
immediately before the selected region. This test passes (as expected).


craigcomstock (approved):
I see the tests and agree! 🎉

vpodzime (commented):
ACK

PR #4963, opened 2022-06-02
CFE-3988: Added test for editing after a selected region

nickanderson:
This change adds a test described by the related ticket for being able to edit
after a selected region.


craigcomstock (approved):

vpodzime (approved):
ACK

PR #4968, opened 2022-06-09
ENT-8818: Added test variables referencing non-existing namespaces do not crash

nickanderson:
This test exercises a bug that causes the agent to crash with an error like:

cf-agent: eval_context.c:2233: GetVariableTableForScope: Assertion `!ns || strcmp("default", ns) == 0' failed.


vpodzime (approved):
ACK, this shouldn't be an assertion. Or, actually, the assertion should stay, but there should be a check before it.

craigcomstock (approved):

PR #4977, opened 2022-06-15
CFE-2367: Added tests that fileexists does not error fatally when given an undefined var parameter

nickanderson:
Adding tests for CFE-2367 posthumously since I found this is no longer happening but the ticket was never closed.

### DIFF
--- a/tests/acceptance/31_tickets/CFE-2367/1/test.cf
+++ b/tests/acceptance/31_tickets/CFE-2367/1/test.cf
@@ -1,0 +1,34 @@
+body file control
+{
+        inputs => {
+                    "../../../default.cf.sub",
+        };
+}
+bundle agent __main__
+# If this is the policy entry (cf-agent --file) then this bundle will be run by default.
+{
+  methods:
+        "bundlesequence"  usebundle => default("$(this.promise_filename)");
+}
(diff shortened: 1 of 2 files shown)


craigcomstock (commented):
cool. do you want to keep the `__main__` bundles in there? Is that an artifact of you running the files directly instead of with the `testall` test runner?

nickanderson:
> cool. do you want to keep the `__main__` bundles in there? Is that an artifact of you running the files directly instead of with the `testall` test runner?

I don't think it matters honestly. I kind of like it because it's very clear to me what happens when I run that file.

testall runs each test directly, so `__main__` takes effect if it's present. In many tests we see body common control with budnelsequence set to default()
https://github.com/cfengine/core/blob/master/tests/acceptance/dcs.cf.sub#L210

So, this does basically the same thing. Except for that it COULD be included in a larger policy set. Which is not something I often do (take test from here, plop into some other policy set and run). But, in the reverse it's many times part of my development process. Wriring policy on some hub as part of a larger set, and then tweaking and masaging it into form for our tests. Using __main__ simply short circuits the work of making body common control since i instead use body file control and bundle __main__

craigcomstock (approved):

PR #4984, opened 2022-06-22
Fixes for the cfengine_action_script_t domain

vpodzime:
It should be entered by a transition from the cfengine_httpd_t
domain not from init_t. And it should be allowed to execute
things from /bin and /usr/bin.

### DIFF
--- a/misc/selinux/cfengine-enterprise.te
+++ b/misc/selinux/cfengine-enterprise.te
@@ -761,7 +761,7 @@ typeattribute cfengine_action_script_exec_t exec_type;
 typeattribute cfengine_action_script_exec_t file_type, non_security_file_type, non_auth_file_type;
 role object_r types cfengine_action_script_exec_t;
 
-type_transition init_t cfengine_action_script_exec_t:process cfengine_action_script_t;
+type_transition cfengine_httpd_t cfengine_action_script_exec_t:process cfengine_action_script_t;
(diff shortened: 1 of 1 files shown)


misc/selinux/cfengine-enterprise.te
```diff
@@ -761,7 +761,7 @@ typeattribute cfengine_action_script_exec_t exec_type;
 typeattribute cfengine_action_script_exec_t file_type, non_security_file_type, non_auth_file_type;
 role object_r types cfengine_action_script_exec_t;
 
-type_transition init_t cfengine_action_script_exec_t:process cfengine_action_script_t;
+type_transition cfengine_httpd_t cfengine_action_script_exec_t:process cfengine_action_script_t;
```
nickanderson, line 764: Just [reading docs](https://selinuxproject.org/page/TypeRules) while I read the diff and making translations.

When a process of type `cfengine_httpd_t` executes a file of type `cfengine_action_script_exec_t` the process type should be changed to `cfengine_action_script_t`.

Makes sense to me. Currently action scripts are triggered by Apache when the conditions are observed.

Wouldn't it make sense to move this to `cf-reactor`?

nickanderson (approved):

misc/selinux/cfengine-enterprise.te
vpodzime: > Wouldn't it make sense to move this to `cf-reactor`?

It surely would!

PR #4987, opened 2022-06-27
Load the legacy provider on OpenSSL 3+

vpodzime:
The legacy encrypted network stuff on community uses Blowfish
which is only provided by the legacy provider in OpenSSL 3+.


olehermanse:
Related to dependency updates PR: https://github.com/cfengine/buildscripts/pull/1075

olehermanse (approved):

PR #5004, opened 2022-07-29
Added examples that illustrate various naming conventions

nickanderson:
Intended for use within the Policy Style Guide

### DIFF
--- a/examples/style_hungarian.cf
+++ b/examples/style_hungarian.cf
@@ -1,0 +1,52 @@
+#+begin_src cfengine3
+bundle agent __main__
+{
+  vars:
+      "s_one"   string => "one";
+      "ITwo"    int => "2";
+      "rThree"  real => "3.0";
+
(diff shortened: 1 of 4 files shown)


vpodzime (approved):
ACK

examples/style_hungarian.cf
```diff
@@ -0,0 +1,52 @@
+#+begin_src cfengine3
+bundle agent __main__
+{
+  vars:
+      "s_one"   string => "one";
+      "ITwo"    int => "2";
+      "rThree"  real => "3.0";
+
+      "lMyList" slist => { "$(s_one)", "$(ITwo)", "$(rThree)" };
```
olehermanse, line 9: I think you should not mix the 3 different styles like this. It seems like you are either encouraging something which users should not do, or your example is noisy (this camelcase vs snakecase vs pascalcase is noise when trying to learn what hungarian notation means).

nickanderson: Feel free to change it.

PR #5010, opened 2022-08-02
ENT-9037: Added cf-support utility for generating support information

craigcomstock:
This script gathers various system information as well
as log files and creates a tarball.

If the host has enterprise CFEngine installed further
details and files will be collected.

Ticket: ENT-9037
Changelog: title

merge together:
https://github.com/cfengine/core/pull/5010
https://github.com/cfengine/nova/pull/1918
https://github.com/cfengine/buildscripts/pull/1106
https://github.com/cfengine/system-testing/pull/445

### DIFF
--- a/misc/Makefile.am
+++ b/misc/Makefile.am
@@ -35,3 +35,4 @@ install-data-hook:
 	-systemctl daemon-reload
 
 endif
+dist_bin_SCRIPTS = cf-support
--- a/misc/cf-support
+++ b/misc/cf-support
@@ -1,0 +1,190 @@
+#!/usr/bin/env bash
+
+export WORKDIR=/var/cfengine
+export BINDIR="$WORKDIR/bin"
+export non_interactive=0
+
+if [ $# -gt 1 ]; then
+  echo "usage: $(basename "$0") [--yes]"
+  echo '	--yes - Non-interactive use. Assume no ticket number and assume include masterfiles.'
+  exit 1
+fi
+
+if [ "$1" = "-M" ]; then
+  cat <<EOF
+.\"Copyright 2022 Northern.tech AS
+.\"
+.\"This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+.\"
+.\"This program is free software; you can redistribute it and/or modify it
+.\"under the terms of the GNU General Public License as published by the
+.\"Free Software Foundation; version 3.
+.\"
+.\"This program is distributed in the hope that it will be useful,
+.\"but WITHOUT ANY WARRANTY; without even the implied warranty of
+.\"MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+.\"GNU General Public License for more details.
+.\"
+.\"You should have received a copy of the GNU General Public License
+.\"along with this program; if not, write to the Free Software
+.\"Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+.\"
+.\"To the extent this program is licensed as part of the Enterprise
+.\"versions of CFEngine, the applicable Commercial Open Source License
+.\"(COSL) may apply to this file if you as a licensee so wish it. See
+.\"included file COSL.txt.
+e COSL.txt.
+.TH CF-SUPPORT 8 "2022-08-10" "CFEngine" "System Administrator"
+.SH NAME
+cf-support \- create tarball of information to submit for support tickets.
+.SH SYNOPSIS
+.B cf-support
+.RI [ OPTION ]
+.SH DESCRIPTION
+cf-support gathers various details about the system and creates a tarball in the current directory to submit to support. If the system is an enterprise hub then additional details will be gathered and included. The utility will prompt for an optional support ticket number as well as prompt whether to include masterfiles in the tarball.
+
+.SH OPTIONS
+.IP "--yes"
+Non-interactive use. Assume no ticket number and assume include masterfiles.
+.SH CFENGINE
+CFEngine provides automated configuration management of large-scale computer systems. A system administrator describes the desired state of a system using CFEngine policy code. The program \fBcf-agent\fR reads policy code and attempts to bring the current system state to the desired state described. Policy code is downloaded by \fBcf-agent\fR from a \fBcf-serverd\fR daemon. The daemon \fBcf-execd\fR is responsible for running \fBcf-agent\fR periodically.
+.br
+Documentation for CFEngine is available at https://docs.cfengine.com/.
+.SH PROMISE THEORY
+CFEngine is built on principles from promise theory, proposed by Mark Burgess in 2004. Promise theory is a model of voluntary cooperation between individual, autonomous actors or agents who publish their intentions to one another in the form of promises. A promise is a declaration of intent whose purpose is to increase the recipient's certainty about a claim of past, present or future behaviour. For a promise to increase certainty, the recipient needs to trust the promiser, but trust can also be built on the verification that previous promises have been kept, thus trust plays a symbiotic relationship with promises.
+.br
+For an introduction to promise theory, please see http://arxiv.org/abs/0810.3294/
+.SH AVAILABILITY
+cf-support is part of CFEngine.
+.br
+Binary packages may be downloaded from https://cfengine.com/download/.
+.br
+The source code is available at https://github.com/cfengine/
+.SH BUGS
+Please see the public bug-tracker at https://tracker.mender.io/projects/CFE/.
+.br
+GitHub pull-requests may be submitted to https://github.com/cfengine/core/.
+.SH "SEE ALSO"
+.BR cf-promises (8),
+.BR cf-agent (8),
+.BR cf-serverd (8),
+.BR cf-execd (8),
+.BR cf-monitord (8),
+.BR cf-runagent (8),
+.BR cf-key (8)
+.SH AUTHOR
+Northern.tech AS
+EOF
+  exit 0
+fi
+
+[ "$1" = "--yes" ] || [ "$1" = "-y" ] && non_interactive=1
+
+if [ "$EUID" -ne 0 ]; then
+  echo "$0 must be run as root"
+  exit 1
+fi
+
+if ! command -v gzip >/dev/null; then
+  echo "Please install gzip. This is required in order to minimize size of support collection."
+  exit 1
+fi
+
+if [ $non_interactive -eq 0 ]; then
+  read -rp "If you have it, please enter your support case number: " case_number
+fi
+case_number="${case_number:-NA}"
+timestamp=$(date +%F-%H%M)
+collection="cfengine_support_case_$case_number-$(hostname -f)-$timestamp"
+
+tmpdir="$(mktemp -d)/$collection"
+export tmpdir
+mkdir -p "$tmpdir"
+
+export info_file="$tmpdir/system-info.txt"
+
+function file_add
+{
+  local filename
+  filename="$(basename "$1")"
+  if [ -f "$1" ]; then
+    cp "$1" "$tmpdir"/"$filename"
+    echo "Added file $1"
+  else
+    echo "$1 file not found" >> "$info_file"
+  fi
+}
+export -f file_add
+
+function gzip_add
+{
+  local filename
+  filename="$(basename "$1")"
+  if [ -f "$1" ]; then
+    gzip -c "$1" > "$tmpdir"/"$filename".gz
+    echo "Added compressed copy of file $1"
+  else
+    echo "$1 file not found" >> "$info_file"
+  fi
+}
+export -f gzip_add
+
+function log_cmd
+{
+  echo "=== $1 ===" >> "$info_file"
+  eval "$1" >> "$info_file" 2>&1 || true
+  echo >> "$info_file" # one newline for easier to read output
+  echo "Captured output of command $1"
+}
+export -f log_cmd
+
+log_cmd "hostname -f"
+log_cmd "uname -a"
+[ -f "/etc/os-release" ] && log_cmd "cat /etc/os-release"
+log_cmd "$BINDIR/cf-promises -V"
+log_cmd "free -m"
+log_cmd "df -h"
+log_cmd "ps auwwx"
+log_cmd "top -b -H -c -n1"
+log_cmd "netstat -ie"
+log_cmd "ifconfig"
+if command -v lsof 2>/dev/null >/dev/null; then
+  lsof > "$tmpdir/lsof.txt"
+  echo "Captured output of command lsof"
+fi
+log_cmd "mount"
+file_add "/etc/fstab"
+log_cmd "$BINDIR/cf-key -p $WORKDIR/ppkeys/localhost.pub"
+log_cmd "grep 'version =' $WORKDIR/inputs/promises.cf"
+log_cmd "$BINDIR/cf-key -s"
+log_cmd "$BINDIR/cf-check diagnose"
+if command -v systemctl >/dev/null; then
+  log_cmd "systemctl status cfengine3"
+else
+  log_cmd "/etc/init.d/cfengine3 status"
+fi
+for f in /var/log/CFEngine-Install*; do
+  gzip_add "$f"
+done
+[ -f /var/log/messages ] && syslog_cmd="cat /var/log/messages"
+[ -f /var/log/syslog ] && syslog_cmd="cat /var/log/syslog"
+[ "$(command -v journalctl >/dev/null)" ] && syslog_cmd="journalctl"
+[ -z "$syslog_cmd" ] && syslog_cmd="dmesg"
+_syslog_filtered="$tmpdir"/syslog-filtered-for-cfengine.log.gz
+$syslog_cmd | grep -E 'cf-|CFEngine' | gzip -c > "$_syslog_filtered" || true
+echo "Captured output of $syslog_cmd filtered for cf-|CFEngine"
+
+gzip_add $WORKDIR/outputs/previous
+file_add $WORKDIR/policy_server.dat
+
+if [ -f $WORKDIR/share/cf-support-nova-hub.sh ]; then
+  # shellcheck source=/dev/null
+  source $WORKDIR/share/cf-support-nova-hub.sh
+fi
+
+# Here we create the tarball one directory up
+# to preserve a top-level of $collection in the tarball.
+# This gives a nice context of timestamp, hostname and support ticket
+# if provided. (see $collection definition above)
+tar czf "$collection.tar.gz" --directory "$tmpdir/../" . && rm -rf "$tmpdir"
+echo "Please send $collection.tar.gz to CFEngine support staff."


cf-bottom:

misc/Makefile.am
```diff
@@ -35,3 +35,4 @@ install-data-hook:
 	-systemctl daemon-reload
 
 endif
+bin_SCRIPTS = cf-support
```
craigcomstock, line 38: `bin_SCRIPTS` seems to work on ubuntu in a local VM but in CI seems to fail, maybe use `dist_bin_SCRIPTS` as that option doesn't expect to "build" `cf-support`? Will try tomorrow.

misc/cf-support
```diff
@@ -0,0 +1,168 @@
...
+.BR cf-runagent (8),
+.BR cf-key (8)
+.SH AUTHOR
+Northern.tech AS
+EOF
+exit 0
+fi
+
+[ "$1" = "--yes" ] || [ "$1" = "-y" ] && non_interactive=1
+# TODO: remove set -x here
```
Lex-2008, line 75: was it removed? I don't see it anywhere near

misc/cf-support
```diff
@@ -0,0 +1,168 @@
...
+log_cmd "hostname -f"
+log_cmd "uname -a"
+[ -f "/etc/os-release" ] && log_cmd "cat /etc/os-release"
+log_cmd "$BINDIR/cf-promises -V"
+log_cmd "free -m"
+log_cmd "df -h"
+log_cmd "ps auwwx"
+log_cmd "top -b -H -c -n1"
+log_cmd "netstat -ie"
+log_cmd "lsof"
```
Lex-2008, line 134: this command produces quite a lot of output! 503269 (half a million) on my desktop machine when running as non-root user, and 22884 (23k) on my private server. Maybe it's better to save it to individual file?

misc/cf-support
```diff
@@ -0,0 +1,168 @@
...
+  else
+    echo "$1 not found" >> "$info_file"
+  fi
+}
+export -f gzip_add
+
+function log_cmd
+{
+  echo "=== $1 ===" | tee -a "$info_file"
+  eval "$1" | tee -a "$info_file"
```
Lex-2008, line 120: worth adding `|| true` because some of commands (notably, netstat) might not be present on some systems.

Also, probably worth adding `2>&1` before the `| tee` - to catch stderr, too. Otherwise, at least "bash: netstat: command not found" message is lost.

misc/cf-support
```diff
@@ -0,0 +1,168 @@
...
+  $syslog_cmd | egrep 'cf-|CFEngine' | gzip -c > "$_syslog_filtered"
+else
+  $syslog_cmd | grep cf- | gzip -c > "$_syslog_filtered"
+fi
+gzip_add $WORKDIR/outputs/previous
+file_add $WORKDIR/policy_server.dat
+[ -f $WORKDIR/share/cf-support-nova-hub ] && source $WORKDIR/share/cf-support-nova-hub
+find "$tmpdir"
+ls -l "$tmpdir"
+cat "$info_file"
```
Lex-2008, line 161: All contents of this file is added via `| tee -a "$info_file"`, so maybe there's no real need to `cat` it afterwards, again?

misc/cf-support
```diff
@@ -0,0 +1,168 @@
...
+
+if [ -f $WORKDIR/share/cf-support-nova-hub ]; then
+  source $WORKDIR/share/cf-support-nova-hub
+fi
+
+tar czf "$collection.tar.gz" --directory $tmpdir/../ . && rm -rf $tmpdir
```
Lex-2008, line 167: I wonder if instead of `find "$tmpdir"` above it might be fancier to add `v` to tar arguments
(and echo a message "creating $collection.tar.gz file" before tar command).

It will serve same purpose (output all files), but also is slightly slower (since tar takes time to compress), and "keeps user entertained" (for a lack of better word) when tar is working.

misc/cf-support
```diff
@@ -0,0 +1,168 @@
...
+export tmpdir="$(mktemp -d)/$collection"
+mkdir -p "$tmpdir"
+
+export info_file="$tmpdir/system-info.txt"
+
+function file_add
+{
+  local filename=$(basename $1)
+  if [ -f "$1" ]; then
+    cp "$1" "$tmpdir"/"$filename"
```
Lex-2008, line 99: maybe it's better to add `echo "adding file $1" here to be more transparent and in line with `log_cmd` which shows what commands we execute?

Lex-2008 (dismissed):
nice! Suggested few improvements

misc/cf-support
craigcomstock: Yes. I removed it ;)

misc/cf-support
craigcomstock: agreed.

misc/cf-support
craigcomstock: true. ;)

misc/cf-support
craigcomstock: odd. yeah. debugs.

misc/cf-support
craigcomstock: The find was debugs for me. I don't think the user will have to wait long. I'll remove the find and keep this as-is.

misc/cf-support
craigcomstock: mmm, an echo to show progress would be fine :+1:

misc/cf-support
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env bash
+
+export WORKDIR=/var/cfengine
+export BINDIR="$WORKDIR/bin"
+export non_interactive=0
+if [ "$1" = "-M" ]; then
```
vpodzime, line 6: You should also check `$#` here.

misc/cf-support
```diff
@@ -0,0 +1,165 @@
...
+.BR cf-agent (8),
+.BR cf-serverd (8),
+.BR cf-execd (8),
+.BR cf-monitord (8),
+.BR cf-runagent (8),
+.BR cf-key (8)
+.SH AUTHOR
+Northern.tech AS
+EOF
+exit 0
```
vpodzime, line 71: This should be indented.

misc/cf-support
```diff
@@ -0,0 +1,165 @@
...
+if [ "$EUID" -ne 0 ]; then
+  echo "$0 must be run as root"
+  exit 1
+fi
+
+
+if [ $non_interactive -eq 0 ]; then
+  read -rp "If you have it, please enter your support case number: " case_number
+fi
+case_number=${case_number:-NA}
```
vpodzime, line 85: Please use quotes around var expansion, especially when the value comes from a user.

misc/cf-support
```diff
@@ -0,0 +1,165 @@
...
+  fi
+}
+export -f file_add
+
+function gzip_add
+{
+  local filename
+  filename="$(basename "$1")"
+  if [ -f "$1" ]; then
+    gzip -c "$1" > "$tmpdir"/"$filename".gz
```
vpodzime, line 112: This can also fail if `gzip` is not available.

misc/cf-support
```diff
@@ -0,0 +1,165 @@
...
+.BR cf-runagent (8),
+.BR cf-key (8)
+.SH AUTHOR
+Northern.tech AS
+EOF
+exit 0
+fi
+
+[ "$1" = "--yes" ] || [ "$1" = "-y" ] && non_interactive=1
+set -e
```
vpodzime, line 75: What's the reason/benefit for/of `set -e` in this script?

misc/cf-support
```diff
@@ -0,0 +1,165 @@
...
+
+log_cmd "hostname -f"
+log_cmd "uname -a"
+[ -f "/etc/os-release" ] && log_cmd "cat /etc/os-release"
+log_cmd "$BINDIR/cf-promises -V"
+log_cmd "free -m"
+log_cmd "df -h"
+log_cmd "ps auwwx"
+log_cmd "top -b -H -c -n1"
+log_cmd "netstat -ie"
```
vpodzime, line 135: Trying `ifconfig` as a fallback might cover some more cases here.

misc/cf-support
```diff
@@ -0,0 +1,165 @@
...
+  log_cmd "systemctl status cfengine3"
+else
+  log_cmd "/etc/init.d/cfengine3 status"
+fi
+for f in /var/log/CFEngine-Install*; do
+  gzip_add "$f"
+done
+[ -f /var/log/messages ] && syslog_cmd="cat /var/log/messages"
+[ -f /var/log/syslog ] && syslog_cmd="cat /var/log/syslog"
+[ -z "$syslog_cmd" ] && syslog_cmd="dmesg"
```
vpodzime, line 153: `dmesg` is not syslog.

misc/cf-support
```diff
@@ -0,0 +1,165 @@
...
+log_cmd "$BINDIR/cf-check diagnose"
+if [ "$(command -v systemctl >/dev/null)" ]; then
+  log_cmd "systemctl status cfengine3"
+else
+  log_cmd "/etc/init.d/cfengine3 status"
+fi
+for f in /var/log/CFEngine-Install*; do
+  gzip_add "$f"
+done
+[ -f /var/log/messages ] && syslog_cmd="cat /var/log/messages"
```
vpodzime, line 151: What about `journalctl`?

misc/cf-support
vpodzime: Plus we should tell use what's being grabbed from their system.

misc/cf-support
```diff
@@ -0,0 +1,165 @@
...
+$syslog_cmd | grep -E 'cf-|CFEngine' | gzip -c > "$_syslog_filtered" || true
+gzip_add $WORKDIR/outputs/previous
+file_add $WORKDIR/policy_server.dat
+
+if [ -f $WORKDIR/share/cf-support-nova-hub.sh ]; then
+  # shellcheck source=/dev/null
+  source $WORKDIR/share/cf-support-nova-hub.sh
+fi
+
+tar czf "$collection.tar.gz" --directory "$tmpdir/../" . && rm -rf "$tmpdir"
```
vpodzime, line 164: Why `/../`?

misc/cf-support
craigcomstock: For what reason would I check `$#`? There are two options supported and not in combination. If I check for number of arguments in relation to `-M` then isn't that the same difference as the current check of `[ "$1" = "-M" ]`?

misc/cf-support
craigcomstock: I am accustomed to using `dmesg` in alpine linux as there is no syslog by default "I think". Hmm.

misc/cf-support
craigcomstock: `tmpdir="$(mktemp -d)/$collection"` so we use `mktemp` and add another dir `$collection` and want the support tarball to have that top-level directory `$collection` in the tarball.

misc/cf-support
vpodzime: If there are no arguments, this will fail because `=` is a binary operator.

craigcomstock: Not sure I understand. The way I understand things and how it seems to test for me is that if no arguments then the expression will still be valid as `[ "" = "-M" ]` (two arguments given to a binary operator `=`) and give the right answer. I put quotes around `$1` to allow for no arguments.

```
[craig@mbp ~]$ x=""
[craig@mbp ~]$ [ $x = "M" ]
-bash: [: =: unary operator expected
[craig@mbp ~]$ [ "$x" = "M" ]
[craig@mbp ~]$ echo $?
1
```

vpodzime: :+1: Cool. I though it would/might fail, maybe based on some past experience with a weird/old shell.

misc/cf-support
vpodzime: OK, makes sense. Please add a comment explaining this.

misc/cf-support
craigcomstock: Right on. It will definitely fail if no quotes because then you would end up with `[ = "-M" ]` like you said, binary versus unary operator. :tada: thanks for watching out for it!

vpodzime:
@cf-bottom jenkins with no tests, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/8526/)

 [NO TESTS]

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/8526/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-8526/

misc/cf-support
```diff
@@ -0,0 +1,189 @@
...
+log_cmd "uname -a"
+[ -f "/etc/os-release" ] && log_cmd "cat /etc/os-release"
+log_cmd "$BINDIR/cf-promises -V"
+log_cmd "free -m"
+log_cmd "df -h"
+log_cmd "ps auwwx"
+log_cmd "top -b -H -c -n1"
+log_cmd "netstat -ie"
+netstat -ie >/dev/null 2>/dev/null || log_cmd "ifconfig"
+log_cmd "ifconfig"
```
vpodzime, line 151: Double `ifconfig` here.

misc/cf-support
```diff
@@ -0,0 +1,189 @@
...
+log_cmd "hostname -f"
+log_cmd "uname -a"
+[ -f "/etc/os-release" ] && log_cmd "cat /etc/os-release"
+log_cmd "$BINDIR/cf-promises -V"
+log_cmd "free -m"
+log_cmd "df -h"
+log_cmd "ps auwwx"
+log_cmd "top -b -H -c -n1"
+log_cmd "netstat -ie"
+netstat -ie >/dev/null 2>/dev/null || log_cmd "ifconfig"
```
vpodzime, line 150: So `netstat` output is not captured?

misc/cf-support
```diff
@@ -0,0 +1,189 @@
...
+  log_cmd "/etc/init.d/cfengine3 status"
+fi
+for f in /var/log/CFEngine-Install*; do
+  gzip_add "$f"
+done
+[ -f /var/log/messages ] && syslog_cmd="cat /var/log/messages"
+[ -f /var/log/syslog ] && syslog_cmd="cat /var/log/syslog"
+[ -z "$syslog_cmd" ] && syslog_cmd="dmesg"
+_syslog_filtered="$tmpdir"/syslog-filtered-for-cfengine.log.gz
+$syslog_cmd | grep -E 'cf-|CFEngine' | gzip -c > "$_syslog_filtered" || true
```
vpodzime, line 171: Why collecting from `$syslog_cmd` even if `journalctl` is available and then collecting from `journalctl`? Why not do `syslog_cmd="journalctl"`?

vpodzime (commented):
Looks good to me otherwise.

misc/cf-support
```diff
@@ -0,0 +1,189 @@
...
+.br
+For an introduction to promise theory, please see http://arxiv.org/abs/0810.3294/
+.SH AVAILABILITY
+cf-support is part of CFEngine.
+.br
+Binary packages may be downloaded from http://cfengine.com/downloads/.
+.br
+The source code is available at http://github.com/cfengine/
+.SH BUGS
+Please see the public bug-tracker at http://bug.cfengine.com/.
```
Lex-2008, line 64: I think, this site doesn't work, at least for me?

misc/cf-support
```diff
@@ -0,0 +1,189 @@
...
+.SH DESCRIPTION
+cf-support gathers various details about the system and creates a tarball in the current directory to submit to support. If the system is an enterprise hub then additional details will be gathered and included. The utility will prompt for an optional support ticket number as well as prompt whether to include masterfiles in the tarball.
+
+.SH OPTIONS
+.IP "--yes"
+Non-interactive use. Assume no ticket number and assume include masterfiles.
+.SH CFENGINE
+CFEngine provides automated configuration management of large-scale computer systems. A system administrator describes the desired state of a system using CFEngine policy code. The program \fBcf-agent\fR reads policy code and attempts to bring the current system state to the desired state described. Policy code is downloaded by \fBcf-agent\fR from a \fBcf-serverd\fR daemon. The daemon \fBcf-execd\fR is responsible for running \fBcf-agent\fR periodically.
+.br
+Documentation for CFEngine is available at http://cfengine.com/documentation/.
```
Lex-2008, line 52: maybe docs.cfengine.com?

Also, what about changing all links to httpS? Most of these links redirect to httpS, anyway.

misc/cf-support
Lex-2008, line 151: above command already executes `log_cmd "ifconfig"`...

misc/cf-support
```diff
@@ -0,0 +1,189 @@
...
+[ -f "/etc/os-release" ] && log_cmd "cat /etc/os-release"
+log_cmd "$BINDIR/cf-promises -V"
+log_cmd "free -m"
+log_cmd "df -h"
+log_cmd "ps auwwx"
+log_cmd "top -b -H -c -n1"
+log_cmd "netstat -ie"
+netstat -ie >/dev/null 2>/dev/null || log_cmd "ifconfig"
+log_cmd "ifconfig"
+command -v lsof 2>/dev/null >/dev/null && lsof > "$tmpdir/lsof.txt" || true
```
Lex-2008, line 152: maybe `echo` a message to the user that we're collecting this?

Lex-2008 (dismissed):

misc/cf-support
craigcomstock: mm, I copy/pasted this from elsewhere so we have more to fix in this case.

I'll log a ticket to fix all instances.

misc/cf-support
craigcomstock: I found that the `-e` option is not always available so I figured if `log_cmd "netstat -ie"` worked in the previous line send that output to `/dev/null`, if it didn't work, run `ifconfig`.

misc/cf-support
craigcomstock: oh. right. I could be less "tricky" and just run both. :)

misc/cf-support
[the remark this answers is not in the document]
craigcomstock: ack

misc/cf-support
craigcomstock: Right, makes sense. :+1:

misc/cf-support
Lex-2008: yes, in some places in our buildscripts we have things like `if [ "x$var" = "xtrue" ]` - but I believe it's for some very exotic exotics, maybe we even don't support them anyway (but I'm not sure, I haven't tested it).

misc/cf-support
craigcomstock: https://tracker.mender.io/browse/ENT-9115

misc/cf-support
```diff
@@ -0,0 +1,188 @@
...
+fi
+for f in /var/log/CFEngine-Install*; do
+  gzip_add "$f"
+done
+[ -f /var/log/messages ] && syslog_cmd="cat /var/log/messages"
+[ -f /var/log/syslog ] && syslog_cmd="cat /var/log/syslog"
+[ "$(command -v journalctl >/dev/null)" ] && syslog_cmd="journalctl"
+[ -z "$syslog_cmd" ] && syslog_cmd="dmesg"
+_syslog_filtered="$tmpdir"/syslog-filtered-for-cfengine.log.gz
+$syslog_cmd | grep -E 'cf-|CFEngine' | gzip -c > "$_syslog_filtered" || true
```
vpodzime, line 174: Missing `echo "Added system log"`.

craigcomstock: excellent.

vpodzime (approved):
ACK. Thanks for fixing all the nitpicks.

craigcomstock:
> ACK. Thanks for fixing all the nitpicks.

No problem. Appreciate the details! :)

craigcomstock:
cherry picked to 3.18.x https://github.com/cfengine/core/pull/5042

PR #5021, opened 2022-08-08
[test/dontmerge] unblock Solaris

Lex-2008:
This temporary reverts commit ef86e711843798157b24d3d38b484a9aef67baf9.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -136,14 +136,6 @@ AC_CONFIG_LIBOBJ_DIR(libntech/libcompat)
 AC_PATH_PROG(GETCONF, getconf, false, $PATH:$prefix/bin:/usr/bin:/usr/local/bin:/sw/bin)
 AM_CONDITIONAL(CROSS_COMPILING, test "x$cross_compiling" = "xyes")
 
-# Check whether `pkg-config' is available
-AC_ARG_VAR([PKG_CONFIG], [path to pkg-config])
-AC_ARG_VAR([PKG_CONFIG_PATH], [directories to add to the pkg-config search path])
-AC_ARG_VAR([PKG_CONFIG_LIBDIR], [path overriding pkg-config's search path])
-
-if test "x$ac_cv_env_PKG_CONFIG_set" != "xset"; then
-    AC_PATH_TOOL([PKG_CONFIG], [pkg-config])
-fi
 
 dnl ######################################################################
 dnl Use pthreads if available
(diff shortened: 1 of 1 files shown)


hughmcmaster:
Hi @Lex-2008, what's the reasoning behind this reversion MR?

Lex-2008:
hi @hughmcmaster, issue is that currently our master branch fails to build Solaris, with errors like this:
```
files_editxml.c: In function 'DeleteTextInNode':
files_editxml.c:1659:9: error: implicit declaration of function 'xmlFirstElementChild' [-Werror=implicit-function-declaration]
         elemnode = xmlFirstElementChild(docnode);
         ^
```
and this:
```
xml-c14nize.c: In function 'xmlC14nizeFile':
xml-c14nize.c:25:37: error: 'XML_C14N_1_0' undeclared (first use in this function)
     if (xmlC14NDocSaveTo(doc, NULL, XML_C14N_1_0, 0, true, out) < 0)
                                     ^
xml-c14nize.c:25:37: note: each undeclared identifier is reported only once for each function it appears in
Makefile:706: recipe for target 'xml_c14nize-xml-c14nize.o' failed
gmake[4]: *** [xml_c14nize-xml-c14nize.o] Error 1
```
I created this PR for two reasons:
a) to confirm that it's indeed caused by this commit
b) to unblock testing some other work that I'm doing on Solaris (https://github.com/cfengine/buildscripts/pull/1107, if you're interested)

I'll work on a proper fix a bit later. No, we're not planning to revert your commit for real :) I should've marked this PR as test/dontmerge from the beginning, sorry for the confusion!

hughmcmaster:
Thanks for clarifying.

It would be good to know what version of libxml2 is available on your Solaris 10 system.

`xmlFirstChildElement` requires 2.7.3 at a minimum.

My Solaris 10 test VM had libxml 2.6.23 – too old to have the needed symbol. But once I installed a more recent version, it was fine.

I can't check your Solaris logs, since I get 403 access denied (same with your CI builds, unless using the GitHub checks tab).

Functionally, there is little difference pre-patch and post-patch. If it is a version issue, I'd recommend adding a minimum version.

vpodzime:
> It would be good to know what version of libxml2 is available on your Solaris 10 system.

We build _libxml2_ for CFEngine on our build machines (including the Solaris 10 ones) in our build root. But for some reason, the pkg-config files for libxml are available on the Solaris 10 machines in the system. And they are picked up and preferred over the definitions using _libxml_ from our build root.

vpodzime:
Thanks for the tips, @hughmcmaster! However, I think the easiest thing will be to remove the problematic `.pc` files from our Solaris hosts. We don't need them there.

hughmcmaster:
That should work too!

Lex-2008:
Indeed, our Solaris 10 build machines have quite a lot of *.pc files in `/usr/lib/pkgconfig`.

Thanks @hughmcmaster and @vpodzime for your input, I managed to get a green build by setting relevant `PKG_CONFIG_` variable - see https://github.com/cfengine/buildscripts/pull/1109

Just deleting a `*.pc` file might be simple, but fragile solution - what if we migrate to another Solaris VM in the future?

FWIW, pkg-config on Solaris 10 doesn't use `PKG_CONFIG_LIBDIR` env variable, only `PKG_CONFIG_PATH`.

Oh, and also to answer @hughmcmaster's question: yes, our Solaris build machine also has libxml version 2.6.23 installed.

PR #5024, opened 2022-08-09
CFE-4022: Redacted useless ifelse

nickanderson:
If this, do that, else, do that anyway. Instead, just do that.

Ticket: CFE-4022
Changelog: None

### DIFF
--- a/cf-monitord/env_monitor.c
+++ b/cf-monitord/env_monitor.c
@@ -640,15 +640,7 @@ static void ArmClasses(EvalContext *ctx, const Averages *const av)
                 strcat(ldt_buff, buff);
             }
 
-            if (CF_THIS[i] > av->Q[i].expect)
-            {
-                snprintf(buff, CF_BUFSIZE, "%s_high_ldt", name);
-            }
-            else
-            {
-                snprintf(buff, CF_BUFSIZE, "%s_high_ldt", name);
-            }
-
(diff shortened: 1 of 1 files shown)


nickanderson:
Making CFEngine smaller, 8 lines at a time.

craigcomstock (approved):
would be interesting to know why the if/else is there but I spelunked the code and found the original commit was like this, so no history to tell us why.

vpodzime (approved):
ACK

cf-monitord/env_monitor.c
```diff
@@ -640,15 +640,7 @@ static void ArmClasses(EvalContext *ctx, const Averages *const av)
                 strcat(ldt_buff, buff);
             }
 
-            if (CF_THIS[i] > av->Q[i].expect)
-            {
-                snprintf(buff, CF_BUFSIZE, "%s_high_ldt", name);
-            }
-            else
-            {
-                snprintf(buff, CF_BUFSIZE, "%s_high_ldt", name);
```
olehermanse, line 649: @nickanderson AFAICT this was intended to be setting the `_low_ldt` class instead. If greater -> high, if lower -> low.

nickanderson: Ah, okay. Here is the better fix then.

https://github.com/cfengine/core/pull/5033

PR #5032, opened 2022-08-15
Check if $with_libxml2/bin/xml2-config exists and is executable

vpodzime:
Otherwise we rely on it and fail to get data from it. Not all our
builds provide xml2-config even when they use
'--with-libxml2=/var/cfengine'.

Ticket: ENT-9070
Changelog: None


vpodzime:
:crossed_fingers: [Build Status](https://ci.cfengine.com/job/pr-pipeline/8489/)

Lex-2008 (approved):

PR #5033, opened 2022-08-15
CFE-4022: Fixed definition of _low_ldt class from cf-monitord

nickanderson:


cf-monitord/env_monitor.c
```diff
@@ -640,7 +640,15 @@ static void ArmClasses(EvalContext *ctx, const Averages *const av)
                 strcat(ldt_buff, buff);
             }
 
-            snprintf(buff, CF_BUFSIZE, "%s_high_ldt", name);
+            if (CF_THIS[i] > av->Q[i].expect)
+            {
+                snprintf(buff, CF_BUFSIZE, "%s_high_ldt", name);
+            }
+            else
```
olehermanse, line 647: Perhaps:

```suggestion
            else if (CF_THIS[i] < av->Q[i].expect)
```

I really don't understand exactly what all of these variables are. @vpodzime any input?

olehermanse: I am guessing that since we are inside the if for:

```c
if (LDT_FULL && (CHI[i] > CHI_LIMIT[i]))
```

This else if is not necessary.

olehermanse (approved):

cf-monitord/env_monitor.c
vpodzime: Do you think the above condition ensures that `CF_THIS[i] > av->Q{i].expect`?

PR #5040, opened 2022-08-30
Bump Travis to Ubuntu 18.04

vpodzime:
Ubuntu 16.04 is no longer supported and it's using OpenSSL that
doesn't even support TLS 1.3.


vpodzime:
This causes multiple failures which I'm not willing to spend time on.

PR #5058, opened 2022-09-15
Fix non-root builds

stweil:
Out of tree builds not working:

$ cd core
$ make distclean
$ cd ..
$ mkdir build
$ cd build
$ ../core/autogen.sh --enable-debug
$ make -j8

Eventually gives me:

Making all in cf-agent
make[2]: *** No rule to make target '../../core/libntech/libutils/libutils.la', needed by 'manifest_file'.  Stop.

### DIFF
--- a/cf-agent/Makefile.am
+++ b/cf-agent/Makefile.am
@@ -129,7 +129,7 @@ TESTS =
 if !NT
 noinst_PROGRAMS = manifest_file
 manifest_file_SOURCES = manifest_file.c
-manifest_file_LDADD = ${top_srcdir}/libntech/libutils/libutils.la \
+manifest_file_LDADD = ../libntech/libutils/libutils.la \
 					  ../libpromises/libpromises.la \
                       libcf-agent.la
 TESTS += manifest_file
--- a/cf-monitord/Makefile.am
+++ b/cf-monitord/Makefile.am
@@ -77,7 +77,7 @@ if LINUX
 TESTS += get_socket_info
 noinst_PROGRAMS = get_socket_info
 get_socket_info_SOURCES = get_socket_info.c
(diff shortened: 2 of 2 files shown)


craigcomstock:
Can you give an example of what problem you are solving? I build core quite often without being the root user. Or are you referring to a different `root`?

cf-agent/Makefile.am
```diff
@@ -129,7 +129,7 @@ TESTS =
 if !NT
 noinst_PROGRAMS = manifest_file
 manifest_file_SOURCES = manifest_file.c
-manifest_file_LDADD = ${top_srcdir}/libntech/libutils/libutils.la \
```
craigcomstock, line 132: libntech should be checked out as a submodule so `${top_srcdir}/libntech` certainly seems correct to me when referring to `libntech`.

craigcomstock: What you have written here seems equivalent unless somehow autotools thinks `top_srcdir` is something else?

Ah, here I think maybe you are building out of tree then? Since then certainly the `libutils.la` would not be in `${top_srcdir}` but in the build directory. I will try your PR out in this way and see if it helps.

craigcomstock: Yes. I think this is what you are doing. I find your change fixes this build error:

```
$ cd core
$ make distclean
$ cd ..
$ mkdir build
$ cd build
$ ../core/autogen.sh --enable-debug
$ make -j8
```

Eventually gives me

```
Making all in cf-agent
make[2]: *** No rule to make target '../../core/libntech/libutils/libutils.la', needed by 'manifest_file'.  Stop.
```

But with your changes, no such errors. Excellent! I was just wishing that out-of-tree builds worked the other day. Thanks!

craigcomstock (approved):

PR #5062, opened 2022-09-20
ENT-9162: Refined testall check for rpmcmpver binary (follow up)

craigcomstock:
Needed to add the check for `rpm` to another section where we check that all set binaries exist and are executable.

Ticket: ENT-9162
Changelog: none


craigcomstock:
trying with canaries only: [Build Status](https://ci.cfengine.com/job/pr-pipeline/8620/)

Since this is really the "same change" in a different place so should be "OK" if passes canaries (ubuntu, redhat, mingw)

nickanderson (approved):

craigcomstock:
cherry picked to
3.15: https://github.com/cfengine/core/pull/5065
3.18: https://github.com/cfengine/core/pull/5066

PR #5081, opened 2022-10-25
AUTHORS: minor fix to grammar.

tdlee:
* "less than": something non-countable
* "fewer than": something countable

So change "people with less than 10 commits..." to "people with fewer than 10 commits...".

Example ref: https://www.grammarly.com/blog/fewer-vs-less/


cf-bottom:

vpodzime (approved):
ACK. Thanks.

PR #5082, opened 2022-10-25
CFE-3020: ps output line ... is shorter than its associated header

tdlee:
A program can overwrite its incoming arguments.  This can include blanking them out, and subsequent "ps" output can be empty.

Add an additional test.

While this test itself does not expose or address the CFE-3020 bug, nevertheless it seems worth having such a test in place.

### DIFF
--- a/tests/unit/split_process_line_test.c
+++ b/tests/unit/split_process_line_test.c
@@ -117,6 +117,42 @@ static void test_split_line_noelapsed(void)
     free_and_null_strings(field);
 }
 
+static void test_split_line_nocmd(void)
+{
+    /* Collect all test data in one array to make alignments visible: */
+    static const char *lines[] = {
+        /* Indent any continuation lines so that it's clear that's what they are. */
+        /* Use the username field as a test name to confirm tests are in sync. */
+
+        "USER       PID STAT    VSZ  NI   RSS NLWP STIME     ELAPSED     TIME COMMAND",
+        "spacecmd   4 S         0   0     0    1 10:30       54:29 00:00:01  ",
+        "nullcmd    4 S         0   0     0    1 10:30       54:29 00:00:01 ",
+    };
+    char *name[CF_PROCCOLS] = { 0 }; /* Headers */
+    char *field[CF_PROCCOLS] = { 0 }; /* Content */
+    int start[CF_PROCCOLS] = { 0 };
+    int end[CF_PROCCOLS] = { 0 };
+    int user = 0, command = 10;
+    time_t pstime = 1410000000;
+
+    /* Prepare data needed by tests and assert things assumed by test: */
+    GetProcessColumnNames(lines[0], name, start, end);
+    assert_string_equal(name[user], "USER");
+    assert_string_equal(name[command], "COMMAND");
+
+    assert_true(SplitProcLine(lines[1], pstime, name, start, end, PCA_AllColumnsPresent, field));
+    assert_string_equal(field[user], "spacecmd");
+    assert_string_equal(field[command], lines[1] + 69);
+
+    assert_true(SplitProcLine(lines[2], pstime, name, start, end, PCA_AllColumnsPresent, field));
+    assert_string_equal(field[user], "nullcmd");
+    assert_string_equal(field[command], lines[1] + 69);
+
+    /* Finally, tidy away headers: */
+    free_and_null_strings(name);
+    free_and_null_strings(field);
+}
+
 static void test_split_line_longcmd(void)
 {
     static const char *lines[] = {
@@ -909,6 +945,7 @@ int main(void)
             unit_test(test_split_line_challenges),
             unit_test(test_split_line_noelapsed),
             unit_test(test_split_line_elapsed),
+            unit_test(test_split_line_nocmd),
             unit_test(test_split_line_longcmd),
             unit_test(test_split_line),
             unit_test(test_split_line_serious_overspill),
--- a/tests/unit/split_process_line_test.c
+++ b/tests/unit/split_process_line_test.c
@@ -126,7 +126,7 @@ static void test_split_line_nocmd(void)
 
         "USER       PID STAT    VSZ  NI   RSS NLWP STIME     ELAPSED     TIME COMMAND",
         "spacecmd   4 S         0   0     0    1 10:30       54:29 00:00:01  ",
-        "nullcmd    4 S         0   0     0    1 10:30       54:29 00:00:01 ",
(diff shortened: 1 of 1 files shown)


cf-bottom:

tests/unit/split_process_line_test.c
```diff
@@ -117,6 +117,42 @@ static void test_split_line_noelapsed(void)
...
+    time_t pstime = 1410000000;
+
+    /* Prepare data needed by tests and assert things assumed by test: */
+    GetProcessColumnNames(lines[0], name, start, end);
+    assert_string_equal(name[user], "USER");
+    assert_string_equal(name[command], "COMMAND");
+
+    assert_true(SplitProcLine(lines[1], pstime, name, start, end, PCA_AllColumnsPresent, field));
+    assert_string_equal(field[user], "spacecmd");
+    assert_string_equal(field[command], lines[1] + 69);
```
olehermanse, line 145: I think it's easier and less error prone if you just compare to what string you expect here, for example:

```suggestion
    assert_string_equal(field[command], "");
```

tests/unit/split_process_line_test.c
```diff
@@ -117,6 +117,42 @@ static void test_split_line_noelapsed(void)
...
+
+    assert_true(SplitProcLine(lines[2], pstime, name, start, end, PCA_AllColumnsPresent, field));
+    assert_string_equal(field[user], "nullcmd");
+    assert_string_equal(field[command], lines[1] + 69);
```
olehermanse, line 149: Same as above, also unclear whether the `lines[1]` here is intentional or not? We are working on `lines[2]` now.

```suggestion
    assert_true(SplitProcLine(lines[2], pstime, name, start, end, PCA_AllColumnsPresent, field));
    assert_string_equal(field[user], "nullcmd");
    assert_string_equal(field[command], "");
```

tests/unit/split_process_line_test.c
```diff
@@ -117,6 +117,42 @@ static void test_split_line_noelapsed(void)
...
+static void test_split_line_nocmd(void)
+{
+    /* Collect all test data in one array to make alignments visible: */
+    static const char *lines[] = {
+        /* Indent any continuation lines so that it's clear that's what they are. */
+        /* Use the username field as a test name to confirm tests are in sync. */
+
+        "USER       PID STAT    VSZ  NI   RSS NLWP STIME     ELAPSED     TIME COMMAND",
+        "spacecmd   4 S         0   0     0    1 10:30       54:29 00:00:01  ",
+        "nullcmd    4 S         0   0     0    1 10:30       54:29 00:00:01 "
```
olehermanse, line 129: Why does line 1 end with 2 spaces and line 2 end with 1 space?

tests/unit/split_process_line_test.c
```diff
@@ -117,6 +117,42 @@ static void test_split_line_noelapsed(void)
...
 }
 
+static void test_split_line_nocmd(void)
+{
+    /* Collect all test data in one array to make alignments visible: */
+    static const char *lines[] = {
+        /* Indent any continuation lines so that it's clear that's what they are. */
+        /* Use the username field as a test name to confirm tests are in sync. */
+
+        "USER       PID STAT    VSZ  NI   RSS NLWP STIME     ELAPSED     TIME COMMAND",
```
olehermanse, line 127: If this is exactly copied from a (ps) command, putting the command in a comment here would be helpful.

If not, is it intentional / important that the labels and values are sometimes misaligned? It could help readability if you added some spaces to align things perfectly.

tests/unit/split_process_line_test.c
```diff
@@ -117,6 +117,42 @@ static void test_split_line_noelapsed(void)
...
+        "spacecmd   4 S         0   0     0    1 10:30       54:29 00:00:01  ",
+        "nullcmd    4 S         0   0     0    1 10:30       54:29 00:00:01 "
+    };
+    char *name[CF_PROCCOLS] = { 0 }; /* Headers */
```
olehermanse, line 131: `headers` or `labels` would be a better name here.

tests/unit/split_process_line_test.c
```diff
@@ -117,6 +117,42 @@ static void test_split_line_noelapsed(void)
...
+static void test_split_line_nocmd(void)
+{
+    /* Collect all test data in one array to make alignments visible: */
+    static const char *lines[] = {
+        /* Indent any continuation lines so that it's clear that's what they are. */
+        /* Use the username field as a test name to confirm tests are in sync. */
+
+        "USER       PID STAT    VSZ  NI   RSS NLWP STIME     ELAPSED     TIME COMMAND",
+        "spacecmd   4 S         0   0     0    1 10:30       54:29 00:00:01  ",
+        "nullcmd    4 S         0   0     0    1 10:30       54:29 00:00:01 ",
```
olehermanse: Why does line 1 end with 2 spaces and line 2 end with 1 space? If this is intentional, it deserves a comment.

olehermanse (changes requested):
Thank you for working on this. I've left some questions and suggestions on individual lines of code. Additionally, this is not passing tests as there is some illegal memory access going on in one of the `assert_string_equal` lines. Feel free to comment out parts of the test to see which line is problematic. I assume you can see the test results and failures in GitHub?

tdlee:
> Marking this PR as stale due to inactivity; it will be closed in 7 days.

Why has this been marked for closure?

Surely it is overdue for the opposite: taking forwards to completion, isn't it?

The long-existing means of getting process data by reverse-engineering idiosyncratic human-readable output of "ps" seems problematic, and there is already one known bug.

I realise that bug-fixing is less glamorous than adding new features, but could I ask the Northern.Tech address the backlog of bugs, including the "ps" one, please?  I have provided an MR with the hardcore work done. Could you take it forwards through your acceptance refining and release procedures, please?  Thanks.

tests/unit/split_process_line_test.c
tdlee: Please feel free to take this forwards as you wish.  I have no particular opinion one way or the other.  So take it forwards, please, in the way you think best, with your own amendments to your styles and taste.

Thanks.

tests/unit/split_process_line_test.c
tdlee: Please feel free to add a comment that you think best addresses the idea.

Thanks.

nickanderson:
Thanks @tdlee we do enthusiastically appreciate your efforts, however I don't know if or when we might pick it up ourselves.

I updated the description of CFE-3020 to highlight this pr in hopes whoever does pick it up in the future can benefit from the work you have already done.

Thanks.

PR #5085, opened 2022-10-27
ENT-9473: SELinux fixes for sending reports via an external SMTP server

vpodzime:


craigcomstock (approved):
Yes! I have already even tested this on a RHEL 8 hub! :) Mail sending was fixed with this patch. :)

PR #5087, opened 2022-11-01
ENT-9496: Allow PostgreSQL to create directories under /var/cfengine

vpodzime:
It needs some space for temporary files when returning large data sets as query results.

Ticket: ENT-9496
Changelog: SELinux no longer breaks SQL queries with large result
           sets on RHEL 8 hubs


craigcomstock (approved):
excellent

PR #5089, opened 2022-11-01
ENT-9557: Allow selinux hub context to send email

craigcomstock:
Found when testing scheduled reports emailing.

Ticket: ENT-9557
Changelog: title


vpodzime (approved):
ACK

PR #5100, opened 2022-11-09
[3.18.x] CFE-4094: Allow any scalar type for a constraint expecting a string

vpodzime:
Scalar values (int, real, string,...) are internally represented as strings. When doing a syntax check of policy, which happens in pre-eval, scalar values are considered OK for constraints (attributes) that expect string values. However, if a function cannot be called in pre-eval, the function's return type is checked instead of the value returned by the function and there's easily a mismatch because function return types are more specific than value types.

Let's consider all function return types the values of which would be scalar values OK for constraints (attributes) requiring string values.

Ticket: CFE-4094
Changelog: Unresolved function calls that return scalar values
           are now considered OK for constraints expecting
           strings during syntax check
(cherry picked from commit a57990e96bc2770607ead3ac450928b1741f5405)


larsewi (approved):
ACK

PR #5102, opened 2022-11-10
CFE-4097: Moved cfengine-code-style to it's own repository

nickanderson:
@tarsius, maintainer of the Emacsmirror asked that we consider moving
this to it's own repository. He provided sensible
reasoning (https://github.com/cfengine/core/discussions/5101), but notably the
current state results in additional costs and burden for the maintainers of
Emacs official package manager, melpa, and Emacsmirror.

So, the C code style for CFEngine has moved to it's own repository,
https://github.com/cfengine/cfengine-code-style.

Ticket: CFE-4097
Changelog: Title


vpodzime (commented):
*its own* in the commit message. And should we add the repo as a submodule? If this was the first submodule, I would vote against opening the gate to hell for it, but having the *libntech* submodule already, there should be nothing new if we add another one.

vpodzime (commented):
I think the submodule path should be something that looks like a directory. Maybe `emacs-code-style`? (and the identifier/name of the submodule can be also just that, not a full path)

nickanderson:
😅 that's what I did initially, then I renamed it to `.el` so it was clear it's the same thing and emacs specific. What about `.el.d`?

tarsius:
This isn't my call, but I would just go with `emacs-code-style`.

olehermanse (approved):

PR #5103, opened 2022-11-10
CFE-192: Added suggestion to use a negative lookahead when non-convergent edits are attempted

nickanderson:
This change simply tries to make it more clear how the log message can be avoided.

Ticket: CFE-192
Changelog: Title


vpodzime (approved):
Makes a lot of sense. ACK.

PR #5109, opened 2022-11-15
Fixed redhat package for yaml dependency

craigcomstock:
Not sure if this was every `yaml-devel` but checked on three systems and `libyaml-devel` seems to be the right answer.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -112,11 +112,11 @@ Here we have examples of command lines for various systems to install dependenci
 needed to build. The version and date checked are noted in parenthesis. Please
 submit PRs with updates to this information.
 
-* CentOS (centos7 2021-07-02)
+* RedHat/CentOS (rhel-8/rhel-9/centos-7 2022-11-15)
 
 Note that first you will need to install epel-release (https://fedoraproject.org/wiki/EPEL) for lmdb-devel to be available.
 
-$ sudo yum install -y gcc gdb make git libtool autoconf automake byacc flex openssl-devel pcre-devel lmdb-devel pam-devel flex-devel yaml-devel
+$ sudo yum install -y gcc gdb make git libtool autoconf automake byacc flex openssl-devel pcre-devel lmdb-devel pam-devel flex-devel libyaml-devel
 
 For SELinux support you will need selinux-policy-devel package and specify `--with-selinux-policy` to `autogen.sh` or `configure`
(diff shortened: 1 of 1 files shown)


nickanderson (approved):

nickanderson:
All the way back to el2, it's been libyaml-devel.

https://www.rpmfind.net/linux/rpm2html/search.php?query=libyaml-devel&submit=Search+...&system=&arch=

INSTALL
```diff
@@ -112,11 +112,11 @@ Here we have examples of command lines for various systems to install dependenci
 needed to build. The version and date checked are noted in parenthesis. Please
 submit PRs with updates to this information.
 
-* CentOS (centos7 2021-07-02)
+* RedHat/CentOS (rhel-8/rhel-9/centos-7 2022-11-15)
 
 Note that first you will need to install epel-release (https://fedoraproject.org/wiki/EPEL) for lmdb-devel to be available.
 
-$ sudo yum install -y gcc gdb make git libtool autoconf automake byacc flex openssl-devel pcre-devel lmdb-devel pam-devel flex-devel yaml-devel
+$ sudo yum install -y gcc gdb make git libtool autoconf automake byacc flex openssl-devel pcre-devel lmdb-devel pam-devel flex-devel libyaml-devel
```
vpodzime, line 119: We also need `expat-devel` now.

vpodzime (approved):
ACK

INSTALL
craigcomstock: > We also need `expat-devel` now.

Actually... why? I don't have it on my rhel dev machines. I do see it in our CI config for redhat 8 and 9 build hosts but not sure exactly why. I don't see any mention in configure.ac in core so maybe a nova/hub dependency?

craigcomstock: @vpodzime I'll go ahead and merge this one and we can fiddle/discuss expat-devel elsewhere.

craigcomstock:
cherry picked to 3.18.x just for fun https://github.com/cfengine/core/pull/5116

INSTALL
vpodzime: Yup. It's a dependency of one of our nova dependencies (PHP?). The confusion on my side comes from my efforts to make `make full-vm-check` work with full nova builds last week.

PR #5115, opened 2022-11-21
Removed duplicate name_connect capability for tcp_socket

craigcomstock:
Ticket: none
Changelog: none


vpodzime (approved):
ACK

craigcomstock:
No need to back-port, this duplication occurred only in master for changes related to cfbs build which is not part of any previous version. Commit where it was introduced: https://github.com/cfengine/core/commit/34d7969170909de505f75ab018e33916ad7a6f41

PR #5119, opened 2022-12-02
SELinux fixes for call-collect

vpodzime:
Extra two rules for call-collect when RHEL 8 is running on the hub not just on the hosts.


craigcomstock (approved):
:lock:

PR #5133, opened 2022-12-20
Added fifo_file write access for cf-reactor

craigcomstock:
This is needed in order for cf-reactor to write out to cfbs process to get module inputs.

Ticket: none
Changelog: none


vpodzime (approved):
ACK

PR #5134, opened 2022-12-20
ENT-9605: Added setopt for cf-hub connecting to cf-serverd

craigcomstock:
cf-hub needs access to set socket options when connecting to cf-serverd (port 5308)

Ticket: ENT-9605
Changelog: none


nickanderson (approved):

vpodzime (approved):
Ack

PR #5135, opened 2022-12-20
ENT-9605: Added needed selinux policy for Policy Analyzer enablement

craigcomstock:
Ticket: ENT-9605
Changelog: none

### DIFF
--- a/misc/selinux/cfengine-enterprise.te
+++ b/misc/selinux/cfengine-enterprise.te
@@ -634,6 +634,11 @@ allow cfengine_httpd_t system_dbusd_var_run_t:dir search;
 allow cfengine_httpd_t system_dbusd_var_run_t:sock_file write;
 allow init_t cfengine_httpd_t:dbus send_msg;
 
+# rhel9 needs to enable policy analyzer
+allow cfengine_httpd_t systemd_userdbd_runtime_t:dir { add_name remove_name read open getattr search write };
+allow cfengine_httpd_t systemd_userdbd_runtime_t:sock_file { write };
+allow cfengine_httpd_t init_t:unix_stream_socket { connectto };
+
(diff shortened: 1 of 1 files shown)


misc/selinux/cfengine-enterprise.te
```diff
@@ -634,6 +634,11 @@ allow cfengine_httpd_t system_dbusd_var_run_t:dir search;
 allow cfengine_httpd_t system_dbusd_var_run_t:sock_file write;
 allow init_t cfengine_httpd_t:dbus send_msg;
 
+# rhel9 needs to enable policy analyzer
+allow cfengine_httpd_t systemd_userdbd_runtime_t:dir { add_name remove_name read open getattr search write };
```
craigcomstock, line 638: these seem to be actually needed in general for many pages in the mission-portal like dashboard, host info page... so not quite sure what these are "just yet" something to do with looking up user/group according to research @nickanderson did but not sure how that relates to dir and sock_file.

vpodzime: My guess would be NSS, but why would httpd or php need this?

craigcomstock: I thought about trying strace on httpd, granted maybe setup to run a single thread to make it easier. Or maybe step debug through the frontend to try and isolate the requests coming back and correlate with the denied message on the backend.

vpodzime: You should (or *might*, unfortunately) be able to see the violating syscall. But a full trace to the request would of course be the best.

craigcomstock:
I am not seeing any problems like this with either master or 3.18.4 or 3.18.5a so am closing this ticket as not needed.

PR #5143, opened 2022-12-29
ENT-9740: Added --help option to cf-support and aligned output with other components

nickanderson:
This change simply added an explicit --help option and aligned output with other components.

### DIFF
--- a/misc/cf-support
+++ b/misc/cf-support
@@ -4,9 +4,15 @@ export WORKDIR=/var/cfengine
 export BINDIR="$WORKDIR/bin"
 export non_interactive=0
 
-if [ $# -gt 1 ]; then
-  echo "usage: $(basename "$0") [--yes]"
-  echo '	--yes - Non-interactive use. Assume no ticket number and assume include masterfiles.'
+if [ $# -gt 1 ] || [ "$1" = "--help" ] || ["$1" = "-h" ]; then
+  echo "usage: $(basename "$0") [OPTIONS]"
+  echo ''
+  echo 'Options:'
+  echo '	--yes, -y - Non-interactive use. Assume no ticket number and assume include masterfiles.'
+  echo '	--help, -h - Print the help message'
+  echo ''
+  echo 'Website: http://www.cfengine.com'
+  echo 'This software is Copyright (C) 2008,2010-present Northern.tech AS.'
   exit 1
 fi
 


misc/cf-support
```diff
@@ -4,9 +4,15 @@ export WORKDIR=/var/cfengine
...
-  echo '	--yes - Non-interactive use. Assume no ticket number and assume include masterfiles.'
+if [ $# -gt 1 ] || [ "$1" = "--help" ] || ["$1" = "-h" ]; then
+  echo "usage: $(basename "$0") [OPTIONS]"
+  echo ''
+  echo 'Options:'
+  echo '	--yes, -y - Non-interactive use. Assume no ticket number and assume include masterfiles.'
+  echo '	--help, -h - Print the help message'
+  echo ''
+  echo 'Website: http://www.cfengine.com'
+  echo 'This software is Copyright (C) 2008,2010-present Northern.tech AS.'
```
craigcomstock, line 15: why 2008,2010-present?

craigcomstock: https://github.com/cfengine/core/blob/master/cf-promises/cf-promises.c and most of those I think have just

```
  Copyright 2022 Northern.tech AS
```

nickanderson: Yeah, i copied from cf-agent help output.

```
root@precision-5570:~# cf-agent -h
Usage: cf-agent [OPTIONS] [FILE]

Options:
  --bootstrap   , -B value - Bootstrap CFEngine to the given policy server IP, hostname or :avahi (automatic detection)
  --bundlesequence, -b value - Set or override bundlesequence from command line
  --workdir     , -w value - Override the default /var/cfengine work directory for testing (same as setting CFENGINE_TEST_OVERRIDE_WORKDIR)
  --debug       , -d       - Enable debugging output
  --define      , -D value - Define a list of comma separated classes to be defined at the start of execution
  --self-diagnostics, -x value - Run checks to diagnose a CFEngine agent installation
  --dry-run     , -n       - All talk and no action mode - make no changes, only inform of promises not kept
  --file        , -f value - Specify an alternative input file than the default. This option is overridden by FILE if supplied as argument.
  --help        , -h       - Print the help message
  --inform      , -I       - Print basic information about changes made to the system, i.e. promises repaired
  --log-level   , -g value - Specify how detailed logs should be. Possible values: 'error', 'warning', 'notice', 'info', 'verbose', 'debug'
  --negate      , -N value - Define a list of comma separated classes to be undefined at the start of execution
  --no-lock     , -K       - Ignore locking constraints during execution (ifelapsed/expireafter) if "too soon" to run
  --verbose     , -v       - Output verbose information about the behaviour of the agent
  --version     , -V       - Output the version of the software
  --timing-output, -t       - Output timing information on console when in verbose mode
  --trust-server, -T value - Possible values: 'yes' (default, trust the server when bootstrapping), 'no' (server key must already be trusted)
  --color       , -C value - Enable colorized output. Possible values: 'always', 'auto', 'never'. If option is used, the default value is 'auto'
  --no-extensions, -E       - Disable extension loading (used while upgrading)
  --timestamp   , -l       - Log timestamps on each line of log output
  --ignore-preferred-augments, -        - Ignore def_preferred.json file in favor of def.json
  --log-modules , - value - Enable even more detailed debug logging for specific areas of the implementation. Use together with '-d'. Use --log-modules=help for a list of available modules
  --show-evaluated-classes, - value - Show *final* evaluated classes, including those defined in common bundles in policy. Optionally can take a regular expression.
  --show-evaluated-vars, - value - Show *final* evaluated variables, including those defined without dependency to user-defined classes in policy. Optionally can take a regular expression.
  --skip-bootstrap-policy-run, -        - Do not run policy as the last step of the bootstrap process
  --skip-db-check, - value - Do not run database integrity checks and repairs at startup
  --simulate    , - value - Run in simulate mode, either 'manifest', 'manifest-full' or 'diff'

Website: http://www.cfengine.com
This software is Copyright (C) 2008,2010-present Northern.tech AS.

```

nickanderson: for the C --help output it seems to come from writer.c in libntech https://github.com/cfengine/libntech/blob/5882865019d7a032d04891d8829a380d69b7d6ad/libutils/writer.c#L351

nickanderson: So, i kinda think we should merge this and backport to 3.21 and 3.18, then if we should adjust libntech we should do that separately.

craigcomstock (approved):

PR #5147, opened 2023-01-02
ENT-9380: Created new policy function isreadable

larsewi:
Merge with:
- https://github.com/cfengine/documentation/pull/2951

TODO:
 - [x] Tested on Windows
 - [x] Tested on NFS
 - [x] Tested by adding `sleep(1000)` in the thread routine (just after attempting to read the file) to test that timeouts works as expected

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -22,6 +22,7 @@
   included file COSL.txt.
 */
 
+#include <platform.h>
 #include <evalfunction.h>
 
 #include <policy_server.h>
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -73,6 +73,7 @@
 #include <string_sequence.h>
 #include <string_lib.h>
 #include <version_comparison.h>
+#include <mutex.h>          /* ThreadWait */
 
 #include <math_eval.h>
 
@@ -8451,6 +8452,212 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const 
 
 /*********************************************************************/
 
+struct IsReadableThreadData
+{
+    pthread_t thread;
+    pthread_cond_t cond;
+    pthread_mutex_t mutex;
+    const char *path;
+    FnCallResult result;
+};
+
+static void *IsReadableThreadRoutine(void *data)
+{
+    assert(data != NULL);
+
+    struct IsReadableThreadData *const thread_data = data;
+
+    // Give main thread time to call pthread_cond_timedwait(3)
+    int ret = pthread_mutex_lock(&thread_data->mutex);
+    if (ret != 0)
+    {
+        ProgrammingError("Failed to lock mutex: %s",
+                         GetErrorStrFromCode(ret));
+    }
+
+    thread_data->result = FnReturnContext(false);
+
+    // Allow main thread to require lock on pthread_cond_timedwait(3)
+    ret = pthread_mutex_unlock(&thread_data->mutex);
+    if (ret != 0)
+    {
+        ProgrammingError("Failed to unlock mutex: %s",
+                         GetErrorStrFromCode(ret));
+    }
+
+    char buf[1];
+    const int fd = open(thread_data->path, O_RDONLY);
+    if (fd < 0) {
+        Log(LOG_LEVEL_DEBUG, "Failed to open file '%s': %s",
+            thread_data->path, GetErrorStr());
+    }
+    else if (read(fd, buf, sizeof(buf)) < 0)
+    {
+        Log(LOG_LEVEL_DEBUG, "Failed to read from file '%s': %s",
+            thread_data->path, GetErrorStr());
+        close(fd);
+    }
+    else
+    {
+        close(fd);
+        thread_data->result = FnReturnContext(true);
+    }
+
+    ret = pthread_cond_signal(&(thread_data->cond));
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to signal waiting thread: %s",
+            GetErrorStrFromCode(ret));
+    }
+
+    return NULL;
+}
+
+static FnCallResult FnCallIsReadable(ARG_UNUSED EvalContext *const ctx,
+                                     ARG_UNUSED const Policy *const policy,
+                                     const FnCall *const fp,
+                                     const Rlist *finalargs)
+{
+    assert(fp != NULL);
+    assert(fp->name != NULL);
+
+    if (finalargs == NULL)
+    {
+        Log(LOG_LEVEL_ERR, "Function %s requires path as first argument",
+            fp->name);
+        return FnFailure();
+    }
+    const char *const path = RlistScalarValue(finalargs);
+
+    long timeout = (finalargs->next == NULL) ? 3L /* default timeout */
+                 : IntFromString(RlistScalarValue(finalargs->next));
+    assert(timeout >= 0);
+
+    if (timeout == 0) // Try read in main thread, possibly blocking forever
+    {
+        Log(LOG_LEVEL_DEBUG,
+            "Checking if file '%s' is readable in main thread, "
+            "possibly blocking forever.", path);
+
+        char buf[1];
+        const int fd = open(path, O_RDONLY);
+        if (fd < 0)
+        {
+            Log(LOG_LEVEL_DEBUG, "Failed to open file '%s': %s", path,
+                GetErrorStr());
+            return FnReturnContext(false);
+        }
+        else if (read(fd, buf, sizeof(buf)) < 0)
+        {
+            Log(LOG_LEVEL_DEBUG, "Failed to read from file '%s': %s", path,
+                GetErrorStr());
+            close(fd);
+            return FnReturnContext(false);
+        }
+
+        close(fd);
+        return FnReturnContext(true);
+    }
+
+    // Else try read in separate thread, possibly blocking for N seconds
+    Log(LOG_LEVEL_DEBUG,
+        "Checking if file '%s' is readable in separate thread, "
+        "possibly blocking for %ld seconds.", path, timeout);
+
+    struct IsReadableThreadData thread_data = {0};
+    thread_data.path = path;
+
+    int ret = pthread_mutex_init(&thread_data.mutex, NULL);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to initialize mutex: %s",
+            GetErrorStrFromCode(ret));
+        return FnFailure();
+    }
+
+    ret = pthread_cond_init(&thread_data.cond, NULL);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to initialize condition: %s",
+            GetErrorStrFromCode(ret));
+        return FnFailure();
+    }
+
+    // Keep thread from doing its thing until we call
+    // pthread_cond_timedwait(3)
+    ret = pthread_mutex_lock(&thread_data.mutex);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to lock mutex: %s",
+            GetErrorStrFromCode(ret));
+        return FnFailure();
+    }
+
+    ret = pthread_create(&thread_data.thread, NULL, &IsReadableThreadRoutine,
+                         &thread_data);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to create thread: %s",
+            GetErrorStrFromCode(ret));
+        return FnFailure();
+    }
+
+    FnCallResult result;
+    // Wait on thread to finish or timeout
+    ret = ThreadWait(&thread_data.cond, &thread_data.mutex, timeout);
+    switch (ret)
+    {
+        case 0: // Thread finished in time
+            result = thread_data.result;
+            break;
+
+        case ETIMEDOUT: // Thread timed out
+            Log(LOG_LEVEL_DEBUG, "File '%s' is not readable: "
+                "Read operation timed out, exceeded %ld seconds.", path,
+                timeout);
+
+            ret = pthread_cancel(thread_data.thread);
+            if (ret != 0)
+            {
+                Log(LOG_LEVEL_ERR, "Failed to cancel thread");
+                return FnFailure();
+            }
+
+            result = FnReturnContext(false);
+            break;
+
+        default:
+            Log(LOG_LEVEL_ERR, "Failed to wait for condition: %s",
+                GetErrorStrFromCode(ret));
+            return FnFailure();
+
+    }
+
+    ret = pthread_mutex_unlock(&thread_data.mutex);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to unlock mutex");
+        return FnFailure();
+    }
+
+    void *status;
+    ret = pthread_join(thread_data.thread, &status);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to join thread");
+        return FnFailure();
+    }
+
+    if (status == PTHREAD_CANCELED)
+    {
+        Log(LOG_LEVEL_DEBUG, "Thread was canceled");
+    }
+
+    return result;
+}
+
+/*********************************************************************/
+
 static FnCallResult FnCallFindfilesUp(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const Policy *policy, const FnCall *fp, const Rlist *finalargs)
 {
     assert(fp != NULL);
@@ -9996,6 +10203,13 @@ static const FnCallArg FINDFILES_UP_ARGS[] =
     {NULL, CF_DATA_TYPE_NONE, NULL}
 };
 
+static const FnCallArg ISREADABLE_ARGS[] =
+{
+    {CF_ABSPATHRANGE, CF_DATA_TYPE_STRING, "Path to file"},
+    {CF_VALRANGE, CF_DATA_TYPE_INT, "Timeout interval"},
+    {NULL, CF_DATA_TYPE_NONE, NULL}
+};
+
 static const FnCallArg DATATYPE_ARGS[] =
 {
     {CF_ANYSTRING, CF_DATA_TYPE_STRING, "Variable identifier"},
@@ -10402,6 +10616,8 @@ const FnCallType CF_FNCALL_TYPES[] =
     FnCallTypeNew("findfiles_up", CF_DATA_TYPE_CONTAINER, FINDFILES_UP_ARGS, &FnCallFindfilesUp, "Find files matching a glob pattern by searching up the directory three from a given point in the tree structure",
                   FNCALL_OPTION_VARARG, FNCALL_CATEGORY_FILES, SYNTAX_STATUS_NORMAL),
     FnCallTypeNew("search_up", CF_DATA_TYPE_CONTAINER, FINDFILES_UP_ARGS, &FnCallFindfilesUp, "Hush... This is a super secret alias name for function 'findfiles_up'",
+                  FNCALL_OPTION_VARARG, FNCALL_CATEGORY_FILES, SYNTAX_STATUS_NORMAL),
+    FnCallTypeNew("isreadable", CF_DATA_TYPE_CONTEXT, ISREADABLE_ARGS, &FnCallIsReadable, "Check if file is readable. Timeout immediately or after optional timeout interval",
                   FNCALL_OPTION_VARARG, FNCALL_CATEGORY_FILES, SYNTAX_STATUS_NORMAL),
 
     // Datatype functions
--- a/examples/isreadable.cf
+++ b/examples/isreadable.cf
@@ -1,0 +1,25 @@
+#+begin_src prep
+#@ ```
+#@ echo "Hello CFEngine!" > /tmp/foo.txt
+#@ ```
+#+end_src
+###############################################################################
+#+begin_src cfengine3
+bundle agent __main__
+{
+  vars:
+      "filename"
+        string => "/tmp/foo.txt";
+      "timeout"
+        int => "3";
+  reports:
+      "File '$(filename)' is readable"
+        if => isreadable("$(filename)", "$(timeout)");
+}
+#+end_src
+###############################################################################
+#+begin_src example_output
+#@ ```
+#@ R: File '/tmp/foo.txt' is readable
+#@ ```
+#+end_src
--- a/tests/acceptance/01_vars/02_functions/isreadable.cf
+++ b/tests/acceptance/01_vars/02_functions/isreadable.cf
@@ -1,0 +1,113 @@
+##############################################################################
+#
+#  Test policy function isreadable
+#
+##############################################################################
+
+body common control
+{
+  inputs => { "../../default.cf.sub" };
+  bundlesequence  => { default("$(this.promise_filename)") };
+  version => "1.0";
+}
+
+##############################################################################
+
+bundle agent init
+{
+  files:
+      # Has content
+      "$(G.testfile)_1"
+        content => "Hello CFEngine!";
+      # No content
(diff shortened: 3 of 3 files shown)


larsewi:

libpromises/evalfunction.c
```diff
@@ -8450,6 +8451,84 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+        Log(LOG_LEVEL_ERR, "Function %s requires path as first argument",
+            fp->name);
+        return FnFailure();
+    }
+    const char *const path = RlistScalarValue(finalargs);
+
+    finalargs = finalargs->next;
+    const long timeout = (finalargs == NULL) ? 0L
+                       : IntFromString(RlistScalarValue(finalargs));
+    assert(timeout >= 0);
```
olehermanse, line 8473: What happens if you specify a negative value in policy?

larsewi: Then you will get the following error message:
```
   error: Fatal CFEngine error: In function 'filereadable', error in variable '-1', 'Integer is out of range'
```

olehermanse: Great!

libpromises/evalfunction.c
```diff
@@ -22,6 +22,7 @@
   included file COSL.txt.
 */
 
+#include<platform.h>
```
olehermanse, line 25:
```suggestion
#include <platform.h>
```

libpromises/evalfunction.c
larsewi: Also if the time limit for some reason would be negative, `select` would return `-1` and set `errno = EINVAL`

olehermanse (approved):
Looks good so far, please test on Windows and NFS :)

libpromises/evalfunction.c
```diff
@@ -8450,6 +8451,84 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+        return FnFailure();
+    }
+    const char *const path = RlistScalarValue(finalargs);
+
+    finalargs = finalargs->next;
+    const long timeout = (finalargs == NULL) ? 0L
+                       : IntFromString(RlistScalarValue(finalargs));
+    assert(timeout >= 0);
+
+    FILE *const file = fopen(path, "r");
```
vpodzime, line 8475: Generally I don't think using the file stream (`FILE`) API is a good choice. It's not needed and it hides certain details. Most importantly, AFAICT, even this or even `open()` can potentially hang forever. So applying the timeout only to the `select()` below is too late.

vpodzime (commented):
We should either not play the "timeout game" or the documentation should be very clear about it being just a best-effort thing.

libpromises/evalfunction.c
larsewi: Also, this did not work on Windows. And Windows does not have the `poll` function where I can listen for events. 
```
   debug: File C:\test1.txt is not readable: Could not poll event: An operation was attempted on something that is not a socket.
```
Any ideas on how to do this?

@vpodzime do you think it would be better to not block at all?

nickanderson (approved):
Tests seem reasonable. Would be nice to actually test inability to read in relation to permissions rather than file existence but I can't think of a good way to do so since we run as root and root can read files that are 000 even when selinux is enforcing.

libpromises/evalfunction.c
olehermanse: @larsewi @vpodzime I guess using `open` with `O_RDONLY` and `O_NONBLOCK` is what we want. Then you can only worry about timeout in `select()` - `open()`, `read()` and `close()` will be non-blocking.

vpodzime: Well, my `man:open(2)` says:
>              Note that this flag has no effect for regular files and block devices; that is, I/O operations will (briefly) block when device activity is required, regardless of whether O_NONBLOCK is  set.
>              Since O_NONBLOCK semantics might eventually be implemented, applications should not depend upon blocking behavior when specifying this flag for regular files and block devices.

Not sure how it works with regular files on NFS, for example.

vpodzime: > And Windows does not have the poll function where I can listen for events.

AFAICT, Windows doesn't have `poll()` and `select()` only works on sockets. We have some workarounds for that in the code already.

An option is to create a thread and try to `join()` it with a timeout and detach it if it fails.

libpromises/evalfunction.c
```diff
@@ -8450,6 +8451,191 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+    ret = pthread_cond_init(&(thread_info.cond), NULL);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to initialize condition: %s",
+            GetErrorStrFromCode(ret));
+        return FnFailure();
+    }
+
+    ret = pthread_cond_timedwait(&(thread_info.cond), &(thread_info.mutex),
+                                 &timespec);
```
vpodzime, line 8599: Can't we use `pthread_timedjoin_np()` or `pthread_tryjoin_np()` instead of the condition?

libpromises/evalfunction.c
```diff
@@ -8450,6 +8451,191 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+            return thread_info.result;
+        }
+
+    case ETIMEDOUT:
+        {
+            Log(LOG_LEVEL_DEBUG, "File '%s' is not readable: "
+                "Read operation timed out, exceeded %ld seconds.",
+                thread_info.path, timeout);
+
+            ret = pthread_detach(thread_info.thread);
```
vpodzime, line 8620: maybe `pthread_kill()` would not hurt here as well

vpodzime (commented):
Looks good to me otherwise. Some test on a slow NFS (or some slow FUSE FS) would be nice.

libpromises/evalfunction.c
larsewi: > These functions are nonstandard GNU extensions; hence the suffix  "_np"
       (nonportable) in the names.

Meaning they will probably not work on exotics?

libpromises/evalfunction.c
larsewi: > But note  also  that POSIX  says that an attempt to use a thread ID whose lifetime has ended produces undefined behavior, and an attempt to use an invalid thread ID in  a  call  to  pthread_kill()  can, for example, cause a segmentation fault.

This sounds kind of dangerous. Unless the thread is still considered alive until you call `pthread_join`? Maybe we can use `pthread_cancel` instead?

libpromises/evalfunction.c
```diff
@@ -22,6 +22,7 @@
   included file COSL.txt.
 */
 
+#include <platform.h>
```
larsewi, line 25: This header is no longer needed

larsewi:
@cf-bottom jenkins with exotics please :)

libpromises/evalfunction.c
```diff
@@ -8450,6 +8451,191 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+    pthread_cond_t cond;
+    pthread_mutex_t mutex;
+    const char *path;
+    FnCallResult result;
+};
+
+static void *ThreadRoutine(void *data)
+{
+    assert(data != NULL);
+    struct ThreadInfo *const thread_info = (struct ThreadInfo *) data;
```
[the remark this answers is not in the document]
vpodzime: No `s/thread_info/thread_data`

libpromises/evalfunction.c
vpodzime: Yes, but woth a try, I think.

libpromises/evalfunction.c
vpodzime: > Unless the thread is still considered alive until you call pthread_join? 

Yes, for a joinable thread, this should be the case. So `pthread_kill()` followed by `pthread_detach()` ought to be OK. :wink: `pthread_cancel()` is an alternative, but AFAICT `pthread_kill()` has the benefit of interrupting the potential hanging syscall (`open()`,...). But again, exotics... I cannot even 100% say that `pthread_kill()` won't kill the whole process there. :grin:

libpromises/evalfunction.c
vpodzime: AFAICT, this should be the first `#include` in all our sources. We have seen problems when it wasn't and I was even considering adding a GH action checking its presence.

larsewi: I'll add it as a separate commit ;)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/build-and-deploy-docs-master/73/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/build-and-deploy-docs-master/73/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-build-and-deploy-docs-master-73/

**Documentation:** http://buildcache.cfengine.com/packages/build-documentation-pr/jenkins-build-and-deploy-docs-master-73/output/_site/

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/build-and-deploy-docs-master/75/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/build-and-deploy-docs-master/75/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-build-and-deploy-docs-master-75/

**Documentation:** http://buildcache.cfengine.com/packages/build-documentation-pr/jenkins-build-and-deploy-docs-master-75/output/_site/

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/build-and-deploy-docs-master/76/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/build-and-deploy-docs-master/76/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-build-and-deploy-docs-master-76/

**Documentation:** http://buildcache.cfengine.com/packages/build-documentation-pr/jenkins-build-and-deploy-docs-master-76/output/_site/

larsewi:
Exot1cs passed on last test. I just changed the name `s/filereadable/isreadable` suggested by @nickanderson. I will run one more test Jenkins without exot1cs, just to make sure I did not break anything.

libpromises/evalfunction.c
larsewi: I don't think it's worth to spend more time on this

libpromises/evalfunction.c
larsewi: Used `pthread_cancel` alternative

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/build-and-deploy-docs-master/77/)

**Jenkins:** https://ci.cfengine.com/job/build-and-deploy-docs-master/77/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-build-and-deploy-docs-master-77/

**Documentation:** http://buildcache.cfengine.com/packages/build-documentation-pr/jenkins-build-and-deploy-docs-master-77/output/_site/

libpromises/evalfunction.c
```diff
@@ -8451,6 +8451,230 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+    pthread_mutex_t mutex;
+    const char *path;
+    FnCallResult result;
+};
+
+static void *IsReadableThreadRoutine(void *data)
+{
+    assert(data != NULL);
+    struct IsReadableThreadData *const thread_data =
+        (struct IsReadableThreadData *) data;
```
vpodzime, line 8467: You can just asign to the variable, the type casting will happen implicitly.

libpromises/evalfunction.c
```diff
@@ -8451,6 +8451,230 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+        success = false;
+    }
+    else if (read(fd, buf, sizeof(buf)) < 0)
+    {
+        Log(LOG_LEVEL_DEBUG, "Failed to read from file '%s': %s",
+            thread_data->path, GetErrorStr());
+        close(fd);
+        success = false;
+    }
+    else
```
vpodzime, line 8506: Is it a success when `read()` returns `0` here?

libpromises/evalfunction.c
```diff
@@ -8451,6 +8451,230 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+    ret = pthread_create(&(thread_data.thread), NULL,
+                         &IsReadableThreadRoutine, &thread_data);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to create thread: %s",
+            GetErrorStrFromCode(ret));
+        return FnFailure();
+    }
+
+    ret = pthread_cond_init(&(thread_data.cond), NULL);
```
vpodzime, line 8619: Shouldn't this be done before the thread is started?

libpromises/evalfunction.c
```diff
@@ -8451,6 +8451,230 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+
+            ret = pthread_cancel(thread_data.thread);
+            if (ret != 0)
+            {
+                Log(LOG_LEVEL_ERR, "Failed to cancel thread : %s",
+                    GetErrorStrFromCode(ret));
+                return FnFailure();
+            }
+
+            ret = pthread_join(thread_data.thread, NULL);
```
vpodzime, line 8658: Can't this hang forever if the thread is hanging and cannot be cancelled?

vpodzime (commented):
I can only repeat myself and once again say that I think it's nearly impossible to do make this work in a really reliable way. And as such, I would rather not promise users we can play the timeout game...

libpromises/evalfunction.c
larsewi: Depends on how you define if a file is readable.

> Upon reading end-of-file, zero is returned.
> - read(2)

To me, that means you've successfully read an empty file. But you can probably argue that an empty file is not readable.

libpromises/evalfunction.c
larsewi: Does not really matter. The thread is blocked on trying to acquire the mutex until after the condition is created. The thread uses the condition only after acquiring the mutex. Having `pthread_cond_init` and `pthread_cond_timedwait` consecutively makes the code more readable in my opinion.

libpromises/evalfunction.c
vpodzime: Yeah, that's why I asked that question :sweat_smile:

libpromises/evalfunction.c
vpodzime: >  Does not work with pthread_detach either. 

What if you create the thread as detached? We don't need to `join()` it, right? It either signals the condition in time or we don't care about it.

libpromises/evalfunction.c
olehermanse: I agree with Lars, an empty file is readable when you read it and get to EOF.

larsewi:
Tried creating detached thread. The agent blocks on `pthread_cond_timedwait` forever, even though it should just wait for 1 sec. I'm not sure what I'm doing wrong. @olehermanse @vpodzime  any ideas?
```
$ cat > ~/test.cf 
bundle agent test
{
    reports:
        "File is readable"
            if => isreadable("/tmp/testfile", "1");
}
$ sudo /var/cfengine/bin/cf-agent -KIf ~/test.cf -b test
    info: MAIN: Init mutex
    info: MAIN: Init condition
    info: MAIN: Lock mutex
    info: MAIN: Init attr
    info: MAIN: Set detach state
    info: MAIN: Create thread
    info: MAIN: Destroy attr
    info: MAIN: Wait condition (timeout 1 sec)
    info: THREAD: Lock mutex
    info: THREAD: Open/Read file
    info: THREAD: Sleep
```

vpodzime:
```
info: MAIN: Lock mutex
```
The `pthread_cond_timedwait()` needs to lock the mutex when the condition is signaled. So it cannot be locked all the time.

Also, you can use `ThreadWait()` from *libutils/mutex.h*.

larsewi:
@cf-bottom jenkins with exotics please :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/build-and-deploy-docs-master/78/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/build-and-deploy-docs-master/78/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-build-and-deploy-docs-master-78/

**Documentation:** http://buildcache.cfengine.com/packages/build-documentation-pr/jenkins-build-and-deploy-docs-master-78/output/_site/

olehermanse (dismissed):
LGTM, but wait for @vpodzime as well :)

libpromises/evalfunction.c
```diff
@@ -8450,6 +8451,222 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+
+    struct timespec ts = {0};
+    if (clock_gettime(CLOCK_REALTIME, &ts) != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to get clock real time: %s", GetErrorStr());
+        return FnFailure();
+    }
+    ts.tv_sec += timeout;
+
+    // Wait on thread to finnish or timeout
```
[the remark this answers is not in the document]
vpodzime: I think waiting for it to build a sauna and drink some Vodka would be reasonable!

larsewi:
I was able to make this work with canceling thread on timeout. This way we ensure there will not be any dangling threads after returning from `FnCallIsReadable()`. @cf-bottom can you please test that this also works in jenkins with exotics :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/build-and-deploy-docs-master/79/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/build-and-deploy-docs-master/79/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-build-and-deploy-docs-master-79/

**Documentation:** http://buildcache.cfengine.com/packages/build-documentation-pr/jenkins-build-and-deploy-docs-master-79/output/_site/

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/build-and-deploy-docs-master/80/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/build-and-deploy-docs-master/80/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-build-and-deploy-docs-master-80/

**Documentation:** http://buildcache.cfengine.com/packages/build-documentation-pr/jenkins-build-and-deploy-docs-master-80/output/_site/

libpromises/evalfunction.c
```diff
@@ -8451,6 +8452,231 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+    assert(data != NULL);
+
+    struct IsReadableThreadData *const thread_data = data;
+
+    int ret = pthread_mutex_lock(&thread_data->mutex);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to lock mutex: %s",
+            GetErrorStrFromCode(ret));
+        return NULL;
```
vpodzime, line 8475: This should never happen, right? I'm almost inclined to do something like `ProgrammingError()` here.

libpromises/evalfunction.c
```diff
@@ -8451,6 +8452,231 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+    }
+
+    thread_data->result = FnFailure();
+
+    ret = pthread_mutex_unlock(&thread_data->mutex);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to unlock mutex: %s",
+            GetErrorStrFromCode(ret));
+        return NULL;
```
vpodzime, line 8485: Same here.

libpromises/evalfunction.c
```diff
@@ -8451,6 +8452,231 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+    FnCallResult result;
+};
+
+static void *IsReadableThreadRoutine(void *data)
+{
+    assert(data != NULL);
+
+    struct IsReadableThreadData *const thread_data = data;
+
+    int ret = pthread_mutex_lock(&thread_data->mutex);
```
vpodzime, line 8470: Please add a comment explaining why this lock-unlock is needed.

libpromises/evalfunction.c
```diff
@@ -8451,6 +8452,231 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+            thread_data->path, GetErrorStr());
+        close(fd);
+    }
+    else
+    {
+        close(fd);
+        result = true;
+    }
+
+    ret = pthread_mutex_lock(&thread_data->mutex);
```
vpodzime, line 8507: Why do we need this second lock-unlock?

libpromises/evalfunction.c
```diff
@@ -8451,6 +8452,231 @@ static FnCallResult FnCallNetworkConnections(EvalContext *ctx, ARG_UNUSED const
...
+
+    ret = pthread_mutex_unlock(&thread_data.mutex);
+    if (ret != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to unlock mutex");
+        return FnFailure();
+    }
+
+    void *status;
+    ret = pthread_join(thread_data.thread, &status);
```
vpodzime, line 8663: What about using `pthread_tryjoin_np()` on platforms that have it?

vpodzime (commented):
Nice! Thanks for your patience working on this! It looks really promising now. But I would still not bet my life on it.

libpromises/evalfunction.c
larsewi: We don't need it. I was initially thinking there could be a race condition where the main thread reads the result while it is being assigned. But the main thread only reads the variable upon receiving a signal on the condition.

larsewi:

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/build-and-deploy-docs-master/81/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/build-and-deploy-docs-master/81/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-build-and-deploy-docs-master-81/

**Documentation:** http://buildcache.cfengine.com/packages/build-documentation-pr/jenkins-build-and-deploy-docs-master-81/output/_site/

olehermanse (changes requested):
Looks like it's not working; it's failing to join thread in GH Action.

libpromises/evalfunction.c
larsewi: If thread does not have time to exit before we call `pthread_tryjoin_np()` it will leave us with a zombie thread.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/build-and-deploy-docs-master/82/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/build-and-deploy-docs-master/82/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-build-and-deploy-docs-master-82/

**Documentation:** http://buildcache.cfengine.com/packages/build-documentation-pr/jenkins-build-and-deploy-docs-master-82/output/_site/

olehermanse (approved):
LGTM :) Green as well 🚀

libpromises/evalfunction.c
vpodzime: You'd have to detach the thread in such cases.

vpodzime: Yes, absolutely.

PR #5153, opened 2023-01-04
CFE-4127: Updated getusers() example to illustrate the filtering capabilities

nickanderson:


examples/getusers.cf
```diff
@@ -29,6 +29,21 @@ body common control
 bundle agent example
 {
   vars:
+
+      # To get users with a uid 1000 and greater we generate a list of uids from
```
craigcomstock, line 33: I would add in here near the beginning an explanation of the arguments and how they are filters.

Something like:

```
The getusers function takes two filtering arguments: exclude_names and exclude_ids, both a comma separated list of usernames and user IDs respectively.
```

craigcomstock (approved):

PR #5172, opened 2023-01-27
CFE-4023: Fixed --with-libxml2=no case in configure.ac

craigcomstock:
The CF3_WITH_LIBRARY and AC_CHECK_HEADERS were moved to outside of the check for with-libxml2=no

Ticket: CFE-4023
Changelog: title

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -659,19 +659,20 @@ if test "x$with_libxml2" != "xno"; then
             LIBXML2_CPPFLAGS=-I$with_libxml2/include/libxml2
         fi
     fi
-fi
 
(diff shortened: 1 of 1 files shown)


Lex-2008 (approved):
Note to other reviewers: Hide whitespace changes in review settings (a cogwheel with aria-label="Diff settings") to see that the change in just moving `fi` a few lines down, to include `CF3_WITH_LIBRARY(libxml2,...` block.

craigcomstock:
cherry pick including the libntech submodule update https://github.com/cfengine/core/pull/5174

PR #5175, opened 2023-02-01
Allow the cfapache user rwx access to the ${WORKDIR}/reports directory

vpodzime:
So that it can store reports there and read them.

Ticket: ENT-9825
Changelog: The cfapache user is now allowed to use the
		   /var/cfengine/reports directory due to ACL rules.

Signed-off-by: Vratislav Podzimek <vratislav.podzimek@northern.tech>

Merge together:
https://github.com/cfengine/core/pull/5175
https://github.com/cfengine/nova/pull/1977


vpodzime:
Turns out this was too ambitious.

PR #5176, opened 2023-02-03
CFE-4146: Defined feature class built_with_enable_fhs when compiled with --enable-fhs

nickanderson:
Naive attempt at having a class defined when a binary is built with FHS is useful for dealing
with different expectations.

Ticket: CFE-4146
Changelog: Title

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -189,7 +189,8 @@ dnl ######################################################################
 AC_ARG_ENABLE([fhs],
         [AS_HELP_STRING([--enable-fhs], [Enable FHS compliance. Defaults to custom CFEngine files layout])])
 
-#
+AM_CONDITIONAL(HAVE_FHS, [test x"$enable_fhs" = xyes])
+
 # pkglibdir/pkgdatadir are not overridable, so use our own invention instead.
 #
 
--- a/libpromises/feature.c
+++ b/libpromises/feature.c
@@ -5,6 +5,9 @@
 #include <buffer.h>
 
 static const char* features[] = {
+#ifdef HAVE_FHS
+    "built_with_enable_fhs",
+#endif
 #ifdef HAVE_LIBYAML
     "yaml",
 #endif


olehermanse (approved):
LGTM but please test that it actually works :)

configure.ac
```diff
@@ -189,7 +189,8 @@ dnl ######################################################################
 AC_ARG_ENABLE([fhs],
         [AS_HELP_STRING([--enable-fhs], [Enable FHS compliance. Defaults to custom CFEngine files layout])])
 
-#
+AM_CONDITIONAL(HAVE_FHS, [test x"$enable_fhs" = xyes])
```
craigcomstock, line 192: If you look at `--enable-debug` you will see that `AC_ARG_ENABLE` takes a first arg of the feature and `--enable-<feature>` is how it works and it sets a variable of just the feature, e.g. `debug` or in this case `fhs`. So a small adjustment fixes it :+1:

see https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Package-Options.html

```suggestion
AM_CONDITIONAL(HAVE_FHS, [test x"$fhs" = xyes])
```

Well, that still doesn't work for me, but seems right in config status/log: ```vagrant@bullseye:~/core$ grep -i FHS config.log 
  $ ./configure --enable-maintainer-mode --enable-debug --enable-fhs
configure:26165: running /bin/bash ./configure --disable-option-checking '--prefix=/usr/local'  '--enable-maintainer-mode' '--enable-debug' '--enable-fhs' --cache-file=/dev/null --srcdir=.
HAVE_FHS_FALSE=''
HAVE_FHS_TRUE='#'
vagrant@bullseye:~/core$ grep -i FHS config.status
ac_cs_config="'--enable-maintainer-mode' '--enable-debug' '--enable-fhs'"
  set X /bin/bash './configure'  '--enable-maintainer-mode' '--enable-debug' '--enable-fhs' $ac_configure_extra_args --no-create --no-recursion
S["HAVE_FHS_FALSE"]=""
S["HAVE_FHS_TRUE"]="#"```

so until next time...

nickanderson:
This is beyond me and we haven't revisited so closing.

PR #5198, opened 2023-03-28
ENT-9786: Added native core dump handling in cf-support for solaris

craigcomstock:
Ticket: ENT-9786
Changelog: title

merge together
https://github.com/cfengine/core/pull/5198
https://github.com/cfengine/system-testing/pull/485

### DIFF
--- a/misc/cf-support
+++ b/misc/cf-support
@@ -126,6 +126,24 @@ make_temp_dir()
   fi
 }
 
+
+# determine operating system flavor
+if command -v swlist 2>/dev/null; then
+  OS="hpux"
+elif command -v oslevel 2>/dev/null; then 
+  OS=aix
+elif [ -f /etc/release ]; then
+  OS=solaris
+  OS_VERSION=$(uname -r)
+elif [ -f /etc/redhat-release ] || [ -f /etc/os-release ] || [ -f /etc/lsb-release ]; then
+  OS=linux
+elif command -v system_profiler 2>/dev/null; then
+  OS=macos
+else
+  echo "unable to determine operating system, will try generic unix commands."
+  OS=unix
+fi
+
 tmpdir="`make_temp_dir`/$collection"
 export tmpdir
 mkdir -p "$tmpdir"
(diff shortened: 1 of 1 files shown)


cf-bottom:

nickanderson (approved):
Seems ok to me.

craigcomstock:
cherry picked to 3.18.x: https://github.com/cfengine/core/pull/5203 and 3.21.x: https://github.com/cfengine/core/pull/5201

misc/cf-support
```diff
@@ -167,15 +167,21 @@ else
     # core: ELF 64-bit LSB core file, x86-64, version 1 (SYSV), SVR4-style, from '/var/cfengine/bin/cf-key', real uid: 0, effective uid: 0, realgid: 0, effective gid: 0, execfn: '/var/cfengine/bin/cf-key', platform: 'x86_64'
     cf_core_files=`find "$response" -name 'core*' -type f -exec file {} \; 2>/dev/null | grep "core file" | grep "cf-" | cut -d' ' -f1 | sed 's/:$//'`
     if [ -n "$cf_core_files" ]; then
-      if ! command -v gdb >/dev/null; then
-        echo "Please install gdb. This is required in order to analyze core dumps."
-        exit 1
+      if [ "$OS" != "solaris" ]; then
+        if ! command -v gdb >/dev/null; then
+          echo "Please install gdb. This is required in order to analyze core dumps."
+          exit 1
```
vpodzime, line 191: Shouldn't we check if `pstack` is available on Solaris?

vpodzime (commented):
Looks good to me otherwise.

misc/cf-support
craigcomstock: I suppose that does make sense. I don't know how to check if it is "built-in".

I tried `pkg search /usr/bin/pstack` but it errored out on the system I had access to. https://docs.oracle.com/cd/E23824_01/html/821-1451/gkunu.html

Searching the net it seems it is available in the core package so "should" always be there like dd, cat, etc.

http://pkg.oracle.com/solaris/release/en/search.shtml?token=pstack&action=Search

[Manifest](http://pkg.oracle.com/solaris/release/manifest/0/system%2Fcore-os@11.4%2C5.11-11.4.42.0.0.111.1%3A20211203T213412Z)

vpodzime: What about `command`, `which` or something like that? After all, we need to check if the binary is available, who cares which package it is coming from? It could have also been removed without the whole package being uninstalled.

PR #5209, opened 2023-04-12
ENT-9891: Symlink rpmvercmp only when needed 

Lex-2008:
Ticket: ENT-9891

Issue was that when "$RPMVERCMP" variable is not set, this command:

	ln "$RPMVERCMP" "$WORKDIR/bin"

was printing an unhelpful error message that it can't overwrite a
directory, thus polluting logs.

So the solution is to not run this command when this variable is not
set, a.k.a. when we don't need rpmvercmp.

### DIFF
--- a/tests/acceptance/testall
+++ b/tests/acceptance/testall
@@ -431,7 +431,10 @@ runtest() {
                 $LN_CMD "$CF_NET" "$WORKDIR/bin"
                 $LN_CMD "$CF_CHECK" "$WORKDIR/bin"
                 $LN_CMD "$CF_RUNAGENT" "$WORKDIR/bin"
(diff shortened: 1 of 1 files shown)


Lex-2008:
no noise build: https://github.com/cfengine/core/actions/runs/4732658642/jobs/8399189764

craigcomstock (changes requested):
The commit comment doesn't make any sense to me.

What are you fixing, why is it a problem, how did you fix it?

craigcomstock (approved):

Lex-2008:
cherry-picks: https://github.com/cfengine/core/pull/5226 and https://github.com/cfengine/core/pull/5227

PR #5212, opened 2023-04-12
cf_remote_demo.sh: Updated with new platforms and convenience changes

olehermanse:
- cf_remote_demo.sh: Commented out masterfiles deploy
- cf_remote_demo.sh: Removed CentOS 6 and RHEL 6
- cf_remote_demo.sh: Updated Debian 9 to Debian 10
- cf_remote_demo.sh: Updated hub to Ubuntu 22

### DIFF
--- a/docs/cf_remote_demo.sh
+++ b/docs/cf_remote_demo.sh
@@ -16,7 +16,7 @@ cf-remote spawn --platform rhel-7-x64 --count 1 --name rhel7 --role client
 
 cf-remote --version master install --demo --bootstrap hub --hub hub
 
-cf-remote deploy --hub hub "$SCRIPTPATH/../../masterfiles"
+# cf-remote deploy --hub hub "$SCRIPTPATH/../../masterfiles"
 
 cf-remote --version master install --demo --bootstrap hub --clients ubuntu18
 cf-remote --version master install --demo --bootstrap hub --clients ubuntu16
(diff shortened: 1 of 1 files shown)


docs/cf_remote_demo.sh
```diff
@@ -5,27 +5,21 @@ set -x
...
 cf-remote spawn --platform centos-7-x64 --count 1 --name centos7 --role client
-cf-remote spawn --platform debian-9-x64 --count 1 --name debian9 --role client
-cf-remote spawn --platform rhel-6-x64 --count 1 --name rhel6 --role client
+cf-remote spawn --platform debian-10-x64 --count 1 --name debian10 --role client
 cf-remote spawn --platform rhel-7-x64 --count 1 --name rhel7 --role client
 
 cf-remote --version master install --demo --bootstrap hub --hub hub
 
-cf-remote deploy --hub hub "$SCRIPTPATH/../../masterfiles"
+# cf-remote deploy --hub hub "$SCRIPTPATH/../../masterfiles"
```
craigcomstock, line 17: why comment this out? why not just remove it?

olehermanse: The commit message explains why it's being commented out.

I am not removing it completely because it's useful for me to sometimes deploy the masterfiles i have in my masterfiles folder. For example if I am testing a new masterfiles policy change. So then I remove those 2 characters. I quite often make small edits to this script before running it (for example changing a count, a platform, or the `--version` argument). All of these things could be put in command line arguments, of course, but I don't think it's really necessary / advantageous to do so. Now it's mostly just a list of copy-pastable shell commands, nothing too complicated.

vpodzime (approved):
ACK

craigcomstock (approved):

PR #5226, opened 2023-04-25
Symlink rpmvercmp only when needed (3.21)

Lex-2008:
Ticket: ENT-9891

Issue was that when "$RPMVERCMP" variable is not set, this command:

	ln "$RPMVERCMP" "$WORKDIR/bin"

was printing an unhelpful error message that it can't overwrite a directory, thus polluting logs.

So the solution is to not run this command when this variable is not set, a.k.a. when we don't need rpmvercmp.

(cherry picked from commit 3cec91067a300bf721610a08eb3909b41439a434)


olehermanse (approved):

craigcomstock (approved):

nickanderson:
@Lex-2008 LMK if you need someone to merge this, otherwise looks like it's approved and ready to be merged.

PR #5228, opened 2023-04-26
Trying to get rid of extra pre blocks rendered around example output

nickanderson:


nickanderson:
Nope. Removing the triple backticks from inside the example output block result in this rendering:
image

nickanderson:
This change didn't have any significant effect. Closing PR.

PR #5243, opened 2023-05-19
ENT-7725: Variables & Classes Modules automatically tagged

larsewi:
According to the docs; all variables and classes will be tagged with `source=module`. This was apparently only the case when specifying tags using `^meta=Tag1,Tag2`. Now this tag is added all the time, even when you don't specify tags.

Furthermore, we added an additional meta tag `derived_from=<SOME_COMMAND>` which holds the path to the module script when using the `usemodule()` policy function or the command when using a commands promise with `module => "true"`.

Example output when using both `usemodule()` and commands promise without specifying meta tags:
```
$ sudo cf-agent -KIf ~/test.cf 
    info: Executing 'no timeout' ... '/bin/echo "=bogus=Hello CFEngine!"'
    info: Completed execution of '/bin/echo "=bogus=Hello CFEngine!"'
R: getvariablemetatags(doofus.doofus) => '{ "source=module", "derived_from=\"/var/cfengine/modules/doofus\" " }'
R: getvariablemetatags(echo.bogus) => '{ "source=module", "derived_from=/bin/echo \"=bogus=Hello CFEngine!\"" }'
```

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -8817,6 +8817,9 @@ void ModuleProtocol(EvalContext *ctx, const char *command, const char *line, int
 {
     assert(tags);
 
+    StringSetAdd(tags, xstrdup("source=module"));
+    StringSetAddF(tags, "derived_from=%s", command);
+
     // see the sscanf() limit below
     if(context_size < 51)
     {
@@ -8879,6 +8882,7 @@ void ModuleProtocol(EvalContext *ctx, const char *command, const char *line, int
 
             StringSetAddSplit(tags, content, ',');
             StringSetAdd(tags, xstrdup("source=module"));
+            StringSetAddF(tags, "derived_from=%s", command);
         }
         else if (sscanf(line + 1, "persistence=%ld", persistence) == 1)
         {
--- a/tests/acceptance/08_commands/01_modules/set-tags.cf
+++ b/tests/acceptance/08_commands/01_modules/set-tags.cf
@@ -62,10 +62,10 @@ bundle agent check
       "jtags2" string => join(",", "tags2");
       "jtags3" string => join(",", "tags3");
 
-      "etags0" string => "xyz,abc=def,,??what is this??,source=module";
-      "etags1" string => "xyz,abc=def,,??what is this??,source=module";
-      "etags2" string => "1,2,3,source=module";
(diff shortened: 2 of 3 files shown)


libpromises/evalfunction.c
```diff
@@ -8879,6 +8884,9 @@ void ModuleProtocol(EvalContext *ctx, const char *command, const char *line, int
 
             StringSetAddSplit(tags, content, ',');
             StringSetAdd(tags, xstrdup("source=module"));
```
vpodzime, line 8886: We can drop the above line now, right? I know it's adding to a set so the resulting set will be the same, but it's extra allocations and deallocations for no reason.

libpromises/evalfunction.c
```diff
@@ -8879,6 +8884,9 @@ void ModuleProtocol(EvalContext *ctx, const char *command, const char *line, int
...
+
+            xasprintf(&derived_from, "derived_from=%s", command);
+            StringSetAdd(tags, derived_from);
```
vpodzime, line 8889: Would be nice to have `StringSetAddF(tags, "derived_from=%s", command);`, but it's a PITA to add in the *libntech* submodule.

vpodzime (commented):
*which hold**s** the path...* in the commit message. Looks good to me otherwise.

libpromises/evalfunction.c
larsewi: I'm not sure I understand. A few lines above, the string set is cleared. Thus, we need to add the "default" tags again.

libpromises/evalfunction.c
larsewi: Indeed https://github.com/cfengine/libntech/pull/182

larsewi:

libpromises/evalfunction.c
vpodzime: I didn't see the `StringSetClear()` call in the diff. Sorry for the noise.

vpodzime (approved):
ACK

olehermanse (approved):
👍

larsewi:
@nickanderson @olehermanse Should this be cherry-picked to 3.21 and 3.18?

nickanderson:
My instinct is yes.

PR #5258, opened 2023-06-06
Fix building on solaris and hpux

Lex-2008:
I had to revert these three commits to make 3.18.x build on Solaris 10 and HP-UX. I believe they're not really necessary?

Failures were:

On Solaris 10 (both SPARC and x86):
```
00:16:20.675 xml-c14nize.c: In function ‘xmlC14nizeFile’:
00:16:21.104 xml-c14nize.c:25:37: error: ‘XML_C14N_1_0’ undeclared (first use in this function)
00:16:21.129      if (xmlC14NDocSaveTo(doc, NULL, XML_C14N_1_0, 0, true, out) < 0)
00:16:21.129                                      ^
```

on HP-UX:
```
00:10:41.199 checking for xmlFirstElementChild in -lxml2... ./configure[15954]: /var/cfengine/bin/xml2-config:  not found.
00:10:41.570 yes
00:10:41.596 checking libxml/xmlwriter.h usability... no
00:10:41.912 checking libxml/xmlwriter.h presence... no
00:10:42.147 checking for libxml/xmlwriter.h... no
00:10:42.147 configure: error: Cannot find libxml2
00:10:42.357 + [ xyes = xyes ]
```

followed by:

```
00:11:11.262 + cd core
00:11:11.287 + /usr/local/bin/gmake V=1 dist
00:11:11.287 gmake: *** No rule to make target 'dist'.  Stop.
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,4 @@
 3.18.5:
-	- The CF3_WITH_LIBRARY and AC_CHECK_HEADERS were moved to outside of the check for with-libxml2=no (CFE-4023)
-	- configure.ac: Enabled use of pkg-config to find libxml2 (CFE-4014)
 	- Added --help option to cf-support and aligned output with other
 	  components (ENT-9740)
 	- Added classes and vars to cf-support (CFE-4160)
--- a/configure.ac
+++ b/configure.ac
@@ -136,14 +136,6 @@ AC_CONFIG_LIBOBJ_DIR(libntech/libcompat)
 AC_PATH_PROG(GETCONF, getconf, false, $PATH:$prefix/bin:/usr/bin:/usr/local/bin:/sw/bin)
 AM_CONDITIONAL(CROSS_COMPILING, test "x$cross_compiling" = "xyes")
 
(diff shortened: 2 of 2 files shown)


craigcomstock:
Why are these changes not needed? They are improvements to how we build the software in various environments such as debian and yocto. 3.18.x lives until mid next year so still at least one more build. Granted, the only debian situation would be bullseye-backports which is at 3.18.2-1 https://packages.debian.org/bullseye-backports/cfengine3 and no version of yocto is at 3.18.x yet, currently at 3.15.x and I have plans/PRs ready to upgrade to 3.21.x.

craigcomstock (approved):
I'd rather have a reason why they are not needed or fix the issue, but I agree that this change is not required for this version that will be EOL next year and the distributions that need this change likely will never include 3.18.x or not for a long time.

Lex-2008:
yes, under "not needed" I meant "we don't support these versions of Debian and Yocto in this version branch".

I'm all in favour of investigating and fixing it properly, but I think that fixing it properly is not a blocker for a release. But having packages for these platforms is :)

PR #5278, opened 2023-06-20
ENT-10397: Added test for with attribute being namespace aware

nickanderson:


olehermanse (dismissed):
Looks like it works, please make the test soft fail :)

craigcomstock (approved):
Makes sense. The body file control which sets example_space should cause my_bundle.My_Variable to be in example_space instead of default namespace.

olehermanse:
> Makes sense. The body file control which sets example_space should cause my_bundle.My_Variable to be in example_space instead of default namespace.

Yea, the problems we are seeing are not really related to where the variable _is_ but rather how the lookup works. (When you try to expand a variable within `with`).

PR #5290, opened 2023-07-14
ENT-10428: Removed push event handling in github actions workflow

craigcomstock:
We don't have easy visibility on the results of these events which only occur after we merge pull requests.
We use /merge refs in pull requests so running the actions again on push events (after the merge) don't really provide any additional information.

Pushes/commits to pull requests are handled by the pull_request event so no change there.

See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push

Ticket: ENT-10428
Changelog: none

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,13 +6,6 @@ on:
   pull_request:
     branches: [ master, 3.21.x, 3.18.x ]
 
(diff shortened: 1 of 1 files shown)


nickanderson (approved):

vpodzime (approved):
ACK

PR #5311, opened 2023-08-25
ENT-10447: Removed references to travis (3.21)

craigcomstock:
- Removed references to travis


craigcomstock:
I re-disabled passopenfile_test, I guess it doesn't work well, sometimes? We'll see how it goes in master.

PR #5313, opened 2023-08-31
CFE-4244: Fix RlistEqual comparison on lists of different lengths

amousset:
The original issue that leads to this is the fact that the caches for execresult and execresult_as_data were mixed ([CFE-4244](https://northerntech.atlassian.net/browse/CFE-4244)). This is caused by two separate issues:

* The function cache uses the args list as key and discards the function itself. This means different functions with the same args are considered identical, and cache is reused.

https://github.com/cfengine/core/blob/57745c7afbd200a26b79a3a4dd79851781ff7ff1/libpromises/eval_context.c#L2840

* The function args are passed as an Rlist, and the comparison used ignores the additional items of the longest list when comparing two lists of different lengths, leading to treating execresult and execresult_as_data as identical when using the same command and shell args.

This PR only fixes the specific case of execresult, but leaves other function cache issues. For example, `host2ip` vs. `ip2host` could be confused:

```cfengine
bundle agent main {
vars:
  "res1" string => host2ip("cfengine.com");
  "res2" string => ip2host("cfengine.com");
  "res3" string => ip2host("northerntech.atlassian.net");

reports:
  "res1: ${res1}";
  "res2: ${res2}";
  "res3: ${res3}";
}
```
still returns:

```
R: res1: 34.107.174.45
R: res2: 34.107.174.45
R: res3: northerntech.atlassian.net
```

I'll try to add the function name to the Rlist used as cache key in a subsequent PR.

_Note_: Even if the function name "Equal" suggests lists of different lengths should be different, it might break other stuff using Rlists, I haven't tried to investigate other use cases.

[CFE-4244]: https://northerntech.atlassian.net/browse/CFE-4244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libpromises/rlist.c
+++ b/libpromises/rlist.c
@@ -1657,8 +1657,8 @@ bool RlistEqual(const Rlist *list1, const Rlist *list2)
             assert(rp1->val.item == NULL && rp2->val.item == NULL);
         }
     }
-
-    return true;
+    // return false if lengths are different
+    return (rp1 == NULL && rp2 == NULL);
 }
 
 bool RlistEqual_untyped(const void *list1, const void *list2)
--- a/tests/acceptance/01_vars/04_containers/execresult_and_as_data.cf
+++ b/tests/acceptance/01_vars/04_containers/execresult_and_as_data.cf
@@ -1,0 +1,41 @@
+#######################################################
+#
+# Test the ability to call the same command with execresult and execresult_as_data
+#
+#######################################################
+
+body common control
+{
+      inputs => { "../../default.cf.sub" };
+      bundlesequence  => { default("$(this.promise_filename)") };
+      version => "1.0";
+}
+
+#######################################################
+
+
+bundle agent init
+{
+}
+
+#######################################################
+
+bundle agent test
+{
+  vars:
+    !windows::
+     "res1" string => execresult("echo test", "useshell");
+     "res2" data   => execresult_as_data("echo test", "useshell", "stdout");
+    windows::
+     "res1" string => execresult("echo test", "powershell");
+     "res2" data   => execresult_as_data("echo test", "powershell", "stdout");
+}
+
+
+#######################################################
+
+bundle agent check
+{
+  methods:
+      "any" usebundle => dcs_check_strcmp("${test.res1}", "${test.res2[output]}", "$(this.promise_filename)", "no");
+}
--- a/tests/unit/rlist_test.c
+++ b/tests/unit/rlist_test.c
@@ -31,6 +31,23 @@ static void test_length(void)
     RlistDestroy(list);
 }
 
+static void test_equality(void)
+{
+    Rlist *list1 = RlistFromSplitString("a,b,c", ',');
+    Rlist *list2 = RlistFromSplitString("a,b,c", ',');
+    Rlist *list3 = RlistFromSplitString("z,b,c", ',');
+    Rlist *list4 = RlistFromSplitString("a,b,c,d", ',');
+
+    assert_true(RlistEqual(list1, list2));
+    assert_false(RlistEqual(list1, list3));
+    assert_false(RlistEqual(list1, list4));
+
+    RlistDestroy(list1);
+    RlistDestroy(list2);
+    RlistDestroy(list3);
+    RlistDestroy(list4);
+}
+
 static void test_prepend_scalar_idempotent(void)
 {
     Rlist *list = NULL;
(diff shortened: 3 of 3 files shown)


libpromises/rlist.c
```diff
@@ -1657,8 +1657,12 @@ bool RlistEqual(const Rlist *list1, const Rlist *list2)
...
         }
     }
-
-    return true;
+    if (rp1 == NULL && rp2 == NULL) {
+        return true;
+    } else {
+        // Lengths are different
+        return false;
+    }
```
larsewi, line 1665: Could be written as a single statement
```suggestion
    // return false if lengths are different
    return (rp1 == NULL && rp2 == NULL);
```

tests/unit/rlist_test.c
```diff
@@ -31,6 +32,25 @@ static void test_length(void)
...
+    assert_false(RlistEqual(list1, list4));
+
+    RlistDestroy(list1);
+    RlistDestroy(list2);
+    RlistDestroy(list3);
+    RlistDestroy(list4);
+}
+
+
+
```
larsewi, line 53:
```suggestion

```

larsewi (changes requested):
Very nice 🚀 I have only a few nitpicks.

Thank you for providing both an acceptance- and unit tests. I see that your acceptance test fails on Windows, and I'm not quite sure why. I will request a build from @cf-bottom in Jenkins, so that I can investigate this further. Feel free to address the suggested changes in the meanwhile.

amousset:
>  I see that your acceptance test fails on Windows, and I'm not quite sure why.

I you think it would be acceptable I could skip this test on Windows and macOS.

larsewi:
Hi @amousset, I just tested this manually on Windows, and you can fix it my using `"powershell"` instead of `"useshell"` as the second argument of `execresult()` and `execresult_as_data()` on Windows. You can achieve this by for example using the `windows::` classguard.

tests/acceptance/01_vars/04_containers/execresult_and_as_data.cf
```diff
@@ -0,0 +1,40 @@
...
+#######################################################
+
+bundle agent test
+{
+  vars:
+     "res1" string => execresult("echo test", "useshell");
+    !windows::
+     "res2" data   => execresult_as_data("echo test", "useshell", "stdout");
+    windows::
+     "res2" data   => execresult_as_data("echo test", "powershell", "stdout");
```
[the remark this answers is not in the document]
amousset: indeed :facepalm:

larsewi:
Awesome, thanks 🚀  Let's run this though CI again, just to make sure everything is good [Build Status](https://ci.cfengine.com/job/pr-pipeline/9743/)

amousset:
Second part addressed in https://github.com/cfengine/core/pull/5317

larsewi:
Could you please squash the commits as described in https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#interactive-rebase?

tests/acceptance/01_vars/04_containers/execresult_and_as_data.cf
larsewi, line 30: Looks like the change disappeared after your force push
```suggestion
    !windows::
     "res1" string => execresult("echo test", "useshell");
     "res2" data   => execresult_as_data("echo test", "useshell", "stdout");
    windows::
     "res1" string => execresult("echo test", "powersell");
     "res2" data   => execresult_as_data("echo test", "powershell", "stdout");
```

larsewi:
Still failing on Windows 🤔  I will create a new build, investigate further, and then get back to you. [Build Status](https://ci.cfengine.com/job/pr-pipeline/9798/)

tests/acceptance/01_vars/04_containers/execresult_and_as_data.cf
```diff
@@ -0,0 +1,41 @@
...
+#######################################################
+
+bundle agent test
+{
+  vars:
+    !windows::
+     "res1" string => execresult("echo test", "useshell");
+     "res2" data   => execresult_as_data("echo test", "useshell", "stdout");
+    windows::
+     "res1" string => execresult("echo test", "powersell");
```
larsewi, line 30: I think I found it. Little typo
```suggestion
     "res1" string => execresult("echo test", "powershell");
```

larsewi:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/9800/) 🤞

larsewi (approved):
Thanks for your contribution 🚀

larsewi:
Seems that the GitHub Action workflow for windows acceptance tests, tests with master nightly build, instead of testing it with the changes you did to core. Thus I merged this, and will see if it passes when we have a new master nightly build.

PR #5320, opened 2023-09-13
CFE-4244: Fix RlistEqual comparison on lists of different lengths (3.21.x)

larsewi:
The original issue that leads to this is the fact that the cache
for execresult and execresult_as_data were mixed. This is caused by two
separate issues:

* The function cache uses the args list as key and discards the function
  itself. This means different function with the same args are considered
  identical, and cache is reused.
* The args are passed as an Rlist, and the used Rlist comparison ignores the
  additional items of the longest list when comparing two lists of
  different lengths, leading to treating execresult and execresult_as_data
  as identical when using the same command and shell args.

This PR only fixes the specific case of execresult, but leaves other
function cache issues (e.g. host2ip vs. ip2host could be confused).

(cherry picked from commit d5a7372bcef9a63dd53f58bd3951f8392ad21e00)

### DIFF
--- a/libpromises/rlist.c
+++ b/libpromises/rlist.c
@@ -1657,8 +1657,8 @@ bool RlistEqual(const Rlist *list1, const Rlist *list2)
             assert(rp1->val.item == NULL && rp2->val.item == NULL);
         }
     }
-
-    return true;
+    // return false if lengths are different
+    return (rp1 == NULL && rp2 == NULL);
 }
 
 bool RlistEqual_untyped(const void *list1, const void *list2)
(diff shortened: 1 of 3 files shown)


larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9815/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9815/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9815/

larsewi:
Cherry picked from https://github.com/cfengine/core/pull/5313

PR #5332, opened 2023-09-22
Do not #include libcrypto-compat

vpodzime:
We require OpenSSL 1.0.0 or newer so we don't need this compatibility code for older versions.

Merge together:
https://github.com/cfengine/core/pull/5332
https://github.com/cfengine/nova/pull/2124


larsewi (dismissed):
Nice 🚀

larsewi (approved):

PR #5337, opened 2023-09-28
ENT-10647: Added changelog entry for ENT-10647 fix (3.18)

craigcomstock:
Fixed and improved postgresql server state handling in package scriptlets

Ticket: ENT-10647
Changelog: none

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,6 @@
 3.18.6:
+	- Fixed and improved postgresql server state handling in package installer
+	  scriptlets (ENT-10647)
 	- Cached policy function results now take into account number of arguments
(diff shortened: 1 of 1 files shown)


ChangeLog
```diff
@@ -1,4 +1,6 @@
 3.18.6:
+	- Fixed and improved postgresql server state handling in package scriptlets
```
craigcomstock, line 2: maybe "package" should be changed to "installer"? @nickanderson @olehermanse @larsewi

larsewi: How about package installer scriptlets?

larsewi (approved):

PR #5342, opened 2023-10-09
Disable win-acceptance tests for now due to two issues

craigcomstock:
- cf-agent.exe is not always installed "in-time" ENT-10699
- job uses latest nightly instead of building the PR code ENT-10754

Ticket: none
Changelog: none


craigcomstock:
re-running valgrind test, tried an empty commit https://github.com/cfengine/core/pull/5343 and it passed.

larsewi (approved):

craigcomstock:
The win acceptance tests were not cherry picked to 3.18.x or 3.21.x so we are good with just this merge. 👍

PR #5346, opened 2023-10-09
ENT-10761: Fixed two acceptance tests which could only run once before failing

craigcomstock:
Their first run they change an original/source file and so on subsequent runs
the tests failed due to the original being incorrect.

Ticket: ENT-10761
Changelog: none


craigcomstock:
cherry picks
https://github.com/cfengine/core/pull/5347
https://github.com/cfengine/core/pull/5348

PR #5358, opened 2023-11-01
A round of SELinux policy updates and fixes

vpodzime:


nickanderson (dismissed):

misc/selinux/cfengine-enterprise.te
```diff
@@ -34,6 +34,7 @@ require {
 	type proc_t;
 	type proc_net_t;
 	type proc_xen_t;
+	type proc_security_t;
```
olehermanse, line 37: IMO, the changelog entry could be more specific / user-oriented. Some loose suggestions;

```
The CFEngine daemons on SELinux systems are now allowed to read security parameters in /proc/system/kernel through the type proc_security_t
```

```
Fixed SELinux errors relating to CFEngine daemons attempting to access `/etc/proc/sys/kernel`
```

misc/selinux/cfengine-enterprise.te
```diff
@@ -166,6 +166,7 @@ require {
 	class association { sendto recvfrom setcontext polmatch };
 	class security setsecparam;
 	class service { start stop status reload enable disable };
+	class system { module_request };
```
olehermanse, line 171: Good changelog entry (but it could also mention SELinux).

olehermanse (approved):
Looks reasonable to me, added some opinions on the changelog entries, but they are "low severity" :)

PR #5375, opened 2023-11-14
ENT-2044: Added acceptance test for IP-address classes

larsewi:
Merge together with https://github.com/cfengine/enterprise/pull/772


larsewi:
But I guess the machinery for leveraging test.description makes it worth it anyways?

olehermanse (approved):

olehermanse:
In addition to the speed argument, I like to not use the dcs stuff when the test is simple. Easier to run (just a policy file, no deps), easier to read / understand.

larsewi:
I'll keep it as is. Because it was really simple to test on Windows. I just copied the file over and ran it directly.

PR #5396, opened 2023-12-08
Do not finalize things before printing digest in cf-key

vpodzime:
Getting the digest needs crypto functions and finalization also
unloads crypto providers. Also, it just doesn't make sense.


vpodzime:
RHEL 7 hub only jenkins run: [Build Status](https://ci.cfengine.com/job/pr-pipeline/10145/)

larsewi (approved):

PR #5408, opened 2023-12-22
CFE-4300: Fixed cf-support call to cf-promises to collect all classes and vars

craigcomstock:
Had a --no-lock argument which is not supported so nothing was collected.

Ticket: CFE-4300
Changelog: title


vpodzime (approved):
ACK

PR #5425, opened 2024-01-11
CFE-4274: In configure.ac, force with-systemd-socket=no when with-systemd-service=no

craigcomstock:
Previously --with-systemd-socket is defaulted to true always.

If a system has libsystemd-dev and configure is called with --with-systemd-service=no the build would fail.

Ticket: CFE-4274
Changelog: none


cf-bottom:

craigcomstock:
I checked this with no changes on master and it seems broken, so investigating and will try to fix.

https://northerntech.atlassian.net/browse/CFE-4310

craigcomstock:
waiting on https://github.com/cfengine/core/pull/5427 to prove things are alright here as well.

configure.ac
```diff
@@ -1611,6 +1598,22 @@ else
...
 AC_SUBST([OS_ENVIRONMENT_PATH])
 
+dnl systemd-socket activation
+
+AC_ARG_WITH([systemd-socket], [AS_HELP_STRING([--with-systemd-socket[[=PATH]]], [support systemd socket activation])], [], [with_systemd_socket=check])
+
+dnl ########################################################################
+dnl systemd socket feature, only available if systemd init scripts requested
+dnl ########################################################################
+if test "x$with_systemd_service" != xno && test "x$with_systemd_socket" != xno
```
craigcomstock, line 1608: moving this whole block AFTER with_systemd_service processing code allows us to include testing with_systemd_service and not check for systemd_socket if systemd_service is not requested.

larsewi (approved):

craigcomstock:
cherry picks:
https://github.com/cfengine/core/pull/5433
https://github.com/cfengine/core/pull/5432

PR #5440, opened 2024-01-29
Add options to skip loading augments and host-specific data to cf-agent

vpodzime:
Ticket: ENT-10792
Changelog: cf-agent has two new options --no-augments and
           --no-host-specific-data to skip loading augments (def.json)
           and host-specific data (host_specific.json), respectively


olehermanse (approved):
LGTM. The changelog entry could also mention how it affects `def_preferred.json`.

vpodzime:
> LGTM. The changelog entry could also mention how it affects `def_preferred.json`.

Added it next to `def.json`.

olehermanse (approved):
👍

PR #5449, opened 2024-02-15
CFE-3152: Fix building and using core on OpenBSD

craigcomstock:
- Adjusted INSTALL instructions for OpenBSD 7.4
- Adjusted openssl ifdefs to support building on OpenBSD

### DIFF
--- a/libpromises/crypto.c
+++ b/libpromises/crypto.c
@@ -887,11 +887,13 @@ static void SetupOpenSSLThreadLocks(void)
     }
 
 #ifndef __MINGW32__
+#if OPENSSL_VERSION_NUMBER < 0x10100000
     CRYPTO_set_id_callback((unsigned long (*)())ThreadId_callback);
 #endif
-    // This is a no-op macro for OpenSSL >= 1.1.0
-    // The callback function is not used (or defined) then
+#endif
+#if OPENSSL_VERSION_NUMBER < 0x10100000
     CRYPTO_set_locking_callback((void (*)())OpenSSLLock_callback);
+#endif
 }
 
 static void CleanupOpenSSLThreadLocks(void)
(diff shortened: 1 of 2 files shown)


libpromises/crypto.c
```diff
@@ -887,19 +887,23 @@ static void SetupOpenSSLThreadLocks(void)
...
     CRYPTO_set_locking_callback((void (*)())OpenSSLLock_callback);
+#endif
 }
 
 static void CleanupOpenSSLThreadLocks(void)
 {
     const int numLocks = CRYPTO_num_locks();
     CRYPTO_set_locking_callback(NULL);
 #ifndef __MINGW32__
+#if OPENSSL_VERSION_NUMBER < 0x10100000
```
SamuelMarks, line 904: Both can be rather than having nested if/defs
```c
#if !defined(__MINGW32__) && OPENSSL_VERSION_NUMBER < 0x10100000
```

Or even:
```c
#if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER < 0x10100000
```

craigcomstock: I do like separate lines to separate concerns here but certainly notice the two `#endif` without any comments of what they are ending looks bad.

Thanks for the suggestion!

craigcomstock:
@cf-bottom jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/10484/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/10484/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-10484/

larsewi (approved):

libpromises/crypto.c
craigcomstock: Looking at it some more the whole section could use some cleanup as we have the same if/ifndef conditions repeated several times. The endifs are so close I disagree with my intuition of adding comments there so will just go with this as-is. Again, @SamuelMarks , appreciate the suggestions and will keep them in mind. Maybe

```
#if defined(OPENSSL_VERSION_NUMBER)
```

is more descriptive than 

```
#ifndef __MINGW32__
```

But we don't currently support an SSL other than OpenSSL so 🤷 . Probably OK as-is.

craigcomstock:
cherry picks
https://github.com/cfengine/core/pull/5451
https://github.com/cfengine/core/pull/5450

PR #5453, opened 2024-02-22
Fixed incorrect sed command for cf-support filtering syslog output

craigcomstock:
sed -n '/cfe/p;/cf-/p' would cause any line matching both expressions to be printed twice

Ticket: CFE-4337
Changelog: none


nickanderson (approved):

craigcomstock:
trying it out on another fix which should always fail and cause cf-support to be called

[Build Status](https://ci.cfengine.com/job/pr-pipeline/10499/)

PR #5461, opened 2024-02-29
Update BSD.md, remove editor install

craigcomstock:
removed annoying opinionated editor install ;) make this easy to use for anyone and not install software that is not completely necessary. A user will have an editor installed already most likely.

### DIFF
--- a/docs/BSD.md
+++ b/docs/BSD.md
@@ -7,7 +7,7 @@ Tested on FreeBSD 12.1 VM in AWS EC2.
 ## Dependencies
 
 ```
-$ su root -c 'pkg update -f && pkg install -y gdb gcc gmake lmdb autoconf automake libtool git python3 emacs-nox'
+$ su root -c 'pkg update -f && pkg install -y gdb gcc gmake lmdb autoconf automake libtool git python3'
 ```
 
 ## Download


docs/BSD.md
```diff
@@ -7,7 +7,7 @@ Tested on FreeBSD 12.1 VM in AWS EC2.
 ## Dependencies
 
 ```
-$ su root -c 'pkg update -f && pkg install -y gdb gcc gmake lmdb autoconf automake libtool git python3 emacs-nox'
+$ su root -c 'pkg update -f && pkg install -y gdb gcc gmake lmdb autoconf automake libtool git python3'
```
nickanderson, line 10: LOL, BOOOOOOO

nickanderson (approved):

docs/BSD.md
craigcomstock: editor 🔥 merge this PR and the 🔥  is gone ;) and I don't have to edit the line when running it with `.w !sh` in `ed(1)` ;)

craigcomstock: while we're at it, we should remove the preceding `$ ` and `# ` from stuff like this... I am always having to delete it with `s/^$//` or `s/^#//` before running `.w !sh` :p

craigcomstock:
no need to cherry pick ;)

docs/BSD.md
olehermanse: @craigcomstock IMO this just makes it slightly more annoying. Whenever I tested something on BSD, I didn't have an editor, so I need to install something, and having this command ready helped :)

PR #5474, opened 2024-03-19
CFE-4341: Adjusted package module inventory to include quotes around fields when needed (3.18)

craigcomstock:
On FreeBSD where package versions can have commas such as joe-4.6,1 this was causing errors with the use of packagesmatching() function.

```
   error: Line from package inventory 'joe,4.6,1,amd64,pkg' did not yield correct number of elements.
```

Used CsvWriterField to quote the field in case it includes " , \r \n. See CsvWriterField() at https://github.com/NorthernTechHQ/libntech/blob/master/libutils/csv_writer.c#L73

Had to add a workaround to convert CRLF written by csv_writer to expected LF in resulting inventory_list.

Ticket: CFE-4341
Changelog: title

Co-authored-by: Lars Erik Wik <lars.erik.wik@northern.tech>
(cherry picked from commit 27fdfb5afd1ccd0cdefd9926a0129f0ca30563ec)

### DIFF
--- a/cf-agent/package_module.c
+++ b/cf-agent/package_module.c
@@ -39,6 +39,7 @@
 #include <eval_context.h>
 #include <changes_chroot.h>     /* RecordPkgOperationInChroot() */
 #include <simulate_mode.h>      /* CHROOT_PKG_OPERATION_* */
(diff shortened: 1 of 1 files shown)


cf-bottom:

PR #5477, opened 2024-03-27
Prevent systemctl from asking for password after installing unit files

vpodzime:
To prevent `make install` from hanging. The rule is fault-tolerant so if the authentication is required the command will fail and things will continue just fine. Authentication required means `make install` was not run neither with `sudo` nor as root so `systemctl daemon-reload` would not work anyway because this means the systemd units were not installed into the standard locations where only the root user has access to.


craigcomstock (approved):
nice use of options :+1:

vpodzime:
smoke test: [Build Status](https://ci.cfengine.com/job/pr-pipeline/)

PR #5480, opened 2024-04-10
ENT-11526: Changed cf-apache systemd unit to reload configuration gracefully

craigcomstock:
apachectl graceful doesn't close currently open connections.

Ticket: ENT-11526
Changelog: title

with https://github.com/cfengine/masterfiles/pull/2864

### DIFF
--- a/misc/systemd/cf-apache.service.in
+++ b/misc/systemd/cf-apache.service.in
@@ -10,6 +10,7 @@ PartOf=cfengine3.service
 Type=forking
 ExecStart=@workdir@/httpd/bin/apachectl start
 ExecStop=@workdir@/httpd/bin/apachectl stop
+ExecReload=@workdir@/httpd/bin/apachectl graceful
 PIDFile=@workdir@/httpd/httpd.pid
 Restart=always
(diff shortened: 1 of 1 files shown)


nickanderson (approved):
I commented in another PR to the MPF for a similar change with concern about aync vs sync. 

I think this is probably OK in a systemd unit where generally management calls have async behavior anyway.

vpodzime (approved):
Nice! I'm surprised we didn't have this. Just make sure this is also updated in MPF, we still have the service files in two places, AFAIK.

larsewi (approved):

craigcomstock:
had a typo in the service file of `@workdir` instead of required `@workdir@`. `@` at end was missing so service file didn't work properly, especially during upgrades and server restarts.

vpodzime (approved):
Still looks good to me :grin:

larsewi (approved):

craigcomstock:
Will wait to cherry pick for a week or so to see how/if this improves our CI. Will set a reminder for myself for 2024-04-19.

PR #5488, opened 2024-04-25
README.md: Simplified hello world example

olehermanse:
There's no need to try to introduce as much syntax as possible
in this example. Hello world can and should be simple.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ bundle agent main
 The following policy code may be executed with cf-agent (the main CFEngine binary) as follows.
 
 ```bash
-$ cf-agent/cf-agent hello.cf
+$ cf-agent/cf-agent ./hello.cf
 R: Hello, world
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,28 +48,24 @@ $ pwd
...
+}
+```
 
 The following policy code may be executed with cf-agent (the main CFEngine binary) as follows.
 
-    $ cf-agent/cf-agent hello.cf
-    R: Hello, world
-
+```bash
+$ cf-agent/cf-agent hello.cf
```
nickanderson, line 66:
```suggestion
$ cf-agent/cf-agent ./hello.cf
```

nickanderson (dismissed):
If you don't supply fullly qualified path, I think it looks for the file in inputs.

nickanderson (approved):

PR #5498, opened 2024-05-06
ENT-11765: Changed cf-execd's sleep behavior so it attempts to wake up at the beginning of every minute

olehermanse:
This could fix issues if things are slow and agent runs end up getting skipped if cf-execd wakes up too late.

TODO:

- [ ] Manually test

### DIFF
--- a/cf-execd/cf-execd.c
+++ b/cf-execd/cf-execd.c
@@ -434,6 +434,23 @@ void ThisAgentInit(void)
     umask(077);
 }
 
+// Return 2 to 62 seconds depending on how much time is left until the next
+// minute starts. Since cf-execd wakes up "every minute" to evaluate its
+// schedule, we want to do this at the start of the minute, to avoid
+// accidentally skipping any runs if things are slow.
+//
+// Why +2? Why not 0 to 60? Not a very good reason, but:
+// Sleep at least 2 seconds to avoid 2 agent runs very close together
+// Target waking up at :02 seconds, to reduce the chances of waking
+// up at :59 in the previous minute, and unintentionally having 2
+// agent runs in the same minute
+static inline time_t GetPulseTime()
+{
+    const time_t current_second = (time(NULL) % CFPULSETIME);
+    const time_t remaining_seconds = CFPULSETIME - current_second;
+    return remaining_seconds + 2;
+}
+
 /*****************************************************************************/
 
 
@@ -534,6 +551,7 @@ static bool HandleRequestsOrSleep(time_t seconds, const char *reason,
     return false;
 }
 
+// Non-windows version of main loop:
 static void CFExecdMainLoop(EvalContext *ctx, Policy **policy, GenericAgentConfig *config,
                             ExecdConfig **execd_config, ExecConfig **exec_config,
                             int runagent_socket)
@@ -564,7 +582,7 @@ static void CFExecdMainLoop(EvalContext *ctx, Policy **policy, GenericAgentConfi
             }
         }
         /* 1 Minute resolution is enough */
-        terminate = HandleRequestsOrSleep(CFPULSETIME, "pulse time", runagent_socket,
+        terminate = HandleRequestsOrSleep(GetPulseTime(), "pulse time", runagent_socket,
                                           (*execd_config)->local_run_command);
         if (terminate)
         {
@@ -699,6 +717,7 @@ static inline unsigned int MaybeSleepLog(LogLevel level, const char *msg_format,
     return sleep(seconds);
 }
 
+// Windows version of main loop:
 static void CFExecdMainLoop(EvalContext *ctx, Policy **policy, GenericAgentConfig *config,
                             ExecdConfig **execd_config, ExecConfig **exec_config,
                             ARG_UNUSED int runagent_socket)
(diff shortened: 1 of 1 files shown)


cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/10713/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/10713/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-10713/

olehermanse:
Backport: https://github.com/cfengine/core/pull/5499

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/10716/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/10716/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-10716/

craigcomstock (approved):
seems reasonable. I do think that two logs might be helpful
- verbose log about how long environment took to load
- warning log about environment load taking more than one minute

cf-execd/cf-execd.c
```diff
@@ -725,7 +736,7 @@ static void CFExecdMainLoop(EvalContext *ctx, Policy **policy, GenericAgentConfi
             }
         }
         /* 1 Minute resolution is enough */
-        MaybeSleepLog(LOG_LEVEL_VERBOSE, "Sleeping for pulse time %u seconds...", CFPULSETIME);
+        MaybeSleepLog(LOG_LEVEL_VERBOSE, "Sleeping for pulse time %u seconds...", GetPulseTime());
```
vpodzime, line 739: The function should only be called once and the result should be used both above and here in the log message. Otherwise the log message can be misleading mentioning  a different value than the code actually does.

cf-execd/cf-execd.c
```diff
@@ -434,6 +434,17 @@ void ThisAgentInit(void)
     umask(077);
 }
 
+// Return 2 to 62 seconds depending on how much time is left until the next
```
vpodzime, line 437: Why 2 to 62 and not the actual remaining number of seconds?

vpodzime (commented):
Looks good to me otherwise.

cf-execd/cf-execd.c
craigcomstock: It is only called once per version of the main loop: windows or !windows so seems OK to me.

cf-execd/cf-execd.c
```diff
@@ -434,6 +434,17 @@ void ThisAgentInit(void)
...
+// minute starts. Since cf-execd wakes up "every minute" to evaluate its
+// schedule, we want to do this at the start of the minute, to avoid
+// accidentally skipping any runs if things are slow.
+static inline time_t GetPulseTime()
+{
+    const time_t current_second = (time(NULL) % 60);
```
craigcomstock, line 443: @larsewi [mentioned](https://github.com/cfengine/core/pull/5499/files#r1592600540) using `CFPULSETIME` here instead of 60 but if we want to start at the beginning of every minute I would say remove `CFPULSETIME` entirely and keep these hard-coded 60s here.

cf-execd/cf-execd.c
nickanderson: I wondered why 2 to 62 as well.

cf-execd/cf-execd.c
olehermanse: What Craig said. This is not just a log message, this is the windows version of the execd loop. 

+1 point to @craigcomstock for getting it right. -2 points to @vpodzime for not understanding his own code from 3 years ago, and -1 points to @larsewi for a wrong thumbsup :)

I've added some extra comments ;)

cf-execd/cf-execd.c
olehermanse: Added more comments.

cf-execd/cf-execd.c
olehermanse: Okay, adjusted to use the constant.

larsewi (approved):

craigcomstock (approved):

cf-execd/cf-execd.c
vpodzime: It would be best to use the monotonic clock on platforms that have it.

cf-execd/cf-execd.c
vpodzime: Who cares about Windows?

cf-execd/cf-execd.c
vpodzime: This change actually doesn't make sense to me. If we want to determine which second in a minute the clock currently is, `% 60` is the right calculation. If `CFPULSETIME` is set to 120 or 180, its use in the calculation won't make sense.

vpodzime (approved):
Definitely an improvement.

nickanderson:
FWIW, word on the street is that this change has a positive impact.

cf-execd/cf-execd.c
olehermanse: That's true, it would give better guarantees about how long we actually sleep. But still it would be a bit awkward, since the time you wake up would maybe not match what is expected. (Same as now).

cf-execd/cf-execd.c
olehermanse: I thought a bit about that, but it does make some sense. If you define a 120 second pulse, it will tell you how many seconds is left of the current pulse, even though that feels less intuitive to think about than the current setup.

cf-execd/cf-execd.c
vpodzime: Yes, there are two separate things -- **how often** this code wakes up (monotonic clock) and **when** it happens (system time). We can only control the former, but the latter actually determines the defined classes and thus the agent run triggering. So... :man_shrugging:

PR #5507, opened 2024-05-10
GH Actions macOS fixes (3.18)

craigcomstock:
- GH Actions macOS: Fixed installation and PATH for libtool and autoconf
- GH Actions macOS: Use brew --prefix to locate LMDB, OpenSSL, and pcre


craigcomstock:
> Already fixed here: #5505

excellent. didn't see it.

PR #5509, opened 2024-05-21
ENT-11771: Adjust windows acceptance test job workflow to use only C: partition

craigcomstock:
There was an issue with 3.18.x version of this job which required this change.

Ticket: ENT-11771
Changelog: none


vpodzime (approved):
This looks like multiple changes in one commit. I'd prefer splitting them up. But ACK as it's just a GH action and it improves things.

PR #5512, opened 2024-05-22
check on presence of visual studio tools

craigcomstock:


craigcomstock:
just a debug pr. real one is https://github.com/cfengine/core/pull/5518

PR #5517, opened 2024-05-23
ENT-11771: Check master for where cf.events.dll is searched for

craigcomstock:
- **Try not copying cf.events.dll and see if tests fail**
- **just run one windows acceptance test**


craigcomstock:
was just debugging here, real PR for 3.18.x is here: https://github.com/cfengine/core/pull/5518

PR #5522, opened 2024-05-29
Fix VCS_TYPE docs in masterfiles-stage/README.org

vpodzime:
Use ~code~ instead of =verbatim= to avoid issues with the highlighted content containing =.

Also link cfbs website for people to easily find out what cfbs is and what it does.


vpodzime:
oh, good job, GitHub! :man_facepalming: 
image

PR #5535, opened 2024-06-05
CFE-4053: Fixed bug where default:sys.fqhost contained many spaces (3.18)

olehermanse:


olehermanse:
Already done by Vratislav here: https://github.com/cfengine/core/pull/5533

PR #5536, opened 2024-06-06
ENT-9491: Fixed bug in double expansion of foreign list variables

larsewi:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/10982/)

Back-ported to:
 - https://github.com/cfengine/core/pull/5560
 - https://github.com/cfengine/core/pull/5559

### DIFF
--- a/libpromises/iteration.c
+++ b/libpromises/iteration.c
@@ -314,8 +314,6 @@ static bool IsMangled(const char *s)
  */
 static void MangleVarRefString(char *ref_str, size_t len)
 {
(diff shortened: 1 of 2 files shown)


cf-bottom:

vpodzime (approved):
Very well done, sir! :clap: :beers: I'd just squash the last two commits because it doesn't make sense to cherry-pick one without the other. We would have to mark the test as soft-failing and then remove that together with the fix, but I think that's just extra unnecessary work.

PR #5537, opened 2024-06-07
ENT-11854: Fixed doc build (3.21)

nickanderson:
Markdown in the changelog (especially single backticks around words) cause the
doc build to break with failure to resolve automatic links. Here we simply
replaced backticks with single quotes.

Ticket: ENT-11854
Changelog: None


vpodzime (approved):
ACK

PR #5542, opened 2024-06-17
Allow depth_search on non-directory files

vpodzime:
So that policy using depth_search with a promiser that is a file
simply handles the file itself. A warning is issued, though.

Ticket: ENT-8996
Changelog: depth_search acting on a non-directory promiser now
           handles such file as if the promise didn't use
           depth_search issuing a warning

### DIFF
--- a/cf-agent/verify_files.c
+++ b/cf-agent/verify_files.c
@@ -446,16 +446,11 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
     }
     else
     {
-        if (!S_ISDIR(osb.st_mode))
+        if (!S_ISDIR(osb.st_mode) && a.havedepthsearch)
         {
-            if (a.havedepthsearch)
-            {
-                /* TODO: PROMISE_RESULT_DENIED */
-                Log(LOG_LEVEL_DEBUG,
-                    "depth_search (recursion) is promised for a base object '%s' that is not a directory",
-                    path);
-                goto exit;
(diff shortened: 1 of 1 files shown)


cf-agent/verify_files.c
```diff
@@ -446,18 +446,6 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
...
     else
     {
-        if (!S_ISDIR(osb.st_mode))
-        {
-            if (a.havedepthsearch)
-            {
-                /* TODO: PROMISE_RESULT_DENIED */
-                Log(LOG_LEVEL_DEBUG,
-                    "depth_search (recursion) is promised for a base object '%s' that is not a directory",
-                    path);
```
nickanderson, line 456: Perhaps a warning, or a verbose log that recursion was promised on a single file, ignoring recursion?

vpodzime: Updated.

nickanderson:
Manually tested these builds and they seem to behave as expected.

I get warnings:

```
warning: depth_search (recursion) is promised for a base object '/tmp/single-stable_remote-file-to-dir/stable-source-file.txt' that is not a directory
 warning: depth_search (recursion) is promised for a base object '/tmp/single-churning_remote-file-to-dir/churning-source-file.txt' that is not a directory
```

And, the file is updated.

nickanderson (approved):

PR #5549, opened 2024-06-20
Updated to newer platforms in demo script

olehermanse:


larsewi (approved):

vpodzime (commented):
What's the reason for the reduction in the number of different platforms? I think the variety was nice for demos. Also, could we add some arm64 machine?

craigcomstock (approved):

olehermanse:
> What's the reason for the reduction in the number of different platforms? I think the variety was nice for demos. Also, could we add some arm64 machine?

Good point! I think at the time, some of the older platforms were really slow / not working properly, so I was a bit liberal with removing. I'll have a look at adding some more platforms back in.

PR #5557, opened 2024-06-27
Added capture of masterfiles-stage log dc-scripts.log to cf-support

craigcomstock:
If something goes wrong with masterfiles-stage then /var/cfengine/outputs/dc-scripts.log will contain additional details about what happened.

This is a cumulative log so compress it.

Ticket: none
Changelog: none


vpodzime (approved):
ACK

larsewi (approved):

PR #5567, opened 2024-07-08
valgrind-checks improvements

vpodzime:


tests/valgrind-check/valgrind.sh
```diff
@@ -102,15 +110,25 @@ print_ps
...
+# some time before we move on and potentially hit the wall if it's actually
+# malfunctioning
+tries=12 # max 2 minutes
+while /var/cfengine/bin/cf-net -H $BOOTSTRAP_IP connect | grep Failed; do
+  tries=$((tries - 1))
+  if [ $tries -le 0 ]; then
+    break;
+  fi
+  sleep 10
+done
```
olehermanse, line 130: Any reason for doing this after starting cf-execd? I guess the reasoning for how it was before (cf-serverd, sleep, cf-execd) was to let cf-serverd start properly before cf-execd has a chance to mess with it.

olehermanse (approved):

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11042/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11042/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11042/

larsewi (approved):

tests/valgrind-check/valgrind.sh
vpodzime: Why would cf-execd mess with cf-serverd?

vpodzime: The logic is to do everything else we can before we potentially `sleep`.

vpodzime:
> Sure, I triggered a build:
> 
> [Build Status](https://ci.cfengine.com//job/pr-pipeline/11042/)
> 
> ****

image
Aborted to not waste resources, this doesn't impact acceptance tests or deployment tests.

tests/valgrind-check/valgrind.sh
olehermanse: @vpodzime `cf-execd` starts all the CFEngine services, right? (indirectly via cf-agent, systemd, or other ways). At least intuitively, it feels like `valgrind + cf-serverd` starting slowly could cause different situations to occur (race conditions) when `valgrind + cf-execd` is started immediately after it. I don't know how realistic / probable it is, though.

vpodzime: > `cf-execd` starts all the CFEngine services, right?

Not in case of systemd, we use a path condition on `policy_server.dat` IIRC. But on non-systemd systems it does, I guess. Not exactly sure what happens in the container. Anyway, cf-execd/cf-agent should see the cf-serverd process running even without a sleep.

olehermanse: @vpodzime Okay, 🤞

PR #5574, opened 2024-07-09
WIP: CFE-4401: Adjusted use of pthreads for Android/Termux platform

craigcomstock:
Android/Termux doesn't have pthread_cancel so use pthred_kill instead

Ticket: CFE-4401
Changelog: title

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -8668,7 +8668,11 @@ static FnCallResult FnCallIsReadable(ARG_UNUSED EvalContext *const ctx,
                 "Read operation timed out, exceeded %ld seconds.", path,
                 timeout);
 
+#ifdef HAVE_PTHREAD_CANCEL
             ret = pthread_cancel(thread_data.thread);
+#else
+            ret = pthread_kill(thread_data.thread, 0);
+#endif
             if (ret != 0)
             {
                 Log(LOG_LEVEL_ERR, "Failed to cancel thread");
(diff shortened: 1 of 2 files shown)


libpromises/evalfunction.c
```diff
@@ -8668,7 +8668,11 @@ static FnCallResult FnCallIsReadable(ARG_UNUSED EvalContext *const ctx,
                 "Read operation timed out, exceeded %ld seconds.", path,
                 timeout);
 
+#ifdef HAVE_PTHREAD_CANCEL
             ret = pthread_cancel(thread_data.thread);
+#else
+            ret = pthread_kill(thread_data.thread, 0);
```
larsewi, line 8674: Is it safe to join a "killed" thread? Will this call the signal handler installed by the agent. If so, is this fine? Also the man pthread_kill(3) says: 

> If sig is 0, then no signal is sent, but error checking is still performed.

craigcomstock: I should read your blog post, man pages and then try out `isreadable()` and see how it goes on termux. :)

craigcomstock: Will read https://cfengine.com/blog/2023/check-if-file-is-readable/ and then proceed.

craigcomstock: Looks like there is already a patch in termux-packages we may be able to pull in.

https://github.com/termux/termux-packages/blob/master/packages/cfengine/pthread_cancel.patch

PR #5576, opened 2024-07-10
ENT-11970: Revert masterfiles-stage --check-only and git_check_is_in_sync()

craigcomstock:
- Revert "ENT-11969: Fixed masterfiles-stage.sh -c|--check-only option parsing"
- Revert "Avoid overwriting the whole policy set in masterfiles-stage.sh if there is no update"
- Revert "Implement --check-only in masterfiles-stage.sh"

In testing with a cfbs project I feel these changes don't work much at all. At least for cfbs projects.

See ENT-11970 for some notes as I tried to debug cfbs project upstream updates not getting deployed.

together
https://github.com/cfengine/nova/pull/2282
https://github.com/cfengine/core/pull/5576

### DIFF
--- a/contrib/masterfiles-stage/common.sh
+++ b/contrib/masterfiles-stage/common.sh
@@ -296,30 +296,9 @@ git_cfbs_deploy_refspec() {
 ######################################################
 ##           VCS_TYPE-based main functions           #
 ######################################################
-git_check_is_in_sync() {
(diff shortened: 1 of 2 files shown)


craigcomstock:
I suspect that I may have fixed this issue so we don't need to revert.

Prefer this PR: https://github.com/cfengine/core/pull/5577

PR #5579, opened 2024-07-12
Fix early git revision comparison in masterfiles-stage.sh

vpodzime:
We need to compare the revision in the mirror against the target deploy directory, not the temporary checkout from the mirror. And we can do this earlier in the process.

Ticket: ENT-11970
Changelog: None

Merge together:
https://github.com/cfengine/core/pull/5579
https://github.com/cfengine/nova/pull/2284


vpodzime:
Looks like this is good enough! [Build Status](https://ci.cfengine.com/job/pr-pipeline/11055/)

olehermanse (approved):

PR #5581, opened 2024-07-16
ENT-11891: Redefined CFEngine roles inventory attribute and added related classes

olehermanse:
The "CFEngine roles" inventory attribute is now tied to the
sys.cfengine_roles variable (string list). This variable describes
what role a host has;

* On an Enterprise hub: { "Reporting hub", "Policy server" }
* On a Community policy server: { "Policy server" }
* On clients (the other hosts): { "Client" }

Similarly, the cfengine_reporting_hub and cfengine_client classes
have been added to more easily make decisions in policy based on role.

A host bootstrapped to itself is considered a "Policy server" /
policy_server (pre-existing class).

A host with the cf-hub binary (from the CFEngine Enterprise Hub
package) is considered a "Reporting hub" / cfengine_reporting_hub.

Any other host (not a reporting hub nor a policy server) is considered
a "Client" / cfengine_client.

### DIFF
--- a/libpromises/generic_agent.c
+++ b/libpromises/generic_agent.c
@@ -32,6 +32,7 @@
 #include <eval_context.h>
 #include <policy.h>
 #include <promises.h>
+#include <file_lib.h> // FileCanOpen()
 #include <files_lib.h>
 #include <files_names.h>
 #include <files_interfaces.h>
@@ -48,7 +49,7 @@
 #include <cleanup.h>
 #include <unix.h>
 #include <client_code.h>
-#include <string_lib.h>
+#include <string_lib.h> // StringCopy()
 #include <regex.h>      // CompileRegex()
 #include <writer.h>
 #include <exec_tools.h>
@@ -72,6 +73,7 @@
 #include <libgen.h>
 #include <cleanup.h>
 #include <cmdb.h>               /* LoadCMDBData() */
+#include "cf3.defs.h"
 
 #define AUGMENTS_VARIABLES_TAGS "tags"
 #define AUGMENTS_VARIABLES_DATA "value"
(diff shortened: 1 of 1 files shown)


olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11060/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11060/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11060/

nickanderson (approved):

libpromises/generic_agent.c
```diff
@@ -132,13 +134,62 @@ ENTERPRISE_VOID_FUNC_2ARG_DEFINE_STUB(void, GenericAgentSetDefaultDigest, HashMe
     *digest_len = CF_MD5_LEN;
 }
 
+void SetCFEngineRoles(EvalContext *ctx)
+{
+    char cf_hub_path[PATH_MAX];
+    StringCopy(GetBinDir(), cf_hub_path, sizeof(cf_hub_path));
+    JoinPaths(cf_hub_path, sizeof(cf_hub_path), "cf-hub");
+
+    const bool is_reporting_hub = FileCanOpen(cf_hub_path, "r");
```
olehermanse, line 143: @nickanderson @vpodzime This was the "best" function I found in our file related functions. Would it be better to check if the file is executable, or if the file exists (using stat)? Do you think that's more correct? Does it make a difference in practice?

nickanderson: My instinct is to use the most cheap check there is. If it does matter later, we can always improve it.

fwiw, when I implemented this in policy I did use the isexecutable() function to check.

olehermanse: https://github.com/NorthernTechHQ/libntech/pull/224

vpodzime: I'd argue that using `access(cf_hub_path, F_OK)` would be better here because a broken symlink would mean this is not a host collecting reports.

olehermanse: @vpodzime Yea, maybe. I was thinking that it does not make sense to do more sanity checks, because once you start, where do you stop? Check if it's a file, check permissions, check if it's executable, check if it's a valid symlink, check if the binary is non-empty, run the binary with `--help` to see that it starts, check if there's also postgreSQL available and running, etc. etc.

I don't really have an opinion on whether `FileCanOpen()` or `PathExists()` or `access(cf_hub_path, F_OK)` is better, or that it makes any difference, but as Nick said maybe the fastest option is the best? `PathExists()` a.k.a. `lstat` sounds the fastest to me, but I have no data on it.

olehermanse:
> Looks generally good to me, but my concern is that this will break existing queries using CFEngine roles = policy_server as a filter/condition on Inventory.

@vpodzime Yes, I agree, however I talked to Nick about it and we said that this is probably not very common, the current state looks quite bad (compared to other inventory) and is not very useful / usable, and now might be a good time to change it.

For sure I think I should write a change in behavior blog post to highlight it, do you think that's enough / acceptable?

vpodzime:
> is not very useful / usable, and now might be a good time to change it.
> 
> For sure I think I should write a change in behavior blog post to highlight it, do you think that's enough / acceptable?

I can't tell how much people use this, but filtering results to the hub only doesn't sound like a super rare thing to do to me. So I'd expect this to be used by users. No data, of course.

vpodzime (approved):
ACK

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11061/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11061/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11061/

PR #5590, opened 2024-07-26
CFE-4401: Adjusted isreadable() to work without pthread_cancel when not available

craigcomstock:
- **Adjusted use of pthreads for Android/Termux platform**
- **Take patch in downstream termux-packages project and adjust to use HAVE_PTHREAD_CANCEL instead of __ANDROID__**


libpromises/evalfunction.c
```diff
@@ -8668,7 +8685,11 @@ static FnCallResult FnCallIsReadable(ARG_UNUSED EvalContext *const ctx,
                 "Read operation timed out, exceeded %ld seconds.", path,
                 timeout);
 
+#ifdef HAVE_PTHREAD_CANCEL
             ret = pthread_cancel(thread_data.thread);
+#else
+            ret = pthread_kill(thread_data.thread, SIGUSR2);
```
craigcomstock, line 8691: @larsewi you had mentioned in my previous PR [comment](https://github.com/cfengine/core/pull/5574#discussion_r1670902889)

> Is it safe to join a "killed" thread? Will this call the signal handler installed by the agent. If so, is this fine? Also the man pthread_kill(3) says:

And so here I am using the downstream patch which sends a kill signal and adds a signal handler to which then calls `pthread_exit` which is the "other option" of how the joined thread could be terminated at the appropriate timeout moment.

> After a canceled thread has terminated, a join with that thread
       using [pthread_join(3)](https://www.man7.org/linux/man-pages/man3/pthread_join.3.html) obtains PTHREAD_CANCELED as the thread's
       exit status.  (Joining with a thread is the only way to know that
       cancelation has completed.)

from https://www.man7.org/linux/man-pages/man3/pthread_cancel.3.html

craigcomstock:
@xtkoba, I took your patch at https://github.com/termux/termux-packages/blob/master/packages/cfengine/pthread_cancel.patch as inspiration to improve my earlier attempt.

I am also working on a "cfengine: Bump 3.24.0" PR. Getting this patch upstream will have to wait for 3.24.1 release if we merge the PR here.

libpromises/evalfunction.c
```diff
@@ -8513,10 +8513,27 @@ struct IsReadableThreadData
...
 {
     assert(data != NULL);
 
+#ifndef HAVE_PTHREAD_CANCEL
+    struct sigaction actions;
+    memset(&actions, 0, sizeof(actions));
+    sigemptyset(&actions.sa_mask);
+    actions.sa_flags = 0;
+    actions.sa_handler = ThreadSignalHandler;
+    sigaction(SIGUSR2, &actions, NULL);
```
vpodzime, line 8534: I'd prefer `SIGHUP` or some other signal that semantically matches the case? In fact, I think we should block many common signals in the new thread, we don't want handlers for `SIGINT` etc. to be handled in it. So I'd extend this to something like [what we do in cf-reactor](https://github.com/cfengine/nova/blob/master/cf-reactor/cf-reactor.h#L66-L82) and used one of the other signals here.

vpodzime (commented):
The cherry-picked lines mention commits that are not in this repository. I don't think we need them. And let's squash the commits, please. The first one is not good enough alone.

larsewi (dismissed):

libpromises/evalfunction.c
craigcomstock: Sure, I agree with SIGHUP/etc.

The reason for splitting commits was to acknowledge and make clear the use of another persons contribution.

I do agree that it makes more sense to have commits which are sufficient and complete on their own but also recognize that sometimes we make "mistakes" in a commit and then have to come back and fix them, sort of like in this case here.

craigcomstock:
> The cherry-picked lines mention commits that are not in this repository. I don't think we need them. And let's squash the commits, please. The first one is not good enough alone.

Actually the "cherry-picked" comments are not needed, these are simply commits and not cherry-picked from someplace permanent or interesting (just another branch or my forked repo so an artifact of git operations). I will remove the cherry-pick comments and consider squashing after I fixup the signal handling.

craigcomstock:
This is a work-in-progress now as I am refactoring a few things based on code review. No need to comment until I un-draft the PR.

vpodzime:
> This is a work-in-progress now as I am refactoring a few things based on code review. No need to comment until I un-draft the PR.

Thanks for the notice and double thanks for the refactoring!

PR #5593, opened 2024-08-01
examples/main: Fix example output

btriller:


nickanderson:
That's what I see it should be. Thanks!

```
cfengine/src/core/examples$ cf-agent -Kf ./main_entry_point.cf 
R: Library: Hello from main_entry_point.cf.
```

nickanderson (approved):

PR #5599, opened 2024-08-09
Log errors for commands promises with exit codes not matching any _re…

vpodzime:
…turncodes

They are marked as failed with the following message:

  Command related to promiser '...' returned code '1' not defined as promise kept, not kept or repaired; setting to failed

Failed promises should log errors not just info
messages. Otherwise running without `--info` or more detailed logging fails silently.

Ticket: CFE-4429
Ticket: ENT-12103
Changelog: commands promises with exit codes not matching any
           _returncodes attributes from classes body now log and
           error message not just an info message

### DIFF
--- a/cf-agent/retcode.c
+++ b/cf-agent/retcode.c
@@ -86,7 +86,7 @@ bool VerifyCommandRetcode(EvalContext *ctx, int retcode, const Attributes *a, co
 
         if (!matched)
         {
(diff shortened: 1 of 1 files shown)


vpodzime:
smoke test: [Build Status](https://ci.cfengine.com/job/pr-pipeline/)

craigcomstock (approved):
This absolutely makes sense. Now the cases where PROMISE_RESULT_FAIL are the result either in case of a classes attribute or not act the same.

👍

PR #5602, opened 2024-08-22
Added logging CFEngine component related SELinux denials in cf-support

craigcomstock:
Ticket: ENT-12137
Changelog: title

### DIFF
--- a/misc/cf-support
+++ b/misc/cf-support
@@ -380,6 +380,11 @@ else
 fi
 echo "Captured output of $syslog_cmd filtered for cf-|CFEngine"
 
+# cf- component related SELinux denials
+if command -v ausearch >/dev/null; then
+  ausearch -m avc -su cfengine > "$tmpdir"/ausearch-cfengine-avcs.log
+fi
+
 gzip_add $WORKDIR/outputs/previous
 gzip_add $WORKDIR/outputs/dc-scripts.log # masterfiles-stage log
 file_add $WORKDIR/policy_server.dat


larsewi (dismissed):

misc/cf-support
```diff
@@ -380,6 +380,11 @@ else
 fi
 echo "Captured output of $syslog_cmd filtered for cf-|CFEngine"
 
+# cf- component related SELinux denials
+if command -v ausearch >/dev/null; then
+  ausearch -m avc -c cf- > "$tmpdir"/ausearch-cfengine-avcs.log
+fi
```
vpodzime, line 386: This might not match all the cases, for example `httpd` or `postgres` don't start with `cf-`. And we may run many other commands that can hit issues. I think a better filter would be on the source context.

craigcomstock: right! I will adjust.

vpodzime (approved):
ACK as long as `-su` works with a partial match.

craigcomstock:
> ACK as long as `-su` works with a partial match.

Yes, I checked with cfen, cfengine_react and it does seem to work with a partial match.

From the `ausearch` man page

```
     -su, --subject SE-Linux-context-string
              Search for event with scontext (subject) matching the string.
```

craigcomstock:
cherry picks
htps://github.com/cfengine/core/pull/5605
htps://github.com/cfengine/core/pull/5604

PR #5607, opened 2024-09-11
tests/static: Install extra tools in build image

btriller:
Install extra tools in build image to get rid of errors in GH workflow

./configure: line 8045: cmp: command not found
./configure: line 8224: /usr/bin/file: No such file or directory
./configure: line 9811: diff: command not found

### DIFF
--- a/tests/static-check/run.sh
+++ b/tests/static-check/run.sh
@@ -16,7 +16,7 @@ fi
 
 function create_image() {
   local c=$(buildah from -q $BASE_IMG)
-  buildah run $c -- dnf -q -y install "@C Development Tools and Libraries" clang cppcheck which >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


tests/static-check/run.sh
```diff
@@ -16,7 +16,7 @@ fi
 
 function create_image() {
   local c=$(buildah from -q $BASE_IMG)
-  buildah run $c -- dnf -q -y install "@C Development Tools and Libraries" clang cppcheck which >/dev/null 2>&1
+  buildah run $c -- dnf -q -y install "@C Development Tools and Libraries" clang cppcheck which diffutils file >/dev/null 2>&1
```
vpodzime, line 19: Interesting that these are suddenly required.

vpodzime (approved):
Thanks! I'll take care of the newly-found `cppcheck` issues (probably false-positives).

PR #5609, opened 2024-09-12
Workaround cppcheck ignoring `-i libpromises/cf3lex.c`

vpodzime:
When running cppcheck in static checks, we use `-i libpromises/cf3lex.c` in order to make cppcheck ignore the generated cf3lex.c file. However, the new version of cppcheck ignores this option and fails on issues found in the file.

However, we don't want this file or any other generated files except for bootstrap.inc which contains the bootstrap policy. So we can run `make clean` and `make -C libpromise/ boostrap.inc` before running cppcheck to get a cleaner environment.

### DIFF
--- a/tests/static-check/run_checks.sh
+++ b/tests/static-check/run_checks.sh
@@ -21,6 +21,8 @@ function check_with_clang() {
 
 function check_with_cppcheck() {
   rm -f config.cache
+  make clean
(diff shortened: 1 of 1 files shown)


vpodzime:
A follow-up to #5607

vpodzime:
Reported as https://bugzilla.redhat.com/show_bug.cgi?id=2311947

PR #5611, opened 2024-10-08
ENT-11988: Atomic copy_from in files promise

larsewi:
Changes to `files` promise in `copy_from` attribute:

- The new file (i.e., `<FILENAME>.cfnew`) is now created with correct
  permission during remote copy. Previously it would be created with
  default permissions.
- The destination file (i.e., `<FILENAME>`) is no longer deleted on
  backup during file copy.  Previously it would be renamed to
  `<FILENAME>.cfsaved`, causing the original file to dissappear. Now an
  actual copy of the original file with the same permissions is created
  instead.

As a result, there will no longer be a brief moment where the original
file is inaccessible.

### DIFF
--- a/cf-agent/verify_files_utils.c
+++ b/cf-agent/verify_files_utils.c
@@ -1552,7 +1552,7 @@ bool CopyRegularFile(EvalContext *ctx, const char *source, const char *dest, con
         }
 
         if (!CopyRegularFileNet(source, ToChangesPath(new),
-                                sstat->st_size, attr->copy.encrypt, conn))
+                                sstat->st_size, attr->copy.encrypt, conn, sstat->st_mode))
         {
             RecordFailure(ctx, pp, attr, "Failed to copy file '%s' from '%s'",
                           source, conn->remoteip);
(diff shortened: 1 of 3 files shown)


larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11252/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11252/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11252/

vpodzime (commented):
Looks good to me except that we should use (the code from) `CopyRegularFileDisk()`, I believe.

larsewi:
> Looks good to me except that we should use (the code from) CopyRegularFileDisk(), I believe.

Thanks :beers:

craigcomstock (approved):
Yeah, makes sense. I had to go looking for the "move cfnew to dest" code and found it at

```
        if (rename(changes_new, changes_dest) == 0)
        {
            RecordChange(ctx, pp, attr, "Moved '%s' to '%s'", new, dest);
            *result = PromiseResultUpdate(*result, PROMISE_RESULT_CHANGE);
        }
```

Interesting that the RecordChange() call uses different variables than the rename() call but they "should" be the same. ;)

vpodzime (approved):
ACK

PR #5614, opened 2024-10-10
ENT-11988: Atomic copy_from in files promise (3.21.x)

larsewi:
Changes to `files` promise in `copy_from` attribute:

- The new file (i.e., `<FILENAME>.cfnew`) is now created with correct
  permission during remote copy. Previously it would be created with
  default permissions.
- The destination file (i.e., `<FILENAME>`) is no longer deleted on
  backup during file copy.  Previously it would be renamed to
  `<FILENAME>.cfsaved`, causing the original file to dissappear. Now an
  actual copy of the original file with the same permissions is created
  instead.

As a result, there will no longer be a brief moment where the original
file is inaccessible.

Back-ported from https://github.com/cfengine/core/pull/5611/

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -41,7 +41,7 @@ jobs:
         
       - name: Install dependencies (C)
         if: ${{ matrix.language == 'cpp' }}
-        run: sudo apt-get update -y && sudo apt-get install -y libssl-dev libpam0g-dev liblmdb-dev byacc curl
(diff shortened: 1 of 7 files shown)


larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11258/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11258/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11258/

PR #5618, opened 2024-10-11
Fixed error in static checks

larsewi:
libncurses5 was not available in the noble repository, but maybe
libncurses6 will work.

```
E: Unable to locate package libncurses5
```

Ticket: None
Changelog: None


larsewi:
Thanks @craigcomstock :rocket: 

> is there a fake package for just libncurses (no version) that works?

Not that I'm aware of. Tried looking briefly with `apt search`

PR #5619, opened 2024-10-14
Adjust cf-support core dump finding to include more CFEngine binary l…

craigcomstock:
…ocations

For example, httpd is outside of /var/cfengine/bin but could still coredump and we want to catch that.

### DIFF
--- a/misc/cf-support
+++ b/misc/cf-support
@@ -154,11 +154,11 @@ _sysctl_kernel_core_pattern="$(sysctl -n kernel.core_pattern)"
 if expr "$_sysctl_kernel_core_pattern" : ".*/systemd-coredump.*" > /dev/null 2>&1; then
   echo "Found systemd-coredump used in sysctl kernel.core_pattern \"$_sysctl_kernel_core_pattern\""
   echo "Using coredumpctl to analyze CFEngine core dumps"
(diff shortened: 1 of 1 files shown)


cf-bottom:

misc/cf-support
```diff
@@ -178,10 +178,10 @@ elif command -v apport-unpack >/dev/null; then
...
   if [ "$non_interactive" -eq 0 ]; then
-    printf "Analyze coredumps found under /var/cfengine/bin? [Y/<enter alternate path>/n]: "
+    printf "Analyze coredumps found under /var/cfengine? [Y/<enter alternate path>/n]: "
     read -r response
   fi
-  response=${response:-/var/cfengine/bin}
+  response=${response:-/var/cfengine}
   if [ "$response" != "n" ]; then
     # file command on core files results in lines like the following which we parse for cf-* binaries
     # core: ELF 64-bit LSB core file, x86-64, version 1 (SYSV), SVR4-style, from '/var/cfengine/bin/cf-key', real uid: 0, effective uid: 0, realgid: 0, effective gid: 0, execfn: '/var/cfengine/bin/cf-key', platform: 'x86_64'
```
craigcomstock, line 187: this portion needs more work, we shouldn't parse for cf-* anymore as we might be dealing with httpd or other binaries from /var/cfengine so more research/testing must be done.

craigcomstock:
let's try updated #5817 instead

PR #5620, opened 2024-10-21
ENT-12033: Made multiple changes to increase robustness of remote file copying

larsewi:
Initially I was trying to reproduce a bug reported in ENT-12033. However, I was not able to reproduce it. While statically analyzing the code, I made some changes to make it more robust. 

- **Added doc string to CopyRegularFileNet**
- **Added assert to check return code of snprintf**
- **Fixed typo in comment for CopyRegularFileNet**
- **Removed check for filename too long**
- **Removed unnecessary assignment of NULL-byte to buffer**
- **Check return code of sscanf**
- **Added sanity check in CopyRegularFileNet**
- **Allocate receive buffer on stack instead of heap**
- **cf-serverd now stats file every read on network transition**
- **Added filestream flushing on error in EncryptCopyRegularFileNet**

[Build Status](https://ci.cfengine.com/job/pr-pipeline/11328/)

### DIFF
--- a/cf-serverd/server_common.c
+++ b/cf-serverd/server_common.c
@@ -403,7 +403,7 @@ static void FailedTransfer(ConnectionInfo *connection)
 void CfGetFile(ServerFileGetState *args)
 {
     int fd;
-    off_t n_read, total = 0, sendlen = 0, count = 0;
+    off_t n_read, total = 0, sendlen = 0;
     char sendbuffer[CF_BUFSIZE + 256], filename[CF_BUFSIZE - 128];
     struct stat sb;
     int blocksize = 2048;
@@ -459,13 +459,6 @@ void CfGetFile(ServerFileGetState *args)
     }
     else
     {
-        int div = 3;
-
-        if (sb.st_size > 10485760L) /* File larger than 10 MB, checks every 64kB */
-        {
-            div = 32;
-        }
-
         while (true)
         {
             memset(sendbuffer, 0, CF_BUFSIZE);
@@ -488,14 +481,11 @@ void CfGetFile(ServerFileGetState *args)
 
                 /* check the file is not changing at source */
 
-                if (count++ % div == 0)   /* Don't do this too often */
+                if (stat(filename, &sb) == -1)
                 {
-                    if (stat(filename, &sb))
-                    {
-                        Log(LOG_LEVEL_ERR, "Cannot stat file '%s'. (stat: %s)",
-                            filename, GetErrorStr());
-                        break;
-                    }
+                    Log(LOG_LEVEL_ERR, "Cannot stat file '%s'. (stat: %s)",
+                        filename, GetErrorStr());
+                    break;
                 }
 
                 if (sb.st_size != savedlen)
--- a/libcfnet/client_code.c
+++ b/libcfnet/client_code.c
@@ -579,12 +579,6 @@ static bool EncryptCopyRegularFileNet(const char *source, const char *dest, off_
 
     snprintf(cfchangedstr, 255, "%s%s", CF_CHANGEDSTR1, CF_CHANGEDSTR2);
 
-    if ((strlen(dest) > CF_BUFSIZE - 20))
-    {
-        Log(LOG_LEVEL_ERR, "Filename too long");
-        return false;
-    }
-
     unlink(dest);                /* To avoid link attacks */
 
     int dd = safe_open_create_perms(dest, O_WRONLY | O_CREAT | O_TRUNC | O_EXCL | O_BINARY, CF_PERMS_DEFAULT);
@@ -768,12 +762,6 @@ bool CopyRegularFileNet(const char *source, const char *dest, off_t size,
 
     NDEBUG_UNUSED int rc = snprintf(cfchangedstr, 255, "%s%s", CF_CHANGEDSTR1, CF_CHANGEDSTR2);
     assert(rc >= 0 && rc < sizeof(cfchangedstr));
-
-    if ((strlen(dest) > CF_BUFSIZE - 20))
-    {
-        Log(LOG_LEVEL_ERR, "Filename too long");
-        return false;
-    }
 
     unlink(dest);                /* To avoid link attacks */
 
--- a/libcfnet/client_code.c
+++ b/libcfnet/client_code.c
@@ -864,7 +864,11 @@ bool CopyRegularFileNet(const char *source, const char *dest, off_t size,
         /* Check for mismatch between encryption here and on server. */
 
         int value = -1;
-        sscanf(buf, "t %d", &value);
+        rc = sscanf(buf, "t %d", &value);
+        if (rc != 1)
+        {
+            value = -1;
+        }
 
         if ((value > 0) && (strncmp(buf + CF_INBAND_OFFSET, "BAD: ", 5) == 0))
         {
--- a/libcfnet/client_code.c
+++ b/libcfnet/client_code.c
@@ -598,8 +598,6 @@ static bool EncryptCopyRegularFileNet(const char *source, const char *dest, off_
         return true;
     }
 
-    workbuf[0] = '\0';
-
     NDEBUG_UNUSED int rc = snprintf(in, CF_BUFSIZE - CF_PROTO_OFFSET, "GET dummykey %s", source);
     assert(rc >= 0 && rc < sizeof(in));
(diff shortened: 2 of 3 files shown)


larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11318/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11318/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11318/

libcfnet/client_code.h
```diff
@@ -47,8 +47,23 @@ AgentConnection *ServerConnection(const char *server, const char *port, const Rl
...
+
+/**
+ * @brief Copy regular file over network.
+ * @param source The source filename.
+ * @param dest The destination filename.
+ * @param size The source file size.
+ * @param encrypt Whether or not to encrypt the transmission.
+ * @param conn The connection object to communicate with server.
+ * @param mode The source file mode (perms).
+ * @return True on success, false on error.
```
vpodzime, line 59: `true`, or even better `%true`, doxygen recognizes `%special_value` IIRC.

libcfnet/client_code.c
```diff
@@ -767,12 +767,6 @@ bool CopyRegularFileNet(const char *source, const char *dest, off_t size,
     NDEBUG_UNUSED int rc = snprintf(cfchangedstr, 255, "%s%s", CF_CHANGEDSTR1, CF_CHANGEDSTR2);
     assert(rc >= 0 && rc < sizeof(cfchangedstr));
 
-    if ((strlen(dest) > CF_BUFSIZE - 20))
-    {
-        Log(LOG_LEVEL_ERR, "Filename too long");
-        return false;
-    }
```
vpodzime, line 773: It has to do with the CFEngine's (file transfer) protocol, right? I believe `GET $SIZE $FILENAME` have to fit into `CF_BUFSIZE`.

cf-serverd/server_common.c
```diff
@@ -488,14 +481,11 @@ void CfGetFile(ServerFileGetState *args)
 
                 /* check the file is not changing at source */
 
-                if (count++ % div == 0)   /* Don't do this too often */
+                if (stat(filename, &sb))
```
vpodzime, line 484: Might as well do `!= -1` while you are moving this.

cf-serverd/server_common.c
```diff
@@ -459,13 +459,6 @@ void CfGetFile(ServerFileGetState *args)
     }
     else
     {
-        int div = 3;
-
-        if (sb.st_size > 10485760L) /* File larger than 10 MB, checks every 64kB */
-        {
-            div = 32;
-        }
```
vpodzime, line 467: Nice. `stat()` is a syscall, but it has pretty good caching mechanisms on modern systems so it's not that expensive. And old crap will always behave like old crap...

libcfnet/client_code.h
```diff
@@ -47,8 +47,23 @@ AgentConnection *ServerConnection(const char *server, const char *port, const Rl
...
 bool CompareHashNet(const char *file1, const char *file2, bool encrypt, AgentConnection *conn);
+
+/**
+ * @brief Copy regular file over network.
+ * @param source The source filename.
+ * @param dest The destination filename.
+ * @param size The source file size.
+ * @param encrypt Whether or not to encrypt the transmission.
+ * @param conn The connection object to communicate with server.
+ * @param mode The source file mode (perms).
```
craigcomstock, line 58: Why is there a parameter for SOURCE file mode? It would make a lot more sense to me if this was the DESTINATION mode.

libcfnet/client_code.c
```diff
@@ -901,6 +903,18 @@ bool CopyRegularFileNet(const char *source, const char *dest, off_t size,
...
+
+        /* A sanity check to make sure we don't write more than the original
+         * filesize acquired by SYNCH ... STAT source. This might happen if
+         * the file was changed while copying. */
+        if (n_wrote_total > size)
+        {
+            Log(LOG_LEVEL_INFO, "Source '%s:%s' changed while copying",
+                conn->this_server, source);
+            close(dd);
+            free(buf);
```
craigcomstock: It seems to me that you need some more cleanup. Before this and the other failure catch above this one we have

```
        bool w_ok = FileSparseWrite(dd, buf, n_read,
                                    &last_write_made_hole);
```

and so it would seem to me you need more than close/free but rather the same as the other failure case catch above, though I am not sure about conn->error = true but seems also reasonable.

```
            free(buf);
            unlink(dest);
            close(dd);
            FlushFileStream(conn->conn_info->sd, size - n_wrote_total - n_read);
            conn->error = true;
```

craigcomstock (commented):
Looks good.

libcfnet/client_code.c
larsewi: But this is the destination file. As far as I can see, it is never sent over the protocol. I don't understand why it is there. It's been there from the beginning of time.

libcfnet/client_code.c
larsewi: Good catch 🚀

larsewi: I'm not sure about the `conn->error = true` either. As far as I can see, it's not really checked after executing this code.

libcfnet/client_code.c
```diff
@@ -748,7 +748,8 @@ bool CopyRegularFileNet(const char *source, const char *dest, off_t size,
 {
     assert(conn != NULL);
 
-    char *buf, workbuf[CF_BUFSIZE], cfchangedstr[265];
+    char buf[CF_BUFSIZE + sizeof(int)]; /* Note CF_BUFSIZE not buf_size !! */
```
larsewi, line 751: The `sizeof(int)` makes me a little worried, as it can be either 4 or 8 Bytes

cf-serverd/server_common.c
larsewi: Added this note to the commit message. Thanks 😉

libcfnet/client_code.h
larsewi: It creates the destination file, based on a remote `stat` of the source file. Extended the description to make this more clear 😉

libcfnet/client_code.c
vpodzime: Oh, you're right! Then I'm not sure, maybe it was part of some reporting data or whatever. Good riddance now, for sure!

libcfnet/client_code.h
```diff
@@ -47,8 +47,24 @@ AgentConnection *ServerConnection(const char *server, const char *port, const Rl
...
+ * @brief Copy regular file over network.
+ * @param source The source filename.
+ * @param dest The destination filename.
+ * @param size The source file size.
+ * @param encrypt Whether or not to encrypt the transmission.
+ * @param conn The connection object to communicate with server.
+ * @param mode The source file mode (perms) to be used when creating the
+ *             destination file.
+ * @return %true on success, %false on error.
+ * @note Note that the %encrypt parameter only has an effect on the CLASSIC
```
vpodzime, line 61: This should be `#encrypt` as an anchor/link.

libcfnet/client_code.c
vpodzime: It should be quite easy to merge the two cleanup blocks into a single one acting on two possible error/failure cases.

libcfnet/client_code.c
vpodzime: Yeah... Do you know what it is there for? Please create a ticket for this if you cannot spend the extra time on it now.

vpodzime (dismissed):
Looks good to me otherwise. Thanks!

craigcomstock (approved):

libcfnet/client_code.c
```diff
@@ -869,40 +859,48 @@ bool CopyRegularFileNet(const char *source, const char *dest, off_t size,
...
+            }
+            else // if (n_wrote_total > size)
+            {
+                Log(LOG_LEVEL_INFO, "Source '%s:%s' changed while copying",
+                    conn->this_server, source);
+            }
             unlink(dest);
             close(dd);
-            FlushFileStream(conn->conn_info->sd, size - n_wrote_total - n_read);
+            FlushFileStream(conn->conn_info->sd, size - n_wrote_total);
```
vpodzime, line 900: Doesn't this depend on which error case we are handling? I'd probably just check `n_wrote_total + n_read` in the condition instead of doing the addition first.

vpodzime (approved):
Hopefully the last nitpick.

PR #5621, opened 2024-10-29
ENT-12414: New network protocol v4 - SAFEGET

larsewi:
The current protocol (v3) uses a `STAT <FILENAME>` request to determine the file size (among other things), followed by a `GET <FILENAME>` request to fetch the file content. However, there is a race condition here. If the file size increases between the two requests, the client would think that the remaining data after the "file size" offset is a new response header.

Here, we introduce a new protocol-version "safeget" to address the race condition. PDUs from the response following a `GET <FILENAME>` request now contain a protocol header consisting of a NULL-byte terminated string of four unsigned integers (`uint64_t`). We will refer to the four integers as ERROR_CODE, FILE_SIZE, FILE_OFFSET, and PAYLOAD_SIZE. The integers will appear in this order in the protocol header.  Each integer is separated by a whitespace character.

A non-zero ERROR_CODE signals that an error occurred. Furthermore, the exact value of ERROR_CODE specifies what went wrong on the server side. FILE_SIZE and FILE_OFFSET are mainly used to determine when the last PDU is received (i.e., when FILE_SIZE = FILE_OFFSET + PAYLOAD_SIZE). However, FILE_SIZE is also used to detect if the file is modified at the source during transmission. The payload starts immediately after the NULL-terminating byte in the protocol header and consists of PAYLOAD_SIZE bytes.

### DIFF
--- a/libcfnet/README.md
+++ b/libcfnet/README.md
@@ -15,6 +15,7 @@ Names of protocol versions:
 1. `"classic"` - Legacy, pre-TLS, protocol. Not enabled or allowed by default.
 2. `"tls"` - TLS Protocol using OpenSSL. Encrypted and 2-way authentication.
 3. `"cookie"` - TLS Protocol with cookie command for duplicate host detection.
+3. `"safeget"` - Same as cookie, but with an atomic and safe `GET` request.
 
 Wanted protocol version can be specified from policy:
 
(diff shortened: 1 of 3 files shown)


larsewi:
Superseded by https://github.com/cfengine/core/pull/5629

PR #5626, opened 2024-11-13
Adjusted peer related examples in regards to hostname -f versus sys.fqhost (3.21)

craigcomstock:
This was breaking on some systems in a way that doesn't benefit the test.

Instead of assuming that `hostname -f` is equal to `sys.fqhost` just use `sys.fqhost` in the first place.

Ticket: ENT-12437
Changelog: none
(cherry picked from commit 471dc0f4b5d468b989b0e42cb0cf3dd0275e7acc)

### DIFF
--- a/examples/peerleader.cf
+++ b/examples/peerleader.cf
@@ -25,7 +25,8 @@
 #@ echo beta >> /tmp/cfe_hostlist
 #@ echo gamma >> /tmp/cfe_hostlist
 #@ echo "Set HOSTNAME appropriately beforehand"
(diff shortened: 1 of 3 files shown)


craigcomstock:
Sadly, my change doesn't work. Even here in github I get

```
Q: "...r/bin/perl /hom": R: Unable to find my fully qualified hostname fv-az1534-168.5pdur5myp1iuthl4u4lgbmrhvg.dx.internal.cloudapp.net in /tmp/cfe_hostlist. Can't determine peers.
```

Will look more carefully at what is going on.

craigcomstock:
just missing windows msi package, not a big deal, will merge.

PR #5629, opened 2024-11-19
ENT-12414: Use `librsync`'s Stream API in `GET <FILENAME>` requests

larsewi:
- **Added file stream API**
- **New network protocol version v4 - filestream**

**TODO:**
 - [ ] Depends on https://github.com/cfengine/buildscripts/pull/1520
 - [x] Depends on https://github.com/NorthernTechHQ/libntech/pull/230
 - [x] Check that errors in different places in the code does not make client/server hang
 - [x] Add `librsync` as dependency in CI
 - [ ] Should be merged before hand https://gitlab.com/Northern.tech/CFEngine/jenkins-vms/-/merge_requests/59

[Build Status](https://ci.cfengine.com/job/pr-pipeline/11510/)


libcfnet/file_stream.c
```diff
@@ -0,0 +1,943 @@
...
+    int ret = TLSRecv(conn, recv_buffer, HEADER_SIZE);
+    if (ret != HEADER_SIZE)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to receive message header during file stream (%d != %d)",
+            ret,
+            HEADER_SIZE);
+        return false;
+    }
+    memcpy(&header, recv_buffer, HEADER_SIZE);
```
vpodzime, line 193: You could use `&header` as the buffer and avoid this `memcpy()`.

libcfnet/file_stream.c
```diff
@@ -0,0 +1,943 @@
...
+    header = ntohs(header);
+
+    /* Extract Error flag */
+    bool err = header & (1 << 0);
+
+    /* Extract End-of-File flag */
+    *eof = header & (1 << 1);
+
+    /* Extract message length */
+    *len = header >> 4;
```
vpodzime, line 203: This requires `len` (`size_t`) to be at least 3 bytes, right? Probably that's guaranteed, but I'd add a static assert for this to avoid potential nasty surprises.

libcfnet/file_stream.c
```diff
@@ -0,0 +1,943 @@
...
+
+    /* Read payload */
+    if (*len > 0)
+    {
+        ret = TLSRecv(conn, recv_buffer, *len);
```
vpodzime, line 208: I take it `TLSRecv()` does some retries to avoid short reads/recvs? Worth a comment.

libcfnet/file_stream.c
```diff
@@ -0,0 +1,943 @@
...
+/**
+ * @brief Send an error message using the file stream protocol
+ *
+ * @param conn The SSL connection object
+ * @param flush Whether or not to flush the stream (see FlushStream())
+ * @param fmt The format string
+ * @param ... The format string arguments
+ * @return true on success, otherwise false
+ */
+static bool SendError(SSL *conn, bool flush, const char *fmt, ...)
```
vpodzime, line 269: There's some `#pragma` to mark this as a printf-like function and let the compiler do all sorts of checks. Please add it.

libcfnet/file_stream.c
```diff
@@ -0,0 +1,943 @@
...
+        fclose(file);
+        return false;
+    }
+
+    /* Setup buffers */
+    rs_buffers_t bufs = {0};
+    bufs.next_out = out_buf;
+    bufs.avail_out = MESSAGE_SIZE; /* We cannot send more using the protocol */
+
+    do
```
vpodzime, line 680: Any chance we could de-duplicate what's in this `do-while` with the one in `SendDelta()`? It's very long anyway...

libcfnet/file_stream.c
```diff
@@ -0,0 +1,943 @@
...
+    FILE *new = safe_fopen_create_perms(dest, "wb", perms);
+    if (new == NULL)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to open/create destination file '%s': %s",
+            dest,
+            GetErrorStr());
+
+        /* At this point the server will not be expecting any more messages
+         * from the client as far as the File Stream API is conserned. Hence,
```
vpodzime, line 809: *concerned*

libcfnet/file_stream.c
```diff
@@ -592,6 +592,31 @@ bool FileStreamServe(SSL *conn, const char *filename)
...
+
+/**
+ * @brief Get the size of a file
+ *
+ * @param file The file pointer
+ * @return the file size or -1 on error
+ * @note -1 on error is quite handy, because rs_sig_args() iterprets it as
+ *       unknown file size
+ */
+static rs_long_t GetFileSize(FILE *file)
```
vpodzime, line 604: Do we need to worry about symlinks in this function?

libcfnet/file_stream.c
```diff
@@ -592,6 +592,31 @@ bool FileStreamServe(SSL *conn, const char *filename)
 
 #define ERROR_MSG_INTERNAL_CLIENT_ERROR "Internal client error"
 
+
+/**
+ * @brief Get the size of a file
+ *
+ * @param file The file pointer
+ * @return the file size or -1 on error
+ * @note -1 on error is quite handy, because rs_sig_args() iterprets it as
```
vpodzime, line 601: *interprets*

libcfnet/file_stream.c
```diff
@@ -592,6 +592,31 @@ bool FileStreamServe(SSL *conn, const char *filename)
...
+        return -1;
+    }
+
+    struct stat sb;
+    if (fstat(fd, &sb) == -1)
+    {
+        return -1;
+    }
+
+    return S_ISREG(sb.st_mode) ? sb.st_size : -1;
```
vpodzime, line 617: `(rs_long_t)` in both cases.

vpodzime (commented):
Looks good to me otherwise. Very nice! Thanks for taking the hard road! :clap: :beers: I'm very excited about this!

libcfnet/file_stream.c
```diff
@@ -0,0 +1,943 @@
...
+
+    /* Set End-of-File flag */
+    if (eof)
+    {
+        header |= (1 << 1);
+    }
+
+    /* Send header */
+    header = htons(header);
+    int ret = TLSSend(conn, (char *) &header, HEADER_SIZE);
```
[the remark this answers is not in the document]
larsewi: The main motivation for having this macro was that I would not have to cast the result to an `int`

libcfnet/file_stream.c
```diff
@@ -0,0 +1,943 @@
...
+ * @param len The length of the message to send (must be less or equal to
+ *            MESSAGE_SIZE Bytes)
+ * @param eof Set to true if this is the last message in a transaction,
+ *            otherwise false
+ * @param err Set to true if transaction must be canceled (e.g., due to an
+ *            unexpected error), otherwise false
+ * @return true on success, otherwise false
+ */
+static bool __SendMessage(
+    SSL *conn, const char *msg, size_t len, bool eof, bool err)
```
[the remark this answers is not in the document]
larsewi: Changing that argument to an `int` would make me have to do a lot of casting where ever these functions are called. The fact that it is an unsigned integer along with the `assert(len <= MESSAGE_SIZE)` also makes sure any values out of range are caught.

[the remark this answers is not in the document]
vpodzime: Shall we change `TLSSend()` to take a `size_t` value instead? That makes most sense, of course.

libcfnet/file_stream.c
larsewi: Yup, I also thought so. But `TLSRecv()` expects a buffer of size `CF_BUFSIZE` and that you never request more than `CF_BUFSIZE - 1` Bytes. And the reason why, is that it actually writes a terminating `\0` at the end of the message. I spent a long time trying to figure out why my program crashed with memory violations :stuck_out_tongue_winking_eye: There is a comment some lines above. But it probably deserves more explanation.

libcfnet/file_stream.c
larsewi: If it was left shift, then yes, 3 bytes. However, with right shift, two bytes should do. The following:
```C
    assert(sizeof(*len) >= 2); /* "It's probably guaranteed, but let's make
                                * sure to avoid potentially nasty surprises."
                                *  - @vpodzime */
```
still would not hurt.

libcfnet/file_stream.c
larsewi: Yes, I think it should be possible. The log messages would have to be more generic and the code would probably be a bit harder to understand. I will create a follow-up ticket to investigate this further.

libcfnet/file_stream.c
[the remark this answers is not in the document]
larsewi: Yes, I can create a follow-up ticket for this :wink:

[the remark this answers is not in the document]
larsewi: Then we can also use `sizeof(header)` without having to cast it to an int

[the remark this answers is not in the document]
larsewi: https://northerntech.atlassian.net/browse/CFE-4455

libcfnet/file_stream.c
larsewi: https://northerntech.atlassian.net/browse/CFE-4456

libcfnet/file_stream.c
vpodzime: Right, it's 12 bits, not 16. I got confused by `uint16_t`.

libcfnet/file_stream.c
vpodzime: Yes. That's why I wanted to suggest `lstat()`, but then I realized it was probably not the desired thing to do. And it made me think how the new protocol handles symlinks. It just copies the contents of the symlink target into a file on the other side, right? Worth a comment.

larsewi: Yup! Added a note in the doc strings :wink:

vpodzime (approved):
Let's see if it works! :grin: :crossed_fingers: :beers:

configure.ac
```diff
@@ -521,6 +521,20 @@ CF3_WITH_LIBRARY(pcre2, [
   )]
 )
 
+dnl librsync
+
+AC_ARG_WITH([librsync], [AS_HELP_STRING([--with-librsync[[=PATH]]], [Spefify librsync path])], [], [with_librsync=yes])
```
craigcomstock, line 526:
```suggestion
AC_ARG_WITH([librsync], [AS_HELP_STRING([--with-librsync[[=PATH]]], [Specify librsync path])], [], [with_librsync=yes])
```

libcfnet/file_stream.c
```diff
@@ -0,0 +1,1013 @@
...
+
+    do
+    {
+        if (bufs.eof_in == 0)
+        {
+            if (bufs.avail_in >= MESSAGE_SIZE)
+            {
+                /* The job requires more data, but the input buffer is full */
+                Log(LOG_LEVEL_ERR,
+                    "Insufficient buffer capacity to compute delta");
```
craigcomstock, line 712: again, maybe include avail_in and MESSAGE_SIZE

libcfnet/file_stream.c
```diff
@@ -0,0 +1,1013 @@
...
+ * @return true on success, otherwise false
+ */
+static bool RecvDelta(
+    SSL *conn, const char *basis, const char *dest, mode_t perms)
+{
+    assert(conn != NULL);
+    assert(basis != NULL);
+    assert(dest != NULL);
+
+    /* The input buffer has to be twice the message size, so that if can fit a
```
craigcomstock, line 819:
```suggestion
    /* The input buffer has to be twice the message size, so that it can fit a
```

libcfnet/file_stream.c
```diff
@@ -0,0 +1,1013 @@
...
+    {
+        /* Fill input buffers */
+        if (bufs.eof_in == 0)
+        {
+            if (bufs.avail_in > MESSAGE_SIZE)
+            {
+                /* The job requires more data, but we cannot fit another
+                 * message into the input buffer */
+                Log(LOG_LEVEL_ERR,
+                    "Insufficient buffer capacity to receive file stream delta");
```
craigcomstock, line 887: add avail_in and MESSAGE_SIZE to error message?

libcfnet/file_stream.h
```diff
@@ -0,0 +1,104 @@
...
+
+#include <tls_generic.h>
+#include <stdbool.h>
+#include <sys/types.h> /* mode_t */
+
+
+/**
+ * @brief Reply with unspecified server refusal
+ *
+ * E.g., use this function to when the resource does not exist or access is
```
craigcomstock, line 69:
```suggestion
 * E.g., use this function when the resource does not exist or access is
```

craigcomstock (commented):
Some comments and little typos.

vpodzime (approved):
ACK

craigcomstock (approved):

libcfnet/file_stream.c
```diff
@@ -0,0 +1,1013 @@
...
+ * @return the file size or -1 on error
+ * @note -1 on error is quite handy, because rs_sig_args() interprets it as
+ *       unknown file size
+ */
+static rs_long_t GetFileSize(FILE *file)
+{
+    /* librsync has rs_file_size() which basically does the same thing.
+     * However, it is only available in the most recent versions of the
+     * library. Hence, we provide our own implementation in order to have
+     * some backwards compatibility in terms of librsync versions. */
```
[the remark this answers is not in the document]
larsewi: Added some more info in the comment. When it comes to minimum version checks, it does not seem like we have  this for any of the other dependencies. If you think we should have these checks, then we should probably make a separate ticket for this.

larsewi:
Just tested this manually on Windows and it works like a charm.
```
PS C:\Program Files\Cfengine\bin> .\cf-agent.exe -KI
    info: Copied file '/var/cfengine/masterfiles/promises.cf' from '172.31.39.27' to 'C:\promises.cf.cfnew'
    info: Moved 'C:\promises.cf.cfnew' to 'C:\promises.cf'
    info: Updated file 'C:\promises.cf' from '172.31.39.27:/var/cfengine/masterfiles/promises.cf'
```
Did however find a separate (surprisingly unrelated) issue in `cf-net.exe` (see [ENT-12511](https://northerntech.atlassian.net/browse/ENT-12511))

libcfnet/file_stream.c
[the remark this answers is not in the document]
craigcomstock: I think what we have is fine. You have a workaround for older versions and we don't have to go down a rabbit hole of trying to figure out where our code-base with the workaround would stop working as we get into older versions of the library.

craigcomstock (approved):

PR #5637, opened 2024-11-26
Adjusted CFEngine SELinux policy to allow cf-execd to run ps command with policy version 33 (3.21)

craigcomstock:
Apparently, ps command running with SELinux kernel policy version 33 requires self:cap_userns sys_ptrace.

Ticket: ENT-12446
Changelog: title
(cherry picked from commit 45ea0fe542f73e8c4358d7c0e23049dee34056d3)


larsewi (approved):

craigcomstock:
Only windows_acceptance failed with a 6 hour time limit maybe?

Yes. https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration

> Job execution time - Each job in a workflow can run for up to 6 hours of execution time. If a job reaches this limit, the job is terminated and fails to complete.

PR #5639, opened 2024-11-27
Disabled windows_acceptance_tests in github CI due to 6 hour time limit

craigcomstock:
Ticket: ENT-12497
Changelog: none


larsewi (approved):
Should we make a ticket for looking into how we can make it run faster?

craigcomstock:
cherry picks: https://github.com/cfengine/core/pull/5642 https://github.com/cfengine/core/pull/5641

PR #5643, opened 2024-12-02
Adjusted SELinux policy to allow components which run cf-promises to getattr everywhere and read symlinks

craigcomstock:
Seen on rhel-8 and rhel-9 with kernels 4.18.0 and 5.14.0 and policy version 33.

Applies to cf-monitord, cf-execd and cf-serverd.

Ticket: ENT-12466
Changelog: title


nickanderson (approved):

craigcomstock:
picked: picked: https://github.com/cfengine/core/pull/5647, https://github.com/cfengine/core/pull/5646

PR #5645, opened 2024-12-03
ENT-10993: Static checks now rebuilds container if run.sh is modified

larsewi:


larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11469/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11469/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11469/

tests/static-check/run.sh
```diff
@@ -25,7 +25,30 @@ function create_image() {
...
+  # If a branch with that version does not exist, then it's probably master
+  VERSION="master"
+  STATUS=`curl -o /dev/null --silent -Iw '%{http_code}' "https://raw.githubusercontent.com/cfengine/core/refs/heads/${VERSION}/tests/static-check/run.sh"`
+fi
+
+# Compare digest of this file with upstream. If it has changed, then we rebuild the container
+if [[ $STATUS = "200" ]]
+then
+  SUM_A=`curl "https://raw.githubusercontent.com/cfengine/core/refs/heads/${VERSION}/tests/static-check/run.sh" | sha256sum | awk '{print $1}'`
+  SUM_B=`sha256sum "$0" | awk '{print $1}'`
```
craigcomstock, line 45: It seems like it might be "easier" to push the sha of the script into the image and then check those against each other.

Checking the sha of this script against upstream seems a bit round-a-bout and not to-the-point.

One case I think would work better with my suggested strategy is if you are iterating on a PR and changing this file, then the first run you would get a new image created. But what if you run this through CI many times after that to make changes to things OTHER than this file? It would still be re-creating the image each time when it doesn't need to.

The scripting looks good just not sure it's quite the right strategy.

larsewi: Yup, that makes sense!

larsewi:
Only static checks [Build Status](https://ci.cfengine.com/job/pr-pipeline/11478/)

craigcomstock (dismissed):
like it!

larsewi:
Let's try again (only static checks) [Build Status](https://ci.cfengine.com/job/pr-pipeline/11479/)

larsewi:
I made one more change [Build Status](https://ci.cfengine.com/job/pr-pipeline/11480/) (again only static checks)

craigcomstock (approved):

tests/static-check/run.sh
```diff
@@ -19,15 +21,37 @@ function create_image() {
...
+
+  # Recreate the image if the checksum of this file has changed 
+  if [[ `buildah run $c ls $SUM_FILE` == $SUM_FILE ]]; then
+    SUM_A=$(sha256sum $0)
+    SUM_B=$(buildah run $c cat $SUM_FILE)
+    if [[ $SUM_A != $SUM_B ]]; then
+      echo "Recreating image due to mismatching checksum..."
+      IMAGE_ID=$(buildah inspect $c | jq -r '.FromImageID')
+      buildah rmi --force $IMAGE_ID >/dev/null
+      c=$(create_image)
```
vpodzime, line 47: Something should `stop` and `rm` the `$c` container, right? Ideally before removing its image.

larsewi: So the `--force` "option will cause Buildah to remove all containers that are using the image before removing the image from the system" (see [buildah-rmi.1.md](https://github.com/containers/buildah/blob/main/docs/buildah-rmi.1.md)). Also, I'm not familiar with any command to stop containers Buildah.

vpodzime: oh, cool! I thought it would leave the container running and just force-remove the image. Perhaps we could add a comment? Thanks!

larsewi: Yup, there should absolutely be a comment explaining this. See https://github.com/cfengine/core/pull/5653

PR #5648, opened 2024-12-03
Granted more access to certificates directory for CFEngine components in SELinux policy

craigcomstock:
Were found to be needed in 3.21.6a and 3.24.1a testing on rhel-9 hubs.
Policy works on rhel-8 as well.

Ticket: ENT-12466
Changelog: title


nickanderson (approved):

craigcomstock:
Same issues were observed in 3.24.1a on rhel-9.
These policies can compile and be loaded on rhel-8 as well even though the AVCs don't occur there.

nickanderson (approved):

PR #5653, opened 2024-12-06
Added comments to explain image recreation logic

larsewi:
I added some comments to explain the logic behind recreating the image in static checks.


vpodzime (approved):
ACK

PR #5663, opened 2024-12-12
Added option to choose protocol version in cf-net

larsewi:
Added option to choose the maximum CFEngine protocol version to use in
cf-net. Maximum, because the server can negotiate on an older version.

Currently, there is no way to test older protocol versions without
targeting an older server. This option will make that possible.
Furthermore, it can be used as a workaround in the potential cases where
we issue a new protocol that does not work.

Ticket: ENT-12519
Changelog: Title


olehermanse:
@larsewi To me, it would make more sense to specify an exact version (the version you want to test) and then have cf-net fail if the server doesn't agree.

larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11519/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11519/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11519/

craigcomstock (approved):

PR #5673, opened 2025-01-03
CFE-4414: `cf-net connect` now exits with 1 in case of error

larsewi:
```
$ sudo /var/cfengine/bin/cf-net -H 192.168.123.456 connect
Failed to connect to '192.168.123.456'
$ echo $?
1
$ sudo /var/cfengine/bin/cf-net -H 127.0.0.1 connect
Connected & authenticated successfully to '127.0.0.1'
$ echo $?
0
```

### DIFF
--- a/cf-net/cf-net.c
+++ b/cf-net/cf-net.c
@@ -472,7 +472,9 @@ static int CFNetRun(CFNetOptions *opts, char **args, char *hostnames)
     int ret = 0;
     char *hostname = strtok(hosts, ",");
     while (hostname != NULL){
-        CFNetCommandSwitch(opts, hostname, args, cmd);
+        if (CFNetCommandSwitch(opts, hostname, args, cmd) != 0) {
+            ret = -1;
+        }
         hostname = strtok(NULL, ",");
     }
     free(hosts);
@@ -636,8 +638,7 @@ static int CFNetConnect(const char *hostname, const char *use_protocol_version, 
         Log(LOG_LEVEL_ERR, "No hostname specified");
         return -1;
     }
-    CFNetConnectSingle(hostname, use_protocol_version, true);
-    return 0;
+    return CFNetConnectSingle(hostname, use_protocol_version, true);
 }
 
 static void CFNetDisconnect(AgentConnection *conn)


larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11572/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11572/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11572/

cf-net/cf-net.c
```diff
@@ -472,7 +472,9 @@ static int CFNetRun(CFNetOptions *opts, char **args, char *hostnames)
     int ret = 0;
     char *hostname = strtok(hosts, ",");
     while (hostname != NULL){
-        CFNetCommandSwitch(opts, hostname, args, cmd);
+        if (CFNetCommandSwitch(opts, hostname, args, cmd) != 0) {
+            ret = -1;
```
craigcomstock, line 476: I definitely think you should add an error log statement here, something like

```
Log(LOG_LEVEL_ERR, "Command %s with %s options and %s args failed for hostname %s", cmd, opts, args, hostname);
```

craigcomstock (changes requested):

cf-net/cf-net.c
larsewi: Error is already logged if you look at https://github.com/cfengine/core/pull/5673#issue-2767274192

cf-net/cf-net.c
```diff
@@ -472,7 +472,9 @@ static int CFNetRun(CFNetOptions *opts, char **args, char *hostnames)
     int ret = 0;
```
vpodzime, line 472: So `ret` was not set to anything else than `0` before? Weird that we haven't been getting any warning/errors for this.

vpodzime (approved):
Thanks!

cf-net/cf-net.c
larsewi: I guess because we return `ret` in the end.

cf-net/cf-net.c
craigcomstock: ah, must be inside CFNetCommandSwitch(), yes, inside of CFNetOpenConnection(). Ack. :+1:

craigcomstock (approved):

olehermanse (approved):

PR #5677, opened 2025-01-09
Upgraded valgrind check container to Ubuntu 24

larsewi:
Recently we've seen segmentation faults while compiling core in valgrind
checks. I'll try to upgrade the container image to Ubuntu 24 to see if
that resolves the issue.

[Build Status](https://ci.cfengine.com/job/pr-pipeline/11587/)

### DIFF
--- a/tests/valgrind-check/Containerfile
+++ b/tests/valgrind-check/Containerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04 AS build
+FROM ubuntu:24.04 AS build
 RUN DEBIAN_FRONTEND=noninteractive apt-get update -y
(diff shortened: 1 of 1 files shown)


craigcomstock (approved):
I don't particularly like upgrading to ubu24 to get rid of segfaults without knowing a root cause of why ubu24 upgrade fixes things.

larsewi:
> I don't particularly like upgrading to ubu24 to get rid of segfaults without knowing a root cause of why ubu24 upgrade fixes things.

I don't know either. Don't even know if this fixes things. However, upgrading the image is a good thing non-the-less

PR #5682, opened 2025-01-16
ENT-11574: Link cf-agent and cf-serverd with leech if available

larsewi:
In order to use leech in the enterprise plugin, we need to link with
leech in core. I added code in configure.ac to link with leech if it is
available. A community version of CFEngine will work just fine without
it. However, an enterprise version will require it.


cf-bottom:

PR #5685, opened 2025-01-23
Added option to print rsync performance stats in cf-net

larsewi:
The performance stats are also logged. This way we can also study the
performance through `cf-agent`'s debug logs.

The following is an example of running `cf-net` with the `--stats` option:

```
$ sudo /var/cfengine/bin/cf-net --stats --host 192.168.56.10 get /var/cfengine/masterfiles/promises.cf
Send signature statistics:
  16970 bytes in (read from 'promises.cf')
  2424 bytes out (sent to server)
Receive delta statistics:
  9 bytes in (received from server)
  16970 bytes out (written to 'promises.cf.cfnew')
```

[Build Status](https://ci.cfengine.com/job/pr-pipeline/11640/)

### DIFF
--- a/cf-net/cf-net.c
+++ b/cf-net/cf-net.c
@@ -56,6 +56,7 @@ typedef struct
     char *min_tls_version;
     char *allow_ciphers;
     char *use_protocol_version;
+    bool print_stats;
 } CFNetOptions;
 
 //*******************************************************************
@@ -114,6 +115,7 @@ static const struct option OPTIONS[] =
     {"tls-version", required_argument,  0, 't'},
     {"ciphers",     required_argument,  0, 'c'},
     {"protocol",    required_argument,  0, 'p'},
+    {"stats",       no_argument,        0, 's'},
     {NULL,          0,                  0, '\0'}
 };
 
@@ -129,6 +131,7 @@ static const char *const HINTS[] =
     "Minimum TLS version to use",
     "TLS ciphers to use (comma-separated list)",
     "Specify CFEngine protocol to use. Possible values: 'classic', 'tls', 'cookie', 'filestream', 'latest' (default)",
+    "Print rsync performance statistics to stderr",
     NULL
 };
 
@@ -234,6 +237,7 @@ static void CFNetSetDefault(CFNetOptions *opts){
     opts->min_tls_version = NULL;
     opts->allow_ciphers   = NULL;
     opts->use_protocol_version = NULL;
+    opts->print_stats = false;
 }
 
 static void CFNetOptionsClear(CFNetOptions *opts)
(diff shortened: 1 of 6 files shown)


libcfnet/file_stream.c
```diff
@@ -1012,18 +1044,36 @@ static bool RecvDelta(
...
+    bool print_stats)
 {
     assert(conn != NULL);
     assert(basis != NULL);
     assert(dest != NULL);
 
-    /* Let's make sure the basis file exists */
-    FILE *file = safe_fopen_create_perms(basis, "wb", perms);
+    /* Let's make sure the basis file exists, but don't truncate it */
+    FILE *file = safe_fopen_create_perms(basis, "ab+", perms);
```
vpodzime, line 1076: Oops. :sweat_smile:  I believe this change deserves its own commit.

libcfnet/file_stream.c
```diff
@@ -829,6 +836,20 @@ static bool SendSignature(SSL *conn, const char *filename)
...
     rs_job_free(job);
 
+    const char *msg =
+        "Send signature statistics:\n"
+        "  %zu bytes in (read from '%s')\n"
+        "  %zu bytes out (sent to server)\n";
+
+    if (print_stats)
+    {
+        printf(msg, bytes_in, filename, bytes_out);
```
vpodzime, line 846: I'd personally do `fprintf(stderr, ...)`, but we generally don't use the error output so this is probably fine.

vpodzime (commented):
Looks good to me otherwise.

cf-net/cf-net.c
```diff
@@ -129,6 +131,7 @@ static const char *const HINTS[] =
     "Minimum TLS version to use",
     "TLS ciphers to use (comma-separated list)",
     "Specify CFEngine protocol to use. Possible values: 'classic', 'tls', 'cookie', 'filestream', 'latest' (default)",
+    "Print RSYNC performance statistics",
```
craigcomstock, line 134: Out of curiosity, why all caps? I don't see that on their website for example.

https://rsync.samba.org/

libcfnet/file_stream.c
```diff
@@ -829,6 +836,20 @@ static bool SendSignature(SSL *conn, const char *filename)
...
+    const char *msg =
+        "Send signature statistics:\n"
+        "  %zu bytes in (read from '%s')\n"
+        "  %zu bytes out (sent to server)\n";
+
+    if (print_stats)
+    {
+        fprintf(stderr, msg, bytes_in, filename, bytes_out);
+    }
+    else
```
craigcomstock, line 848: I wonder if we should always log and only optionally print to stderr. Probably doesn't matter much but seems more right to me somehow.

craigcomstock (approved):
cool.

cf-net/cf-net.c
larsewi: You're right :+1:

libcfnet/file_stream.c
larsewi: Either way is fine by me :+1:

craigcomstock (approved):
right on!

PR #5688, opened 2025-01-24
Adjusted cf-secret help, encrypt defaults to localhost key like decrypt does

craigcomstock:
// Default to localhost on encryption
    char *localhost = "127.0.0.1";
    if (encrypt && key_path_arg == NULL && host_arg == NULL)
    {
        host_arg = localhost;
    }

Ticket: none
Changelog: none

### DIFF
--- a/cf-secret/cf-secret.c
+++ b/cf-secret/cf-secret.c
@@ -126,14 +126,14 @@ static const char *const HINTS[] =
     "Specify how detailed logs should be. Possible values: 'error', 'warning', 'notice', 'info', 'verbose', 'debug'",
     "Enable basic information output",
     "Comma-separated list of key files to use (one of -k/-H options is required for encryption)",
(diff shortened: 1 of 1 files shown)


larsewi (approved):

cf-secret/cf-secret.c
```diff
@@ -126,14 +126,14 @@ static const char *const HINTS[] =
     "Specify how detailed logs should be. Possible values: 'error', 'warning', 'notice', 'info', 'verbose', 'debug'",
     "Enable basic information output",
     "Comma-separated list of key files to use (one of -k/-H options is required for encryption)",
-    "Comma-separated list of hosts to encrypt/decrypt for (defaults to 'localhost' for decryption)",
+    "Comma-separated list of hosts to encrypt/decrypt for (defaults to 'localhost')",
```
vpodzime, line 129: The line above says there's no default for encryption.

olehermanse: https://github.com/cfengine/core/pull/5690

PR #5694, opened 2025-02-06
ENT-10961, CFE-1840: Files promise can now modify immutable bit in file system attributes

larsewi:
Merge together with https://github.com/NorthernTechHQ/libntech/pull/235

[Build Status](https://ci.cfengine.com/job/pr-pipeline/11785/)

### DIFF
--- a/libpromises/mod_files.c
+++ b/libpromises/mod_files.c
@@ -227,6 +227,15 @@ static const ConstraintSyntax delete_constraints[] =
 
 static const BodySyntax delete_body = BodySyntaxNew("delete", delete_constraints, NULL, SYNTAX_STATUS_NORMAL);
 
+static const ConstraintSyntax fsattrs_constraints[] =
+{
+    CONSTRAINT_SYNTAX_GLOBAL,
+    ConstraintSyntaxNewBool("immutable", "true to set / false to clear the immutable flag", SYNTAX_STATUS_NORMAL),
+    ConstraintSyntaxNewNull()
+};
+
+static const BodySyntax fsattrs_body = BodySyntaxNew("fsattrs", fsattrs_constraints, NULL, SYNTAX_STATUS_NORMAL);
+
 static const ConstraintSyntax rename_constraints[] =
 {
     CONSTRAINT_SYNTAX_GLOBAL,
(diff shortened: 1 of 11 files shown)


cf-bottom:
Thanks for submitting a pull request! Maybe @craigcomstock can review this?

larsewi:
I'm struggling with acceptance tests. So I moved them to a separate [PR](https://github.com/cfengine/core/pull/5705) so that we can hopefully make this one ready for review.

tests/acceptance/10_files/14_immutable/00_immutable.cf
```diff
@@ -0,0 +1,64 @@
+body common control
+{
+  bundlesequence  => { "init", "test", "check", "clean" };
```
craigcomstock, line 3: doesn't the dcs framework handle this sequence?

There is "destroy" which you could use instead of clean. FWIW.

https://github.com/cfengine/core/blob/master/tests/acceptance/dcs.cf.sub#L208-L221

tests/acceptance/10_files/14_immutable/00_immutable.cf
```diff
@@ -0,0 +1,64 @@
+body common control
```
craigcomstock, line 1: I would find it VERY helpful to have a comment near the top of each test saying what use case it is testing like:

- test that immutable file cannot be changed by a files promise without fsattrs attribute
- test that mutable file can be changed by a files promise without fsattrs attribute
- test that immutable file CAN be changed by a files promise which includes the fsattrs attribute

libpromises/mod_files.c
```diff
@@ -227,6 +227,15 @@ static const ConstraintSyntax delete_constraints[] =
 
 static const BodySyntax delete_body = BodySyntaxNew("delete", delete_constraints, NULL, SYNTAX_STATUS_NORMAL);
 
+static const ConstraintSyntax fsattrs_constraints[] =
+{
+    CONSTRAINT_SYNTAX_GLOBAL,
+    ConstraintSyntaxNewBool("immutable", "true/false whether or not to immutable flag should be set", SYNTAX_STATUS_NORMAL),
```
craigcomstock, line 233: s/or not to/or not the/

or

"true/false whether or not to set the immutable flag"

craigcomstock (commented):
looks good! A few nitpicks.

tests/acceptance/10_files/14_immutable/00_immutable.cf
larsewi: I think it's nice to be able to execute a stand-alone file. A bit unnecessary to include the dcs framework just to get a predefined bundle sequence. But I'm open to include it, if you think that's better. Maybe there are some other benefits that I'm not aware of.

tests/acceptance/10_files/14_immutable/00_immutable.cf
larsewi: Yup, totally agree 🚀

tests/acceptance/10_files/14_immutable/00_immutable.cf
nickanderson: dcs does some other things with test metadata IIRC. It's not critical to have and it does add some overhead. Probably the nice things dcs does should be elsewhere.

PR #5703, opened 2025-03-02
CFE-4019: Added getbundlemetatags policy function

victormlg:
Merge together:
https://github.com/cfengine/core/pull/5703
https://github.com/cfengine/documentation/pull/3402

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -1563,9 +1563,48 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
 }
 
 /*********************************************************************/
-
-static FnCallResult FnCallGetMetaTags(EvalContext *ctx, ARG_UNUSED const Policy *policy, const FnCall *fp, const Rlist *finalargs)
+static Bundle *GetBundleFromPolicy(const Policy *policy, const char *namespace, const char *bundlename)
 {
+    assert(policy != NULL);
+    const size_t bundles_length = SeqLength(policy->bundles);
+
+    for (size_t i = 0; i < bundles_length; i++)
+    {
+        Bundle *bp = SeqAt(policy->bundles, i);
+        if (StringEqual(bp->name, bundlename) && StringEqual(bp->ns, namespace))
+        {
+            return bp;
+        }
+    }
+    return NULL;
+}
+
+static Promise *GetPromiseFromBundle(const Bundle *bundle, const char *promise_type, const char *promiser)
+{
+    assert(bundle != NULL);
+    const size_t sections_length = SeqLength(bundle->sections);
+    for (size_t i = 0; i < sections_length; i++)
+    {
+        BundleSection *bsection = SeqAt(bundle->sections, i);
+        if (StringEqual(bsection->promise_type, promise_type))
+        {
+            const size_t promises_length = SeqLength(bsection->promises);
+            for (size_t i = 0; i < promises_length; i++)
+            {
+                Promise *promise = SeqAt(bsection->promises, i);
+                if (StringEqual(promise->promiser, promiser))
+                {
+                    return promise;
+                }
+            }
+        }
+    }
+    return NULL;
+}
+
+static FnCallResult FnCallGetMetaTags(EvalContext *ctx, const Policy *policy, const FnCall *fp, const Rlist *finalargs)
+{
+    assert(fp != NULL);
     if (!finalargs)
     {
         FatalError(ctx, "Function '%s' requires at least one argument", fp->name);
@@ -1574,17 +1613,62 @@ static FnCallResult FnCallGetMetaTags(EvalContext *ctx, ARG_UNUSED const Policy 
     Rlist *tags = NULL;
     StringSet *tagset = NULL;
 
-    if (strcmp(fp->name, "getvariablemetatags") == 0)
+    if (StringEqual(fp->name, "getvariablemetatags"))
     {
         VarRef *ref = VarRefParse(RlistScalarValue(finalargs));
         tagset = EvalContextVariableTags(ctx, ref);
         VarRefDestroy(ref);
     }
-    else if (strcmp(fp->name, "getclassmetatags") == 0)
+    else if (StringEqual(fp->name, "getclassmetatags"))
     {
         ClassRef ref = ClassRefParse(RlistScalarValue(finalargs));
         tagset = EvalContextClassTags(ctx, ref.ns, ref.name);
         ClassRefDestroy(ref);
+    }
+    else if (StringEqual(fp->name, "getbundlemetatags"))
+    {
+        const char *bundleref = RlistScalarValue(finalargs);
+        assert(bundleref != NULL);
+        const Rlist *args = RlistFromSplitString(bundleref, ':');
+        const char *namespace = (args->next == NULL) ? "default" : RlistScalarValue(args);
+        const char *name = RlistScalarValue((args->next == NULL) ? args : args->next);
+
+        const Bundle *bundle = GetBundleFromPolicy(policy, namespace, name);
+        if (bundle == NULL)
+        {
+            Log(LOG_LEVEL_ERR,
+                "Function %s couldn't find bundle '%s' with namespace '%s'",
+                fp->name,
+                name,
+                namespace);
+            return FnFailure();
+        }
+        const Promise *promise = GetPromiseFromBundle(bundle, "meta", "tags");
+        if (bundle == NULL)
+        {
+            Log(LOG_LEVEL_ERR,
+                "Function %s couldn't find meta tags in '%s:%s'",
+                fp->name,
+                namespace,
+                name);
+            return FnFailure();
+        }
+        Rlist *start = PromiseGetConstraintAsList(ctx, "slist", promise);
+        if (start == NULL)
+        {
+            Log(LOG_LEVEL_ERR,
+                "Function %s couldn't find meta tags constraint string list",
+                fp->name);
+            return FnFailure();
+        }
+
+        tagset = StringSetNew();
+        Rlist *temp = start;
+        while (temp != NULL)
+        {
+            StringSetAdd(tagset, temp->val.item);
+            temp = temp->next;
+        }
     }
     else
     {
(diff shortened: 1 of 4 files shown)


libpromises/evalfunction.c
```diff
@@ -1563,8 +1563,45 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
...
 /*********************************************************************/
+static Bundle *GetBundleFromPolicy(const Policy *policy, const char *namespace, const char *bundlename)
+{
+    const size_t bundles = SeqLength(policy->bundles);
 
-static FnCallResult FnCallGetMetaTags(EvalContext *ctx, ARG_UNUSED const Policy *policy, const FnCall *fp, const Rlist *finalargs)
+    for (size_t i = 0; i < bundles; i++)
+    {
+        Bundle *bp = SeqAt(policy->bundles, i);
+        if ((strcmp(bp->name, bundlename) || strcmp(bp->ns, namespace)) == 0)
```
larsewi, line 1573:
```suggestion
        if (StringEqual(bp->name, bundlename) && StringEqual(bp->ns, namespace))
```
`StringEqual()` and `&&` would be less confusing in my opinion.

libpromises/evalfunction.c
```diff
@@ -1563,8 +1563,45 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
...
+        {
+            return bp;
+        }
+    }
+    return NULL;
+}
+
+static Promise *GetPromiseFromBundle(const Bundle *bundle, const char *promise_type, const char *promiser)
+{
```
larsewi, line 1582:
```suggestion
{
    assert(bundle != NULL);
```

libpromises/evalfunction.c
```diff
@@ -1563,8 +1563,45 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
...
+    const size_t sections_length = SeqLength(bundle->sections);
```
larsewi, line 1583: I would use `num_sections`, but either way is fine by me ;)

libpromises/evalfunction.c
```diff
@@ -1563,8 +1563,45 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
...
+    for (size_t i = 0; i < sections_length; i++)
+    {
+        BundleSection *bsection = SeqAt(bundle->sections, i);
+        if (strcmp(bsection->promise_type, promise_type) == 0 )
```
larsewi, line 1587:
```suggestion
        if (StringEqual(bsection->promise_type, promise_type))
```

libpromises/evalfunction.c
```diff
@@ -1563,8 +1563,45 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
...
+        {
+            const size_t promises_length = SeqLength(bsection->promises);
```
larsewi, line 1589: Same here; `num_promises`

libpromises/evalfunction.c
```diff
@@ -1563,8 +1563,45 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
...
+            for (size_t i = 0; i < promises_length; i++)
+            {
+                Promise *promise = SeqAt(bsection->promises, i);
+                if (strcmp(promise->promiser, promiser) == 0)
+                {
+
+                    return promise;
```
larsewi, line 1596:
```suggestion
                    return promise;
```

libpromises/evalfunction.c
```diff
@@ -1563,8 +1563,45 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
...
+                }
+            }
+        }
+    }
+    return NULL;
+}
+
+static FnCallResult FnCallGetMetaTags(EvalContext *ctx, const Policy *policy, const FnCall *fp, const Rlist *finalargs)
 {
```
larsewi, line 1605:
```suggestion
{
    assert(fp != NULL);
```

libpromises/evalfunction.c
```diff
@@ -1586,6 +1623,44 @@ static FnCallResult FnCallGetMetaTags(EvalContext *ctx, ARG_UNUSED const Policy
         tagset = EvalContextClassTags(ctx, ref.ns, ref.name);
         ClassRefDestroy(ref);
     }
+    else if (strcmp(fp->name, "getbundlemetatags") == 0)
```
larsewi, line 1626:
```suggestion
    else if (StringEqual(fp->name, "getbundlemetatags"))    
```

libpromises/evalfunction.c
```diff
@@ -1586,6 +1623,44 @@ static FnCallResult FnCallGetMetaTags(EvalContext *ctx, ARG_UNUSED const Policy
         tagset = EvalContextClassTags(ctx, ref.ns, ref.name);
         ClassRefDestroy(ref);
     }
+    else if (strcmp(fp->name, "getbundlemetatags") == 0)    
+    {
+        const char *bundleref = RlistScalarValue(finalargs);
+        const Rlist *args = RlistFromSplitString(bundleref, ':');
```
larsewi, line 1629: `RlistFromSplitString()` can return `NULL` if the first argument is `NULL`. This should not really happen, because `RlistScalarValue()` does not return `NULL`. However, an `assert()` would be nice, to test this assumption.

libpromises/evalfunction.c
```diff
@@ -1586,6 +1623,44 @@ static FnCallResult FnCallGetMetaTags(EvalContext *ctx, ARG_UNUSED const Policy
...
+        const char *namespace;
+        const char *name;
+        if (!args->next)
+        {
+            namespace = "default";
+            name = RlistScalarValue(args);
+        } else {
+            namespace = RlistScalarValue(args);
+            name = RlistScalarValue(args->next);
+        }
```
larsewi, line 1639:
```suggestion
        const char *namespace = (!args->next) ? "default" : RlistScalarValue(args);
        const char *name = (!args->next) ? RlistScalarValue(args) : RlistScalarValue(args->next);
```
You can consider using the ternary operator here (since we are not payed per line of code).

libpromises/evalfunction.c
```diff
@@ -1586,6 +1623,44 @@ static FnCallResult FnCallGetMetaTags(EvalContext *ctx, ARG_UNUSED const Policy
...
+        const Bundle *bundle =  GetBundleFromPolicy(policy, namespace, name);
+        if (!bundle)
+        {
+            FatalError(ctx, "Couldn't find bundle '%s' with namespace '%s'", name, namespace);
```
larsewi, line 1643: Returning `FnFailure()` is probably more appropriate here. `FatalError()` will stop the execution of the agent, and the remaining policy will not be evaluated. I believe `FatalError()` should only be used when it is not safe to continue execution.

libpromises/evalfunction.c
```diff
@@ -1586,6 +1623,44 @@ static FnCallResult FnCallGetMetaTags(EvalContext *ctx, ARG_UNUSED const Policy
...
+        if (!args->next)
+        {
+            namespace = "default";
+            name = RlistScalarValue(args);
+        } else {
+            namespace = RlistScalarValue(args);
+            name = RlistScalarValue(args->next);
+        }
+        const Bundle *bundle =  GetBundleFromPolicy(policy, namespace, name);
+        if (!bundle)
```
larsewi, line 1641:
```suggestion
        if (bundle == NULL)
```
Please use explicit comparisons as described in [CONTRIBUTING.md](https://github.com/cfengine/core/blob/master/CONTRIBUTING.md)

libpromises/evalfunction.c
```diff
@@ -1586,6 +1623,44 @@ static FnCallResult FnCallGetMetaTags(EvalContext *ctx, ARG_UNUSED const Policy
...
+            FatalError(ctx, "Couldn't find meta tags in '%s:%s'", namespace, name);
+        }
+        Rlist *start = PromiseGetConstraintAsList(ctx, "slist", promise);
+        if (!start)
+        {
+            FatalError(ctx, "Couldn't find meta tags constraint string list");
+        }
+
+        tagset = StringSetNew();
+        Rlist *temp=start;
```
larsewi, line 1657:
```suggestion
        Rlist *temp = start;
```

libpromises/evalfunction.c
```diff
@@ -1586,6 +1623,44 @@ static FnCallResult FnCallGetMetaTags(EvalContext *ctx, ARG_UNUSED const Policy
...
+        while(temp != NULL)
+        {
+            StringSetAdd(tagset, temp->val.item);
+            temp= temp->next;
```
larsewi, line 1661:
```suggestion
            temp = temp->next;
```

larsewi (changes requested):
Great work :rocket: Added some comments that you will need to address

libpromises/evalfunction.c
```diff
@@ -1563,8 +1563,45 @@ static FnCallResult FnCallVariablesMatching(EvalContext *ctx, ARG_UNUSED const P
 }
 
 /*********************************************************************/
+static Bundle *GetBundleFromPolicy(const Policy *policy, const char *namespace, const char *bundlename)
+{
```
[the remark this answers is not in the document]
olehermanse: @victormlg some context; When you are writing this code, you are assuming that `policy` is a pointer (address) of a `Policy` struct which you can dereference to access the fields of `Policy`. Something before this, whoever is calling this function, should ensure that is the case, and that `policy` is not uninitialized, or `NULL` for example.

Our way of codifying this assumption is to put this at the start of the function:

```
assert(policy != NULL);
```

It both acts as an indication to the reader (human or tool) that this function expects a valid non-NULL pointer, and as a check, in debug builds, to try to find mistakes. (It's not "perfect", since non-initialized memory can be something other than NULL, and also it's skipped in release builds, for performance).

libpromises/evalfunction.c
olehermanse: This one is not the same as the others, this one says `_length`, that's nice and specific, arguably better than `num`.

libpromises/evalfunction.c
olehermanse: Or even use it inside params;

```
const char *name = RlistScalarValue((args->next == NULL) ? args : args->next);
```

(Also, our code style guidelines say to explicitly compare with NULL).

larsewi (approved):
Great work :rocket:

larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11821/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11821/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11821/

PR #5706, opened 2025-03-06
2025 03 06 static check fixes

craigcomstock:


tests/static-check/run_checks.sh
```diff
@@ -5,25 +5,25 @@ set -x
...
+  ./configure --enable-debug CC=gcc
   local gcc_exceptions="-Wno-sign-compare -Wno-enum-int-mismatch"
   make -j -l${n_procs} --keep-going CFLAGS="-Werror -Wall -Wextra $gcc_exceptions"
 }
 
 function check_with_clang() {
-  rm -f config.cache
   make clean
-  ./configure -C --enable-debug CC=clang
+  ./configure --enable-debug CC=clang
```
craigcomstock, line 16: I have added more about why this change in the commit comment but don't want to waste GH job resources on that just yet.

larsewi (dismissed):
Thanks for doing this :rocket:

tests/static-check/run_checks.sh
vpodzime: Curious to see that because `-C` after `config.cache` should only ensure the checks for *core* are reused for *libntech*, AFAICT.

vpodzime (commented):
Looks good to me, assuming the drop of `-C` is justified :wink:

tests/static-check/run_checks.sh
craigcomstock: ah, right, I didn't think of libntech benefiting from core, does it? or would the cache be separate since it is a subdir? I will test actually. :+1:

craigcomstock:
> Looks good to me, assuming the drop of `-C` is justified 😉

Right. It is only justified in libntech, not here. Thanks for mentioning libntech re-using core's config results.

larsewi (approved):
Great work :rocket:

craigcomstock:
cherry picks
https://github.com/cfengine/core/pull/5711
https://github.com/cfengine/core/pull/5710

tests/static-check/run_checks.sh
```diff
@@ -43,6 +52,13 @@ cd "$(dirname $0)"/../../
 
 failure=0
 failures=""
+
+# in jenkins the workdir is already autogen'd
+# in github it is not, so do that work here
+if [ ! -f configure ]; then
+  ./autogen.sh --enable-debug
```
vpodzime, line 59: This runs `./configure` as well. We don't need that, AFAIK, so it should be prefixed with `NO_CONFIGURE=1`.

craigcomstock: right.

PR #5702, opened 2025-02-28
CFE-3429: Use current process ID to investigate proc filesystem to workaround in-container non-root owned symlinks

craigcomstock:
together
https://github.com/cfengine/masterfiles/pull/2990
https://github.com/cfengine/core/pull/5702

PR #5707, opened 2025-03-06
ENT-12551: ensure that community CFEngine works well in a container without errors

craigcomstock:
- Minimized resources for static-check (package dependencies) and added cppcheck version printout
- address static-check complaints
- Added librsync-dev package dependency for alpine linux build from source
- Use current process ID to investigate proc filesystem to workaround in-container non-root owned symlinks
- Added error message when netstat file is not found during UNIX interface discovery.
- Move GetRelocatedProcdirRoot() from libntech and add GetProcdirPid()
- Added env var pid override in proc-net acceptance test
- sync with libntech pr commit, delete me after that is merged
- Added test for running community in alpine container

### DIFF
--- a/libenv/unix_iface.c
+++ b/libenv/unix_iface.c
@@ -1286,7 +1286,7 @@ static JsonElement* GetNetworkingStatsInfo(const char *filename)
     }
     else
     {
-        Log(LOG_LEVEL_ERROR, "netstat file not found at '%s'", filename);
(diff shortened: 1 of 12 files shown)


craigcomstock:
Includes commits from two other PRs :( which will disappear when/if those are merged

https://github.com/cfengine/core/pull/5702
https://github.com/cfengine/core/pull/5706

PR #5711, opened 2025-03-10
2025 03 06 static check fixes/3.21

craigcomstock:
- Minimized resources for static-check (package dependencies) and added cppcheck version printout
- address static-check complaints

Back ported from https://github.com/cfengine/core/pull/5706

### DIFF
--- a/.github/workflows/job-static-check.yml
+++ b/.github/workflows/job-static-check.yml
@@ -46,13 +46,7 @@ jobs:
     - name: Prepare Environment
       run: |
         sudo apt-get update && \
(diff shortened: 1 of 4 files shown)


craigcomstock:
bumping libntech to master in 3.21.x here is tricky due to this move of some code from libntech/libcompat to core/libcfecompat: https://github.com/cfengine/core/commit/a238bec62a716d154be830f42f4c8209aa2746ee Will investigate a bit.

larsewi:
Yup, it will be difficult. Maybe you can cherry-pick the commits that you need or manually reapply them in the current context?

PR #5729, opened 2025-03-21
cf-agent/verify_files: Drop duplicate check on present select attribute

btriller:


cf-bottom:
Thanks for submitting a pull request! Maybe @craigcomstock can review this?

larsewi:
Thanks @btriller :rocket: I wonder if the check you removed has a more understandable error message than the one you kept. I.e.:

```C
    if ((a->haveselect) && (!a->select.result))
    {
        Log(LOG_LEVEL_ERR, "File select constraint body promised no result (check body definition)");
        PromiseRef(LOG_LEVEL_ERR, pp);
        return false;
    }
```

Then again, by removing the last check (like you did) nothing really changes for the users. @nickanderson what do you think?

larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11881/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11881/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11881/

craigcomstock:
I have some doubts about our documentation versus this code. The `issymlinkto` example seems to not include any `file_result` and so currently causes an error but according to the docs it would seem a "valid" `file_select` body.

https://docs.cfengine.com/docs/3.24/reference-promise-types-files.html#issymlinkto

I am looking into it.

nickanderson:
> I have some doubts about our documentation versus this code. The `issymlinkto` example seems to not include any `file_result` and so currently causes an error but according to the docs it would seem a "valid" `file_select` body.
> 
> https://docs.cfengine.com/docs/3.24/reference-promise-types-files.html#issymlinkto
> 
> I am looking into it.

My instinct is that if `file_result` is not specified it would be an AND of each specified file select attribute

PR #5751, opened 2025-04-01
libenv/sysinfo: Fix detected os description

btriller:
Use platform context variable also for logging the os class description. Before this change, it were using loop counter variable, which is incremented again. This led to verbose logging message:
 verbose: CFEngine detected operating system description is qnx
on Darwin/macOS.

Also use already lowercased sysname from utsname struct.

Fixes: e2c9f00c0 ("Use a proper 'bool' variable for a condition in GetNameInfo3()")

### DIFF
--- a/libenv/sysinfo.c
+++ b/libenv/sysinfo.c
@@ -549,13 +549,9 @@ static void GetNameInfo3(EvalContext *ctx)
     bool found = false;
     for (i = 0; !found && (i < PLATFORM_CONTEXT_MAX); i++)
     {
-        char sysname[CF_BUFSIZE];
-        strlcpy(sysname, VSYSNAME.sysname, CF_BUFSIZE);
-        ToLowerStrInplace(sysname);
-
         /* FIXME: review those strcmps. Moved out from StringMatch */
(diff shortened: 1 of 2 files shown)


libenv/sysinfo.c
```diff
@@ -549,13 +549,9 @@ static void GetNameInfo3(EvalContext *ctx)
     bool found = false;
     for (i = 0; !found && (i < PLATFORM_CONTEXT_MAX); i++)
     {
```
larsewi, line 551: It would comfort me if we also test the assumption that `VSYSNAME.sysname` is in fact already lower case.
```suggestion
    {
#ifndef NDEBUG
        for (const char *ch = VSYSNAME.sysname; *ch != '\0'; ch++)
        {
            assert(islower(*ch));
        }
#endif /* NDEBUG */
```

larsewi (commented):
Great :rocket: Thanks!

larsewi (approved):
Thanks :rocket:

craigcomstock (approved):
seems reasonable. :) thanks!

olehermanse:
@cf-bottom jenkins, please

PR #5752, opened 2025-04-01
ENT-10961, CFE-1840: Files promise can now modify immutable bit in file system attributes

larsewi:
TODO:
 - [x] handle `content` attribute
 - [x] handle `copy_from` attribute
 - [x] handle `delete` attribute
 - [x] handle `edit_line` attribute
 - [x] handle `edit_xml` attribute
 - [x] handle `perms` attribute
 - [x] handle `touch` attribute
 - [x] handle `edit_template` attribute
 - [x] handle `acl` attribute
 - [x] handle `transformer` attribute
 - [x] handle `rename` attribute

Is there any other ones that I've missed?

[Build Status](https://ci.cfengine.com/job/pr-pipeline/12186/)

### DIFF
--- a/cf-agent/verify_files.c
+++ b/cf-agent/verify_files.c
@@ -60,6 +60,7 @@
 #include <evalfunction.h>
 #include <changes_chroot.h>     /* PrepareChangesChroot(), RecordFileChangedInChroot() */
 #include <cf3.defs.h>
+#include <fsattrs.h>
 
 static PromiseResult FindFilePromiserObjects(EvalContext *ctx, const Promise *pp);
 static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promise *pp);
@@ -345,6 +346,62 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
         changes_path = chrooted_path;
     }
 
+    bool is_immutable = false; /* We assume not in case of failure */
+    FSAttrsResult res = FSAttrsGetImmutableFlag(changes_path, &is_immutable);
+    if (res != FS_ATTRS_SUCCESS)
+    {
+        Log((res == FS_ATTRS_FAILURE) ? LOG_LEVEL_ERR : LOG_LEVEL_VERBOSE,
+            "Failed to get the state of the immutable bit from file '%s': %s",
+            changes_path, FSAttrsErrorCodeToString(res));
+    }
+
+    if (a.havefsattrs && a.fsattrs.haveimmutable && !a.fsattrs.immutable)
+    {
+        /* Here we only handle the clearing of the immutable the immutable
+         * bit. Later we'll handle the setting of the immutable bit. */
+        if (is_immutable)
+        {
+            res = FSAttrsUpdateImmutableFlag(changes_path, false);
+            switch (res)
+            {
+            case FS_ATTRS_SUCCESS:
+                RecordChange(ctx, pp, &a,
+                             "Cleared the immutable bit on file '%s'",
+                             changes_path);
+                result = PromiseResultUpdate(result, PROMISE_RESULT_CHANGE);
+                break;
+            case FS_ATTRS_FAILURE:
+                RecordFailure(ctx, pp, &a,
+                              "Failed to clear the immutable bit on file '%s'",
+                              changes_path);
+                result = PromiseResultUpdate(result, PROMISE_RESULT_FAIL);
+                break;
+            case FS_ATTRS_NOT_SUPPORTED:
+                /* We will not treat this as a promise failure because this
+                 * will happen on many platforms and filesystems. Instead we
+                 * will log a verbose message to make it apparent for the
+                 * users. */
+                Log(LOG_LEVEL_VERBOSE,
+                    "Failed to clear the immutable bit on file '%s': %s",
+                    changes_path, FSAttrsErrorCodeToString(res));
+                break;
+            case FS_ATTRS_DOES_NOT_EXIST:
+                /* File does not exist. Nothing to do really, but let's log a
+                 * debug message for good measures */
+                Log(LOG_LEVEL_DEBUG,
+                    "Failed to clear the immutable bit on file '%s': %s",
+                    changes_path, FSAttrsErrorCodeToString(res));
+                break;
+            }
+        }
+        else
+        {
+            RecordNoChange(ctx, pp, &a,
+                           "The immutable bit is not set on file '%s' as promised",
+                           changes_path);
+        }
+    }
+
     if (lstat(changes_path, &oslb) == -1)       /* Careful if the object is a link */
     {
         if ((a.create) || (a.touch))
@@ -583,6 +640,51 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
             Log(LOG_LEVEL_VERBOSE,
                 "Cannot render file '%s': file does not exist", path);
             goto exit;
+        }
+    }
+
+    if (a.havefsattrs && a.fsattrs.haveimmutable && a.fsattrs.immutable)
+    {
+        /* Here we only handle the setting of the immutable bit. Previously we
+         * handled the clearing of the immutable bit. */
+        if (is_immutable)
+        {
+            RecordNoChange(ctx, pp, &a,
+                           "The immutable bit is already set on file '%s' as promised",
+                           changes_path);
+        }
+        else
+        {
+            res = FSAttrsUpdateImmutableFlag(changes_path, true);
+            switch (res)
+            {
+            case FS_ATTRS_SUCCESS:
+                Log(LOG_LEVEL_VERBOSE, "Set the immutable bit on file '%s'",
+                    changes_path);
+                break;
+            case FS_ATTRS_FAILURE:
+                /* Things still may be fine as long as the agent does not try to mutate the file */
+                Log(LOG_LEVEL_VERBOSE,
+                    "Failed to set the immutable bit on file '%s': %s",
+                    changes_path, FSAttrsErrorCodeToString(res));
+                break;
+            case FS_ATTRS_NOT_SUPPORTED:
+                /* We will not treat this as a promise failure because this
+                 * will happen on many platforms and filesystems. Instead we
+                 * will log a verbose message to make it apparent for the
+                 * users. */
+                Log(LOG_LEVEL_VERBOSE,
+                    "Failed to set the immutable bit on file '%s': %s",
+                    changes_path, FSAttrsErrorCodeToString(res));
+                break;
+            case FS_ATTRS_DOES_NOT_EXIST:
+                /* File does not exist. Nothing to do really, but let's log a
+                 * debug message for good measures */
+                Log(LOG_LEVEL_DEBUG,
+                    "Failed to set the immutable bit on file '%s': %s",
+                    changes_path, FSAttrsErrorCodeToString(res));
+                break;
+            }
         }
     }
 
--- a/cf-agent/verify_files.c
+++ b/cf-agent/verify_files.c
@@ -61,11 +61,12 @@
 #include <changes_chroot.h>     /* PrepareChangesChroot(), RecordFileChangedInChroot() */
 #include <cf3.defs.h>
 #include <fsattrs.h>
+#include <override_fsattrs.h>
 
 static PromiseResult FindFilePromiserObjects(EvalContext *ctx, const Promise *pp);
 static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promise *pp);
 static PromiseResult WriteContentFromString(EvalContext *ctx, const char *path, const Attributes *attr,
-                                            const Promise *pp);
+                                            const Promise *pp, bool override_immutable);
 
 /*****************************************************************************/
 
@@ -402,6 +403,11 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
         }
     }
 
+    /* If we encounter any promises to mutate the file and the immutable
+     * attribute in body fsattrs is "true", we will override the immutable bit
+     * by temporarily clearing it when ever needed. */
+    const bool override_immutable = a.havefsattrs && a.fsattrs.haveimmutable && a.fsattrs.immutable && is_immutable;
+
     if (lstat(changes_path, &oslb) == -1)       /* Careful if the object is a link */
     {
         if ((a.create) || (a.touch))
@@ -611,7 +617,7 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
         Log(LOG_LEVEL_VERBOSE, "Replacing '%s' with content '%s'",
             path, a.content);
 
-        PromiseResult render_result = WriteContentFromString(ctx, path, &a, pp);
+        PromiseResult render_result = WriteContentFromString(ctx, path, &a, pp, override_immutable);
         result = PromiseResultUpdate(result, render_result);
 
         goto exit;
@@ -760,7 +766,7 @@ skip:
 /*****************************************************************************/
 
 static PromiseResult WriteContentFromString(EvalContext *ctx, const char *path, const Attributes *attr,
-                                            const Promise *pp)
+                                            const Promise *pp, bool override_immutable)
 {
     assert(path != NULL);
     assert(attr != NULL);
@@ -795,20 +801,28 @@ static PromiseResult WriteContentFromString(EvalContext *ctx, const char *path, 
             return result;
         }
 
-        FILE *f = safe_fopen(changes_path, "w");
-        if (f == NULL)
+        char override_path[PATH_MAX];
+        if (!OverrideImmutableBegin(changes_path, override_path, sizeof(override_path), override_immutable))
         {
-            RecordFailure(ctx, pp, attr, "Cannot open file '%s' for writing", path);
+            RecordFailure(ctx, pp, attr, "Failed to override immutable bit on file '%s'", changes_path);
             return PromiseResultUpdate(result, PROMISE_RESULT_FAIL);
         }
 
+        FILE *f = safe_fopen(override_path, "w");
+        if (f == NULL)
+        {
+            RecordFailure(ctx, pp, attr, "Cannot open file '%s' for writing", path);
+            OverrideImmutableCommit(changes_path, override_path, override_immutable, true);
+            return PromiseResultUpdate(result, PROMISE_RESULT_FAIL);
+        }
+
+        bool override_abort = false;
         Writer *w = FileWriter(f);
         if (WriterWriteLen(w, attr->content, bytes_to_write) == bytes_to_write )
         {
             RecordChange(ctx, pp, attr,
                          "Updated file '%s' with content '%s'",
                          path, attr->content);
-
             result = PromiseResultUpdate(result, PROMISE_RESULT_CHANGE);
         }
         else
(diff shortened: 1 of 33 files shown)


nickanderson:
MMM, the transformer attribute comes to mind.

```
  bundle agent __main__
  {
    files:
        "/tmp/example.txt"
          content => "Hello World";
  
        "/tmp/example.txt"
          transformer => "/usr/bin/gzip $(this.promiser)";
  
    reports:
        "$(with)" with => join( ",", lsdir( "/tmp", "example.*", "true" ));
  }
```
Looking at the files promise docs I imagine that `create` would also be affected:

```cfengine3
bundle agent __main__
{
  vars:
      "check_file" string => "/tmp/check-file.txt";

  files:
      "$(check_file)"
        handle => "create_file",
        create => "true",
        if => "file_absent_report_kept";

   reports:
      "$(check_file) does not exist"
        classes => results( "namespace", "file_absent_report" );

      "$(check_file) was repaired"
        depends_on => { "create_file" };

      "$(check_file) exists"
        if => fileexists( "/tmp/check-file.txt" );
}
```

    R: /tmp/check-file.txt does not exist
    R: /tmp/check-file.txt was repaired
    R: /tmp/check-file.txt exists
    
And the `delete` attribute would seem to be affected as well.

And `rename` too

```cfengine3
bundle agent __main__
{
  files:
    "/tmp/example.txt"
      content => "Disable me!";

    "/tmp/example.txt"
      rename => default:disable;
}
# For reference:
# body rename disable
# # @brief Disable the file
# {
#         disable => "true";
# }
```

       info: Created file '/tmp/example.txt', mode 0600
       info: Updated file '/tmp/example.txt' with content 'Disable me!'
    warning: File object '/tmp/example.txt' exists, contrary to promise
       info: Changed permissions of '/tmp/example.txt' to 'mode 0600'
       info: Disabled file '/tmp/example.txt' by renaming to '/tmp/example.txt.cfdisabled' with mode 0600

larsewi:

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12196/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12196/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12196/

larsewi:

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12213/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12213/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12213/

libpromises/override_fsattrs.c
```diff
@@ -0,0 +1,140 @@
...
+
+    return true;
+}
+
+bool OverrideImmutableCommit(const char *orig, const char *copy)
+{
+    struct stat sb;
+    if (lstat(orig, &sb) == -1)
+    {
+        Log(LOG_LEVEL_ERR, "Failed to stat file '%s'", orig);
```
craigcomstock: Seems like this message is a bit too generic. Maybe something like "Failed to stat original file during immutable operations: '%s'".

libpromises/override_fsattrs.c
```diff
@@ -0,0 +1,140 @@
...
+        unlink(copy);
+        return false;
+    }
+
+    if (chmod(copy, sb.st_mode) == -1)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to change mode bits on file '%s' to %04jo: %s",
```
craigcomstock: same here, seems like adding a bit of context to the message would be helpful when debugging.

libpromises/override_fsattrs.c
```diff
@@ -0,0 +1,140 @@
...
+            GetErrorStr());
+        unlink(copy);
+        return false;
+    }
+
+    /* If the operations on the file system attributes fails for any reason,
+     * we can still proceed to try to replace the original file. We will only
+     * log an actual error in case of an unexpected failure (i.e., when
+     * FS_ATTRS_FAILURE is returned). Other failures will be logged as verbose
+     * messages because they can be useful, but are be quite verbose. */
```
craigcomstock: "but can be quite verbose."  e.g. s/but are/but can/ or... "but are quite verbose."

libpromises/override_fsattrs.c
```diff
@@ -0,0 +1,140 @@
...
+        Log((res == FS_ATTRS_FAILURE) ? LOG_LEVEL_ERR : LOG_LEVEL_VERBOSE,
+            "Failed to get immutable bit from file '%s': %s",
+            orig,
+            FSAttrsErrorCodeToString(res));
+    }
+
+    if (rename(copy, orig) == -1)
+    {
+        Log(LOG_LEVEL_ERR,
+            "Failed to replace original file '%s' with copy '%s'",
```
craigcomstock: maybe add "immutabe" context to the message like "During immutable bit manipulations failed to replace ..."

cf-agent/verify_files.c
```diff
@@ -344,6 +345,62 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
...
+                              changes_path);
+                result = PromiseResultUpdate(result, PROMISE_RESULT_FAIL);
+                break;
+            case FS_ATTRS_NOT_SUPPORTED:
+                /* We will not treat this as a promise failure because this
+                 * will happen on many platforms and filesystems. Instead we
+                 * will log a verbose message to make it apparent for the
+                 * users. */
+                Log(LOG_LEVEL_VERBOSE,
+                    "Failed to clear the immutable bit on file '%s': %s",
```
craigcomstock, line 386: seems this message should mention "not supported" and maybe not say Failed since nothing failed sort of? Maybe

"Platform does not support immutable bit so could not clear the immutable bit on file "%s': %s"

cf-agent/verify_files.c
```diff
@@ -344,6 +345,62 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
...
+                    changes_path, FSAttrsErrorCodeToString(res));
+                break;
+            case FS_ATTRS_DOES_NOT_EXIST:
+                /* File does not exist. Nothing to do really, but let's log a
+                 * debug message for good measures */
+                Log(LOG_LEVEL_DEBUG,
+                    "Failed to clear the immutable bit on file '%s': %s",
```
craigcomstock, line 393: shouldn't the file not existing cause a promise failure? I suppose you could have create => "true" that could create the file in a later agent run... Maybe add context here like "Failed to clear the immutable bit on non-existent file: '%s': %s"

cf-agent/verify_files.c
```diff
@@ -585,6 +642,51 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
...
+                    "Failed to set the immutable bit on file '%s': %s",
+                    changes_path, FSAttrsErrorCodeToString(res));
+                break;
+            case FS_ATTRS_DOES_NOT_EXIST:
+                /* File does not exist. Nothing to do really, but let's log a
+                 * debug message for good measures */
+                Log(LOG_LEVEL_DEBUG,
+                    "Failed to set the immutable bit on file '%s': %s",
+                    changes_path, FSAttrsErrorCodeToString(res));
+                break;
```
craigcomstock, line 692: again, for these three it seems useful to differentiate the messages and add context of failure, not supported, doesn't exist.

libpromises/override_fsattrs.h
```diff
@@ -33,26 +33,32 @@
  * @param orig The original file (may be immutable)
  * @param copy Updated to contain the filename of the mutable copy
  * @param copy_len The size of the buffer to store the filename of the copy
+ * @param override Whether to actually do override (original filename is
```
craigcomstock: do you mean overwrite? as in directly manipulate the file instead of making the temp copy?

libpromises/override_fsattrs.c
```diff
@@ -177,3 +177,50 @@ bool OverrideImmutableRename(
...
+                    "The immutable bit is not set on file '%s'",
+                    filename);
+            }
+        }
+        else
+        {
+            Log((res == FS_ATTRS_FAILURE) ? LOG_LEVEL_ERR : LOG_LEVEL_VERBOSE,
+                "Failed to get immutable bit from file '%s': %s",
+                filename,
+                FSAttrsErrorCodeToString(res));
```
craigcomstock: I think adding the context of "For file delete operation, ..." context would be nice. And remove "Failed" in preference for something else unless you will log as an error.

libpromises/evalfunction.c
```diff
@@ -9776,6 +9817,13 @@ static const FnCallArg AND_ARGS[] =
     {NULL, CF_DATA_TYPE_NONE, NULL}
 };
 
+static const FnCallArg GET_ACLS_ARGS[] =
+{
+    {CF_ABSPATHRANGE, CF_DATA_TYPE_STRING, "Path to file or directory"},
+    {"default,access", CF_DATA_TYPE_OPTION, "Whether to get default- or access ACL"},
```
craigcomstock: why "default-" here? and not just "Whether to get default or access ACL"?

craigcomstock (commented):
Yeah, this is a big one. There are a lot of commits here and a few refactorings. I know it would be some work but I wonder if you want to squash the commits a bit. Maybe not worth the effort though... Mostly I would say some logs need context if that is possible and I understand "override" now but you might want to add some comments around that to be clear what that is doing.

libpromises/override_fsattrs.h
craigcomstock: I see what override means now... override the immutable bit while you do the operation.

cf-agent/verify_files.c
larsewi: `FSAttrsErrorCodeToString(res)` will return `"Operation is not supported"`

cf-agent/verify_files.c
larsewi: `FSAttrsErrorCodeToString(res)` will return `"File does not exist"`

cf-agent/verify_files.c
larsewi: Yup, `FSAttrsErrorCodeToString(res)` will take care of it ;)

cf-agent/verify_files.c
larsewi:
```
        /* Here we only handle the clearing of the immutable the immutable
         * bit. Later we'll handle the setting of the immutable bit. */
```

Here we only clear the immutable bit. Later we set it. If the file does not exist, then there is no immutable bit to clear.

libpromises/override_fsattrs.c
larsewi: This rename is not related to the immutable bit operations/manipulations. It's more like the action that you want to perform after you have temporarily cleared the immutable bit. E.g., in the `rename` promise attribute.

libpromises/override_fsattrs.c
larsewi: This comment was removed at some point. So no point in fixing the typo

larsewi:

libpromises/override_fsattrs.h
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * @brief Temporarily clears the immutable bit of the original file and
+ * replaces it with the mutated copy
+ * @param orig The original file (may be immutable)
+ * @param copy The mutated copy to replace the original
+ * @return false in case of failure
+ * @note The immutable bit is reset to it's original state
+ */
+bool OverrideImmutableCommit(const char *orig, const char *copy);
```
craigcomstock: @larsewi in the design of these functions is there any safety in checking if the original file was modified between the time the Begin() was called and Commit() is called? I would think that might result in Commit() failing with an error that the original file changed while the immutable "lock" was being held.

libpromises/override_fsattrs.c
```diff
@@ -0,0 +1,140 @@
...
+    }
+    else
+    {
+        Log((res == FS_ATTRS_FAILURE) ? LOG_LEVEL_ERR : LOG_LEVEL_VERBOSE,
+            "Failed to get immutable bit from file '%s': %s",
+            orig,
+            FSAttrsErrorCodeToString(res));
+    }
+
+    if (rename(copy, orig) == -1)
```
craigcomstock: What I said in the function declaration in the header file. Maybe right here where you will rename() you could check the access time of the file "right now" (I know timing is everything here and there is still a window of non-atomicity) and if newer than a saved access time when running Begin() you could fail.

I am not sure how to get truly atomic here, other than possibly FORCING the originial file to be immutable! ;) Then nobody could change it until you Commit(). That's funny and ironic a bit to use immutable in the service of operating on an immutable file.

libpromises/override_fsattrs.c
```diff
@@ -0,0 +1,140 @@
...
+        Log(LOG_LEVEL_ERR,
+            "Failed to generate name for temporary copy of '%s': Filename is too long (%d >= %zu)",
+            orig,
+            ret,
+            copy_len);
+        return false;
+    }
+
+    /* We'll match the original file permissions on commit */
+    if (!CopyRegularFileDiskPerms(orig, copy, 0600))
```
craigcomstock, line 46: I think you should change the orig file to immutable just before this copy. Not 100% atomic but pretty close so that between this Begin() and the Commit() no one else can "easily" modify the file without also modifying the immutable bit of course.

libpromises/override_fsattrs.c
craigcomstock: I also understand that there is more non-atomic moments between when making the orig immutable, copying, then making it immutable again.

libpromises/override_fsattrs.h
```diff
@@ -0,0 +1,123 @@
...
+ * @brief Temporarily clears the immutable bit of the original file and
+ * replaces it with the mutated copy
+ * @param orig The original file (may be immutable)
+ * @param copy The mutated copy to replace the original
+ * @param override Whether to actually do override (i.e. temporarily clear it)
+ * @param abort Whether to abort the override
+ * @return false in case of failure
+ * @note The immutable bit is reset to it's original state
+ */
+bool OverrideImmutableCommit(
```
craigcomstock, line 56: Here it seems that we should have something like:

OverrideImmutableBegin(struct utimbuf *times) to capture the modified/access times when we begin the "transaction"

Also, during this Begin() we should make the file immutable to attempt to restrict access as much as possible while we do "work" and before we call Commit().

And then OverrideImmutableCommit() takes that as a parameter and if the original file has changed mod/access since Begin() we should fail to commit saying that the file changed inside of the window of time of the transaction to modify the immutable file.

cf-agent/verify_files.c
```diff
@@ -345,6 +347,67 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
...
+    if (res != FS_ATTRS_SUCCESS)
+    {
+        Log((res == FS_ATTRS_FAILURE) ? LOG_LEVEL_ERR : LOG_LEVEL_VERBOSE,
+            "Failed to get the state of the immutable bit from file '%s': %s",
+            changes_path, FSAttrsErrorCodeToString(res));
+    }
+
+    if (a.havefsattrs && a.fsattrs.haveimmutable && !a.fsattrs.immutable)
+    {
+        /* Here we only handle the clearing of the immutable the immutable
```
craigcomstock, line 361:
```suggestion
        /* Here we only handle the clearing of the immutable
```

cf-agent/verify_files.c
```diff
@@ -345,6 +347,67 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
...
+        {
+            RecordNoChange(ctx, pp, &a,
+                           "The immutable bit is not set on file '%s' as promised",
+                           changes_path);
+        }
+    }
+
+    /* If we encounter any promises to mutate the file and the immutable
+     * attribute in body fsattrs is "true", we will override the immutable bit
+     * by temporarily clearing it when ever needed. */
```
craigcomstock, line 408:
```suggestion
     * by temporarily clearing it whenever needed. */
```

libpromises/evalfunction.c
```diff
@@ -9776,6 +9817,13 @@ static const FnCallArg AND_ARGS[] =
     {NULL, CF_DATA_TYPE_NONE, NULL}
 };
 
+static const FnCallArg GET_ACLS_ARGS[] =
+{
+    {CF_ABSPATHRANGE, CF_DATA_TYPE_STRING, "Path to file or directory"},
+    {"default,access", CF_DATA_TYPE_OPTION, "Whether to get default or access ACL"},
```
craigcomstock, line 9823: So this is our "menu" item list for an attribute right? Is this required and if not, defaulted to "default"? I will probably find out later in the review. :)

craigcomstock (commented):
Looks good! A high-level question about atomicity and maybe one typo. :)

libpromises/override_fsattrs.h
craigcomstock: stale, added this comment elsewhere to current changes.

libpromises/override_fsattrs.c
craigcomstock: stale, added this comment elsewhere to current changes.

cf-agent/verify_files.c
```diff
@@ -406,7 +406,7 @@ static PromiseResult VerifyFilePromise(EvalContext *ctx, char *path, const Promi
     /* If we encounter any promises to mutate the file and the immutable
      * attribute in body fsattrs is "true", we will override the immutable bit
      * by temporarily clearing it when ever needed. */
-    const bool override_immutable = a.havefsattrs && a.fsattrs.haveimmutable && a.fsattrs.immutable && is_immutable;
+    EvalContextOverrideImmutableSet(ctx, a.havefsattrs && a.fsattrs.haveimmutable && a.fsattrs.immutable && is_immutable);
```
vpodzime, line 409: It's already in the attributes and we do pass those into gazillions functions already, right?

cf-agent/verify_files_utils.c
```diff
@@ -2000,7 +2000,7 @@ static bool TransformFile(EvalContext *ctx, char *file, const Attributes *attr,
 
     if (!IsExecutable(CommandArg0(BufferData(command))))
     {
-        RecordFailure(ctx, pp, attr, "Transformer '%s' for file '%s' failed", attr->transformer, file);
+        RecordFailure(ctx, pp, attr, " '%s' for file '%s' failed", attr->transformer, file);
```
vpodzime, line 2003: Why this change?

larsewi: Yup, I don't know how that happened. Maybe some typos in normal mode :laughing:

libpromises/override_fsattrs.h
larsewi: > Also, during this Begin() we should make the file immutable to attempt to restrict access as much as possible while we do "work" and before we call Commit().

If we make the file immutable while another process is writing it, we can maybe cause it to be left in an inconsistent state?

larsewi: I think this very quickly is becoming like the yak shaving. The task is to implement immutable bit support. Not to make CFEngine file operations atomic (they never have been). I created a value-add project [CFE-4551](https://northerntech.atlassian.net/browse/CFE-4551) to try to achieve this.

nickanderson: All CFEngine file operations are supposed to be atomic.

larsewi: > All CFEngine file operations are supposed to be atomic.

They unfortunately are not. However, this code uses the same attempt to achieve "atomic" file operations as before. I.e., write changes to a temporary file and swap it with the original.

libpromises/override_fsattrs.c
larsewi: Not sure if this is a good idea. Let's discuss in digital F2F format :)

libpromises/evalfunction.c
larsewi: If me make it optional and default to one of them, then we cannot add any more arguments in the future. Maybe this is OK?

Also it does not make sense to default to `"default"`, since `"default"` only works on directories. `"default"` is like a template ACL  that is inherited by all child objects.

cf-agent/verify_files.c
larsewi: It's not passed everywhere. We also would need to pass the `is_immutable` variable each function.

larsewi:
Superseded by https://github.com/cfengine/core/pull/5833

PR #5754, opened 2025-04-01
CFE-3629 - just poking around

btriller:
Set CFLAGS before AC_PROG_CC as it might add -g/-O options if CFLAGS is unset.
Don't add CFLAGS to CORE_FLAGS because I don't know, but maybe CFLAGS is populated otherwise.


btriller:
NorthernTechHQ/libntech#246
@olehermanse 🤷

PR #5756, opened 2025-04-02
ENT-12721: Added a new policy function useringroup

victormlg:


libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,33 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+static FnCallResult FnCallUserInGroup(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const Policy *policy, ARG_UNUSED const FnCall *fp, const Rlist *finalargs)
+{
+    assert(finalargs != NULL);
+
+    const char *user_name = RlistScalarValue(finalargs);
+    const char *group_name = RlistScalarValue(finalargs->next);
+
+    StringSet *groups = StringSetNew();
+    
+    struct group *grent = getgrnam(group_name);
```
larsewi, line 1202: It would be nice if you checked for error. How do you differentiate between error and no entry? You can set
```C
errno = 0;
```
before the call to `getgrnam()`. And then check the value of `errno` after the call. Check man page to see which error codes means that the given name or gid was not found. For all other codes, you can print an error message.

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,33 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+
+    if (grent == NULL) {
+        return FnReturnContext(false);
```
larsewi, line 1205: Remember to destroy the string set

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,33 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+    }
+
+    while (grent->gr_mem[0] != NULL)
+    {
+        StringSetAdd(groups, grent->gr_mem[0]);
+        grent->gr_mem++;
+    }   
+    
+    bool result = StringSetContains(groups, user_name);
+    return FnReturnContext(result);
```
larsewi, line 1215: Also there the string set should be destroyed

libpromises/evalfunction.c
```diff
@@ -9673,6 +9700,13 @@ static const FnCallArg GETUID_ARGS[] =
     {NULL, CF_DATA_TYPE_NONE, NULL}
 };
 
+static const FnCallArg USERINGROUP_ARGS[] =
+{
+    {CF_ANYSTRING, CF_DATA_TYPE_STRING, "User name in text"},
+    {CF_ANYSTRING, CF_DATA_TYPE_STRING, "Group name in text"},
```
larsewi, line 9706: Not sure what you mean by "in text". Do you mean "as string"? Anyways, I think this would be sufficient:
```suggestion
    {CF_ANYSTRING, CF_DATA_TYPE_STRING, "User name"},
    {CF_ANYSTRING, CF_DATA_TYPE_STRING, "Group name"},
```

larsewi (changes requested):
Nice :rocket: Some minor comments.

Also, it would be cool if there was an additional optional boolean parameter `id` that would let you specify user- and group ID. Or maybe it's not necessary. Idk.

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,50 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
 #endif
 }
 
+/*********************************************************************/ 
+
+static FnCallResult FnCallUserInGroup(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const Policy *policy, const FnCall *fp, const Rlist *finalargs)
+{
+    assert(fp != NULL);
+#ifdef __MINGW32__
```
larsewi, line 1196: I would use `_WIN32`, which covers all Windows systems.

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,50 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+    const char *group_name = RlistScalarValue(finalargs->next);
+
+    StringSet *groups = StringSetNew();
+    
+    errno = 0;
+    struct group *grent = getgrnam(group_name);
+
+    if (errno != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Couldn't open group database in function '%s'. Error number %d", fp->name, errno);
```
larsewi, line 1212: The error number is not very user friendly. Instead I would use `GetErrorStr()`

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,65 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+        {
+            case ENOENT:
+            case ESRCH:
+            case EBADF:
+            case EPERM:
+                StringSetDestroy(groups);
+                return FnReturnContext(false);
+            default:
+                break;
+        }
```
larsewi, line 1234: This code will never be run, because `errno` is only set when `getgrnam()` returns `NULL`

larsewi (changes requested):

olehermanse (changes requested):

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,56 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+            case 0:
+            case ENOENT:
+            case ESRCH:
+            case EBADF:
+            case EPERM:
+                StringSetDestroy(groups);
+                return FnReturnContext(false);
+            default:
+                const char *error_msg = GetErrorStr();
+                Log(LOG_LEVEL_ERR, "Couldn't open group database in function '%s': %s", fp->name, error_msg);
```
olehermanse, line 1222: This part does not compile on macOS (see GH Actions).

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,95 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+    if (grent == NULL)
+    {
+        switch(ret)
+        {
+            case 0:
+            case ENOENT:
+            case ESRCH:
+            case EBADF:
+            case EPERM:
+                break;
```
craigcomstock, line 1223: Would a debug message be nice here saying "couldn't find group %s as secondary group in group database", group_name?

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,95 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+        if (pwent == NULL)
+        {
+            switch(ret)
+            {
+                case 0:
+                case ENOENT:
+                case ESRCH:
+                case EBADF:
+                case EPERM:
+                    return FnReturnContext(false);
```
craigcomstock, line 1262: maybe a debug like above saying "could not find group %s as primary group for user %s", group_name, user_name.

craigcomstock (approved):
looks good, maybe add debug messages for "can't find" and where it "couldn't find" for some clarity in odd cases.

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,95 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+            default:
+            {
+                const char *error_msg = GetErrorStrFromCode(ret);
+                Log(LOG_LEVEL_ERR, "Couldn't open group database in function '%s': %s", fp->name, error_msg);
+                StringSetDestroy(users);
+                return FnFailure();
+            }
+        }
+    }
+    else {
```
olehermanse, line 1233: Our code style is to put the curly brace on a separate line;
```suggestion
    else
    {
```

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,95 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+/*********************************************************************/ 
+
+static FnCallResult FnCallUserInGroup(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const Policy *policy, const FnCall *fp, const Rlist *finalargs)
+{
+    assert(fp != NULL);
+#ifdef _WIN32
+    Log(LOG_LEVEL_ERR, "Function '%s' is POSIX specific", fp->name);
+    return FnFailure();  
+#else
+    assert(finalargs != NULL);
```
olehermanse, line 1200: This could be put before the ifdef. Asserts are testing our assumptions as programmers and looking for programmer errors / mistakes. This function should always be called with `finalargs != NULL`, if not, that's an error, I think, even on Windows.

It's also quite nice to have these asserts at the very top of the function - a bit nicer to see what is the acceptable way of calling this function.

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,95 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+
+    const char *user_name = RlistScalarValue(finalargs);
+    const char *group_name = RlistScalarValue(finalargs->next);
+
+    StringSet *users = StringSetNew();
+    int ret;
+
+    // check secondary group
```
olehermanse, line 1208: Here you could mention that this is what we expect the function to be used for the most, which is why we are checking it first (before primary group).

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,95 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+                return FnFailure();
+            }
+        }
+    }
+    else {
+        while (grent->gr_mem[0] != NULL)
+        {
+            char *group_member = xstrdup(grent->gr_mem[0]);
+            StringSetAdd(users, group_member);
+            grent->gr_mem++;
```
olehermanse, line 1238: Here it would be more optimized to just check if the group member is equal to the user we are looking for. If it is we can return early. There is no need to build a string set, and then iterate over it again for `StringSetContains()`. Also there is no need to iterate over the rest of the members after we have found our match. Just return early here if we find the user. Remember to clean up / free before returning.

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,95 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+        }
+    }
+    
+    bool result = StringSetContains(users, user_name);
+    StringSetDestroy(users);
+
+    // check primary group
+    if (!result && grent != NULL)
```
olehermanse, line 1246: This if can be avoided with an early return.

libpromises/evalfunction.c
```diff
@@ -1188,6 +1188,95 @@ static FnCallResult FnCallGetGid(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const P
...
+    {
+        struct passwd pwd;
+        struct passwd *pwent;
+        char pw_buf[GETPW_R_SIZE_MAX] = {0};
+        ret = getpwnam_r(user_name, &pwd, pw_buf, GETPW_R_SIZE_MAX, &pwent);
+
+        if (pwent == NULL)
```
olehermanse, line 1253: Same here, this if can be avoided with an early return.

Read this for some nice examples of this; https://refactoring.guru/replace-nested-conditional-with-guard-clauses

olehermanse (changes requested):
Behavior looks correct, but let's improve / refactor the code a bit to be faster and easier to read.

olehermanse:
@cf-bottom please test this in Jenkins :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12022/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12022/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12022/

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12025/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12025/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12025/

PR #5780, opened 2025-04-23
CFE-3429: Use current process ID to investigate proc filesystem to workaround in-container non-root owned symlinks (3.24)

craigcomstock:
- Use current process ID to investigate proc filesystem to workaround in-container non-root owned symlinks
- Added error message when netstat file is not found during UNIX interface discovery.
- Move GetRelocatedProcdirRoot() from libntech and add GetProcdirPid()
- Added env var pid override in proc-net acceptance test To align with change to reading /proc/<pid>/...
- bump libntech to master

### DIFF
--- a/libenv/unix_iface.c
+++ b/libenv/unix_iface.c
@@ -1283,6 +1283,10 @@ static JsonElement* GetNetworkingStatsInfo(const char *filename)
 
         fclose(fin);
     }
+    else
+    {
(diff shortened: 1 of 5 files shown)


cf-bottom:

craigcomstock:
the libntech bump in this PR was stale and caused macos unit test failures, so rebased libntech to current 3.24.x version.

PR #5779, opened 2025-04-23
Upgrade workflow from unsupported platform (3.21.x)

larsewi:

PR #5782, opened 2025-04-24
Do OSSL_PROVIDER_unload() on the providers that we load (3.21)

craigcomstock:
To avoid memory leaks.

(cherry picked from commit 504c938670d62e0a1ce9e092e182d35173c8f96b)


craigcomstock:
instead, incorporate into https://github.com/cfengine/core/pull/5779 which includes the workflow OS bumps which are needed.

PR #5656, opened 2024-12-09
CFE-4454: Fixed ODR violations in ASAN Unit Tests

larsewi:
- **Fixed ODR violations in ASAN Unit Tests**
- **Revert "Disabled ODR violation from ASAN Unit Tests"**

PR #5785, opened 2025-04-28
ENT-12766: Fixed ODR violations in ASAN Unit Tests (3.24.x)

larsewi:
Ticket: ENT-12766
(cherry picked from commit 88e6052728345daa18d9cb59478ea6b8f2771423)

Back-ported from https://github.com/cfengine/core/pull/5656

### DIFF
--- a/tests/unit/Makefile.am
+++ b/tests/unit/Makefile.am
@@ -188,6 +188,7 @@ if MACOSX
 XFAIL_TESTS = set_domainname_test
 XFAIL_TESTS += process_test
 XFAIL_TESTS += mon_processes_test
+XFAIL_TESTS += rlist_test
 endif
 
 if HPUX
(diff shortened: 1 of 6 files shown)


tests/unit/Makefile.am
```diff
@@ -188,6 +188,7 @@ if MACOSX
 XFAIL_TESTS = set_domainname_test
 XFAIL_TESTS += process_test
 XFAIL_TESTS += mon_processes_test
+XFAIL_TESTS += rlist_test
```
craigcomstock, line 191: can we put a reference to the ticket here? would be nice if we already did that previously so we have context on WHY these tests are allowed to fail.

craigcomstock (commented):
yeah, one comment, wondering if all these commits are cherry picks (and if so, fix that up so the references are there) and like you said, squash the last commit of 4 present right now.

PR #5786, opened 2025-04-29
ENT-12751: Fixed bug in parsing process_select for Windows

larsewi:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/12061/)

Signed-off-by: Lars Erik Wik <lars.erik.wik@northern.tech>

PR #5776, opened 2025-04-21
Added exception for solaris platform for findlocalusers* tests

craigcomstock:
The test needs work.
The expected data is not accurate for solaris as well as windows and aix.

Ticket: ENT-12767
Changelog: none

Solaris only build: [Build Status](https://ci.cfengine.com/job/pr-pipeline/12071/)

PR #5787, opened 2025-04-29
Fixed assertion triggered in OS detection code

larsewi:
- **Revert "libenv/sysinfo: Check on lowercase"**
- **Fixed bug in system detection code caused by wrong assumption**


craigcomstock:
This PR needs https://github.com/cfengine/core/pull/5786 merged which in turn depends on https://github.com/cfengine/core/pull/5776. I will work from there backward to here. :)

PR #5799, opened 2025-05-08
CFE-2815: Added `isnewerthantime` function

jakub-nt:
corresponding docs PR: https://github.com/cfengine/documentation/pull/3430

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -3068,7 +3068,7 @@ static FnCallResult FnCallIsConnectable(ARG_UNUSED EvalContext *ctx,
     }
 
     return FnReturnContext(sd > -1);
-} 
+}
 
 /*********************************************************************/
 
@@ -4494,6 +4494,28 @@ static FnCallResult FnCallIsNewerThan(ARG_UNUSED EvalContext *ctx, ARG_UNUSED co
     }
 
     return FnReturnContext(frombuf.st_mtime > tobuf.st_mtime);
+}
+
+static FnCallResult FnCallIsNewerThanTime(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const Policy *policy, ARG_UNUSED const FnCall *fp, const Rlist *finalargs)
+{
+    assert(finalargs != NULL);
+
+    const char *arg_file = RlistScalarValue(finalargs);
+    // a comment in `FnCallStrftime`: "this will be a problem on 32-bit systems..."
+    const time_t arg_mtime = IntFromString(RlistScalarValue(finalargs->next));
+
+    struct stat file_buf;
+    int exit_code = stat(arg_file, &file_buf);
+    if (exit_code == -1)
+    {
+        return FnFailure();
+    }
+
+    time_t file_mtime = file_buf.st_mtime;
+
+    bool result = file_mtime > arg_mtime;
+
+    return FnReturnContext(result);
 }
 
 /*********************************************************************/
@@ -7349,7 +7371,7 @@ static FnCallResult ValidateDataGeneric(const char *const fname,
     {
         isvalid = isvalid && JsonGetElementType(json) != JSON_ELEMENT_TYPE_PRIMITIVE;
     }
-    
+
     FnCallResult ret = FnReturnContext(isvalid);
     JsonDestroy(json);
     return ret;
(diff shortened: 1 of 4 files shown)


larsewi (changes requested):
Great work!

Since the title will be part of the changelog you can make it a little bit more specific. E.g., "Added isnewerthantime() policy function". Also backticks in the changelog can cause som issues later on. So better omit them in the case of changelog entries.

It would also be nice if you included an acceptance test. Here you can add more test cases that are not suitable in an example. E.g., what happens if the timestamp is not a string, what happens in the timestamp is `CF_NOINT`, what happens if timestamp is `CF_INFINITY` etc.

libpromises/evalfunction.c
```diff
@@ -4496,6 +4496,27 @@ static FnCallResult FnCallIsNewerThan(ARG_UNUSED EvalContext *ctx, ARG_UNUSED co
     return FnReturnContext(frombuf.st_mtime > tobuf.st_mtime);
 }
 
+static FnCallResult FnCallIsNewerThanTime(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const Policy *policy, ARG_UNUSED const FnCall *fp, const Rlist *finalargs)
+{
+    const char *arg_file = RlistScalarValue(finalargs);
+    // a comment in `FnCallStrftime`: "this will be a problem on 32-bit systems..."
+    const time_t arg_mtime = IntFromString(RlistScalarValue(finalargs->next));
+
```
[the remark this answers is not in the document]
larsewi: It's a bit hard to see, but here I want you to remove trailing whitespace

larsewi:
You will also have to sing the CLA

victormlg (approved):
Nice!

libpromises/evalfunction.c
```diff
@@ -4496,6 +4496,27 @@ static FnCallResult FnCallIsNewerThan(ARG_UNUSED EvalContext *ctx, ARG_UNUSED co
...
+    const char *arg_file = RlistScalarValue(finalargs);
+    // a comment in `FnCallStrftime`: "this will be a problem on 32-bit systems..."
+    const time_t arg_mtime = IntFromString(RlistScalarValue(finalargs->next));
+    
+    struct stat file_buf;
+    int exit_code = stat(arg_file, &file_buf);
+    
+    if (exit_code == -1)
+    {
+        return FnFailure();
```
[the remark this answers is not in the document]
olehermanse: In general it's a good idea to add helpful error messages, but looking at the other similar policy functions below, this seems consistent with them - they also `return FnFailure()` without printing an error, when `stat` fails, indicating:

- Maybe it's desirable to not print anything? Would adding errors become too noisy?
- Maybe it's intentional, since something else is responsible for printing the error automatically when you return `FnFailure()`?
- Maybe it is really a mistake, and we should update all of these to have error messages?

I haven't looked into it further, so I don't know the answer.

tests/acceptance/02_classes/02_functions/isnewerthantime.cf
```diff
@@ -0,0 +1,64 @@
+body common control
+{
+  inputs => { "../../default.cf.sub" };
+  bundlesequence => { default("$(this.promise_filename)") };
```
larsewi, line 4: Since you are not using anything from _default.cf.sub_ you can consider not including it and instead write your own bundle sequence.

```suggestion
  bundlesequence => { "check" };
```

tests/acceptance/02_classes/02_functions/isnewerthantime.cf
```diff
@@ -0,0 +1,64 @@
...
+  version => "1.0";
+}
+#######################################################
+
+bundle agent check
+{
+  vars:
+    "test_file"
+      string => "$(this.promise_dirname)/isnewerthantime_file.txt";
```
larsewi, line 13: You can consider creating this file in an `init` bundle and setting the `mtime`, all using the `files` promise.

larsewi (approved):
Looks good :rocket:

tests/acceptance/02_classes/02_functions/isnewerthantime.cf
nickanderson: Alternatively, you can omit body common control and just have bundle agent __main__ and then use methods promises to call each of the other bundles you want.

```
bundle agent __main__
{
  methods:
    "init";
    "test";
    "check";
    "cleanup";
```

So, many ways to skin the cat.

tests/acceptance/02_classes/02_functions/isnewerthantime.cf
nickanderson: Implementation is using the derived mtime via filestat()

PR #5816, opened 2025-05-27
testall: Added option to not run with --no-locks

larsewi:
While debugging an issue concerning locks, I found that I needed an
option to not use the `--no-lock` option when running the agent through
the testall script.


jakub-nt (approved):

larsewi:
> no-no-lock is funny, I almost suggested --with-locks but this is fine as-is.

Yup, it's quite funny indeed. Considered `--with-lock` but though `--no-no-lock` would better evoke (bring to mind) the `--no-lock` option.

PR #5831, opened 2025-06-25
Adjusted valgrind-check auto_destruct timeout from 30m to 40m

craigcomstock:
Some build hosts weren't finishing in 30 minutes so see if 40 is enough.

Ticket: ENT-13080
Changelog: none


olehermanse (approved):

larsewi:
Thanks @craigcomstock. Hope it does the trick

PR #5850, opened 2025-08-05
CFE-4562: classfilterdata(): added support of object of arrays and object of objects

larsewi:
- **classfilterdata() now supports object of objects**
- **classfilterdata() now supports object of arrays**
- **classfilterdata(): deduped code for getting class expression**


larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12452/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12452/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12452/

jakub-nt (approved):

victormlg (approved):
Cool!

libpromises/evalfunction.c
```diff
@@ -7864,6 +7865,13 @@ static bool ClassFilterDataArray(
     JsonElement *child,
     bool *remove)
 {
+    if (key_or_index == NULL) {
+        Log(LOG_LEVEL_VERBOSE,
+            "Function %s(): At least 3 arguments are required when the data structure is an array of arrays or array of objects",
```
craigcomstock: maybe you could mention the third parameter is key or index so they don't have to check the manual.

libpromises/evalfunction.c
```diff
@@ -7900,17 +7908,131 @@ static bool ClassFilterDataArray(
     return false;
 }
 
+static bool ClassFilterDataObjectOfArrays(
+    ARG_UNUSED EvalContext *ctx,
+    ARG_UNUSED const char *fn_name,
+    ARG_UNUSED JsonElement *json_array,
+    ARG_UNUSED const char *class_expr_index,
+    ARG_UNUSED const char *exogenous_key,
```
craigcomstock: I wouldn't prefer the word exogenous as it is too specifically linguistically technical. Maybe `self_key`? 🚲 🏠 (bikeshed todo). Ideally you want to write code that a less smart maintainer can handle :)

libpromises/evalfunction.c
```diff
@@ -7900,17 +7908,131 @@ static bool ClassFilterDataArray(
...
+    ProgrammingError("classfilterdata() with data structure object of arrays is not yet implemented");
+    return false;
+}
+
+static bool ClassFilterDataObjectOfObjects(
+    EvalContext *ctx,
+    const char *fn_name,
+    JsonElement *json_object,
+    const char *class_expr_key,
+    const char *exogenous_key,
```
craigcomstock: maybe `self_key`?

libpromises/evalfunction.c
```diff
@@ -7900,17 +7908,131 @@ static bool ClassFilterDataArray(
...
+    /* Use the key of the object itself as the class expression, unless a class
+     * expression key is specified */
+    const char *class_expr = exogenous_key;
+    if (class_expr_key != NULL)
+    {
+        JsonElement *json_child = JsonObjectGet(json_object, class_expr_key);
+        if (json_child == NULL)
+        {
+            Log(LOG_LEVEL_VERBOSE,
+                "Function %s(): Bad class expression key '%s': Key not found",
```
craigcomstock: if we aren't validating if it is a "bad class expression" aka somehow not a well formed class expression then this message should only say "Function %s(): Key not found with class expression '%s'".

Do we have a function that can validate a class expression? Should add that to a static checking tool like cf-promises?

craigcomstock: ah, I see, asserting that the class expression is not empty, maybe just say "Looking for key in the object itself and found none." or similar.

libpromises/evalfunction.c
```diff
@@ -7900,17 +7908,131 @@ static bool ClassFilterDataArray(
...
+    {
+    case JSON_TYPE_ARRAY:
+        if (StringEqual(data_structure, "auto") ||
+            StringEqual(data_structure, "object_of_arrays"))
+        {
+            return ClassFilterDataObjectOfArrays(
+                ctx, fn_name, child, key_or_index, exogenous_key, remove);
+        }
+        Log(LOG_LEVEL_VERBOSE,
+            "Function %s(): Expected child element to be of container type array",
```
craigcomstock: would it be helpful to mention the type the item WAS? aka "to be of container type array but was %s", data_structure

libpromises/evalfunction.c
```diff
@@ -7900,17 +7908,131 @@ static bool ClassFilterDataArray(
...
+
+    if ((n_args < 2) || (n_args > 3))
+    {
+        Log(LOG_LEVEL_VERBOSE,
+            "Function %s(): Require 2 to 3 arguments, got %d",
+            fp->name, n_args);
+        return FnFailure();
+    }
+
+    /* CodeQL does not understand that this is implied from above */
```
craigcomstock, line 8062: why the comment? does CodeQL complain about this code? Wouldn't we add some meta info here for CodeQL to ignore the code in that case?

libpromises/evalfunction.c
```diff
@@ -7922,43 +8044,93 @@ static FnCallResult FnCallClassFilterData(
...
+                assert(child != NULL);
 
-        bool remove;
-        if (!ClassFilterDataArray(ctx, fp->name, data_structure, key_or_index, child, &remove))
-        {
-            /* Error is already logged */
+                bool remove;
+                if (!ClassFilterDataArray(ctx, fp->name, data_structure, key_or_index, child, &remove))
+                {
+                    /* Error is already logged */
```
craigcomstock, line 8117: maybe "Error is already logged by ClassFilterDataArray()" to be specific and aid in refactoring in the future.

libpromises/evalfunction.c
```diff
@@ -7922,43 +8044,93 @@ static FnCallResult FnCallClassFilterData(
...
+                }
+            }
+            break;
+        case JSON_TYPE_OBJECT:
+            iter = JsonIteratorInit(parent);
+            child = JsonIteratorNextValue(&iter);
+
+            while ((child != NULL))
+            {
+                const char *exogenous_key = JsonIteratorCurrentKey(&iter);
```
craigcomstock: again, I think you should change exogenous_key to self_key or something that is easier to understand.

libpromises/evalfunction.c
```diff
@@ -7922,43 +8044,93 @@ static FnCallResult FnCallClassFilterData(
...
+                {
+                    /* Error is already logged */
+                    JsonDestroy(parent);
+                    return FnFailure();
+                }
+
+                if (remove)
+                {
+                    NDEBUG_UNUSED bool success = JsonObjectRemoveKey(parent, exogenous_key);
+                    assert(success);
```
craigcomstock, line 8148: why no production-time error handling here?

libpromises/evalfunction.c
```diff
@@ -7922,43 +8044,93 @@ static FnCallResult FnCallClassFilterData(
...
+                else
+                {
+                    child = JsonIteratorNextValue(&iter);
+                }
+            }
+            break;
+        default:
+            /* Only parent type array and object is supported */
+            Log(LOG_LEVEL_VERBOSE,
+                "Function %s(): Expected parent element to be of container type array or object",
```
craigcomstock: maybe remove the comment since that is in the log message and remove "array or object" and just leave it at "container type"? Or maybe even remove "container type" and just specify "array or object type" since those are more generally understandable than our special use of the term "container".

tests/acceptance/01_vars/02_functions/classfilterdata_object_of_objects_endogenous.cf
```diff
@@ -0,0 +1,73 @@
+body common control
+{
+  bundlesequence => { "test", "check" };
+}
+
+bundle agent test
+{
+  meta:
+    "description" -> { "CFE-4562" }
+      string => "Test for expected results from classfilterdata() with object of objects using endogenous key";
```
craigcomstock: replace endogenous key with something less technical, as I suggested "self_key" for "exogenous_key", maybe this would be "internal_key"?

libpromises/evalfunction.c
```diff
@@ -7909,17 +7909,52 @@ static bool ClassFilterDataArray(
...
+    /* Use the key of the object itself as the class expression, unless a class
+     * expression index is specified */
+    const char *class_expr = exogenous_key;
+    if (class_expr_index != NULL)
+    {
+        size_t index;
+        assert(SIZE_MAX >= ULONG_MAX); /* make sure returned value can fit in size_t */
+        if (StringToUlong(class_expr_index, &index) != 0) {
+            Log(LOG_LEVEL_VERBOSE,
+                "Function %s(): Bad class expression index '%s': Failed to parse integer",
```
craigcomstock: is an index a class expression? shouldn't that just be an index, aka integer and specific to this function as an argument.

libpromises/evalfunction.c
larsewi: It fails if the key-value pair you are trying to remove does not exist. However, we already know that it does. Thus, it should never return false.

libpromises/evalfunction.c
larsewi: CodeQL complains that we dereference `args` without checkout for `NULL`. The asserts will silence it. It does not understand that the if statement above implicitly proves that it cannot be `NULL` at this point.

libpromises/evalfunction.c
larsewi: This is stating the obvious in my opinion. How does it aid refactoring?

larsewi:
@cf-bottom Jenkins please :)

craigcomstock (approved):
cool!

PR #5859, opened 2025-08-20
CFE-4494: Fixed CIFS mounts on Linux

nickanderson:
Ticket: CFE-4494

### DIFF
--- a/cf-agent/nfs.c
+++ b/cf-agent/nfs.c
@@ -481,6 +481,9 @@ int VerifyInFstab(EvalContext *ctx, char *name, const Attributes *a, const Promi
     }
 
 #if defined(__QNX__) || defined(__QNXNTO__)
+    // QNX documents 4 fstab fields : https://www.qnx.com/developers/docs/7.1/index.html#com.qnx.doc.neutrino.utilities/topic/f/fstab.html
+    // specialdevice mountpoint type mountoptions
+    // TODO Remove DUMP and PASS options used here (unsupported)?
     NDEBUG_UNUSED int ret = snprintf(fstab, CF_BUFSIZE, "%s \t %s %s\t%s 0 0", device, mountpt, fstype, opts);
     assert(ret >= 0 && ret < CF_BUFSIZE);
 #elif defined(_CRAY)
@@ -492,9 +495,13 @@ int VerifyInFstab(EvalContext *ctx, char *name, const Attributes *a, const Promi
     assert(ret >= 0 && ret < CF_BUFSIZE);
     break;
 #elif defined(__hpux)
+    // HP-UX documents 7 fstab fields: https://nixdoc.net/man-pages/HP-UX/man4/fstab.4.html
+    // deviceSpecialFile directory type options backupFrequency passNumber comment
+    // TODO Bring promise comment in as the 7th comment field # promise comment (stripped of newlines)
     NDEBUG_UNUSED int ret = snprintf(fstab, CF_BUFSIZE, "%s %s \t %s \t %s 0 0", device, mountpt, fstype, opts);
     assert(ret >= 0 && ret < CF_BUFSIZE);
 #elif defined(_AIX)
+    // AIX uses /etc/filesystems: https://www.ibm.com/docs/en/aix/7.2.0?topic=files-filesystems-file
     NDEBUG_UNUSED int ret = snprintf(fstab, CF_BUFSIZE,
                                      "%s:\n\tdev\t= %s\n\ttype\t= %s\n\tvfs\t= %s\n\tnodename\t= %s\n\tmount\t= true\n\toptions\t= %s\n\taccount\t= false\n",
                                      mountpt, rmountpt, fstype, fstype, host, opts);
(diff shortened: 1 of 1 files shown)


cf-agent/nfs.c
```diff
@@ -484,7 +484,14 @@ int VerifyInFstab(EvalContext *ctx, char *name, const Attributes *a, const Promi
              "%s:\n\tdev\t= %s\n\ttype\t= %s\n\tvfs\t= %s\n\tnodename\t= %s\n\tmount\t= true\n\toptions\t= %s\n\taccount\t= false\n",
              mountpt, rmountpt, fstype, fstype, host, opts);
 #elif defined(__linux__)
-    snprintf(fstab, CF_BUFSIZE, "%s:%s \t %s \t %s \t %s", host, rmountpt, mountpt, fstype, opts);
+    if (StringEqual(fstype, "cifs") || StringEqual(fstype, "panfs"))
+    {
+        snprintf(fstab, CF_BUFSIZE, "%s%s \t %s \t %s \t %s", host, rmountpt, mountpt, fstype, opts);
```
larsewi: This is not strictly necessary. But here we assume the buffer is large enough. In this case I personally like to test this assumption with an `assert`. This way, we would catch it in debug builds if the assumption is wrong.
```suggestion
        NDEBUG_UNUSED int ret = snprintf(fstab, CF_BUFSIZE, "%s%s \t %s \t %s \t %s", host, rmountpt, mountpt, fstype, opts);
        assert(ret >= 0 && ret < CF_BUFSIZE);
```

cf-agent/nfs.c
```diff
@@ -484,7 +484,14 @@ int VerifyInFstab(EvalContext *ctx, char *name, const Attributes *a, const Promi
...
+    }
+    else
+    {
+        snprintf(fstab, CF_BUFSIZE, "%s:%s \t %s \t %s \t %s", host, rmountpt, mountpt, fstype, opts);
```
larsewi:
```suggestion
        NDEBUG_UNUSED int ret = snprintf(fstab, CF_BUFSIZE, "%s:%s \t %s \t %s \t %s", host, rmountpt, mountpt, fstype, opts);
        assert(ret >= 0 && ret < CF_BUFSIZE);
```

cf-agent/nfs.c
```diff
@@ -468,38 +468,49 @@ int VerifyInFstab(EvalContext *ctx, char *name, const Attributes *a, const Promi
...
     strlcpy(fstype_upper, fstype, CF_BUFSIZE);
     ToUpperStrInplace(fstype_upper);
 
-    snprintf(fstab, CF_BUFSIZE, "%s:%s \t %s %s\t%s", host, rmountpt, mountpt, fstype_upper, opts);
+    NDEBUG_UNUSED int ret = snprintf(fstab, CF_BUFSIZE, "%s \t %s %s\t%s", device, mountpt, fstype_upper, opts);
+    assert(ret >= 0 && ret < CF_BUFSIZE);
     break;
 #elif defined(__hpux)
-    snprintf(fstab, CF_BUFSIZE, "%s:%s %s \t %s \t %s 0 0", host, rmountpt, mountpt, fstype, opts);
+    NDEBUG_UNUSED int ret = snprintf(fstab, CF_BUFSIZE, "%s %s \t %s \t %s 0 0", device, mountpt, fstype, opts);
```
craigcomstock, line 501: with all of these platforms, it might be nice to include some authoritative resources for the different formats.

cf-agent/nfs.c
```diff
@@ -523,8 +523,16 @@ int VerifyInFstab(EvalContext *ctx, char *name, const Attributes *a, const Promi
     {
         AppendItem(&FSTABLIST, fstab, NULL);
         FSTAB_EDITS++;
-        cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_CHANGE, pp, a, "Adding file system '%s:%s' to '%s'", host, rmountpt,
-             VFSTAB[VSYSTEMHARDCLASS]);
+        if (StringEqual(fstype, "cifs") || StringEqual(fstype, "panfs"))
```
craigcomstock: can't you include this LATER after all the various platforms so you don't repeat the ':' or not ':' choice?

nickanderson: I think i got it, lmk what you think

cf-agent/nfs.c
```diff
@@ -469,14 +469,17 @@ int VerifyInFstab(EvalContext *ctx, char *name, const Attributes *a, const Promi
     fstype = a->mount.mount_type;
 
     char device[CF_BUFSIZE];
+    const char *device_format;
     if (StringEqual(fstype, "cifs") || StringEqual(fstype, "panfs"))
     {
-        NDEBUG_UNUSED int ret = snprintf(device, sizeof(device), "%s%s", host, rmountpt);
+        device_format = "%s%s";
+        NDEBUG_UNUSED int ret = snprintf(device, sizeof(device), device_format, host, rmountpt);
```
craigcomstock: You don't need the `device_format` variable I think as you are not using it anywhere else.

Just put different values in the sprintf command as you had before and that should be fine.

cf-agent/nfs.c
```diff
@@ -523,16 +526,8 @@ int VerifyInFstab(EvalContext *ctx, char *name, const Attributes *a, const Promi
...
-        {
-            cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_CHANGE, pp, a, "Adding file system '%s%s' to '%s'", host, rmountpt,
-                 VFSTAB[VSYSTEMHARDCLASS]);
-        }
-        else
-        {
-            cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_CHANGE, pp, a, "Adding file system '%s:%s' to '%s'", host, rmountpt,
-                 VFSTAB[VSYSTEMHARDCLASS]);
-        }
+        cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_CHANGE, pp, a, "Adding file system '%s' to '%s'", device,
```
craigcomstock: It seems you might want to switch the message to something like "Adding file system entry '%s' to '%s'", fstab, VFSTAB[VSYSTEMHARDCLASS] since the fstab entry has a LOT of interesting differences by platform and may be quite useful for debugging by logs without inspecting the /etc/fstab results if they are even still there.

cf-agent/nfs.c
nickanderson: OK, i added some references.

larsewi:
Thanks :rocket: Please squash commits

craigcomstock (commented):
nice. you should probably squash a few commits in the middle around the logging.

craigcomstock (approved):
🥇

PR #5866, opened 2025-08-27
ENT-4664: Added sys.policy_release_id

victormlg:


cf-agent/cf-agent.c
```diff
@@ -321,6 +342,7 @@ int main(int argc, char *argv[])
 
     BeginAudit();
 
+    SysPolicyReleaseId(ctx, policy);
```
larsewi: Not sure what the policy release ID is. But if you put it here, the variable will only be created in `cf-agent`, non of the other components. Is this intended. Furthermore, I feel like there should be a better place to put this. Where are other `sys` variables created?

victormlg: The problem is that I need a reference to the current policy loaded, and EvalContext (the only argument of `DetectEnvironment`, the function responsible for setting most sys variables) doesn't hold any reference to the policy. Moreover, `DetectEnvironment` is called before the policy is loaded in all cf components

I moved `SysPolicyReleaseId` into `LoadPolicy`, which means that the release_id will be set everytime a policy is loaded.

I could try to give a reference to the policy in  `DetectEnvironment`, but it seems a bit overkill and could potentially break things

larsewi:
Could you add an acceptance test to check that this variable is defined?

larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12571/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12571/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12571/

larsewi:
> logging.c:651:28: error: 'strchrnul' is only available on macOS 15.4 or newer [-Werror,-Wunguarded-availability-new]
  651 |         char *next_token = strchrnul(token, ',');
      |                            ^~~~~~~~~

This is unrelated to you PR, but you can fix it by doing what I did in this PR https://github.com/NorthernTechHQ/libntech/pull/255

libenv/sysinfo.c
```diff
@@ -3803,3 +3803,26 @@ void DetectEnvironment(EvalContext *ctx)
...
+                              "policy_release_id",
+                              release_id,
+                              CF_DATA_TYPE_STRING, "source=agent,attribute_name=Policy Release Id");
+    free(release_id);
+}
+
+void DetectEnvironmentFromPolicy(EvalContext *ctx, Policy *policy)
+{
+    SysPolicyReleaseId(ctx, policy);
+}
```
larsewi: Please add trailing newline

larsewi (changes requested):
The `libntech` should already updated on the master branch (see https://github.com/cfengine/core/pull/5867). Try to fetch all and rebase with upstream master. In the third commit, please explain why you specified the deployment target in the commit message. Otherwise the commit message is not very useful.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12598/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12598/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12598/

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12622/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12622/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12622/

craigcomstock:
@victormlg @larsewi this is the kind of thing we need to backport. I will back port it to 3.24.x here: https://github.com/cfengine/core/pull/5780 to get that PR green. :+1: I will do it also for 3.21 separately as I am not back-porting my other change to 3.21.

PR #5722, opened 2025-03-16
CFE-30: Add policy variable sys.cpusockets

btriller:
Ticket: CFE-30
Changelog: Title
Merge together:
cfengine/core#5722
cfengine/documentation#3408

PR #5877, opened 2025-09-11
tests/valgrind: Check for leaking file descriptors

btriller:
Let valgrind track fds to catch unclosed file descriptors. That would have catched bug in #5722.


craigcomstock (approved):

larsewi:
There's still a leaked file descriptor. I will have to get back to this next week @btriller

PR #5882, opened 2025-09-22
CFE-4550: Added findlocalgroups() policy function

victormlg:


libpromises/evalfunction.c
```diff
@@ -1152,6 +1152,159 @@ static FnCallResult FnCallFindLocalUsers(ARG_UNUSED EvalContext *ctx, ARG_UNUSED
 
 /*********************************************************************/
 
+#if defined(HAVE_GETPWENT) && !defined(__ANDROID__)
+
+static bool GroupMembersContain(const char *member, struct group *gr)
```
larsewi: It looks like you are checking if a member is a part of the group. If that is the case, then you should probably rename the function. 
```suggestion
static bool GroupContainsMember(const char *member, struct group *gr)
```
You should also use `const struct group *gr` as the second parameter and not mutate it.

libpromises/evalfunction.c
```diff
@@ -1152,6 +1152,159 @@ static FnCallResult FnCallFindLocalUsers(ARG_UNUSED EvalContext *ctx, ARG_UNUSED
...
+{
+    assert(gr != NULL);
+    bool contains = false;
+    while (gr->gr_mem[0] != NULL)
+    {
+        if (StringMatchFull(gr->gr_mem[0], member))
+        {
+            contains = true;
+        }
+        gr->gr_mem++;
```
larsewi: Here you are changing the value of `gr->gr_mem`. This is not necessary. You can instead create a new local pointer that you can mutate instead, to avoid any side effects. Or even better, you could use a for loop with an incrementing counter. This would be easier to read and understand.

libpromises/evalfunction.c
```diff
@@ -11598,6 +11751,8 @@ const FnCallType CF_FNCALL_TYPES[] =
                   FNCALL_OPTION_NONE, FNCALL_CATEGORY_UTILS, SYNTAX_STATUS_NORMAL),
     FnCallTypeNew("findlocalusers", CF_DATA_TYPE_CONTAINER, FIND_LOCAL_USERS_ARGS, &FnCallFindLocalUsers, "Find matching local users",
                   FNCALL_OPTION_VARARG, FNCALL_CATEGORY_DATA, SYNTAX_STATUS_NORMAL),
+    FnCallTypeNew("findlocalgroups", CF_DATA_TYPE_CONTAINER, FIND_LOCAL_USERS_ARGS, &FnCallFindLocalGroups, "Find matching local groups",
+                  FNCALL_OPTION_VARARG, FNCALL_CATEGORY_DATA, SYNTAX_STATUS_NORMAL),
```
larsewi: No reason to use `FNCALL_OPTION_VARARG` here?

libpromises/evalfunction.c
```diff
@@ -1152,6 +1152,159 @@ static FnCallResult FnCallFindLocalUsers(ARG_UNUSED EvalContext *ctx, ARG_UNUSED
...
+        Log(LOG_LEVEL_ERR, "Bad argument '%s' in function '%s': Expected data container or slist",
+            RlistScalarValueSafe(finalargs), fp->name);
+        JsonDestroyMaybe(json, allocated);
+        return FnFailure();
+    }
+
+    JsonElement *parent = JsonObjectCreate(10);
+    setgrent();
+    struct group *gr;
+    while ((gr = getgrent()) != NULL)
```
larsewi, line 1275: Would be nice this part into multiple functions. It gets very hard to read when you go too many scopes deep.

libpromises/evalfunction.c
```diff
@@ -1152,6 +1152,159 @@ static FnCallResult FnCallFindLocalUsers(ARG_UNUSED EvalContext *ctx, ARG_UNUSED
...
+            if (JsonGetElementType(element) != JSON_ELEMENT_TYPE_PRIMITIVE)
+            {
+                Log(LOG_LEVEL_ERR, "Bad argument '%s' in function '%s': Filter cannot include nested data",
+                    RlistScalarValueSafe(finalargs), fp->name);
+                JsonDestroyMaybe(json, allocated);
+                JsonDestroy(parent);
+                return FnFailure();
+            }
+            const char *field = JsonPrimitiveGetAsString(element);
+            const Rlist *tuple = RlistFromSplitString(field, '=');
```
larsewi:
```
$ sudo groupadd test=
$ grep test /etc/group
test=:x:1001:
```
The `=` character can be part of the group name.

victormlg: Do you have any alternative? Split on a different character or include quotes around the group name?

larsewi: Maybe it's fine. If it becomes a problem, we can try to add a way to escape it later, or parse it as CSV.

libpromises/evalfunction.c
```diff
@@ -1152,6 +1152,161 @@ static FnCallResult FnCallFindLocalUsers(ARG_UNUSED EvalContext *ctx, ARG_UNUSED
...
+#if defined(HAVE_GETPWENT) && !defined(__ANDROID__)
+
+static bool GroupContainsMember(const char *member, const struct group *gr)
+{
+    // if the group doesn't have any members
+    assert(gr != NULL);
+    if (gr->gr_mem[0] == NULL)
+    {
+        return StringMatchFull(member, "");
+    }
```
larsewi, line 1164: Would you not get the same outcome if you just removed this code?

libpromises/evalfunction.c
```diff
@@ -1152,6 +1152,161 @@ static FnCallResult FnCallFindLocalUsers(ARG_UNUSED EvalContext *ctx, ARG_UNUSED
...
+            if (!StringMatchFull(value, gid_string))
+            {
+                can_add_to_json = false;
+            }
+        }
+        else if (StringEqual(attribute, "members"))
+        {
+            Rlist *members = RlistFromSplitString(value, ',');
+
+            // for each member in the filter, if they all are contained inside the group's members, match
```
larsewi: It's a little bit hard to understand this comment. Maybe you could make it a little bit more clear?

libpromises/evalfunction.c
```diff
@@ -1152,6 +1152,161 @@ static FnCallResult FnCallFindLocalUsers(ARG_UNUSED EvalContext *ctx, ARG_UNUSED
...
+        }
+    }
+    return contains;
+}
+
+static bool GroupMatchesFilter(const struct group *gr, JsonElement *filter)
+{
+    assert(gr != NULL);
+
+    bool can_add_to_json = true;
```
larsewi: This variable name does not make that much sense in the context of this function. Should it be named `group_matches_filter`?

larsewi (changes requested):

libpromises/evalfunction.c
```diff
@@ -1152,6 +1152,161 @@ static FnCallResult FnCallFindLocalUsers(ARG_UNUSED EvalContext *ctx, ARG_UNUSED
...
+    JsonDestroyMaybe(filter, allocated);
+
+    return FnReturnContainerNoCopy(parent);
+}
+
+#else
+
+static FnCallResult FnCallFindLocalGroups(ARG_UNUSED EvalContext *ctx, ARG_UNUSED const Policy *policy, ARG_UNUSED const FnCall *fp, ARG_UNUSED const Rlist *finalargs)
+{
+    Log(LOG_LEVEL_ERR, "findlocalgroups is not implemented");
```
craigcomstock: Maybe add to the message "on this platform"? Since that is the condition where it is not implemented.

tests/acceptance/01_vars/02_functions/findlocalgroups.cf
```diff
@@ -0,0 +1,51 @@
+body common control
```
craigcomstock: I think you have some funny spaces in here, like trailing space on first line and spaces on an empty line after "number_filter" vars promise.

tests/acceptance/01_vars/02_functions/findlocalgroups.cf
```diff
@@ -0,0 +1,51 @@
...
+      
+      # longer filters
+      "slist_filter" slist => { "gid=0", "name=root" };
+
+      # using data
+      "data_filter" data => '[ "gid=0", "name=root" ]';
+
+      # using regex
+      "simple_regex" slist => { "name=roo.*" };
+      "number_regex" slist => { "gid=0.*" };
```
craigcomstock: I see the number regex, what about character classes like [0-9] and such?

tests/acceptance/01_vars/02_functions/findlocalgroups.cf
```diff
@@ -0,0 +1,51 @@
...
+      "longer_regex" slist => { "name=ro.*", "gid=0.*" };
+
+      # non-existent group
+      "unknown" slist => { "name=thisgroupdoesntexist" };
+}
+bundle agent test 
+{
+  meta:
+      "test_soft_fail" string => "windows|aix|solaris",
```
craigcomstock: maybe !linux would be better here? Maybe even !linux|(termux|android) as both of those hard-classes (termux and android) are defined in a termux build of CFEngine.

tests/acceptance/01_vars/02_functions/findlocalgroups.cf
```diff
@@ -0,0 +1,51 @@
...
+        meta => { "CFE-2318" };
+
+  vars:
+    "ulist1" data => findlocalgroups("@(init.simple_filter)");
+    "ulist2" data => findlocalgroups("init.number_filter");
```
craigcomstock: why no "@(...)" here?

craigcomstock (approved):
Yeah, looks good with a few comments.

tests/acceptance/01_vars/02_functions/findlocalgroups.cf
larsewi: I guess to test both reference and inline JSON? If that is what `@(...)` does (i.e., expand it beforehand)? IDK

libpromises/evalfunction.c
victormlg: You wouldn't be able to match on groups without members (.*) or do negative matching (all groups that don't contain X)

tests/acceptance/01_vars/02_functions/findlocalgroups.cf
victormlg: The test is very limited because I want it to be predictable. For now the only matching group is root

libpromises/evalfunction.c
larsewi: So this returns `true` if `member` is an empty string? If that is the case, then why not just check if the string is empty, instead of invoking an expensive regex engine?

larsewi (approved):
🚀

larsewi (changes requested):
Looks like you have a failing acceptance test

libpromises/evalfunction.c
victormlg: I thought about cases like `r"(?!a).*"`: if we want to match everything but some group, then we need to compare the regex to something, in that case an empty string

larsewi:
@cf-bottom Jenkins please :)

PR #5884, opened 2025-09-22
Added example using nth() with key

nickanderson:


nickanderson:
@cf-bottom jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12697/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12697/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12697/

examples/nth-key.cf
```diff
@@ -0,0 +1,20 @@
...
+bundle agent main
+{
+  vars:
+      "d" data => '{ "key1": "dvalue1", "key2": "dvalue2" }';
+      "a[key1]" string => "value1";
+      "a[key2]" string => "value2";
+
+  reports:
+      "$(with)" with => nth( a, key1 );
+      "$(with)" with => nth( d, key1 );
```
craigcomstock: I know that tokens in policy are really strings and sometimes magically they will match a proper thing like a variable as well but... not putting quotes around key1 here seems confusing without some explanation of this "magic" in the policy language.

Certainly the data container d and classic array a can also be referred to with quotes around them since that parameter to nth() is also "just a string" and not some non-string variable reference but in that case it "makes sense" that it wouldn't have quotes around it.

```suggestion
      "$(with)" with => nth( a, "key1" );
      "$(with)" with => nth( d, "key1" );
```

craigcomstock (changes requested):

PR #5931, opened 2025-11-04
CFE-4590, ENT-13239: Override default directory permissions

larsewi:
- **Added test for overriding default directory create mode**
- **Skip remaining string comparisons after successful match**
- **Added override_default_directory_create_mode to body agent control**
- **Replaced hardcoded directory create mode with DEFAULTMODE**


cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12934/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12934/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12934/

libpromises/mod_common.c
```diff
@@ -321,6 +321,7 @@ const ConstraintSyntax CFA_CONTROLBODY[] =
     ConstraintSyntaxNewBool("select_end_match_eof", "Set the default behavior of select_end_match_eof in edit_line promises. Default: false", SYNTAX_STATUS_NORMAL),
     ConstraintSyntaxNewStringList("copyfrom_restrict_keys", ".*", "A list of key hashes to restrict copy_from to", SYNTAX_STATUS_NORMAL),
     ConstraintSyntaxNewString("evaluation_order", "(classic|top_down)", "Order of evaluation of promises of agent", SYNTAX_STATUS_NORMAL),
+    ConstraintSyntaxNewString("override_default_directory_create_mode", ".*", "Override default directory create mode (defaults to 0700 if not specified)", SYNTAX_STATUS_NORMAL),
```
olehermanse: I think putting "override default" is unnecessary - aren't all the values here overriding some default?

```suggestion
    ConstraintSyntaxNewString("default_directory_create_mode", ".*", "Override default directory create mode (defaults to 0700 if not specified)", SYNTAX_STATUS_NORMAL),
```

olehermanse (changes requested):

cf-agent/cf-agent.c
```diff
@@ -1337,6 +1337,7 @@ static void KeepControlPromises(EvalContext *ctx, const Policy *policy, GenericA
             {
                 Log(LOG_LEVEL_VERBOSE, "SET select_end_match_eof %s", (char *) value);
                 EvalContextSetSelectEndMatchEof(ctx, BooleanFromString(value));
+                continue;
```
craigcomstock, line 1340: Why? I don't understand the context or the reason for the change and what the change does.

I noticed around line 1141 there is another block that doesn't `continue` I wonder if this whole block of code needs to be audited for correct logic?

```
if (strcmp(cp->lval, CFA_CONTROLBODY[AGENT_CONTROL_ALLCLASSESREPORT].lval) == 0)
```

craigcomstock (approved):
looks good.

cf-agent/cf-agent.c
larsewi: We are checking if the string matches any of the possible attributes. Once we find a match, we don't need to check if it matches the remaining attributes, because we know implicitly that they will not match.

There is no change in behavior, just less CPU cycles wasted.

nickanderson: Feels like i am just showing what i dont know about the code

but, @larsewi your description there made me wonder if this would be affected:

```
body common control
{
   default_directory_create_mode => "000";
   default_directory_create_mode => "770";
}
```
It's a silly example, but in some policies the same attribtue is set differently in different contexts and multiple contexts might apply. 

I just wanted to be sure that not continuing to check for the attribute will not affect this (in the above simple example, expect the last one to win).

larsewi: I don't think these `break`s will affect multiple configurations of the default directory create mode. However, I added a line in the policy to test this, just in case. I will merge once the acceptance test workflow from GitHub Actions have passed.

olehermanse (approved):

PR #6009, opened 2026-01-07
013.cf: Skipped failing test on HP-UX

larsewi:
Test was added here https://github.com/cfengine/core/pull/5996
[Build Status](https://ci.cfengine.com/job/pr-pipeline/13280/)

Back ported to https://github.com/cfengine/core/pull/6011

### DIFF
--- a/tests/acceptance/10_files/01_create/013.cf
+++ b/tests/acceptance/10_files/01_create/013.cf
@@ -31,6 +31,9 @@ body delete init_delete
 
 bundle agent test
 {
+  meta:
+    "test_skip_needs_work" string => "hpux";
+
   vars:
       "owner" string => "nobody";
 


tests/acceptance/10_files/01_create/013.cf
```diff
@@ -31,6 +31,9 @@ body delete init_delete
 
 bundle agent test
 {
+  meta:
+    "test_skip_needs_work" string => "hpux";
```
craigcomstock, line 35: We should really have a "why" and a reference to a ticket.

craigcomstock (approved):

tests/acceptance/10_files/01_create/013.cf
larsewi: I will investigate

PR #6034, opened 2026-02-09
Make expandrange able to output a single value

amousset:
- Allow ranges with a unique value. It looks to me like an expected behavior compared to other implementations of ranges (python, etc.). This should not be a compatibility problem as it would currently crash the agent.
- Make invalid step a soft fail. I don't see a reason to stop the agent in case the value is invalid (but I may be missing something).

Rationale: This avoids some special case handling when computing ranges where the bounds are equal (or lower than the step), with a consistent interface.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -4558,9 +4558,11 @@ static FnCallResult FnCallExpandRange(EvalContext *ctx, ARG_UNUSED const Policy 
         sscanf(template, "%[^[\[][%d-%d]%[^\n]", before, &from, &to, after);
     }
 
-    if (step_size < 1 || abs(from-to) < step_size)
+    if (step_size < 1)
     {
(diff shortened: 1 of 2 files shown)


libpromises/evalfunction.c
```diff
@@ -4558,9 +4558,11 @@ static FnCallResult FnCallExpandRange(EvalContext *ctx, ARG_UNUSED const Policy
         sscanf(template, "%[^[\[][%d-%d]%[^\n]", before, &from, &to, after);
     }
 
-    if (step_size < 1 || abs(from-to) < step_size)
+    if (step_size < 1)
     {
-        FatalError(ctx, "EXPANDRANGE Step size cannot be less than 1 or greater than the interval");
+        assert(fp->name != NULL);
```
larsewi: I see you tried to fix the CodeQL issue. Can you change it to
```C
assert(fp != NULL);
```
and put it in the top of the function?

larsewi (approved):
Thanks @amousset! This makes sense to me 🚀

PR #6071, opened 2026-04-14
fix: ci/install.sh should also generate a hostkey with cf-key

craigcomstock:


nickanderson (approved):
that will return 1 if there is a key there and probably fail fast which is preferable.

olehermanse (approved):

olehermanse:
Cherry-pick: https://github.com/cfengine/core/pull/6137

PR #6079, opened 2026-04-20
ENT-13879: Expanded syntax-description to include if param is optional

SimonThalvorsen:
Ticket: ENT-13879
Changelog: title


SimonThalvorsen:
https://github.com/cfengine/cfengine-cli/pull/74

libpromises/syntax.h
```diff
@@ -121,8 +121,11 @@ JsonElement *SyntaxToJson(void);
 #define PromiseTypeSyntaxNew(agent_type, promise_type, constraints, check_fn, status) { agent_type, promise_type, constraints, check_fn, status }
 #define PromiseTypeSyntaxNewNull() PromiseTypeSyntaxNew(NULL, NULL, NULL, NULL, SYNTAX_STATUS_NORMAL)
 
-#define FnCallTypeNew(name, return_type, arguments, implementation, description, opts, category, status) { name, return_type, arguments, implementation, description, .options = opts, category, status }
-#define FnCallTypeNewNull() FnCallTypeNew(NULL, CF_DATA_TYPE_NONE, NULL, NULL, NULL, false, FNCALL_CATEGORY_UTILS, SYNTAX_STATUS_NORMAL)
+#define DEFAULT_ARGC -1, -1
+#define ARGC(min, max) min, max
+
+#define FnCallTypeNew(name, return_type, arguments, implementation, description, opts, category, status, argc_range) { name, return_type, arguments, implementation, description, .options = opts, category, status, .min_args = argc_range}
```
olehermanse: Here you are relying on overwriting `.max_args` after `.min_args`? I think you could probably make this a bit nicer by using a `union` in the FnCallType struct

libpromises/syntax.c
```diff
@@ -1291,6 +1291,9 @@
         }
         JsonObjectAppendArray(json_fn, "parameters", params);
     }
+    char range_buffer[32];
+    snprintf(range_buffer, sizeof(range_buffer), "%d,%d", fn_syntax->min_args, fn_syntax->max_args);
```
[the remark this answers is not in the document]
olehermanse: @SimonThalvorsen add `assert(fn_syntax != NULL)` to the beginning of this function.

libpromises/evalfunction.c
```diff
@@ -11521,429 +11521,433 @@ static const FnCallArg FIND_LOCAL_USERS_ARGS[] =
 #ifndef CFENGINE_EVALFUNCTION_TEST_C
 const FnCallType CF_FNCALL_TYPES[] =
 {
+    // -1, -1 // DEFAULT_ARGC, all required
+    // 1, -1  // ARGC(1, -1), 1-n
+    // 0, -1  // 0-n
+    // 2, 3   // 2-3
```
olehermanse: This comment looks a bit out of place. Maybe add a sentence above it about what it is, and/or move it to the place where `DEFAULT_ARGC` and `ARGC` are defined?

libpromises/syntax.c
```diff
@@ -1291,6 +1291,9 @@ static JsonElement *FnCallTypeToJson(const FnCallType *fn_syntax)
...
+    JsonObjectAppendString(json_fn, "argRange", range_buffer);
```
olehermanse: I was thinking we could make this field optional, skip this for all the cases where it is `-1, -1`.

olehermanse: Also split it up to `minArgs` and `maxArgs` JSON ints. Probably makes it easier to read, and easier to use, and avoids the need to use any additional parsing / checking. Both should be optional so it's easy to specify only one or the other.

olehermanse (commented):
Looks good overall, just some smaller suggestions.

olehermanse:
@SimonThalvorsen When this one is done and merged, you should try to adjust cf-promises / cf-agent to also use these minimums and maximums. Right now, before this change, there is no way to know if a call has too many args without actually running it (IIRC). With this new info, cf-promises could detect more errors earlier.

(Do this in a separate PR).

libpromises/fncall.h
```diff
@@ -54,6 +54,11 @@ typedef struct
     const char *description;
 } FnCallArg;
 
+typedef union {
+    struct { int min, max; };
+    int raw[2];
```
larsewi: Please explain why the `int raw[2]` is there :)

olehermanse: @SimonThalvorsen now you made it into a struct and you're not using the union / raw name. I guess you can simply do;

```c
typedef struct {
  int mix;
  int max;
} FnArgcRange;
```

olehermanse (changes requested):
Please adjust `FnArgcRange` as discussed above - union can be removed or `raw` could be used and there could be an explaining comment.

olehermanse (approved):

PR #6085, opened 2026-04-22
changelog-generator: Emit markdown bullets without leading tabs

larsewi:
The CHANGELOG.md format no longer uses tab indentation for entries, so drop the tab from the bullet prefix and the continuation indent in `changelog-generator`. This makes the tool's output drop-in compatible with the current markdown CHANGELOG.md produced after ENT-13497.

Before:
```
	- Added foo (CFE-1234)
	- Added bar with
	  a continuation line (CFE-5678)
```

After:
```
- Added foo (CFE-1234)
- Added bar with
  a continuation line (CFE-5678)
```

### DIFF
--- a/misc/changelog-generator/changelog-generator
+++ b/misc/changelog-generator/changelog-generator
@@ -254,12 +254,12 @@ for sha_entry in ENTRIES:
 if SORT_CHANGELOG:
     entry_list.sort()
 for entry in entry_list:
-    entry = "\t- " + entry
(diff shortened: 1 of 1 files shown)


olehermanse (approved):

larsewi:
Backported to:
- https://github.com/cfengine/core/pull/6086
- https://github.com/cfengine/core/pull/6087

PR #6092, opened 2026-04-24
Remove maxArg from syntax-description in cases where it is unbound

SimonThalvorsen:


libpromises/syntax.c
```diff
@@ -1293,9 +1293,12 @@ static JsonElement *FnCallTypeToJson(const FnCallType *fn_syntax)
         JsonObjectAppendArray(json_fn, "parameters", params);
     }
```
olehermanse, line 1295: I think you could also have an assert here;

```
if (fn_syntax->argc.min == -1)
{
    assert(fn_syntax->argc.max == -1);
}
```

olehermanse (approved):

PR #6109, opened 2026-05-04
Added mpf_masterfiles_stage_run_always class to enable legacy behavior of masterfiles-stage to always run cfbs build

craigcomstock:
This enables the previous unwanted behavior of running cfbs regardless of whether changes exist.
In some cases this behavior was used to run scripts which updated information dynamically on a periodic basis.
The goal of cfbs is to make reproducible builds so this strategy is not recommended and the behavior in masterfiles-stage is deprecated.

Ticket: ENT-13756
Changelog: title

### DIFF
--- a/contrib/masterfiles-stage/README.org
+++ b/contrib/masterfiles-stage/README.org
@@ -98,6 +98,16 @@ are pre-installed when using CFEngine Enterprise.
 
 Next, configure the default parameter set (=DEFAULT_PARAMS_FILE=).
 
+*** Legacy behavior
+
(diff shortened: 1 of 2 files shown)


nickanderson (approved):
Makes sense. Should this be elevated to the README?

craigcomstock:
> Makes sense. Should this be elevated to the README?

yes, already done, was very draft when you reviewed :)

olehermanse (approved):

craigcomstock:
cherry picks: https://github.com/cfengine/core/pull/6111 and https://github.com/cfengine/core/pull/6112

PR #6140, opened 2026-05-26
ENT-14139: Fixed cf-agent SIGABRT on SIGTERM during early policy validation

larsewi:
When SIGTERM arrives before `ThisAgentInit()`, the signal handler calls `TerminateCustomPromises()` while `custom_modules` is still `NULL`, causing an assert to be triggered in `MapIteratorInit()`.

Ticket: ENT-14139

Backported to:
- https://github.com/cfengine/core/pull/6144
- https://github.com/cfengine/core/pull/6145

### DIFF
--- a/libpromises/mod_custom.c
+++ b/libpromises/mod_custom.c
@@ -1199,6 +1199,11 @@ static void PromiseModule_Terminate_untyped(void *data)
 
 void TerminateCustomPromises(void)
 {
+    if (custom_modules == NULL)
+    {
+        return;
+    }
+
     MapIterator iter = MapIteratorInit(custom_modules);
 
     for (const MapKeyValue *item = MapIteratorNext(&iter); item != NULL; item = MapIteratorNext(&iter))


larsewi:
@cf-bottom Jenkins please :)

craigcomstock (approved):
The change looks smart and right.

I wonder about how this situation happens or how it could be reproduced. I think that would be a good addition to the commit message.

larsewi:
> The change looks smart and right.
> 
> I wonder about how this situation happens or how it could be reproduced. I think that would be a good addition to the commit message.

@craigcomstock, I feel like the commit message already says this. Maybe it could be stated even clearer. It happens if you kill the agent very early.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/13851/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/13851/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-13851/

PR #6158, opened 2026-06-04
 ENT-3868: Improved persistent class logging and plugged memory leak

nickanderson:
This change makes it easier to see when a persistent class timer is being reset. It highlighted a gap in classes promises, they are not able to specify the timer_policy, the ability to specify timer_policy is currently only available in classes bodies. Thus, currently, classes promises *always* cause a timer reset.

Ticket: ENT-3868

### DIFF
--- a/libpromises/eval_context.c
+++ b/libpromises/eval_context.c
@@ -753,6 +753,18 @@ void EvalContextHeapPersistentSave(EvalContext *ctx, const char *name, unsigned 
                     free(new_info);
                     return;
                 }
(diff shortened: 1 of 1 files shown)


craigcomstock (approved):
seems reasonable enough. How did you see the memory issue? Might be useful information for that commit.

nickanderson:
> seems reasonable enough. How did you see the memory issue? Might be useful information for that commit.

LLM pointed it out to me wile I was looking for these logs.

PR #6163, opened 2026-06-05
ENT-14104: Added GETPATCH protocol command for serving leech2 patches

larsewi:
The hub sends 'GETPATCH <hash>' over the established TLS connection, and the server replies with a leech2 patch using a new patch stream API, which reuses the wire format from the file stream protocol. The patch is created by an enterprise hook; CFEngine Community refuses the request. This introduces protocol version 5 - "leech2".

Ticket: ENT-14104

Together with:
- https://github.com/cfengine/nova/pull/2612
- https://github.com/cfengine/enterprise/pull/927

### DIFF
--- a/cf-serverd/server_tls.c
+++ b/cf-serverd/server_tls.c
@@ -36,11 +36,13 @@
 #include <lastseen.h>                 /* LastSaw1 */
 #include <net.h>                      /* SendTransaction,ReceiveTransaction */
 #include <tls_generic.h>              /* TLSSend */
+#include <patch_stream.h>             /* PatchStreamRefuse */
 #include <cf-serverd-enterprise-stubs.h>
 #include <connection_info.h>
 #include <regex.h>                                       /* StringMatchFull */
 #include <known_dirs.h>
 #include <file_lib.h>                                           /* IsDirReal */
+#include <string_lib.h>
 
 #include "server_access.h"          /* access_CheckResource, acl_CheckExact */
 
@@ -706,7 +708,7 @@ bool BusyWithNewProtocol(EvalContext *ctx, ServerConnectionState *conn)
         char *args = &recvbuffer[EXEC_len];
         args += strspn(args, " \t");                       /* bypass spaces */
 
-        Log(LOG_LEVEL_VERBOSE, "%14s %7s %s",
+        Log(LOG_LEVEL_VERBOSE, "%14s %8s %s",
             "Received:", "EXEC", args);
 
         bool b = DoExec2(ctx, conn, args,
@@ -736,7 +738,7 @@ bool BusyWithNewProtocol(EvalContext *ctx, ServerConnectionState *conn)
             goto protocol_error;
         }
 
-        Log(LOG_LEVEL_VERBOSE, "%14s %7s %s",
+        Log(LOG_LEVEL_VERBOSE, "%14s %8s %s",
             "Received:", "GET", filename);
 
         /* TODO batch all the following in one function since it's very
@@ -760,7 +762,7 @@ bool BusyWithNewProtocol(EvalContext *ctx, ServerConnectionState *conn)
 
         PathRemoveTrailingSlash(filename, strlen(filename));
 
-        Log(LOG_LEVEL_VERBOSE, "%14s %7s %s",
+        Log(LOG_LEVEL_VERBOSE, "%14s %8s %s",
             "Translated to:", "GET", filename);
 
         if (acl_CheckPath(paths_acl, filename,
@@ -799,7 +801,7 @@ bool BusyWithNewProtocol(EvalContext *ctx, ServerConnectionState *conn)
             goto protocol_error;
         }
 
-        Log(LOG_LEVEL_VERBOSE, "%14s %7s %s",
+        Log(LOG_LEVEL_VERBOSE, "%14s %8s %s",
             "Received:", "OPENDIR", filename);
 
         /* sizeof()-1 because we need one extra byte for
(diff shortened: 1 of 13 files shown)


larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/13939/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/13939/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-13939/

cf-serverd/server_tls.c
```diff
@@ -1116,6 +1117,43 @@ bool BusyWithNewProtocol(EvalContext *ctx, ServerConnectionState *conn)
...
+        {
+            goto protocol_error;
+        }
+
+        char last_hash[64];
+        int ret = sscanf(recvbuffer, "GETPATCH %63s", last_hash);
+        if (ret != 1)
+        {
+            goto protocol_error;
+        }
```
olehermanse: Could check that `last_hash` only contains the allowed characters here. I.e. since hex hashes can only be a-f + 0-9, we could reject it if it contains other characters like `/.;'\` etc.

libcfnet/patch_stream.c
```diff
@@ -0,0 +1,104 @@
...
+{
+    assert(conn != NULL);
+    assert(data != NULL || len == 0);
+
+    const char *buf = data;
+    size_t offset = 0;
+    bool eof = false;
+
+    while (!eof)
+    {
```
olehermanse, line 52: We should probably set some limit here and handle the case where client is trying to send patches "forever"?

cf-serverd/server_tls.c
```diff
@@ -1116,6 +1117,43 @@ bool BusyWithNewProtocol(EvalContext *ctx, ServerConnectionState *conn)
...
+        const bool access_to_query_delta = acl_CheckExact(
+            query_acl, "delta", conn->ipaddr, conn->revdns, hostkey);
+
+        if (!access_to_query_delta)
+        {
+            Log(LOG_LEVEL_INFO, "access denied to GETPATCH: %s", recvbuffer);
+            /* Refuse using the stream protocol, as opposed to
+             * RefuseAccess(), because the client expects stream protocol
+             * messages after sending the GETPATCH request. */
+            PatchStreamRefuse(ConnectionInfoSSL(conn->conn_info));
```
olehermanse: Here I think you should handle the return value of `PatchStremRefuse`, or comment on why it is not necessary.

cf-serverd/server_tls.c
```diff
@@ -1116,6 +1117,43 @@ bool BusyWithNewProtocol(EvalContext *ctx, ServerConnectionState *conn)
...
+            return true;
+        }
+
+        /* A failed patch stream is not fatal to the connection. The
+         * enterprise hook does all relevant Log()ing. */
+        ServeLeech2Patch(conn, last_hash);
```
olehermanse: AI review:

> The comment says a failed stream "is not fatal to the connection," which is fine for the application-level failure, but if the refusal message itself fails to send the TLS connection may be in a broken state and the busy-loop will keep trying to serve it. Existing handlers (e.g., COOKIE → ReturnCookies) at least break to the protocol-error path on failure.

libcfnet/patch_stream.c
```diff
@@ -0,0 +1,104 @@
...
+        if (!ProtocolRecvMessage(conn, msg, &msg_len, &eof))
+        {
+            /* Error is already logged in ProtocolRecvMessage() */
+            free(buf);
+            return false;
+        }
+
+        if (msg_len > 0)
```
olehermanse, line 93: When do we receive a 0 length message?

libcfnet/patch_stream.h
```diff
@@ -0,0 +1,93 @@
...
+ * @return true on success, otherwise false
+ */
+bool PatchStreamServe(SSL *conn, const void *data, size_t len);
+
+/**
+ * @brief Fetch a patch using the stream API
+ *
+ * @param conn The SSL connection object
+ * @param data Is set to the received patch buffer (caller takes ownership
+ *             and must free it with free())
```
olehermanse: Highlight the fact that it may be set to NULL and must be checked for NULL before access?

cf-serverd/server_tls.c
```diff
@@ -1116,6 +1117,43 @@ bool BusyWithNewProtocol(EvalContext *ctx, ServerConnectionState *conn)
...
+
+        char last_hash[64];
+        int ret = sscanf(recvbuffer, "GETPATCH %63s", last_hash);
+        if (ret != 1)
+        {
+            goto protocol_error;
+        }
+
+        Log(LOG_LEVEL_VERBOSE, "%14s %7s %.7s...",
+            "Received:", "GETPATCH", last_hash);
```
olehermanse: This won't be formatted correctly / as expected. `%7s` is meant to pad anything below 7 characters up to 7, so all of them are aligned - however `GETPATCH` is 8 characters, so all those log messages will be misaligned. We should probably change this and all the other log message format strings to `%8s`.

olehermanse (approved):
Looks good overall. Some comments / suggestions below.

libcfnet/patch_stream.c
larsewi: This is `cf-serverd`, not `cf-hub`. The code is here only in case `TLSSend()` is interrupted, causing parts of the buffer to be sent. However, there should probably be such a limit in `cf-hub`. Hence, it should be implemented in `PatchStreamFetch()`.

libcfnet/patch_stream.c
larsewi: When you send a PDU with a header but no payload attached. E.g., when you send the End-of-File flag, or the Error flag without an error message. Or any other flags that we may support later on.

olehermanse: I would expect that kind of information to be in a comment.

larsewi: https://github.com/cfengine/core/pull/6184

PR #6167, opened 2026-06-05
CFE-4681: timer_policy support for classes: promises

nickanderson:
## Summary

- Add `timer_policy` attribute to `classes:` promise type, allowing users to control whether a persistent class timer resets on re-evaluation (`reset`) or preserves the original expiry (`absolute`)
- Default `timer_policy` for `classes:` promises is `absolute` (PRESERVE) for backward compatibility — historically these promises skip re-evaluation when the class is already defined, so the timer was never reset
- Bypass the `ExpandDeRefPromise` skip when `timer_policy => "reset"` is explicitly set with `persistence > 0`, allowing the promise to reach `VerifyClassPromise` so the DB timer can be updated
- Add second-layer bypass in `VerifyClassPromise` for the case where `EvalClassExpression` returns false (class already in context from persistent DB) but `timer_policy` is `reset`
- Fix `ValueSizeDB` key length to include null terminator (`strlen+1`), matching `ReadDB`/`WriteDB` conventions

## Test plan
- [x] Manual test: `timer_policy => "absolute"` creates class with "policy preserve", second run skips the promise
- [x] Manual test: `timer_policy => "reset"` creates class with "policy reset", second run resets the timer
- [x] Acceptance test `persistent_timer_policy.cf`: verifies absolute/preserve log output
- [x] Acceptance test `persistent_timer_policy_reset.cf`: verifies timer reset across two agent runs
- [ ] CI acceptance tests pass

Ticket: [CFE-4681](https://northerntech.atlassian.net/browse/CFE-4681)

[CFE-4681]: https://northerntech.atlassian.net/browse/CFE-4681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libpromises/attributes.c
+++ b/libpromises/attributes.c
@@ -1133,6 +1133,26 @@ ContextConstraint GetContextConstraints(const EvalContext *ctx, const Promise *p
     a.persistent = PromiseGetConstraintAsInt(ctx, "persistence", pp);
 
     {
+        const char *tp = PromiseGetConstraintAsRval(pp, "timer_policy", RVAL_TYPE_SCALAR);
+        if (StringEqual(tp, "reset"))
+        {
+            a.timer = CONTEXT_STATE_POLICY_RESET;
+        }
+        else
+        {
+            /* Default to PRESERVE (absolute) for backward compatibility:
+             * classes: promises historically skip re-evaluation when the
+             * class is already defined, so the timer was never reset.
+             *
+             * NOTE: this default is planned to change to RESET in 3.28.0
+             * (CFE-4681) to align with classes bodies, whose timer_policy
+             * already defaults to reset.  The absolute default is retained on
+             * the 3.24 and 3.27 LTS backports. */
+            a.timer = CONTEXT_STATE_POLICY_PRESERVE;
+        }
+    }
+
+    {
         const char *context_scope = PromiseGetConstraintAsRval(pp, "scope", RVAL_TYPE_SCALAR);
         a.scope = ContextScopeFromString(context_scope);
     }
@@ -1143,7 +1163,9 @@ ContextConstraint GetContextConstraints(const EvalContext *ctx, const Promise *p
 
         for (int k = 0; CF_CLASSBODY[k].lval != NULL; k++)
         {
-            if (strcmp(cp->lval, "persistence") == 0 || strcmp(cp->lval, "scope") == 0)
+            if (StringEqual(cp->lval, "persistence") ||
+                StringEqual(cp->lval, "scope") ||
+                StringEqual(cp->lval, "timer_policy"))
             {
                 continue;
             }
--- a/libpromises/eval_context.c
+++ b/libpromises/eval_context.c
@@ -735,7 +735,7 @@ void EvalContextHeapPersistentSave(EvalContext *ctx, const char *name, unsigned 
 
     // first see if we have an existing record, and if we should bother to update
     {
-        int existing_info_size = ValueSizeDB(dbp, key, strlen(key));
+        int existing_info_size = ValueSizeDB(dbp, key, strlen(key) + 1);
         if (existing_info_size > 0)
         {
             PersistentClassInfo *existing_info = xcalloc(existing_info_size, 1);
--- a/libpromises/cf3.defs.h
+++ b/libpromises/cf3.defs.h
@@ -1210,6 +1210,7 @@ typedef struct
     ContextScope scope;
     int nconstraints;
     int persistent;
+    PersistentClassPolicy timer;
 } ContextConstraint;
 
 /*************************************************************************/
--- a/libpromises/mod_common.c
+++ b/libpromises/mod_common.c
@@ -226,6 +226,7 @@ const ConstraintSyntax CF_CLASSBODY[] =
     ConstraintSyntaxNewContext("not", "Evaluate the negation of string expression in normal form", SYNTAX_STATUS_NORMAL),
     ConstraintSyntaxNewContextList("select_class", "Select one of the named list of classes to define based on host identity. Default value: random_selection", SYNTAX_STATUS_NORMAL),
     ConstraintSyntaxNewContextList("xor", "Combine class sources with XOR", SYNTAX_STATUS_NORMAL),
+    ConstraintSyntaxNewOption("timer_policy", "absolute,reset", "Whether a persistent class restarts its counter when rediscovered. Default value: absolute", SYNTAX_STATUS_NORMAL),
     ConstraintSyntaxNewNull()
 };
 
--- a/libpromises/promises.c
+++ b/libpromises/promises.c
@@ -705,16 +705,31 @@ Promise *ExpandDeRefPromise(EvalContext *ctx, const Promise *pp, bool *excluded)
     pcopy->org_pp = pp->org_pp;
 
     // if this is a class promise, check if it is already set, if so, skip
+    // Exception: persistent classes with timer_policy => "reset" must not
+    // be skipped — the promise needs to fire so the timer gets reset.
     if (strcmp("classes", PromiseGetPromiseType(pp)) == 0)
     {
         if (IsDefinedClass(ctx, CanonifyName(pcopy->promiser)))
         {
-            Log(LOG_LEVEL_DEBUG,
-                "Skipping evaluation of classes promise as class '%s' is already set",
-                CanonifyName(pcopy->promiser));
+            const char *tp = PromiseGetConstraintAsRval(pp, "timer_policy", RVAL_TYPE_SCALAR);
+            int persistence = PromiseGetConstraintAsInt(ctx, "persistence", pp);
 
-            *excluded = true;
-            return pcopy;
+            if (StringEqual(tp, "reset") && persistence > 0)
+            {
+                Log(LOG_LEVEL_DEBUG,
+                    "Class '%s' is already set but timer_policy is reset"
+                    " — allowing promise evaluation to reset persistence timer",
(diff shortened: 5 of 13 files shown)


cf-bottom:

nickanderson:
@cf-bottom jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/13940/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/13940/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-13940/

libpromises/attributes.c
```diff
@@ -1132,6 +1132,21 @@ ContextConstraint GetContextConstraints(const EvalContext *ctx, const Promise *p
     a.expression = NULL;
     a.persistent = PromiseGetConstraintAsInt(ctx, "persistence", pp);
 
+    {
+        const char *tp = PromiseGetConstraintAsRval(pp, "timer_policy", RVAL_TYPE_SCALAR);
+        if (tp != NULL && strcmp(tp, "reset") == 0)
```
vpodzime: Please use `StringEqual()` (more places below)

libpromises/eval_context.c
```diff
@@ -735,7 +735,7 @@ void EvalContextHeapPersistentSave(EvalContext *ctx, const char *name, unsigned
 
     // first see if we have an existing record, and if we should bother to update
     {
-        int existing_info_size = ValueSizeDB(dbp, key, strlen(key));
+        int existing_info_size = ValueSizeDB(dbp, key, strlen(key) + 1);
```
vpodzime, line 738: This is an unrelated change that should go into a separate commit, isn't it? And it doesn't look correct at the first glance. AFAICT, we don't include the terminating NUL-bytes in LMDB data (but I'd have to double-check).

vpodzime (commented):
So the idea here is that the `class` promise would be in a file that's not loaded at all? How else would one make sure the promise is not evaluated? Classes are part of pre-eval, right? I think we should have a test case for this as well.

nickanderson:
Thanks for taking a look at my :robot: :sparkles: Feel free to be harsh.

> So the idea here is that the `class` promise would be in a file that's not loaded at all? How else would one make sure the promise is not evaluated? 

So, the thing is that currently/prior to this PR there are two ways to define a persistent class.

1) Via a classes promise.

```
classes:
  "persistent"
    persistence => "100";
```

Currently there is no way to extend the time remaining that the persistent class should be defined. It The classes promise itself won't even trigger if the promiser is a defined class (optimization for not doing work you don't need to do), so you have to wait for the class to go beyond the timer and then be undefined, so it may then not be defined for some time during next policy evaluation (e.g. maybe it was guarded behind a separate class that was the result of a promise so that it wasnt resolved during pre-eval) until the classes promise is encountered and evaluated.

2) Via a classes body / promise result.

```
body classes results_persist_timer(class_prefix, persist_min, timer)
{
      inherit_from => results( "namespace", "$(class_prefix)" );
      persist_time => "$(persist_min)";

      #  (extend|absolute) whether to extend the time left on persistence or not
      timer_policy => "extend"; 
}
```

Classes bodies already have the ability to either extend the timer or not. Extending the timer helps to avoid that hole where a class might not be defined for a period of time until the promise re-defines it.

> Classes are part of pre-eval, right? I think we should have a test case for this as well.

Yes, but only classes in common bundles are part of pre-eval: https://docs.cfengine.com/docs/lts/reference/language-concepts/policy-evaluation/#agent-pre-evaluation-step

The following steps are executed per-bundle for each file parsed, in this order:

if it’s a common bundle, evaluate vars promises
if it’s a common bundle, evaluate classes promises
evaluate vars promises (for details see PolicyResolve() in the C code)

libpromises/attributes.c
nickanderson: Done in 4c500a81a — switched to `StringEqual()` here, in the `persistence`/`scope`/`timer_policy` skip list just below, and in the `timer_policy => "reset"` check in `ExpandDeRefPromise`.

libpromises/eval_context.c
nickanderson: Pulled this into its own commit (627d5e2b1). Digging in, it turns out to be correct *and* a prerequisite rather than unrelated.

`WriteDB`/`ReadDB`/`DeleteDB` all store and look up keys with `strlen(key) + 1` (the terminating NUL is included — see `dbm_api.c`), and LMDB matches keys by exact byte length. So `ValueSizeDB(dbp, key, strlen(key))` searched for a key one byte shorter than the stored one and never matched — the whole existing-record branch in `EvalContextHeapPersistentSave` was dead code.

Verified by building both ways and running the reset policy twice:
- **without** `+1`: run 2 logs `Creating persistent class … policy reset` (existing record never found)
- **with** `+1`: run 2 logs `Resetting persistent class … (was 120 minutes remaining)`

The timer still physically resets either way (the `WriteDB` is unconditional), but without this fix the `absolute`/preserve early-return (don't overwrite an already-preserved, unexpired timer) is unreachable, and the reset acceptance test can't pass. Happy to move it to a standalone PR instead if you'd prefer.

nickanderson:
On testing that the promise isn't evaluated: added `persistent_timer_policy.cf` as the `absolute` (default) counterpart. It runs the agent twice and asserts that on the second run the classes: promise is skipped in `ExpandDeRefPromise` (`Skipping evaluation of classes promise … is already set`) and the timer is **not** reset — i.e. the promise is genuinely not evaluated once the class is loaded from the persistent DB.

You're right that classes are pre-eval, but only for common bundles; the skip itself happens in `ExpandDeRefPromise`, which applies to both pre-eval and normal agent-bundle evaluation, so the test uses an agent bundle.

I also tightened the reset test so it actually catches a broken existing-record lookup (confirmed: it fails without the `ValueSizeDB` fix and passes with it), and moved both tests' two sub-agent runs into the `test` bundle to remove a spurious `FAIL` report during the agent's convergence passes.

nickanderson:
@cf-bottom jenkins please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/13942/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/13942/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-13942/

vpodzime (commented):
Nice! Looks good to me except for the first commit message which is too long and confusing (at least to me) -- I'd not refer to code that was never executed due to a bug in an `if` condition as dead code. It can easily just say that it fixes how the key is looked up which actually enables the behavior relying on its existence in the DB and the related value.

nickanderson:
@vpodzime Good call — reworded the first commit. Dropped the "dead code" framing and shortened it: the subject now points at the actual bug (the key length used in the lookup), and the body explains it in terms of the behavior the fix enables — finding the existing record so the preserve/reset logic can act on it — rather than describing a dead branch. Force-pushed (diff unchanged).

nickanderson:
@cf-bottom jenkins please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/13943/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/13943/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-13943/

vpodzime (approved):
Still too wordy to my taste, but much better.

nickanderson:
@olehermanse can I get a :green_apple:  ?

libpromises/attributes.c
```diff
@@ -1132,6 +1132,21 @@ ContextConstraint GetContextConstraints(const EvalContext *ctx, const Promise *p
...
+        if (StringEqual(tp, "reset"))
+        {
+            a.timer = CONTEXT_STATE_POLICY_RESET;
+        }
+        else
+        {
+            /* Default to PRESERVE (absolute) for backward compatibility:
+             * classes: promises historically skip re-evaluation when the
+             * class is already defined, so the timer was never reset. */
+            a.timer = CONTEXT_STATE_POLICY_PRESERVE;
```
olehermanse, line 1151: I would maybe mention it here and/or in the commit message that we're planning to change this default.

olehermanse (dismissed):
I think we should error if someone tries to use `timer_policy` without `persistence`. AFAICT you have not added any error for this(?)

nickanderson:
Thanks @olehermanse — both addressed in 53afadc4f:

- **Error without `persistence`:** `VerifyClassPromise` now fails with "Cannot use 'timer_policy' without 'persistence' …" when `timer_policy` is set but `persistence` is not (it only governs the persistence timer). New acceptance test `timer_policy_requires_persistence.cf` covers it.
- **Planned default change:** noted at the `PRESERVE` default in `attributes.c` and in the commit message — 3.28.0 flips the default to `reset` (stacked #6174), keeping `absolute` on the 3.24/3.27 backports.

olehermanse (approved):

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/13947/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/13947/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-13947/

PR #6172, opened 2026-06-10
Reject directory listing entries containing a path separator

aizu-m:
Found while tracing cf-agent's recursive copy from a remote server. The
directory listing entries pass through ConsiderFile() and are then joined
to the local path by PathAppend(), which only concatenates:

    entry  = "../../../etc/cron.d/x"
    joined = "<dest>/../../../etc/cron.d/x"

ConsiderFile() already rejects ".", ".." and "..." (the last with a
CWE-32 note) but not a name that carries a separator. A genuine dirent is
one path component and never contains '/', so a listing entry that does
is a hostile server stepping outside the destination. cf-agent normally
runs as root, so that is an arbitrary file write.

cf-net "getdir" walks a remote listing the same way in
process_dir_recursive(); it skips "." and ".." and nothing else.

Both now reject an entry that contains a separator (IsFileSep). The unit
test fails before the change and passes after.

### DIFF
--- a/cf-agent/files_properties.c
+++ b/cf-agent/files_properties.c
@@ -26,6 +26,7 @@
 
 #include <files_names.h>
 #include <files_interfaces.h>
+#include <file_lib.h>                                      /* IsFileSep */
 #include <item_lib.h>
 
 static Item *SUSPICIOUSLIST = NULL; /* GLOBAL_P */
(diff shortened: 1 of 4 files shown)


larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/13969/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/13969/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-13969/

larsewi (approved):

PR #6177, opened 2026-06-12
CFE-4687: Fixed off-by-one in protocol recv buffers

nickanderson:
ProtocolOpenDir()/ProtocolGet() receive up to CF_MSGSIZE bytes into a char buf[CF_MSGSIZE], but the receive primitives NUL-terminate at buf[received] where received can equal CF_MSGSIZE -- one past the array (TLSRecv/RecvSocketStream both write toget+1 bytes).

Drives the real ProtocolOpenDir() over a classic-protocol socketpair with a record-filling reply. Aborts under AddressSanitizer on current code, passes once the buffers are sized CF_BUFSIZE (cfengine/core#6171).

Ref: CFE-4687, cfengine/core#6171

### DIFF
--- a/libcfnet/protocol.c
+++ b/libcfnet/protocol.c
@@ -38,7 +38,7 @@ Seq *ProtocolOpenDir(AgentConnection *conn, const char *path)
     assert(conn != NULL);
     assert(path != NULL);
 
-    char buf[CF_MSGSIZE] = {0};
+    char buf[CF_BUFSIZE] = {0};
     int tosend = snprintf(buf, CF_MSGSIZE, "OPENDIR %s", path);
     if (tosend < 0 || tosend >= CF_MSGSIZE)
     {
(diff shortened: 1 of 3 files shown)


nickanderson:
@cf-bottom jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/13946/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/13946/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-13946/

olehermanse (approved):
At first glance, I'd say it sounds more appropriate that the buffers should be +1 (`CF_MSGSIZE + 1`) to account for the NUL-byte. However, after reading the comment in your test, I see that `CF_BUFSIZE` is already used in another place for the same thing (`STAT`), so it makes some sense to do the same as there.

vpodzime (approved):
ACK

PR #6178, opened 2026-06-13
Fixed out-of-bounds read in module protocol when a '%' line has no '='

aizu-m:
AddressSanitizer, parsing a "%" module-protocol line with no "=":

    ERROR: AddressSanitizer: SEGV on a READ
        #0 strlen
        #1 BufferNewFrom buffer.c:68
        #2 ModuleProtocol evalfunction.c:10252

Came across this whilst reading the module protocol parser. The "%" case
builds the JSON payload with

    BufferNewFrom(line + strlen(name) + 1 + 1,
                  length - strlen(name) - 1 - 1);

which assumes a well-formed "%name=<json>" line. But sscanf("%256[^=]=",
name) also returns with name set when the line carries no "=": it leaves
the whole tail in name. So strlen(name) == length - 1, the length argument
underflows to SIZE_MAX, and the data pointer steps one past the terminating
NUL. BufferAppend() then scans from there and walks off the heap line
buffer.

read_module_protocol() hands arbitrary file lines straight to
ModuleProtocol(), so a single line "%x" is enough to reproduce it.

The other directives ("=", "@", ...) already guard their input; the "%"
case did not. Only do the arithmetic once the "=" delimiter is present.

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -10246,7 +10246,10 @@ void ModuleProtocol(EvalContext *ctx, const char *command, const char *line, int
         // context name once it's in the vartable.  Maybe this can be relaxed.
         sscanf(line + 1, "%256[^=]=", name);
 
-        if (CheckID(name))
+        /* A well-formed line is "%name=<json>". Without the '=' delimiter
+         * sscanf() leaves the whole tail in name, so strlen(name) == length-1
+         * and the "length - strlen(name) - 1 - 1" below underflows. */
+        if (CheckID(name) && line[strlen(name) + 1] == '=')
         {
             JsonElement *json = NULL;
             Buffer *holder = BufferNewFrom(line+strlen(name)+1+1,
--- a/tests/unit/evalfunction_test.c
+++ b/tests/unit/evalfunction_test.c
@@ -126,6 +126,39 @@ static void test_basename(void)
     basename_single_testcase("//a//b///c.csv////", ".csv", "c");
 }
 
+static void test_module_protocol_percent_no_delimiter(void)
+{
+    EvalContext *ctx = EvalContextNew();
+    StringSet *tags = StringSetNew();
+    long persistence = 0;
+    char context[CF_BUFSIZE] = "test";
+
+    /* Lines come from CfReadLine() on the heap, so allocate them that way:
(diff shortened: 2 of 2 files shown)


cf-bottom:

libpromises/evalfunction.c
```diff
@@ -10246,7 +10246,10 @@ void ModuleProtocol(EvalContext *ctx, const char *command, const char *line, int
         // context name once it's in the vartable.  Maybe this can be relaxed.
         sscanf(line + 1, "%256[^=]=", name);
 
-        if (CheckID(name))
+        /* A well-formed line is "%name=<json>". Without the '=' delimiter
+         * sscanf() leaves the whole tail in name, so strlen(name) == length-1
+         * and the "length - strlen(name) - 1 - 1" below underflows. */
+        if (CheckID(name) && line[strlen(name) + 1] == '=')
```
larsewi, line 10252: Your test passes without this change

larsewi (changes requested):

libpromises/evalfunction.c
aizu-m: Good catch, you are right. The OOB read only trips under ASan. In a plain build the scan past the line just reads adjacent heap, JsonParse() rejects the garbage, and the variable ends up undefined either way, so the behavioural check could not tell the patched and unpatched paths apart.

Reworked the test to be deterministic instead of leaning on ASan. The "%bad" line now sits with its terminating NUL on the last readable byte before a PROT_NONE guard page, so the underflowed scan walks straight into it. Confirmed both ways locally:

    without the fix: Bus error -> test_module_protocol_percent_no_delimiter: Test failed -> 1 out of 4 tests failed
    with the fix:    All 4 tests passed

Pushed in 5f94cc1.

larsewi (approved):
Thanks 🚀

larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/13979/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/13979/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-13979/

PR #6183, opened 2026-06-20
ENT-14105: Served leech2 patches from cf-testd via the enterprise hook

larsewi:
cf-testd now handles `GETPATCH` through the enterprise `ServeLeech2Patch()` function, the same code path as cf-serverd, instead of replying with a hardcoded empty patch. In a pure core build the stub refuses; with the enterprise plugin loaded a real patch is served from the state directory.

Ticket: ENT-14105


cf-bottom:

olehermanse (approved):

PR #6186, opened 2026-06-23
ENT-14193: Binary for event driven cfengine

victormlg:
* Used `cf-watchd` as a placeholder name for this new binary. This can be modified
* Placeholder daemon that simply prints every second
* Placeholder options copied from other components

### DIFF
--- a/cf-watchd/cf-watchd.c
+++ b/cf-watchd/cf-watchd.c
@@ -1,0 +1,8 @@
+#include <stdio.h>
+
+int main()
+{
+  printf("Hello world!\n");
+  return 0;
+}
+
--- a/cf-watchd/cf-watchd.c
+++ b/cf-watchd/cf-watchd.c
@@ -1,8 +1,257 @@
 #include <stdio.h>
+#include <unistd.h>
 
-int main()
+#include <exec_tools.h>
+#include <signals.h>
+#include <generic_agent.h>
+#include <logging.h>
+#include <eval_context.h>
+#include <cleanup.h>
+#include <man.h>
+#include <known_dirs.h>
+
+
+static GenericAgentConfig *CheckOpts(int argc, char **argv);
+static void WatchStart();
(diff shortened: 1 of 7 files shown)


cf-watchd/cf-watchd
```diff
@@ -0,0 +1,210 @@
+#! /bin/bash
```
olehermanse: This and the Makefile are generated files and should not be committed. Take a look at what is in the .gitignore file and add similar entries for the new daemon.

olehermanse (changes requested):

libpromises/cf3.defs.h
```diff
@@ -409,6 +409,7 @@ typedef enum
     AGENT_TYPE_RUNAGENT,
     AGENT_TYPE_KEYGEN,
     AGENT_TYPE_HUB,
+    AGENT_TYPE_WATCH,
```
larsewi, line 413: You will need a matching entry in `CF_AGENTTYPES[]` in `libpromises/constants.c`

cf-watchd/cf-watchd.c
```diff
@@ -0,0 +1,255 @@
...
+}
+
+static GenericAgentConfig *CheckOpts(int argc, char **argv)
+{
+    extern char *optarg;
+    int c;
+    GenericAgentConfig *config = GenericAgentConfigNewDefault(AGENT_TYPE_WATCH, GetTTYInteractive());
+
+    int longopt_idx;
+    while ((c = getopt_long(argc, argv, "dvIf:g:VSKMFhC::l",
```
larsewi: `-S` is accepted but not handled as a `case` in the `switch` below. `-M` is not documented in `OPTIONS`/`HINTS` above.

larsewi (changes requested):

larsewi (approved):

PR #6193, opened 2026-06-25
2026 06 25 workflow maintenance 3.24

craigcomstock:
- **Bump libntech to latest**
- **Added -Wextra and -Wno-sign-compare to github workflow gcc options**
- **Refactored variable length array code to use alloca()**

### DIFF
--- a/.github/workflows/asan_unit_tests.yml
+++ b/.github/workflows/asan_unit_tests.yml
@@ -15,6 +15,6 @@ jobs:
     - name: Run autotools / configure
       run: ./autogen.sh --enable-debug
     - name: Compile and link (make)
-      run: make -j8 CFLAGS="-Werror -Wall -fsanitize=address" LDFLAGS="-fsanitize=address"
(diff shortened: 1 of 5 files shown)


craigcomstock:
This includes a libntech bump to master. Not sure that is truly needed so that commit could be removed.

craigcomstock:
> This includes a libntech bump to master. Not sure that is truly needed so that commit could be removed.

The bump is needed as libntech has fixes for newer compilers.

larsewi (approved):

craigcomstock:
I see 2 acceptance tests need work here. Will work on them.

Strangely both of these tests PASSED elsewhere: https://github.com/cfengine/enterprise/actions/runs/29437361997/job/87429241655

olehermanse (changes requested):
Changes look good to me, but judging by your comment more work is needed here;

> I see 2 acceptance tests need work here. Will work on them.

PR #6220, opened 2026-07-07
Added dynamic lint workflow

victormlg:
https://github.com/cfengine/cfengine-cli/pull/172

### DIFF
--- a/.github/workflows/cfengine_cli.yml
+++ b/.github/workflows/cfengine_cli.yml
@@ -11,13 +11,26 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.14"
-      - name: Checkout masterfiles
+      - name: Checkout core
         uses: actions/checkout@v7
+        with:
+          submodules: recursive
       - name: Install CFEngine CLI
         run: pipx install cfengine
+      - name: Install dependencies
+        run: ./ci/dependencies.sh
+      - name: Run autotools / configure
+        run: ./autogen.sh --enable-debug
+      - name: Compile and link (make)
+        run: make -j8 CFLAGS="-Werror -Wall"
+      - name: Install CFEngine
+        run: ./ci/install.sh
+      
       - name: Run cfengine lint
         run: |
-          cfengine lint --strict no ./
+          /var/cfengine/bin/cf-promises --syntax-description json > /tmp/syntax.json
+          cfengine lint --syntax-description /tmp/syntax.json --strict no ./
+
       - name: Run cfengine format --check
         run: |
           cfengine format --check


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,33 @@
+name: Lint policy with CFEngine CLI
+on:
+  workflow_call:
```
olehermanse: This means you have to call this workflow from somewhere else, you haven't done that so it will not run.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Install dependencies
+        run: ./ci/dependencies.sh
+      - name: Run autotools / configure
+        run: ./autogen.sh --enable-debug
+      - name: Compile and link (make)
+        run: make -j8 CFLAGS="-Werror -Wall"
+      
+      - name: Run cfengine lint
+        run: |
+          cf-promises --syntax-description json > syntax.json && cfengine lint --syntax-description syntax.json --strict no ./
```
olehermanse: Please put these steps into the existing `cfengine_cli.yml` workflow instead of making a new one.

olehermanse (changes requested):

.github/workflows/cfengine_cli.yml
```diff
@@ -13,11 +13,29 @@ jobs:
           python-version: "3.14"
       - name: Checkout masterfiles
         uses: actions/checkout@v7
+        with:
+          persist-credentials: false
```
olehermanse: Why was this added?

.github/workflows/cfengine_cli.yml
```diff
@@ -13,11 +13,29 @@ jobs:
           python-version: "3.14"
       - name: Checkout masterfiles
         uses: actions/checkout@v7
```
olehermanse: name was already wrong here - this checks out core

.github/workflows/cfengine_cli.yml
```diff
@@ -13,11 +13,29 @@ jobs:
...
+        with:
+          persist-credentials: false
       - name: Install CFEngine CLI
         run: pipx install cfengine
+
+      - name: Clone CFEngine
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+          persist-credentials: false
```
olehermanse: No need to check out core again, just add the submodules part to the checkout above.

.github/workflows/cfengine_cli.yml
```diff
@@ -13,11 +13,29 @@ jobs:
...
+        run: ./autogen.sh --enable-debug
+      - name: Compile and link (make)
+        run: make -j8 CFLAGS="-Werror -Wall"
+      - name: Install CFEngine
+        run: ./ci/install.sh
+      
       - name: Run cfengine lint
         run: |
-          cfengine lint --strict no ./
+          /var/cfengine/bin/cf-promises --syntax-description json > /tmp/syntax.json && cfengine lint --syntax-description /tmp/syntax.json --strict no ./
```
olehermanse: Nicer if you split it across 2 lines instead of `&&`

olehermanse (changes requested):
Looks like it's working now - just some things which can be improved / cleaned up above.

.github/workflows/cfengine_cli.yml
victormlg: Because I checked out twice with different versions. I will remove it

olehermanse (approved):

victormlg:
Cherry-pick: https://github.com/cfengine/core/pull/6221

PR #6221, opened 2026-07-07
Added dynamic lint workflow (3.27)

victormlg:


victormlg:
With the changes from 3.28, the syntax of the output of `cf-promises --syntax-description json` has changed. However, `cfengine lint` expects the newest syntax description syntax. Thus, when we cannot generate `syntax-description.json` in the workflow for 3.27, as it would be an invalid syntax for cfengine lint, and would cause the tests to fail. The syntax description saved inside cfengine lint is more accurate.

PR #6222, opened 2026-07-09
CFE-90: reconcile storage mount options when they drift (opt-in remount)

nickanderson:
Storage/mount promises only applied `mount_options` to the initial mount and (if `edit_fstab` is enabled) the fstab entry. A filesystem already mounted with different options was never corrected, because `mount -a` skips already-mounted filesystems. This PR adds opt-in reconciliation of a live mount when its options drift from the promise, and along the way fixes several long-standing storage-promise bugs (CFE-1539, CFE-1863, CFE-2350, CFE-3366).

### Opt-in live remount reconciliation (CFE-90, CFE-1864)

New `mount` body attributes:
- `remount` (default `false`) — reconcile a live mount's options when they differ. When false, a mounted filesystem with the correct source is kept regardless of option drift (backwards compatible; options still drive the initial mount and, with `edit_fstab`, the fstab entry).
- `remount_methods` (default `{ remount }`) — mechanisms tried in order, re-reading the live mount after each (the kernel returns success from a remount even when it silently ignores NFS-negotiated options). Defaults to the non-disruptive in-place `remount` only; the disruptive `unmount_mount` (which tears the filesystem down and back up) is opt-in, and is required to change options a live remount cannot — e.g. NFS-negotiated `vers=`/`rsize=`, or the server.
- `remount_timeout` — protect against a hung or unreachable server.

**Option comparison (live mount).** Only the options the promise names are enforced; any option it does *not* name is left unpoliced — its provenance is unknown (kernel-negotiated like `vers=`/`rsize=`/`proto=`/`sec=`, or added by a prior manual `mount -o`, indistinguishable). The promise is first resolved with util-linux **"last wins"** semantics — exactly as `mount -o` applies a list, a later option overrides an earlier conflicting one, so `defaults,ro` is a read-only mount and `ro,rw` is `rw`. Each surviving option must then hold on the live mount: its inverse absent, and it either present or a default-on flag. Inverse pairs (`noatime`/`relatime`, `hard`/`soft`, `ro`/`rw`, `sync`/`async`, generic `no<opt>`/`<opt>`) and `tcp`/`udp`↔`proto=` aliases are recognized; an option the promise specifies (including a negotiated one such as `rsize=8192`) must be present. A correctly-mounted filesystem converges instead of being reported changed every run, and an override (`ro` shadowing an earlier `rw`) is logged at verbose.

**The `defaults` pseudo-option.** `defaults` (= `rw,suid,dev,exec,auto,nouser,async`, per mount(8)) is never echoed by the kernel, so it's expanded to its checkable components `rw,suid,dev,exec,async` and subjected to the same last-wins resolution. It holds unless one of the violating negatives is present — `ro` (vs `rw`), `nosuid`/`nodev`/`noexec` (vs `suid`/`dev`/`exec`), or `sync` (vs `async`) — and a later explicit option overrides the matching component (so `defaults,ro` requires read-only, `defaults,nosuid` allows `nosuid`). `auto`/`nouser` are fstab / mount-permission concepts, not runtime state, so they aren't enforced. When reconciling a drifted `defaults` mount in place, the remount command likewise uses `rw,suid,dev,exec,async` (util-linux doesn't apply the options implied by a bare `defaults` on a remount), and mount's own last-wins applies any trailing override — so a mount that drifted to `ro`/`nosuid`/etc. is restored non-disruptively rather than only via `unmount_mount`.

### fstab maintenance for already-mounted filesystems (CFE-1539)

Previously a filesystem already mounted with the correct source was reported "mounted as promised" and fstab was never consulted, so a missing fstab entry was not restored and an options change was not written until the mount happened to be redone. `VerifyInFstab` now runs on the mounted-correctly path too (when `edit_fstab => "true"`), deliberately independent of the opt-in live `remount`: keeping fstab correct is the documented behavior of `mount_options`. fstab option comparison uses `strcmp` because option order matters.

### Surgical single-filesystem mount, not `mount -a` (CFE-1863)

A storage promise for a not-yet-mounted filesystem used to arm `mount -a` (`mount -va` on Linux), which mounts *every* unmounted fstab entry — unrelated devices and foreign filesystem types included — as a side effect of a single promise. The not-mounted path now mounts just the promised filesystem surgically (`VerifyMount`), then persists it to fstab. The `mount -a` mechanism (`MountAll`) is retained only for the explicit `mountfilesystems` agent-control attribute, which still means "mount everything in fstab".

### Target a specific mount on unmount (CFE-2350)

An unmount promise that named `mount_source`/`mount_server` was logged as "probably an error", and the server was never used to pick which mount to act on — so you couldn't unmount one specific mount (e.g. from a server being migrated away) without affecting others. The bogus warning is removed, and the server (host) is now part of the "mounted correctly" identity check, gated on `remount` **or** `unmount` so it only engages when the promise opts into disruptive mount management. An unmount promise that finds a *different* filesystem at the mount point leaves it — and its fstab entry — untouched, and `LiveMountConverged` treats the server as identity so a remount-in-place that can't change it escalates to `unmount_mount`.

### Correct dry-run / warn reporting (CFE-3366)

The mount, unmount and remount outcomes are based on the promise action (`MakingInternalChanges`) rather than a bare `!DONTDO`. A dry-run (`-n`) or `warn` promise now reports `WARN` without defining `promise_repaired`, so dependent promises no longer fire on a no-op run.

### Supporting mount-info fixes

- `GetFstabEntryOptions` returned the fstab type field instead of the options field (spurious rewrite every run).
- `ReplaceFstabEntry` leaked the previous entry string.
- Foreign-filesystem detection is restored by keeping the fstype separate from the kernel-resolved options (`options` vs `raw_opts`).
- The mount point directory is again created unconditionally before mounting.
- "device busy" interruptions are logged at `LOG_LEVEL_ERR` (the outcome is `INTERRUPTED`), and leaked options strings on the error paths are freed.

### Testing

Unit coverage in `tests/unit/nfs_test.c` (option subset matching, inverse/alias pairs, the `defaults` negative-violation check, contradiction detection, and the `defaults`→`rw,suid,dev,exec,async` remount expansion), runnable unprivileged via `make -C tests/unit check`. The behavioral NFS reconcile/escalation, fstab maintenance, and surgical single-filesystem mount — which need root and a real NFS server — are covered by the system-testing PR and were exercised against a loopback NFS export during development.

### Commits

1. Added opt-in remount reconciliation for storage mount options (CFE-90, CFE-1864, CFE-3366)
2. Added unit tests for mount option matching and contradiction detection
3. Maintained fstab entries for already-mounted filesystems (CFE-1539)
4. Mounted a single filesystem surgically instead of running mount -a (CFE-1863)
5. Allowed mount promises to target a specific server (CFE-2350)

### References

- `mount -a` skipping already-mounted filesystems, and `defaults` = `rw,suid,dev,exec,auto,nouser,async` — [mount(8)](https://man7.org/linux/man-pages/man8/mount.8.html)
- NFS options negotiated by client/server (reported in `/proc/mounts`) and NFS-specific options not modifiable on remount — [nfs(5)](https://man7.org/linux/man-pages/man5/nfs.5.html)

---

Resolves CFE-90 (and its duplicate CFE-1864), CFE-1539 (fstab maintenance), CFE-1863 (`mount -a` scope), CFE-2350 (target a specific mount on unmount), and CFE-3366 (dry-run/warn defining `promise_repaired`) for storage mount promises.

Ticket: https://northerntech.atlassian.net/browse/CFE-90

Together with: https://github.com/cfengine/system-testing/pull/693

### DIFF
--- a/cf-agent/nfs.c
+++ b/cf-agent/nfs.c
@@ -851,7 +851,7 @@ int VerifyNotInFstab(EvalContext *ctx, char *name, const Attributes *a, const Pr
 
                 if (strstr(line, "busy"))
                 {
-                    cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_INTERRUPTED, pp, a, "The device under '%s' cannot be removed from '%s'",
+                    cfPS(ctx, LOG_LEVEL_ERR, PROMISE_RESULT_INTERRUPTED, pp, a, "The device under '%s' cannot be removed from '%s'",
                          mountpt, VFSTAB[VSYSTEMHARDCLASS]);
                     *result = PromiseResultUpdate(*result, PROMISE_RESULT_INTERRUPTED);
                     free(line);
--- a/cf-agent/nfs.c
+++ b/cf-agent/nfs.c
@@ -753,7 +753,14 @@ int VerifyInFstab(EvalContext *ctx, char *name, const Attributes *a, const Promi
     }
     else
     {
-        /* CFE-90: Entry exists - check if options differ and update if needed */
+        /* CFE-90: Entry exists - rewrite it if the options differ.  The compare
+         * is exact (order-sensitive) on purpose: for duplicated/conflicting
+         * options the kernel uses the last one, so option order is significant
+         * in an fstab line and must not be normalized away.  Since
+         * GetFstabEntryOptions now reads the real options field (it previously
+         * returned the fstype), this converges - a differently-ordered entry is
+         * rewritten once to the promised form, then matches - rather than being
+         * rewritten on every run. */
         char *existing_opts = GetFstabEntryOptions(mountpt);
         if (existing_opts != NULL && !StringEqual(existing_opts, opts))
         {
--- a/cf-agent/nfs.c
+++ b/cf-agent/nfs.c
@@ -50,9 +50,16 @@ static Item *FSTABLIST = NULL; /* GLOBAL_X */
 
 static void GetHostAndSource(const char *buf, char *host, char *source);
 
-static void AugmentMountInfo(Seq *list, char *host, char *source, char *mounton, char *options);
+static bool ConflictingOptions(const char *a, const char *b);
+static bool OptionPresent(const char *opt, const Seq *actual);
+static bool ConflictingOptionPresent(const char *opt, const Seq *actual);
+static char *OptionStringExpandDefaults(const char *opts);
+
+static void AugmentMountInfo(Seq *list, char *host, char *source, char *mounton, char *fstype, char *options);
 static bool MatchFSInFstab(char *match);
 static void DeleteThisItem(Item **liststart, Item *entry);
+static char *GetFstabEntryOptions(const char *mountpt);
+static void ReplaceFstabEntry(Item *liststart, const char *mountpt, const char *new_entry);
 
 static const char *const VMOUNTCOMM[] =
 {
@@ -172,6 +179,208 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
         source[source_index++] = buf[index++];
     }
     source[source_index] = '\0';
+}
+
+/**
+ * True if 'a' and 'b' are mutually exclusive mount options that cannot both
+ * hold: ro/rw, hard/soft, sync/async, noatime/relatime, or a "no"-prefixed
+ * option and its bare form (e.g. noexec/exec).
+ */
+static bool ConflictingOptions(const char *a, const char *b)
+{
+    /* A "no"-prefixed option vs its bare form, in either direction. */
+    if (StringEqualN(a, "no", 2) && StringEqual(a + 2, b))
+    {
+        return true;
+    }
+    if (StringEqualN(b, "no", 2) && StringEqual(b + 2, a))
+    {
+        return true;
+    }
+
+    /* Inverse pairs that do not share the "no" prefix. */
+    static const char *const pairs[][2] = {
+        { "noatime", "relatime" },
+        { "hard", "soft" },
+        { "sync", "async" },
+        { "ro", "rw" },
+    };
+    for (size_t i = 0; i < sizeof(pairs) / sizeof(pairs[0]); i++)
+    {
+        if ((StringEqual(a, pairs[i][0]) && StringEqual(b, pairs[i][1])) ||
+            (StringEqual(a, pairs[i][1]) && StringEqual(b, pairs[i][0])))
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
+/**
+ * True if 'opt' is present in the live options, directly or via a
+ * tcp/udp<->proto= alias.
+ */
+static bool OptionPresent(const char *opt, const Seq *actual)
+{
+    const size_t len = SeqLength(actual);
+    for (size_t i = 0; i < len; i++)
+    {
+        const char *act = SeqAt(actual, i);
+        if (StringEqual(opt, act))
+        {
+            return true;
+        }
+        if ((StringEqual(opt, "tcp") && StringEqual(act, "proto=tcp")) ||
+            (StringEqual(opt, "udp") && StringEqual(act, "proto=udp")) ||
+            (StringEqual(opt, "proto=tcp") && StringEqual(act, "tcp")) ||
+            (StringEqual(opt, "proto=udp") && StringEqual(act, "udp")))
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
+/**
+ * True if an option contradicting 'opt' is present in the live mount.
+ */
+static bool ConflictingOptionPresent(const char *opt, const Seq *actual)
+{
+    const size_t len = SeqLength(actual);
+    for (size_t i = 0; i < len; i++)
+    {
+        if (ConflictingOptions(opt, SeqAt(actual, i)))
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
+/**
+ * Expand a bare "defaults" in an option string to its positives
+ * rw,suid,dev,exec,async: util-linux ignores the flags implied by a bare
+ * "defaults" on a `mount -o remount,...`, so the explicit form is needed both to
+ * restore a drifted mount in place and to compare against the live options.
+ * Newly allocated (caller frees), NULL if opts empty.
+ */
+static char *OptionStringExpandDefaults(const char *opts)
+{
+    if ((opts == NULL) || (opts[0] == '\0'))
+    {
+        return NULL;
+    }
+
+    Seq *in = SeqStringFromString(opts, ',');
+    const size_t len = SeqLength(in);
+    for (size_t i = 0; i < len; i++)
+    {
+        if (StringEqual(SeqAt(in, i), "defaults"))
+        {
+            SeqSet(in, i, xstrdup("rw,suid,dev,exec,async"));
+        }
+    }
+    char *result = StringJoin(in, ",");
+    SeqDestroy(in);
+    return result;
+}
+
+/**
+ * Extract the options from the parenthesized tail of a mount(8) output line
+ * ("host:/source on /mountpoint type nfs (rw,noatime)"), without the enclosing
+ * parentheses and with surrounding whitespace trimmed. The closing parenthesis
+ * must not be left on the last option or it will never compare equal to the
+ * promised one.
+ * Newly allocated (caller frees), NULL if the line has no "(...)" section.
+ */
+static char *MountOptionsFromLine(const char *line)
+{
+    const char *paren = strchr(line, '(');
+    if (paren == NULL)
+    {
+        return NULL;
+    }
+    const char *end = strchr(paren, ')');
+    if (end == NULL)
+    {
+        return NULL;
+    }
+
+    /* TrimWhitespace() can return an interior pointer, which is not free()-able. */
+    char *tail = xstrndup(paren + 1, (size_t) (end - paren - 1));
+    char *opts = xstrdup(TrimWhitespace(tail));
+    free(tail);
+    return opts;
+}
+
+/**
+ * Whether the live mount (actual_opts, from /proc/mounts) satisfies the promise.
+ * Only named options are enforced; unnamed ones are left alone. The promise is
+ * resolved "last wins" like `mount -o` (defaults,ro is read-only, ro,rw is rw),
+ * with "defaults" expanded to rw,suid,dev,exec,async (auto/nouser aren't runtime
+ * state). Each surviving option must then hold: its inverse absent, and it
+ * present or a default-on flag (suid/dev/exec/async only show when negated).
+ */
+bool OptionsSubsetMatches(const char *promised_opts, const char *actual_opts)
+{
+    if (promised_opts == NULL || promised_opts[0] == '\0')
+    {
+        return true;
+    }
+
+    Seq *actual = SeqStringFromString(actual_opts, ',');
+
+    /* Expand a bare "defaults" to its positives (rw,suid,dev,exec,async),
+     * preserving order for the last-wins pass, then split into options. */
+    char *expanded = OptionStringExpandDefaults(promised_opts);
+    Seq *eff = SeqStringFromString(expanded, ',');
+    free(expanded);
+
+    /* On-by-default flags the kernel does not echo (only their negatives show). */
+    static const char *const default_on[] = { "suid", "dev", "exec", "async" };
+
+    bool mismatch = false;
+    const size_t n_eff = SeqLength(eff);
+    for (size_t i = 0; (i < n_eff) && !mismatch; i++)
+    {
+        const char *x = SeqAt(eff, i);
+
+        /* Last wins: skip this option if a later one overrides it (its inverse)
+         * or repeats it. */
+        bool overridden = false;
+        for (size_t j = i + 1; j < n_eff; j++)
+        {
+            const char *y = SeqAt(eff, j);
+            if (ConflictingOptions(x, y))
+            {
+                Log(LOG_LEVEL_VERBOSE, "Mount option '%s' overridden by later '%s'", x, y);
+                overridden = true;
+                break;
+            }
+            if (StringEqual(x, y))
+            {
+                overridden = true;
+                break;
+            }
+        }
+        if (overridden)
+        {
+            continue;
+        }
+
+        const bool is_default_on = IsStringInArray(
+            x, default_on, sizeof(default_on) / sizeof(default_on[0]));
+
+        if (ConflictingOptionPresent(x, actual) || !(OptionPresent(x, actual) || is_default_on))
+        {
+            mismatch = true;
+        }
+    }
+
+    SeqDestroy(eff);
+    SeqDestroy(actual);
+
+    return !mismatch;
 }
 
 bool LoadMountInfo(Seq *list)
@@ -346,22 +555,26 @@ bool LoadMountInfo(Seq *list)
 
         Log(LOG_LEVEL_DEBUG, "LoadMountInfo: host '%s', source '%s', mounton '%s'", host, source, mounton);
 
+        char *mountopts = MountOptionsFromLine(vbuff);
+
         if (panfs)
         {
-            AugmentMountInfo(list, host, source, mounton, "panfs");
+            AugmentMountInfo(list, host, source, mounton, "panfs", mountopts);
         }
         else if (nfs)
         {
-            AugmentMountInfo(list, host, source, mounton, "nfs");
+            AugmentMountInfo(list, host, source, mounton, "nfs", mountopts);
         }
         else if (cifs)
         {
-            AugmentMountInfo(list, host, source, mounton, "cifs");
+            AugmentMountInfo(list, host, source, mounton, "cifs", mountopts);
         }
         else
         {
-            AugmentMountInfo(list, host, source, mounton, NULL);
+            AugmentMountInfo(list, host, source, mounton, NULL, mountopts);
         }
+
+        free(mountopts);
     }
 
     free(vbuff);
@@ -373,7 +586,7 @@ bool LoadMountInfo(Seq *list)
 
 /*******************************************************************/
 
-static void AugmentMountInfo(Seq *list, char *host, char *source, char *mounton, char *options)
+static void AugmentMountInfo(Seq *list, char *host, char *source, char *mounton, char *fstype, char *options)
 {
     Mount *entry = xcalloc(1, sizeof(Mount));
 
@@ -392,9 +605,19 @@ static void AugmentMountInfo(Seq *list, char *host, char *source, char *mounton,
         entry->mounton = xstrdup(mounton);
     }
 
-    if (options)
+    /* Store the fstype in options so IsForeignFileSystem can detect
+     * foreign filesystems via strstr(entry->options, "nfs"/"panfs"/"cifs"). */
+    if (fstype != NULL)
     {
-        entry->options = xstrdup(options);
+        entry->options = xstrdup(fstype);
+    }
+
+    /* Store the full kernel-resolved options in raw_opts.
+     * For unmounted filesystems (options == NULL or empty), raw_opts stays NULL
+     * and will be checked in FileSystemMountedCorrectly as "not mounted". */
+    if (options != NULL && options[0] != '\0')
+    {
+        entry->raw_opts = xstrdup(options);
     }
 
     SeqAppend(list, entry);
@@ -408,25 +631,11 @@ void DeleteMountInfo(Seq *list)
     {
         Mount *entry = SeqAt(list, i);
 
-        if (entry->host)
-        {
-            free(entry->host);
-        }
-
-        if (entry->source)
-        {
-            free(entry->source);
-        }
-
-        if (entry->mounton)
-        {
-            free(entry->mounton);
-        }
-
-        if (entry->options)
-        {
-            free(entry->options);
-        }
+        free(entry->host);
+        free(entry->source);
+        free(entry->mounton);
+        free(entry->options);
+        free(entry->raw_opts);
     }
 
     SeqClear(list);
@@ -534,12 +743,29 @@ int VerifyInFstab(EvalContext *ctx, char *name, const Attributes *a, const Promi
 
     if (!MatchFSInFstab(mountpt))
     {
+        /* CFE-90: Entry not in fstab - add it */
         AppendItem(&FSTABLIST, fstab, NULL);
         FSTAB_EDITS++;
         cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_CHANGE, pp, a, "Adding file system entry '%s' to '%s'", fstab,
              VFSTAB[VSYSTEMHARDCLASS]);
         *result = PromiseResultUpdate(*result, PROMISE_RESULT_CHANGE);
         changes += 1;
+    }
+    else
+    {
+        /* CFE-90: Entry exists - check if options differ and update if needed */
+        char *existing_opts = GetFstabEntryOptions(mountpt);
+        if (existing_opts != NULL && !StringEqual(existing_opts, opts))
+        {
+            /* Replace the entire fstab entry with the corrected options */
+            ReplaceFstabEntry(FSTABLIST, mountpt, fstab);
+            FSTAB_EDITS++;
+            cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_CHANGE, pp, a, "Updating file system entry for '%s' in '%s' (options: '%s' -> '%s')",
+                 mountpt, VFSTAB[VSYSTEMHARDCLASS], existing_opts, opts);
+            *result = PromiseResultUpdate(*result, PROMISE_RESULT_CHANGE);
+            changes += 1;
+        }
+        free(existing_opts);
     }
 
     free(opts);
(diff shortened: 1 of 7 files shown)


nickanderson:
@cf-bottom jenkins please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/build-and-deploy-docs-master/10/)

**Jenkins:** https://ci.cfengine.com/job/build-and-deploy-docs-master/10/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-build-and-deploy-docs-master-10/

**Documentation:** http://buildcache.cfengine.com/packages/build-documentation-pr/jenkins-build-and-deploy-docs-master-10/output/_site/

nickanderson:
@cf-bottom jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14143/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14143/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14143/

nickanderson:
@cf-bottom jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14145/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14145/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14145/

nickanderson:
@cf-bottom jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14148/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14148/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14148/

nickanderson:
@cf-bottom jenkins please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14151/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14151/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14151/

nickanderson:
@cf-bottom jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14155/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14155/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14155/

nickanderson:
@cf-bottom jenkins please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14170/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14170/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14170/

cf-agent/nfs.c
```diff
@@ -174,6 +183,234 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
...
+    free(arr);
+}
+
+static bool InverseOptions(const char *a, const char *b)
+/* True if 'a' and 'b' are mutually exclusive mount options that cannot both
+ * hold: ro/rw, hard/soft, sync/async, noatime/relatime, or a "no"-prefixed
+ * option and its bare form (e.g. noexec/exec). */
+{
+    /* A "no"-prefixed option vs its bare form, in either direction. */
+    if ((strncmp(a, "no", 2) == 0) && (strcmp(a + 2, b) == 0))
```
larsewi: Use `StringEqual()`/`StringNEqual()`

nickanderson:
@cf-bottom jenkins please.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14254/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14254/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14254/

cf-agent/nfs.c
```diff
@@ -174,6 +183,234 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
     source[source_index] = '\0';
 }
 
+/* Parse comma-separated options string into an array of individual option strings.
+ * Returns a newly allocated array of char* pointers, terminated by NULL.
+ * Caller must free the array itself (not the individual strings). */
+static char **ParseOptionsList(const char *opts)
```
larsewi: See if you can use
```C
Seq *SeqStringFromString(const char *str, char delimiter);
```
or
```C
char **String2StringArray(const char *str, char separator);
void FreeStringArray(char **strs);
```
instead.

cf-agent/nfs.c
```diff
@@ -174,6 +183,234 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
...
+static void FreeOptionsList(char **arr)
+{
+    if (arr == NULL)
+        return;
+    for (int i = 0; arr[i] != NULL; i++)
+        free(arr[i]);
+    free(arr);
+}
+
+static bool InverseOptions(const char *a, const char *b)
```
larsewi: Maybe `ConflictingOptions` would be a better name for this function?

cf-agent/nfs.c
```diff
@@ -174,6 +183,234 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
...
+    {
+        const char *tok = (strcmp(arr[i], "defaults") == 0)
+                          ? "rw,suid,dev,exec,async" : arr[i];
+        int w = snprintf(buf + n, sizeof(buf) - n, "%s%s", (i > 0) ? "," : "", tok);
+        if ((w < 0) || ((size_t) w >= sizeof(buf) - n))
+        {
+            break; /* option list longer than CF_BUFSIZE; stop appending */
+        }
+        n += (size_t) w;
+    }
```
larsewi: Consider using 
```C
char *StringJoin(const Seq *seq, const char *sep);
```
instead.

cf-agent/nfs.c
```diff
@@ -174,6 +183,234 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
...
+    Seq *eff = SeqNew(16, free);
+    for (int p = 0; promised[p] != NULL; p++)
+    {
+        if (strcmp(promised[p], "defaults") == 0)
+        {
+            static const char *const comps[] = { "rw", "suid", "dev", "exec", "async" };
+            for (size_t c = 0; c < sizeof(comps) / sizeof(comps[0]); c++)
+            {
+                SeqAppend(eff, xstrdup(comps[c]));
+            }
```
larsewi: This you could one line with something like
```C
Seq *eff = SeqStringFromString("rw,suid,dev,exec,async", ",");
```

cf-agent/nfs.c
```diff
@@ -174,6 +183,234 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
...
+
+    bool mismatch = false;
+    for (size_t i = 0; (i < SeqLength(eff)) && !mismatch; i++)
+    {
+        const char *x = SeqAt(eff, i);
+
+        /* Last wins: skip this option if a later one overrides it (its inverse)
+         * or repeats it. */
+        bool overridden = false;
+        for (size_t j = i + 1; j < SeqLength(eff); j++)
```
larsewi: Not a big deal, but calling `SeqLength()` once before the loop is more efficient then every iteration of the loop.

cf-agent/nfs.c
```diff
@@ -679,55 +949,65 @@ PromiseResult VerifyMount(EvalContext *ctx, char *name, const Attributes *a, con
...
+    }
 
-        if ((pfp = cf_popen(comm, "r", true)) == NULL)
-        {
-            Log(LOG_LEVEL_ERR, "Failed to open pipe from '%s'", CommandArg0(VMOUNTCOMM[VSYSTEMHARDCLASS]));
-            return PROMISE_RESULT_FAIL;
-        }
+    if (StringEqual(a->mount.mount_type, "panfs"))
+    {
+        snprintf(comm, CF_BUFSIZE, "%s -t panfs -o %s %s%s %s", CommandArg0(VMOUNTCOMM[VSYSTEMHARDCLASS]), opts, host, rmountpt, mountpt);
```
larsewi: It would be nice if you check the returned value of `snprintf`. Check that the buffer was not truncated. If truncation is impossible, then at least an `assert` would be nice to test that assumption in DEBUG builds.

cf-agent/nfs.c
```diff
@@ -679,55 +949,65 @@ PromiseResult VerifyMount(EvalContext *ctx, char *name, const Attributes *a, con
...
             cf_pclose(pfp);
             free(line);
-            return 1;
+            free(opts);
+            return PROMISE_RESULT_FAIL;
         }
-
-        free(line);
+    }
+    else if ((strstr(line, "busy")) || (strstr(line, "Busy")))
```
larsewi: Please use explicit comparisons as described in CONTRIBUTING.md. E.g. `strstr(line, "busy") != NULL`

cf-agent/verify_storage.c
```diff
@@ -470,18 +492,59 @@ static PromiseResult VerifyMountPromise(EvalContext *ctx, char *name, const Attr
                     dir);
             }
 
-            if (a->mount.editfstab)
+            /* CFE-90: distinguish "not mounted at all" from "mounted, but not as
+             * promised" (wrong source, or drifted options with remount enabled). */
+            for (size_t i = 0; i < SeqLength(GetGlobalMountedFSList()); i++)
```
larsewi: Consider calling `GetGlobalMountedFSList` only once. And consider moving `SeqLength` out of the loop condition to avoid it being called every iteration like mentioned before.

larsewi (changes requested):

cf-agent/verify_storage.c
nickanderson: Done — `GetGlobalMountedFSList()` is now called once into a local and `SeqLength` is hoisted out of the loop condition.

While here: this file uses raw `strcmp` throughout (no `StringEqual` anywhere). I switched the new `strcmp`s to `StringEqual` in `nfs.c` per your other comment, but left `verify_storage.c` alone to match its own style. Convert them here too while I'm in this PR, or leave as-is?

cf-agent/nfs.c
nickanderson: Done in ab145c1bf: switched to `StringEqual`/`StringEqualN` here (and the other new comparisons in this file).

cf-agent/nfs.c
nickanderson: Done in ab145c1bf: replaced the hand-rolled `ParseOptionsList`/`FreeOptionsList` with `SeqStringFromString`.

cf-agent/nfs.c
nickanderson: Done: renamed to `ConflictingOptions` in ab145c1bf.

cf-agent/nfs.c
nickanderson: Done in ab145c1bf: builds a `Seq` and joins it with `StringJoin` instead of the manual `snprintf` append.

cf-agent/nfs.c
nickanderson: Done in ab145c1bf: `SeqStringFromString("rw,suid,dev,exec,async", ',')` appended via `SeqAppendSeq` + `SeqSoftDestroy`.

cf-agent/nfs.c
nickanderson: Done in ab145c1bf: `SeqLength` is taken once into a local before the loop (here and in the option-matching helper).

cf-agent/nfs.c
nickanderson: Done in ab145c1bf: the mount/unmount/remount command builds now use `NDEBUG_UNUSED int ret = snprintf(...); assert(ret >= 0 && ret < CF_BUFSIZE);`, matching the fstab builders in this file.

cf-agent/nfs.c
nickanderson: Done in ab145c1bf: both busy checks now use `strstr(line, "busy") != NULL`.

cf-agent/nfs.c
```diff
@@ -174,6 +181,184 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
     source[source_index] = '\0';
 }
 
+static bool ConflictingOptions(const char *a, const char *b)
+/* True if 'a' and 'b' are mutually exclusive mount options that cannot both
+ * hold: ro/rw, hard/soft, sync/async, noatime/relatime, or a "no"-prefixed
+ * option and its bare form (e.g. noexec/exec). */
```
larsewi: Put docs above the function signature and use the doxygen (double star) syntax. This way editors will show this text when you hover over the symbol name.
```suggestion
/**
 * True if 'a' and 'b' are mutually exclusive mount options that cannot both
 * hold: ro/rw, hard/soft, sync/async, noatime/relatime, or a "no"-prefixed
 * option and its bare form (e.g. noexec/exec).
 */
static bool ConflictingOptions(const char *a, const char *b)
```

cf-agent/nfs.c
```diff
@@ -174,6 +181,184 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
...
+            (StringEqual(opt, "proto=tcp") && StringEqual(act, "tcp")) ||
+            (StringEqual(opt, "proto=udp") && StringEqual(act, "udp")))
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
+static bool InversePresent(const char *opt, const Seq *actual)
```
larsewi: This should probably also be renamed, since the other one was?

cf-agent/nfs.c
```diff
@@ -956,6 +1206,278 @@ static void DeleteThisItem(Item **liststart, Item *entry)
...
+            {
+                ret = snprintf(comm, CF_BUFSIZE, "%s -o remount,%s %s",
+                               CommandArg0(VMOUNTCOMM[VSYSTEMHARDCLASS]), ropts, name);
+            }
+            else
+            {
+                ret = snprintf(comm, CF_BUFSIZE, "%s -o remount %s",
+                               CommandArg0(VMOUNTCOMM[VSYSTEMHARDCLASS]), name);
+            }
+            assert(ret >= 0 && ret < CF_BUFSIZE);
```
larsewi: I would move the `ret` declaration and put the `assert` inside the `if` branches. This way the `ret` variable goes out of scope afterwards. Furthermore, you will see which branch triggered the assertion. And you don't save any lines of code by having it outside.

cf-agent/verify_storage.c
```diff
@@ -454,13 +481,30 @@ static PromiseResult VerifyMountPromise(EvalContext *ctx, char *name, const Attr
...
+        /* Whether something is mounted at the promiser at all (vs. nothing
+         * mounted there).  Computed for both mount and unmount: the unmount
+         * path uses it to tell "a different filesystem is mounted here" from
+         * "nothing is mounted here". */
+        bool already_mounted = false;
+        const size_t n_mounted = SeqLength(mounted_fs);
+        for (size_t i = 0; i < n_mounted; i++)
+        {
+            Mount *mp = SeqAt(mounted_fs, i);
+            if (mp != NULL && mp->mounton != NULL && strcmp(name, mp->mounton) == 0)
```
larsewi:
```suggestion
            if (mp != NULL && mp->mounton != NULL && StringEqual(name, mp->mounton))
```

larsewi (commented):
Thanks! Some minor nit picks. It would also be nice if you could amend the fixes from the last commit to the relevant commits that introduced them. It would make it easier for me to review this PR commit for commit. It's already very large / difficult to review / hard to keep all the context in my head.

cf-agent/nfs.c
nickanderson: Done — moved the docs above the signature and switched to `/**`. Applied the same to the other new helpers in this file (`OptionPresent`, `RemountOptionString`, `OptionsSubsetMatches`, `LiveMountConverged`, `ReconcileMountOptions`) for consistency.

cf-agent/nfs.c
nickanderson: Done — renamed to `ConflictingOptionPresent`.

cf-agent/nfs.c
nickanderson: Done — `ret` is now declared and `assert`ed inside each branch.

cf-agent/verify_storage.c
nickanderson: Done. Per my earlier question, I used `StringEqual` on the PR-touched lines here and left the other (untouched) `strcmp`s in this file alone.

nickanderson:
Thanks! Nits addressed, and the trailing fixup commit is gone — its changes, the new nits, and the CodeQL NULL-guards are all folded into the commits that introduced the code, so the PR is now five clean commits that read commit-by-commit. Each commit builds and `nfs_test` passes.

cf-agent/verify_storage.c
larsewi: Yepp, I would keep untouched code alone for the sake of the git history.

cf-agent/nfs.c
```diff
@@ -174,6 +181,194 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
...
+}
+
+/**
+ * True if 'opt' is present in the live options, directly or via a
+ * tcp/udp<->proto= alias.
+ */
+static bool OptionPresent(const char *opt, const Seq *actual)
+{
+    const size_t len = SeqLength(actual);
+    for (size_t a = 0; a < len; a++)
```
larsewi: The convention is to use `i` for the index variable, then `j`, `k`, etc. for nested loops.

cf-agent/nfs.c
```diff
@@ -174,6 +181,194 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
...
+static char *RemountOptionString(const char *opts)
+{
+    if ((opts == NULL) || (opts[0] == '\0'))
+    {
+        return NULL;
+    }
+
+    Seq *in = SeqStringFromString(opts, ',');
+    const size_t len = SeqLength(in);
+    Seq *out = SeqNew(len + 4, free);
```
larsewi: Adding four to the initial capacity does not help. I guess the thought was that you expect at most one occurrences of `defaults`. We add five options below (`rw,suid,dev,exec,async`) while removing one (`defaults`). Hence, we need four more slots. However, the five options below replaces the `defaults` option as a single element in the sequence. Thus, the final capacity is the exact length of the original sequence.

Furthermore, instead of creating two sequences, you should use
```C
void SeqSet(Seq *set, size_t index, void *item);
```
on the original one to mutate it in place.

cf-agent/nfs.c
```diff
@@ -174,6 +181,194 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
...
+        {
+            Seq *comps = SeqStringFromString("rw,suid,dev,exec,async", ',');
+            SeqAppendSeq(eff, comps);
+            SeqSoftDestroy(comps); /* eff now owns the expanded components */
+        }
+        else
+        {
+            SeqAppend(eff, xstrdup(opt));
+        }
+    }
```
larsewi: Can't you use the
```C
static char *RemountOptionString(const char *opts);
```
function above? It appears to be doing the exact same thing.

It should probably be renamed to `OptionStringExpandDefaults()` if you decide re-use it.

cf-agent/nfs.c
```diff
@@ -534,13 +745,30 @@ int VerifyInFstab(EvalContext *ctx, char *name, const Attributes *a, const Promi
...
         cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_CHANGE, pp, a, "Adding file system entry '%s' to '%s'", fstab,
              VFSTAB[VSYSTEMHARDCLASS]);
         *result = PromiseResultUpdate(*result, PROMISE_RESULT_CHANGE);
         changes += 1;
     }
+    else
+    {
+        /* CFE-90: Entry exists - check if options differ and update if needed */
+        char *existing_opts = GetFstabEntryOptions(mountpt);
+        if (existing_opts != NULL && strcmp(existing_opts, opts) != 0)
```
larsewi: Use `StringEqual()` here

cf-agent/nfs.c
```diff
@@ -618,7 +846,7 @@ int VerifyNotInFstab(EvalContext *ctx, char *name, const Attributes *a, const Pr
 
                 if (strstr(line, "busy"))
                 {
-                    cfPS(ctx, LOG_LEVEL_INFO, PROMISE_RESULT_INTERRUPTED, pp, a, "The device under '%s' cannot be removed from '%s'",
+                    cfPS(ctx, LOG_LEVEL_ERR, PROMISE_RESULT_INTERRUPTED, pp, a, "The device under '%s' cannot be removed from '%s'",
```
larsewi, line 854: This seems like it belongs in a separate commit?

cf-agent/nfs.c
```diff
@@ -679,55 +907,68 @@ PromiseResult VerifyMount(EvalContext *ctx, char *name, const Attributes *a, con
...
             cf_pclose(pfp);
             free(line);
-            return 1;
+            free(opts);
+            return PROMISE_RESULT_FAIL;
         }
-
-        free(line);
+    }
+    else if ((strstr(line, "busy") != NULL) || (strstr(line, "Busy") != NULL))
```
larsewi: Prefer
```C
bool StringContains(const char *const haystack, const char *const needle);
```
instead of `strstr` for the sake of readability.

cf-agent/verify_storage.c
```diff
@@ -470,18 +500,59 @@ static PromiseResult VerifyMountPromise(EvalContext *ctx, char *name, const Attr
                     dir);
             }
 
-            if (a->mount.editfstab)
+            /* CFE-90: distinguish "not mounted at all" from "mounted, but not as
+             * promised" (wrong source, or drifted options with remount enabled). */
+            for (size_t i = 0; i < SeqLength(GetGlobalMountedFSList()); i++)
+            {
+                Mount *mp = SeqAt(GetGlobalMountedFSList(), i);
+                if (mp != NULL && mp->mounton != NULL && strcmp(name, mp->mounton) == 0)
```
larsewi: Prefer `StringEqual()`

cf-agent/verify_storage.c
larsewi: I don't see that this was fixed. Maybe it ended up in the wrong commit?

tests/unit/nfs_test.c
```diff
@@ -33,11 +33,84 @@ static void test_MatchFSInFstab(void)
...
+
+    /* "defaults" expands to its checkable positives so an in-place remount
+     * restores a mount that drifted to ro/nosuid/etc (util-linux does not honor
+     * the options implied by "defaults" on a remount). */
+    s = RemountOptionString("defaults");
+    assert_true(strcmp(s, "rw,suid,dev,exec,async") == 0); free(s);
+    s = RemountOptionString("defaults,noatime");
+    assert_true(strcmp(s, "rw,suid,dev,exec,async,noatime") == 0); free(s);
+    /* Everything else is passed through unchanged and order-preserved. */
+    s = RemountOptionString("rw,noatime");       assert_true(strcmp(s, "rw,noatime") == 0); free(s);
```
larsewi: `StringEqual()`

tests/unit/nfs_test.c
```diff
@@ -33,11 +33,84 @@ static void test_MatchFSInFstab(void)
...
+}
+
+static void test_RemountOptionString(void)
+{
+    char *s;
+
+    /* "defaults" expands to its checkable positives so an in-place remount
+     * restores a mount that drifted to ro/nosuid/etc (util-linux does not honor
+     * the options implied by "defaults" on a remount). */
+    s = RemountOptionString("defaults");
```
larsewi: Remember to `free` memory

cf-agent/verify_storage.c
```diff
@@ -372,6 +365,19 @@ static bool FileSystemMountedCorrectly(Seq *list, char *name, const Attributes *
                 return false;
             }
 
+            /* CFE-2350: the server is part of mount identity, but acting on a
+             * mismatch is disruptive (umount+mount), so only check it when the
+             * promise opts in via remount or unmount. */
+            if ((a->mount.remount || a->mount.unmount)
+                && (a->mount.mount_server != NULL)
+                && ((mp->host == NULL) || (strcmp(mp->host, a->mount.mount_server) != 0)))
```
larsewi: `StringEqual()`

larsewi (changes requested):

cf-agent/nfs.c
nickanderson: Done — counters renamed to `i`/`j`.

cf-agent/nfs.c
nickanderson: Done — `SeqNew(len)` and mutate in place with `SeqSet`; the second Seq is gone.

cf-agent/nfs.c
nickanderson: Done — extracted `OptionStringExpandDefaults()` (renamed from `RemountOptionString`) and reused it here.

cf-agent/nfs.c
nickanderson: Done — `StringContains()` here and in the `VerifyUnmount` busy check.

cf-agent/nfs.c
nickanderson: This is `!StringEqual(existing_opts, opts)`, folded into the commit that introduces the line.

cf-agent/nfs.c
nickanderson: Agreed — split into its own commit.

cf-agent/verify_storage.c
nickanderson: Already `StringEqual`; folded into the commit that introduces the line.

cf-agent/verify_storage.c
nickanderson: Right — it was fixed at HEAD but sat in a later commit; folded back into the commit that introduces it. (This thread's anchor SHA was rebased away in round 2, hence the stale view.)

tests/unit/nfs_test.c
nickanderson: Done — `StringEqual` in the asserts.

tests/unit/nfs_test.c
nickanderson: The frees were there, tucked after the `assert_true(...)` on the same line — moved each `free()` to its own line.

cf-agent/verify_storage.c
nickanderson: Done — `!StringEqual(mp->host, a->mount.mount_server)`.

cf-agent/verify_storage.c
nickanderson: Agreed — left the pre-existing `strcmp`s alone; only the PR-touched lines use `StringEqual`.

nickanderson:
Nits addressed, each folded into the commit that introduces the line. INFO→ERR is split out, so this is six commits.

Also applied the same conventions where you hadn't reached: `strchr`, explicit `!= NULL` (incl. two log ternaries), Doxygen `/**` on the two fstab helpers, `SeqLength` hoisted out of two more loop conditions, `SeqNew` sized via `RlistLen`, and `nfs.h` back to plain declarations.

cf-agent/nfs.c
```diff
@@ -174,6 +181,209 @@ static void GetHostAndSource(const char *buf, char *host, char *source)
...
+    {
+        return NULL;
+    }
+
+    char *opts = xstrndup(paren + 1, (size_t) (end - paren - 1));
+    size_t len = strlen(opts);
+    while ((len > 0) && ((opts[len - 1] == ' ') || (opts[len - 1] == '\t')))
+    {
+        opts[--len] = '\0';
+    }
```
larsewi: Maybe you could use
```C
char *TrimWhitespace(char *s);
```
here?

larsewi (approved):

cf-agent/nfs.c
nickanderson: OK, done.

cf-agent/verify_storage.c
nickanderson: It's in the introducing commit, not a later one — `348722a06`, the first commit of the PR: `GetGlobalMountedFSList()` into a local, `const size_t n_mounted = SeqLength(mounted_fs)` hoisted out of the condition, and `StringEqual`. At head that's `verify_storage.c:500-510` (`841d48738` hoisted the block out of the `!unmount` branch so the unmount path can use it too).

The stale view is my rebase's fault — this thread is anchored to a SHA round 2 rewrote away, so GitHub still renders the old `SeqLength(GetGlobalMountedFSList())` line. That line isn't in any commit on the PR.

If what you spotted was `verify_storage.c:452`, that one is real but pre-existing (`master:421`, in `IsForeignFileSystem`) and untouched here. Happy to take it separately.

nickanderson:
@cf-bottom jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14344/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14344/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14344/

nickanderson:
@cf-bottom jenkins please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14367/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14367/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14367/

craigcomstock:
retry, no acceptance tests should make things a bit faster.

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14387/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14387/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14387/

PR #6235, opened 2026-07-14
Fix DBPrivClean() for LMDB 1.0.0 mdb_drop() semantics

aleksandrychev:
LMDB 1.0.0 changed mdb_drop(): emptying the main DBI marks the transaction MDB_TXN_DROPPED, so any subsequent read/write on that same transaction fails with MDB_BAD_TXN. CFEngine reuses one write transaction per thread and callers typically repopulate the DB right after cleaning it, which triggered the failure.


cf-bottom:
Thanks for submitting a PR! Maybe @craigcomstock can review this?

craigcomstock (approved):

PR #6238, opened 2026-07-17
ENT-14297: Improve error message when cf-agent cannot write to an immutable file (3.27)

nickanderson:
Backport of ENT-14297 to `3.27.x` (master PR: #6236).

Improves the error message when cf-agent cannot write to a file whose immutable
bit is set and is not being overridden. The actual system error is always
reported, and `(the immutable bit is set)` is appended only when confirmed via
`FSAttrsGetImmutableFlag()` — never guessed from errno (an immutable file
typically fails with `EPERM`, but that is not exclusive to immutability), so
permission/MAC/read-only/full-filesystem failures are never mislabeled.

Detection lives at the two points where the write actually fails:
- `WriteContentFromString()` — the in-place open for the `content` attribute.
- `OverrideImmutableRename()` — the temporary-file rename that is the common
  landing point for `edit_line`, `edit_xml`, `copy_from`, and templating.

It is skipped when overriding the immutable bit (the agent then writes to a
temporary copy renamed into place after clearing the bit).

Cherry-picked cleanly from master (`verify_files.c` auto-merges; the shared
`override_fsattrs.c` is identical between 3.27.x and master).

**Tests** (`tests/acceptance/10_files/unsafe/14..20_immutable_*`) cover content,
edit_line, copy_from, mustache, cfengine-template, edit_xml, and inline_mustache.
They are `unsafe` tests (they set the immutable bit), so they run in the chroot
acceptance job (`UNSAFE_TESTS=1`), not the GitHub Actions
`--tests=common,errorlog,errorexit` run.

Ticket: ENT-14297

### DIFF
--- a/cf-agent/verify_files.c
+++ b/cf-agent/verify_files.c
@@ -831,7 +831,29 @@ static PromiseResult WriteContentFromString(EvalContext *ctx, const char *path, 
         FILE *f = safe_fopen(override_path, "w");
         if (f == NULL)
         {
-            RecordFailure(ctx, pp, attr, "Cannot open file '%s' for writing", path);
+            const int saved_errno = errno;
+
+            /* Always report the actual system error. When not overriding the
+             * immutable bit we write in place, so also flag it if the file is
+             * immutable - confirmed via the flag, not guessed from errno (which
+             * is EPERM but not exclusive to immutability), so other causes are
+             * not mislabeled. When overriding we write to a temporary copy, so
+             * the immutable bit is not the cause. */
+            bool is_immutable = false;
+            if (!override_immutable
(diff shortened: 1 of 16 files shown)


nickanderson:
@cf-bottom jenkins please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14225/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14225/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14225/

larsewi (approved):

PR #6239, opened 2026-07-20
ENT-6511: Increase cf-monitord measurement slots from 100 to 300

nickanderson:
Raises `CF_OBSERVABLES` from 100 to 300. Previously out of the box ~80 of the 100 slots are consumed by default measurements, leaving only ~19 for custom `measurements:` promises.

Notably, this change does increase the disk footprint required by an agent by ~60MB (~.27MB per slot). That storage is required even if the slots are not in active use (Ref: ENT-14329)

Ticket: ENT-6511


nickanderson:
@cf-bottom jenkins please.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14244/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14244/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14244/

cf-check/dump.c
```diff
@@ -162,18 +162,23 @@ static void print_struct_lock_data(
 static void print_struct_averages(
     const MDB_val value, const bool strip_strings, const char *tskey_filename)
 {
-    assert(sizeof(Averages) == value.mv_size);
-    if (sizeof(Averages) != value.mv_size)
+    // A record written before an increase of CF_OBSERVABLES is shorter than the
+    // current struct (e.g. after a package upgrade, until cf-monitord rewrites
+    // it). Accept any size up to the full struct and copy into a zeroed struct
+    // so the un-stored trailing slots read back as zero.
```
larsewi: Please try to do a DB migration like `libpromises/dbm_migration_lastseen.c` instead if its possible.

larsewi (commented):
Otherwise LGTM

cf-check/dump.c
nickanderson: OK ... this makes me wonder about the other DB backends like tcdb and qmdb and bdb, those aren't covered at all. I guess that's ok we don't vendor any packages with those

nickanderson:
@cf-bottom jenkins please.

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14253/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14253/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14253/

cf-check/dump.c
```diff
@@ -165,7 +165,9 @@ static void print_struct_averages(
     assert(sizeof(Averages) == value.mv_size);
     if (sizeof(Averages) != value.mv_size)
     {
-        // Fall back to simple printing in release builds:
+        // Fall back to simple printing in release builds. Records written
+        // before CF_OBSERVABLES was raised are shorter, but a running agent
+        // migrates them (see dbm_migration_observations.c) before they are read.
```
larsewi: Feels like this change does not belong here anymore

cf-check/dump.c
```diff
@@ -288,7 +290,15 @@ static void print_struct_or_string(
 {
     if (structs)
     {
-        if (StringContains(file, "cf_lastseen.lmdb")
+        if ((StringContains(file, "cf_observations.lmdb")
+             || StringContains(file, "history.lmdb"))
+            && StringEqual(key.mv_data, "version"))
+        {
+            // Migration version scalar (dbm_migration_observations.c), a short
+            // string rather than an Averages struct.
```
larsewi: I'm not sure I understand this comment

libpromises/dbm_migration_observations.c
```diff
@@ -0,0 +1,113 @@
+/*
+  Copyright 2024 Northern.tech AS
```
larsewi:
```suggestion
  Copyright 2026 Northern.tech AS
```

cf-check/observables.c
```diff
@@ -67,10 +67,23 @@ char **GetObservableNames(const char *ts_key_path)
...
+                /* The ts_key has fewer entries than CF_OBSERVABLES -- e.g. it
+                 * was written by an older agent with a smaller CF_OBSERVABLES.
+                 * Fill the remaining names so the returned array keeps its
+                 * documented non-NULL guarantee (callers index all
+                 * CF_OBSERVABLES entries). */
+                for (int j = i; j < CF_OBSERVABLES; ++j)
+                {
+                    snprintf(buf, CF_MAXVARSIZE, "spare[%d]", j);
+                    temp[j] = xstrdup(buf);
+                }
```
larsewi, line 86: Is this still needed post migration fix ?

tests/unit/observations_migration_test.c
```diff
@@ -0,0 +1,238 @@
...
+/*
+ * Provides an observations DB in a pre-migration (version 0) state: OpenDB()
+ * runs the migration and writes a "version" marker, so wipe every entry to get
+ * back to how an old, never-migrated database looks on disk.
+ */
+static DBHandle *setup_unversioned(void)
+{
+    char cmd[CF_BUFSIZE];
+    xsnprintf(cmd, CF_BUFSIZE, "rm -rf '%s'/*", GetStateDir());
+    system(cmd);
```
larsewi: I know, I know... Recursively deleting a directory is not as easy as one would think in C. However, there probably is a function you can use in libntech or libpromises so that you don't have to run a command.

tests/unit/observations_migration_test.c
```diff
@@ -0,0 +1,238 @@
...
+
+    DBHandle *db;
+    assert_int_equal(OpenDB(&db, dbid_observations), true);
```
larsewi: Use `assert_true` instead of treating `bool` as an `int`, even though it really is. It will also give more descriptive error messages.
```suggestion
    assert_true(OpenDB(&db, dbid_observations));
```

tests/unit/observations_migration_test.c
```diff
@@ -0,0 +1,238 @@
...
+    PRINT_TEST_BANNER();
+    int ret = run_tests(tests);
+
+    tests_teardown();
+    return ret;
+}
+
+/* STUBS */
+
+void FatalError(ARG_UNUSED char *s, ...)
```
larsewi, line 236: Please add a comment explaining why this STUB is here

tests/unit/observations_migration_test.c
larsewi: Although I get the argument that the unit tests should be as lightweight as possible. Either way is fine by me.

libpromises/dbm_migration_observations.c
nickanderson: Done.

cf-check/dump.c
nickanderson: Done — added `dbm_migration_observations.c` following the `dbm_migration_lastseen.c` pattern (shared by both `cf_observations.lmdb` and `history.lmdb`), and restored the size assertion in `print_struct_averages` now that records are migrated. Re: the other backends — `dbm_migration.c` is LMDB-only (`#ifndef LMDB` → no-op), and we only vendor LMDB, so nothing to do for tcdb/qmdb/bdb.

cf-check/dump.c
nickanderson: Reverted to the original `// Fall back to simple printing in release builds:`.

cf-check/dump.c
nickanderson: Reworded. The migration stores a `version` bookkeeping key in these DBs; its value is a short version string, not an `Averages` struct, so it is printed as a string rather than falling through to `print_struct_averages()` (which would trip the struct-size assertion).

cf-check/observables.c
nickanderson: Yes — this is about the `ts_key` *text* file, which the DB migration does not touch. `GetObservableNames()` is documented to return a `CF_OBSERVABLES`-sized array of non-NULL strings. A `ts_key` written by an older agent has only 100 lines, so after the bump `fgets` hits EOF at slot 100 and, without this fill, `temp[100..299]` stay uninitialized (`xmalloc`, not zeroed); the sole caller (`dump.c:179`) then indexes and `free()`s all 300 → uninitialized read + invalid free. So it is an independent correctness fix.

tests/unit/observations_migration_test.c
nickanderson: Switched to `DeleteDirectoryTree()` from libntech (empties the tree, keeps the top dir), plus a trailing `rmdir()` in teardown.

tests/unit/observations_migration_test.c
nickanderson: Done — converted the bool-returning assertions (`OpenDB`/`WriteDB`/`ReadDB`/`HasKeyDB`/`NewDBCursor`/`DeleteDBCursor`) to `assert_true`.

tests/unit/observations_migration_test.c
nickanderson: Added a comment: `OpenDB()` pulls in libpromises DB code that references `FatalError()`, whose real implementation lives in the daemons, not the test binary — so the stub is there to let the test link, and reaching it is itself a failure.

cf-check/observables.c
nickanderson: And, just FYI, I have a follow-up PR waiting to changes the storage to only require what is used

cf-check/dump.c
```diff
@@ -162,23 +162,18 @@ static void print_struct_lock_data(
...
     const MDB_val value, const bool strip_strings, const char *tskey_filename)
 {
-    // A record written before an increase of CF_OBSERVABLES is shorter than the
-    // current struct (e.g. after a package upgrade, until cf-monitord rewrites
-    // it). Accept any size up to the full struct and copy into a zeroed struct
-    // so the un-stored trailing slots read back as zero.
-    assert(value.mv_size <= sizeof(Averages));
-    if (value.mv_size > sizeof(Averages))
+    assert(sizeof(Averages) == value.mv_size);
+    if (sizeof(Averages) != value.mv_size)
```
larsewi: Looks like this got messed up. It should have been reverted in the previous commit

nickanderson: Good catch — re-split the two commits so the bump no longer touches `cf-check/dump.c` at all (it left `print_struct_averages` strict), and the migration commit carries only the new `"version"`-key handling. No more introduce-then-revert across the two commits; the net diff is unchanged. Force-pushed.

cf-check/dump.c
```diff
@@ -288,7 +288,18 @@ static void print_struct_or_string(
 {
     if (structs)
     {
-        if (StringContains(file, "cf_lastseen.lmdb")
+        if ((StringContains(file, "cf_observations.lmdb")
```
larsewi, line 291: Pre-existing. But, should probably be `StringEndsWith()`? Could be a follow-up commit.

larsewi (approved):

larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14272/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14272/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14272/

cf-check/dump.c
nickanderson: in this PR or completely separately?

larsewi: Up to you. I'm OK with leaving it as is as well

nickanderson: I think I will leave it for now. Seems to have been intentionally changed from StringsEndsWith to StringContains https://github.com/cfengine/core/commit/3e98366c3

nickanderson:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14345/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14345/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14345/

PR #6241, opened 2026-07-21
CFE-3725: fix double change detection with hash=>"best"

nickanderson:
`hash => "best"` in a `changes` body now uses SHA-512.

In community builds, `best` hashed files with both MD5 and SHA1. A bug in how the two were checked meant that after a file changed, only one of the two stored hashes was updated. The stale one then reported the same change again on the next run, so a single change showed up as "changed" on two runs in a row.

Instead of fixing the two-hash bookkeeping, `best` now uses one strong hash, SHA-512. That is what CFEngine Enterprise has always used for `best`.

On community, the first agent run after upgrading re-records `best`-tracked files under SHA-512 and file changes would be realized.

Documentation stats that `best` cross correlates the two best options available. That doesn't seem to have ever been true for Enterprise. Since community could explicitly already select sha512 I thought it made the most sense to align on Enterpise behavior rather than add additional hashing to enterprise `best`.

### DIFF
--- a/cf-agent/verify_files_utils.c
+++ b/cf-agent/verify_files_utils.c
@@ -3690,7 +3690,6 @@ static PromiseResult VerifyFileIntegrity(EvalContext *ctx, const char *file, con
 {
     assert(attr != NULL);
     unsigned char digest1[EVP_MAX_MD_SIZE + 1];
-    unsigned char digest2[EVP_MAX_MD_SIZE + 1];
 
     if ((attr->change.report_changes != FILE_CHANGE_REPORT_CONTENT_CHANGE) && (attr->change.report_changes != FILE_CHANGE_REPORT_ALL))
     {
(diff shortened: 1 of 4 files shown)


nickanderson:
@cf-bottom jenkins please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14257/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14257/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14257/

nickanderson:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14258/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14258/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14258/

larsewi (approved):
Makes sense to me 🚀

nickanderson:
@olehermanse do you agree with this. I was on the fence between making Enterprise do 2 hashes vs changing best to pick the singular best.

olehermanse (approved):

PR #6244, opened 2026-07-28
Added enterprise client report dump collection to cf-support

nickanderson:
When `/var/cfengine/enable_report_dumps` is present cf-serverd will log
information provided to cf-hub in `/var/cfengine/diagnostics/report_dumps`. This
change makes sure those are collected for review by Support.

Ticket: ENT-5083
Changelog: None


craigcomstock (approved):
I would mention cf-support in the commit. I changed the title of the PR to mention it.

PR #6252, opened 2026-07-29
Fixed m4/cf3_gcc_flags.m4 to actually test compiler flags

craigcomstock:
This was broken in CFE-3629 in commit fc40dd612d32139bee32e657b10fea82159999eb

Ticket: ENT-14382
Changelog: none

together
https://github.com/cfengine/core/pull/6252
https://github.com/cfengine/enterprise/pull/988
https://github.com/cfengine/nova/pull/2670


larsewi (changes requested):
you `save_CFLAGS="$CFLAGS"` conditionally (inside the if block), and restore it unconditionally (outside the if block). Meaning you will remove all the flags unless `test x"$GCC" = "xyes" && test x"$HP_UX_AC" != x"yes"`.

larsewi (approved):

PR #6255, opened 2026-07-30
Adjusted ci/install.sh to use cf-key from different locations other than /var/cfengine

craigcomstock:
e.g. on Termux platform cf-key is in PATH after install but not in /var/cfengine

Ticket: none
Changelog: none

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -18,4 +18,13 @@ if [ ! -n "$TERMUX_VERSION" ]; then
 fi
 
 $GAINROOT make install
-$GAINROOT /var/cfengine/bin/cf-key # to generate the hostkey
+CFKEY=$(command -v cf-key)
+if ! command -v cf-key; then
+  CFKEY=/var/cfengine/bin/cf-key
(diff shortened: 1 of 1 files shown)


cf-bottom:

ci/install.sh
```diff
@@ -18,4 +18,13 @@ if [ ! -n "$TERMUX_VERSION" ]; then
...
-$GAINROOT /var/cfengine/bin/cf-key # to generate the hostkey
+CFKEY=$(command -v cf-key)
+if ! command -v cf-key; then
+  CFKEY=/var/cfengine/bin/cf-key
+  if [ ! -f /var/cfengine/bin/cf-key ]; then
+    echo "Can't find cf-key in PATH or at /var/cfengine/bin/cf-key. You will need to generate a hostkey yourself."
+    exit 0
+  fi
+fi
+echo Generating hostkey with $CFKEY
```
larsewi, line 29: shellcheck will complain without the quotes

larsewi (approved):

PR #6281, opened 2026-08-11
Added support on SuSE for user promises without lib PAM

craigcomstock:
On older SuSE distributions the PAM modules are linked to an older version of openssl/libcrypt than we vendor and so when cf-agent tries to use PAM the modules fail to load and operations fail.

Ticket: ENT-14406
Changelog: title


cf-bottom:
Thank you for submitting a PR! Maybe @larsewi can review this?

PR #6265, opened 2026-08-04
CFE-3988: Fixed insert_lines ignoring include_end_delimiter when locating the region

nickanderson:
The insertion path treats the selected region as the half-open range `[begin_ptr, end_ptr)`, so the end delimiter was never inside it. Inserting after the last line of a region built with `include_end_delimiter => "true"` landed *above* the delimiter instead of below it. Insertion now moves the bound the way `DeletePromisedLinesMatching()` already does.

Nothing could have depended on the old behavior: `include_end_delimiter` had no effect at all on `insert_lines`, and all ~90 existing tests using it are under `07_delete_lines`.

### Tests

- `31_tickets/CFE-3988/1` — the ticket's reproducer, no longer `test_soft_fail`
- `31_tickets/CFE-3988/2` — new, pins the unchanged `include_end_delimiter => "false"` default

Verified in a local starter_pack VM: acceptance 1648 passed / 0 regressions / soft failures 44 → 43, unsafe `10_files` 24/24, unit 70/70. With the fix reverted, case 1 fails and case 2 still passes.

Ticket: https://northerntech.atlassian.net/browse/CFE-3988

Backports to 3.27.x and 3.24.x to follow, matching CFE-3987.

PR #6292, opened 2026-08-15
CFE-3988: Fixed insert_lines ignoring include_end_delimiter when locating the region (3.24)

nickanderson:
Inserting after the last line of a region built with `include_end_delimiter => "true"` landed above the delimiter instead of below it, because the insertion path treats the region as the half-open range `[begin_ptr, end_ptr)`.

Backport of #6265. The new test's `inputs` uses `default.cf.sub`, which is what that file is called on this branch.

Ticket: CFE-3988


cf-bottom:

PR #6305, opened 2026-08-18
CFE-4729: Kill a timed-out command's whole process group

djbclark:
A `commands:` promise's descendants survive `exec_timeout`. `GracefulTerminate()` reaches only the direct child; if that child is a shell running `sleep 30; exit 0`, killing the shell does not stop the `sleep`, which keeps the pipe open, so the agent stays blocked reading a command it already gave up on. `exec_timeout` bounds nothing in that shape.

## The change

The timed command's child now leads a process group of its own, but only while a timeout is armed (`TimeOutIsArmed()`, checked in `cf_popen()`'s child via `setpgid(0, 0)`) — not unconditionally. Unconditional grouping was tried and reverted: it detaches every piped child from the terminal's foreground group, so an interactive `type => "r"` command that reads stdin is stopped by `SIGTTIN` the moment it does, with nothing to end the resulting hang.

`TimeOut()` now signals that group (`kill(-ALARM_PID, SIGKILL)`) after `GracefulTerminate()`, guarded on `getpgid(ALARM_PID) == ALARM_PID` — only when the child actually led its own group, so a `setpgid()` failure degrades to the pre-existing single-process kill rather than signalling something unrelated.

`SetTimeOut()`/`ClearTimeOut()`/`TimeOut()` track the armed state in `TIMEOUT_ARMED` (`volatile sig_atomic_t`, written from the signal handler). The four open-coded `alarm(0); signal(SIGALRM, SIG_DFL);` call sites now go through `ClearTimeOut()` instead, so a command finishing in time cannot leave the flag set for the next, unrelated, child.

Windows has no POSIX process groups; `getpgid()`/`kill(-pid, ...)` are guarded `#ifndef __MINGW32__`, matching `pipes_unix.c` (the `setpgid()` half) already being `if !NT` only.

## Supersedes #6299

This branch is #6299 (`fix/exec-timeout-promise-outcome`, `TIMEOUT_FIRED`/`TIMEOUT_SIGNALLED` outcome reporting) merged with this process-group fix — both touch `SetTimeOut()`/`TimeOut()`/`cf-agent/verify_exec.c`, and #6299's `ClearTimeOut()` had to survive the merge without disturbing what it samples. If #6299 lands separately, this rebases down to the process-group commits alone; if not, this supersedes it and #6299 can close in favor of it.

## Tests

Unit: `tests/unit/timeout_test.c`, 6 cases pinning `SetTimeOut()`/`ClearTimeOut()`/`TimeOut()`'s armed/fired/signalled contract, including one that forks a real child so a `ClearTimeOut()` that wipes a *true* `TimeOutSignalledProcess()` (not just a false one) is caught. `if !NT` — waits on a real `SIGALRM`.

Acceptance: the six tests already in `tests/acceptance/08_commands/04_exec_timeout/` (from #6299) pass unchanged, including `timeout_kills_descendants.cf`.

Discrimination on the two new commits, individually: `ClearTimeOut()`/`TimeOut()` each failing to disarm, `ClearTimeOut()` clearing `TIMEOUT_FIRED`/`TIMEOUT_SIGNALLED`, and `ClearTimeOut()` clearing a *true* `TIMEOUT_SIGNALLED` — each breaks exactly one unit test. Full rebuild clean (0 warnings), source restores byte-identical by sha256 throughout.

Cut from master `17eb78e6d`. Tracked as [CFE-4729](https://northerntech.atlassian.net/browse/CFE-4729).

[CFE-4729]: https://northerntech.atlassian.net/browse/CFE-4729?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cf-agent/verify_exec.c
+++ b/cf-agent/verify_exec.c
@@ -451,23 +451,17 @@ static ActionResult RepairExec(EvalContext *ctx, const Attributes *a,
         {
             int ret = cf_pclose(pfp);
 
-            /* Sample only now, and never earlier. The read loop above ends as
-             * soon as the command closes its output, which it can do long
-             * before it exits -- so the alarm may not fire until cf_pclose()
-             * is already waiting for the child. Reading the flag before that
-             * wait misses exactly the case this is here to catch. It is still
-             * read before the alarm is disarmed further down. */
+            /* Only after cf_pclose(): the read loop ends when the command
+             * closes its output, which can be long before it exits, so the
+             * alarm may not fire until cf_pclose() is already waiting. */
             timed_out = (a->contain.timeout != CF_NOINT) && TimeOutHasFired();
 
             if (timed_out)
             {
-                /* The command exceeded exec_timeout and was signalled, so its
-                 * exit status cannot be trusted to say so. A command killed
-                 * after it has written its last output, or one that exits
-                 * normally while only its children are killed, is reaped with a
-                 * status VerifyCommandRetcode() reads as success -- and the
-                 * promise is then reported kept or repaired even though the
-                 * command never completed. Classify on the timeout instead. */
+                /* Classify on the timeout, not the exit status: a command
+                 * killed after its last output, or one whose children alone
+                 * are killed, is reaped with a status VerifyCommandRetcode()
+                 * reads as success. */
                 cfPS(ctx, LOG_LEVEL_ERR, PROMISE_RESULT_TIMEOUT, pp, a,
                      TimeOutSignalledProcess()
                          ? "Command '%s' exceeded exec_timeout of %d seconds and was terminated"
--- a/cf-agent/verify_exec.c
+++ b/cf-agent/verify_exec.c
@@ -288,6 +288,9 @@ static ActionResult RepairExec(EvalContext *ctx, const Attributes *a,
 
     CommandPrefix(cmdline, comm);
 
+    /* Set once the command has been reaped, if its exec_timeout fired. */
+    bool timed_out = false;
+
     bool do_work_here = true;
 
 #ifndef __MINGW32__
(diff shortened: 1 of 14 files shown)


cf-agent/verify_exec.c
```diff
@@ -448,7 +451,31 @@ static ActionResult RepairExec(EvalContext *ctx, const Attributes *a,
         {
             int ret = cf_pclose(pfp);
 
-            if (ret == -1)
+            /* Sample only now, and never earlier. The read loop above ends as
```
nickanderson: It feels a bit comment heavy. Probably "terse" comments will be sufficient in most cases.

djbclark: Agreed. Trimmed across the whole diff in 0e06ad3d7 — comment volume roughly halved, keeping only the non-obvious constraints (the sampling order, the pgid guard, the SIGTTIN reason for the `TimeOutIsArmed()` condition).

No code change: with comments stripped, the token stream is byte-identical to the previous head.

Left it as a separate commit so your comment thread stays anchored; say the word and I will squash it into the series.

tests/acceptance/08_commands/04_exec_timeout/timeout_after_output_closed.cf
```diff
@@ -0,0 +1,61 @@
...
+bundle agent check
+{
+  methods:
+    "any"
+      usebundle => dcs_passif_expected(
+        "output_closed_repair_timeout",
+        "output_closed_promise_kept,output_closed_promise_repaired",
+        $(this.promise_filename)
+      ),
+      inherit => "true";
```
nickanderson: What are you inheriting here? I don't see any bundle scoped classes defined within this bundle so I don't think inherit is doing anything useful,

tests/acceptance/08_commands/04_exec_timeout/timeout_does_not_leak_to_next_promise.cf
```diff
@@ -0,0 +1,66 @@
...
+bundle agent check
+{
+  methods:
+    "any"
+      usebundle => dcs_passif_expected(
+        "leak_first_repair_timeout,leak_second_promise_repaired",
+        "leak_first_promise_kept,leak_first_promise_repaired,leak_second_repair_timeout",
+        $(this.promise_filename)
+      ),
+      inherit => "true";
```
nickanderson: another i believe un-necessary inherit.

tests/acceptance/08_commands/04_exec_timeout/timeout_overrides_exit_zero.cf
```diff
@@ -0,0 +1,58 @@
...
+bundle agent check
+{
+  methods:
+    "any"
+      usebundle => dcs_passif_expected(
+        "timeout_exit0_repair_timeout",
+        "timeout_exit0_promise_kept,timeout_exit0_promise_repaired",
+        $(this.promise_filename)
+      ),
+      inherit => "true";
```
nickanderson: also un-necessary i believe.

tests/acceptance/08_commands/04_exec_timeout/timeout_overrides_kept_returncodes.cf
```diff
@@ -0,0 +1,66 @@
...
+bundle agent check
+{
+  methods:
+    "any"
+      usebundle => dcs_passif_expected(
+        "timeout_kept0_repair_timeout",
+        "timeout_kept0_promise_kept,timeout_kept0_promise_repaired",
+        $(this.promise_filename)
+      ),
+      inherit => "true";
```
nickanderson: un-necessary

tests/acceptance/08_commands/04_exec_timeout/timeout_overrides_kept_returncodes.cf
```diff
@@ -0,0 +1,66 @@
...
+# dcs_all_classes (dcs.sub.cf) carries no returncode attributes and
+# classes bodies do not compose, so this is dcs_all_classes plus
+# kept_returncodes.
+body classes all_classes_kept_0(prefix)
+{
+  promise_kept => { "$(prefix)_promise_kept" };
+  promise_repaired => { "$(prefix)_promise_repaired" };
+  repair_failed => { "$(prefix)_repair_failed" };
+  repair_denied => { "$(prefix)_repair_denied" };
+  repair_timeout => { "$(prefix)_repair_timeout" };
```
nickanderson: Instead of duplicating th full body you could `inherit_from => dcs_all_classes;`

https://docs.cfengine.com/docs/lts/reference/promise-types/#inherit_from

tests/acceptance/08_commands/04_exec_timeout/within_timeout_normal_outcomes.cf
```diff
@@ -0,0 +1,59 @@
...
+bundle agent check
+{
+  methods:
+    "any"
+      usebundle => dcs_passif_expected(
+        "within_exit0_promise_repaired,within_exit3_repair_failed",
+        "within_exit0_repair_timeout,within_exit0_repair_failed,within_exit3_repair_timeout,within_exit3_promise_repaired",
+        $(this.promise_filename)
+      ),
+      inherit => "true";
```
nickanderson: unnecessary inherit.

olehermanse:

PR #6308, opened 2026-08-18
CFE-4732: Fixed ReconcileMountOptions() leaving its remount timeout armed

djbclark:
Ticket: [CFE-4732](https://northerntech.atlassian.net/browse/CFE-4732)

`ReconcileMountOptions()` arms a timeout at the remount and unmount_mount sites (`nfs.c:1436`, `:1461`) but never disarms it; the only reason it doesn't leak past the function is that `LiveMountConverged()` -> `LoadMountInfo()` incidentally re-arms and clears its own alarm as a side effect. Disarm locally after the dispatch, before `LiveMountConverged()`, matching the file's own idiom at `:581` and `:1178`.

No unit test: this function returns early under a dry run before reaching the fix, and its live path shells out to `mount`/`umount`, which CI can't run. Happy to extract a testable helper if you'd rather have the coverage.

AI-assisted, reviewed and verified by me before submitting.

[CFE-4732]: https://northerntech.atlassian.net/browse/CFE-4732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cf-agent/nfs.c
+++ b/cf-agent/nfs.c
@@ -1468,6 +1468,10 @@ PromiseResult ReconcileMountOptions(EvalContext *ctx, char *name, const Attribut
             continue;
         }
 
+        /* Clear the mount reconciliation timeout before moving on to other things */
+        alarm(0);
+        signal(SIGALRM, SIG_DFL);
+
         /* Verify-after-act: confirm the live mount actually satisfies the promise. */
         if (LiveMountConverged(name, a))
         {


nickanderson:
Hi @djbclark thanks for your attention to CFEngine core, I will take a look at this, but can you please clamp down more on the verbosity of the comments and description. This one is like 100 characters of prose (combined pr description and code comments) per character of code.

Human developers still read all our code and we can not keep pace with the volume the llms like to emit. Bite sized PRs that are demonstrated issues (existing bugs with reports from users) are likely to go in easier than PRs that are very large which have not been reported by users.

cf-agent/nfs.c
```diff
@@ -1468,6 +1468,15 @@ PromiseResult ReconcileMountOptions(EvalContext *ctx, char *name, const Attribut
             continue;
         }
 
+        /* Both methods above arm a timeout and neither disarms it. It bounds
+         * the reconcile command, so retire it here rather than letting it run
+         * on into the convergence check and the next method. Until now it only
+         * ever stopped because LiveMountConverged() -> LoadMountInfo() happens
+         * to arm and disarm one of its own; nothing should depend on a callee
+         * for that. */
```
nickanderson: One or
```suggestion
        /* Clear the mount reconciliation timeout before moving on to other things */

nickanderson (changes requested):
I think this comment can be made much more concise.

Can you please adjust your commit message to match the cfengine core style? past-tense subject, no fix: prefix, and Ticket: CFE-4732 + Changelog: trailers. (The PR title is fine.)

djbclark:
Done — force-pushed to 9dd5eb51c.

- Commit message rewritten to the cfengine style (past-tense subject, no `fix:` prefix, `Changelog:`/`Ticket:` trailers).
- Comment shrunk per your suggestion.
- PR description cut down to defect + fix + why-no-test.

Will keep future PRs smaller and lower-prose, and prioritize demonstrated user-reported bugs over self-found ones. Thanks for the direct feedback.

djbclark:
Correction to the "no unit test" note in the description — I gave the wrong reason.

I said the live path shells out to `mount` so CI can't reach it. Not true: forcing `cf_popen()` to fail (`RLIMIT_NOFILE`) reaches both the arm and the disarm without running `mount`.

The real blocker is that there is nothing to observe. `LiveMountConverged()` always calls `LoadMountInfo()`, which arms `SetTimeOut(RPCTIMEOUT)` at `nfs.c:403` — unconditionally, and ahead of its early returns at `:408`, `:427`, `:487`. The alarm state is overwritten on every path, so a build with this patch is indistinguishable from one without it anywhere outside `ReconcileMountOptions()`.

That makes this fragility removal rather than a bug fix: the current non-leak depends on a callee happening to re-arm. If that's not worth the churn, closing it is a reasonable call and I won't re-open.

olehermanse:
See: https://github.com/cfengine/core/pull/6293#issuecomment-5340989869

PR #6310, opened 2026-08-18
CFE-4727: Fixed exec_timeout never terminating a command that closed its output

djbclark:
Ticket: [CFE-4727](https://northerntech.atlassian.net/browse/CFE-4727)

> **Stacked on #6305.** This branch is cut from that PR's series, so the commits below it are the ones already under review there. **The only new commit here is `8f4ebedbd`**, and it is the one to review. If #6305 merges first this becomes a single-commit PR.

`cf_pclose()` and `cf_pclose_full_duplex()` cleared `ALARM_PID` *before* waiting for the child. A command that closed its output but kept running was therefore unreachable by the time the alarm fired: `TimeOut()` found `ALARM_PID == -1`, terminated nothing, and the command ran to completion entirely unbounded by `exec_timeout`.

This is the termination-half companion to CFE-4726 (#6299), which covers the reporting fail-open. CFE-4726's "was NOT terminated and ran to completion" wording exists because of this gap.

New `ClearAlarmedPid()` in `pipes_unix.c` clears `ALARM_PID` only *after* `cf_pwait()` has reaped, with `pthread_sigmask()` around the compare-and-clear so the alarm cannot land mid-sequence.

**Testing** — measured on Darwin arm64: unfixed ~30 s and never signalled; fixed ~2 s and signalled. The acceptance test was rewritten to assert termination rather than mere detection: 19.41 s with the fix against 31.96 s reverted, 6/6 in `04_exec_timeout/`. Unit tests 7/7 including a new `test_pclose_leaves_the_alarm_its_process`.

[CFE-4727]: https://northerntech.atlassian.net/browse/CFE-4727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cf-agent/nfs.c
+++ b/cf-agent/nfs.c
@@ -578,8 +578,7 @@ bool LoadMountInfo(Seq *list)
     }
 
     free(vbuff);
-    alarm(0);
-    signal(SIGALRM, SIG_DFL);
+    ClearTimeOut();
     cf_pclose(pp);
     return true;
 }
@@ -1175,8 +1174,7 @@ void MountAll()
     }
 
     free(line);
-    alarm(0);
-    signal(SIGALRM, SIG_DFL);
+    ClearTimeOut();
     cf_pclose(pp);
 }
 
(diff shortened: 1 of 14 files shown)


olehermanse:

PR #6311, opened 2026-08-18
CFE-4734: Fixed exec_timeout arming its alarm before the child existed

djbclark:
Ticket: [CFE-4734](https://northerntech.atlassian.net/browse/CFE-4734)

> **Stacked on #6305.** Cut as a sibling from that PR's series, not stacked on #6310 or #6312. **The only new commit here is `ab98a5df4`.** If #6305 merges first this becomes a single-commit PR.

`SetTimeOut()` called `alarm()` immediately, but `cf_popen()` only registers the child in `ALARM_PID` after `fork()`. A `SIGALRM` delivered in that window — which spans `umask()`, logging and the `cf_popen` dispatch — found no process, burned the timeout on nothing, and the command then ran to completion unbounded by `exec_timeout`. Observed as a unit-test flake under parallel `make check` load.

`SetTimeOut()` now stashes the timeout and leaves the clock stopped; `cf_popen()` starts it via `StartTimeOutClock()` immediately after publishing the pid, so no alarm can be pending without a process to kill. One-shot per `SetTimeOut()`: a second fork under the same timeout runs on the remaining time, as before. `TIMEOUT_ARMED` is still set before the fork, so the child's process-group decision is unchanged, and the agent's own setup work is no longer charged against the command's budget.

mingw and cygwin keep arming in `SetTimeOut()` — they are exactly the platforms that do not build `pipes_unix.c`, so nothing there would reach the starter. (Guarding on `__MINGW32__` alone would have left Cygwin on the deferred path with no starter at all, since `AM_CONDITIONAL([NT])` is `mingw|cygwin`; `timeout_test` is `!NT` too, so no CI job would have caught it.)

This does not make the existing `SetTimeOut()` leaks safe and does not try to. A timeout armed but never followed by a successful `cf_popen` now leaves `TIMEOUT_PENDING` set with no timer running, and the next `GenericCreatePipeAndFork()` starts that *full* budget against whatever it forks — where before, the leaked alarm was already ticking and could only spend its remainder. Those leak sites are pre-existing and handled separately (CFE-4732, #6308).

**Testing** — unit 6/6 → 10/10; acceptance `08_commands/04_exec_timeout` 6/6 both sides (62 s → 74 s). Discrimination proved rather than asserted: rebuilding with the pre-fix arming order emulated fails exactly 2 of 10 (`test_clock_does_not_run_before_the_fork`, `test_set_leaves_the_clock_stopped`), both passing again on revert.

[CFE-4734]: https://northerntech.atlassian.net/browse/CFE-4734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cf-agent/nfs.c
+++ b/cf-agent/nfs.c
@@ -578,8 +578,7 @@ bool LoadMountInfo(Seq *list)
     }
 
     free(vbuff);
-    alarm(0);
-    signal(SIGALRM, SIG_DFL);
+    ClearTimeOut();
     cf_pclose(pp);
     return true;
 }
@@ -1175,8 +1174,7 @@ void MountAll()
     }
 
     free(line);
-    alarm(0);
-    signal(SIGALRM, SIG_DFL);
+    ClearTimeOut();
     cf_pclose(pp);
 }
 
--- a/cf-agent/verify_exec.c
+++ b/cf-agent/verify_exec.c
@@ -288,6 +288,9 @@ static ActionResult RepairExec(EvalContext *ctx, const Attributes *a,
 
     CommandPrefix(cmdline, comm);
 
+    /* Set once the command has been reaped, if its exec_timeout fired. */
+    bool timed_out = false;
+
     bool do_work_here = true;
 
 #ifndef __MINGW32__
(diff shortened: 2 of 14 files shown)


olehermanse:

PR #6312, opened 2026-08-18
CFE-4735: Fixed ALARM_PID staying armed against a reaped pid on fdopen() failure

djbclark:
Ticket: [CFE-4735](https://northerntech.atlassian.net/browse/CFE-4735)

> **Stacked on #6310, which is itself stacked on #6305.** This change uses `ClearAlarmedPid()`, introduced by #6310, so it genuinely depends on it. **The only new commit here is `453cc264d`.** Merging #6305 and #6310 first reduces this to a single-commit PR.

`cf_popen*()`'s eight `fdopen()`-failure paths reap the child but never clear `ALARM_PID` (`pipes_unix.c:458,470,588,600,670,681,789,800` — all `cf_pwait(pid); return NULL;`), and `RepairExec()`'s `pfp == NULL` branch returns `ACTION_RESULT_FAILED` without ever reaching the function's only `ClearTimeOut()` at `:502`.

An armed alarm can then stay pointed at a reaped, recyclable pid for up to the full `exec_timeout` — the same defect class as CFE-4727, but a window of seconds-to-minutes rather than instructions. Pre-existing and byte-identical to the pre-CFE-4727 parent; gated on rare `fdopen()` failures (ENOMEM/EMFILE).

Mechanical fix: `ClearAlarmedPid(pid)` after each of the eight `cf_pwait()` calls, plus `ClearTimeOut()` on `RepairExec()`'s `pfp == NULL` path under the same `a->contain.timeout != CF_NOINT` gate the normal-completion path already uses.

**Testing** — the original "untestable" claim was wrong. `tests/unit/timeout_test.c` gains `test_leftover_alarm_does_not_kill_next_child`, portable POSIX using `RLIMIT_NOFILE` to force a `pipe()` failure, no interposition needed. Discriminated by hand: it fails exactly at `TimeOutHasFired()` with the fix's `ClearTimeOut()` call removed.

Two further real-but-out-of-scope leaks were found while doing this and are deliberately not in this SHA: a raw fd leak on `fdopen()` failure, and two more `RepairExec()` early returns that skip `ClearTimeOut()`. Both belong to the CFE-4732 family (#6308).

[CFE-4735]: https://northerntech.atlassian.net/browse/CFE-4735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cf-agent/nfs.c
+++ b/cf-agent/nfs.c
@@ -578,8 +578,7 @@ bool LoadMountInfo(Seq *list)
     }
 
     free(vbuff);
-    alarm(0);
-    signal(SIGALRM, SIG_DFL);
+    ClearTimeOut();
     cf_pclose(pp);
     return true;
 }
@@ -1175,8 +1174,7 @@ void MountAll()
     }
 
     free(line);
-    alarm(0);
-    signal(SIGALRM, SIG_DFL);
+    ClearTimeOut();
     cf_pclose(pp);
 }
 
--- a/cf-agent/verify_exec.c
+++ b/cf-agent/verify_exec.c
@@ -288,6 +288,9 @@ static ActionResult RepairExec(EvalContext *ctx, const Attributes *a,
 
     CommandPrefix(cmdline, comm);
 
+    /* Set once the command has been reaped, if its exec_timeout fired. */
+    bool timed_out = false;
+
     bool do_work_here = true;
 
 #ifndef __MINGW32__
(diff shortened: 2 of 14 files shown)


olehermanse:

PR #6313, opened 2026-08-18
CFE-4736: Fixed short option strings disagreeing with their long option tables

djbclark:
Ticket: [CFE-4736](https://northerntech.atlassian.net/browse/CFE-4736)

Several binaries declare a long option and implement its handler, but omit the short character from the `getopt_long()` option string — or disagree with their own option table about whether the short form takes an argument. The advertised short form is then rejected as unknown, or consumes the wrong argument.

Concretely on master: `cf-check` declares `{"version", no_argument, 0, 'V'}` (`cf-check.c:54`) and handles `case 'V'` (`:159`), but its option string is `"+hMg:dvI"`. `cf-check -V` exits with "unrecognized option".

**How the list was produced** — a checker pairs each `getopt_long()` call site with the `struct option` table it actually references, resolving the option-string variable to its nearest preceding definition. Pairing naively per file gives false positives: `cf-net` has three call sites with three different tables and option strings, and its `-o/--output` and `-j/--jobs` are correctly declared in subcommand-local tables.

One commit per binary:

| binary | change |
|---|---|
| `cf-check` | added missing `V` |
| `cf-net` | added missing `t:` and `c:` |
| `cf-secret` | added missing `v`, `I`, `g:` |
| `cf-testd` | added missing `r:` |
| `cf-serverd` | `-L` (`--ld-library-path`) now consumes its argument |
| `cf-runagent` | added `--no-lock` long form for the existing `-K`; removed vestigial `q:` and stray `S` |
| `cf-promises` | removed stray `S` and vestigial `i:`; removed duplicate `g:` |
| `cf-monitord` | removed stray `S` |
| `cf-execd` | removed stray `1`; removed duplicate `V` and bare `g` |

Duplicate removals are behaviour-preserving — `getopt_long()` matches the first occurrence, so the second was already dead in every case.

**Verified by building and running**, not only by reading:

- `cf-check -V` → `cf-check BETA version 3.29.0a` (previously "unrecognized option")
- `cf-secret -v` → `verbose: Verbose log level enabled` (previously rejected)
- `cf-testd -r` → `option requires an argument -- r`, the correct response for `r:` (previously "unknown option")
- `cf-execd -V` and `cf-execd -g info` unchanged after the duplicate removal, as expected

After this branch the checker reports no missing entries and no duplicates in any shipped binary.

**Two things deliberately left out, discussed in the ticket** — `cf-agent`'s `-x/--self-diagnostics` and `cf-check`'s `-h/--help` both declare `optional_argument` while the option string says none, *and neither handler reads `optarg`*. Fixing the table (start erroring on `--help=foo`) and fixing the option string (keep accepting an argument that is discarded) are both user-visible changes, so I would rather you pick. Likewise `-M` is handled in 9 of 10 binaries but only 3 declare `--manpage`.

Happy to split this per binary or narrow it if that suits review better.

AI-assisted and multi-model vetted before submitting, reviewed by me, per CONTRIBUTING.

[CFE-4736]: https://northerntech.atlassian.net/browse/CFE-4736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


olehermanse:

PR #6314, opened 2026-08-18
CFE-4737: Fixed JSON reals truncated to two decimals and large JSON integers terminating the agent

djbclark:
Ticket: [CFE-4737](https://northerntech.atlassian.net/browse/CFE-4737)

Two call sites take a JSON number, convert it to a C numeric type, and format it back. Both conversions are wrong in the same two ways.

**Reals lose precision silently.** `RlistAppendJson()` and the iteration equivalent ran reals through `StringFromDouble()`, which formats with `"%.2f"`. Any real with more than two decimals came back as a *different* value, not a rounded one: `0.00049` becomes `0.00`. That is how a rate or threshold in JSON data silently becomes zero once it reaches a CFEngine variable.

**Large integers terminate the agent.** The same paths used `JsonPrimitiveGetAsInteger()`, which reaches `StringToLongExitOnError()` and so `DoCleanupAndExit()`. JSON places no limit on a number's magnitude, so a valid document may hold one no `long` can represent — and merely *rendering* it killed the process rather than reporting anything.

Two further callers have the same exit-on-overflow shape and take their input from JSON rather than from our own code: `ReadTimestampFromPolicyValidatedFile()` and the interface-data path in `libenv/unix_iface.c`.

For the list and iteration cases the repair is to stop converting and use the number as parsed, via `JsonPrimitiveGetAsString()`. For the two that genuinely want an integer, it is to stop using a converter that exits — a timestamp that cannot be read now leaves `validated_at` at 0, which already means "not validated" and is the safe direction.

**Independence from CFE-4724.** This branch was originally developed against an unmerged libntech commit, and I had assumed it was blocked on that. It is not, and I verified rather than assumed: `JsonPrimitiveGetAsString()` already exists on libntech master (`libutils/json.h:246`), which already creates primitives from the parsed text buffer (`json.c:2377`). This branch is rebased onto current master with the `libntech` submodule pointer **identical to master's** (`0c0620d`) — the diff is 5 source files, no submodule change — and it builds clean and passes its test against libntech master. CFE-4724 fixes complementary defects on the libntech side; neither blocks the other.

**Testing.** New `test_from_container_numbers` goes through the public `RlistFromContainer()` with ten numbers — `LONG_MAX` and one past it, 10^30, a value that fits a `long` but not an `int`, reals with more than two decimals, and exponent forms — requiring each back exactly as written. Discrimination verified both directions: with `libpromises/rlist.c` reverted the test does not merely fail, it takes the process with it —

```
error: Conversion error (34 - Overflow) on '9223372036854775808' (StringToLongExitOnError)
```

— which is the defect itself. With the fix, all ten round-trip.

*Note for macOS reviewers:* `rlist_test` aborts later in `test_rval_to_scalar2` at `rlist.c:135`. Pre-existing and unrelated — it reproduces identically on unpatched master, and is why `rlist_test` is on the macOS `XFAIL_TESTS` list.

AI-assisted and multi-model vetted before submitting, reviewed by me, per CONTRIBUTING.

[CFE-4737]: https://northerntech.atlassian.net/browse/CFE-4737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libenv/unix_iface.c
+++ b/libenv/unix_iface.c
@@ -1435,11 +1435,19 @@ void GetNetworkingInfo(EvalContext *ctx)
                 JsonElement *metric = JsonObjectGet(route, "metric");
                 if (metric != NULL &&
                     JsonGetElementType(metric) == JSON_ELEMENT_TYPE_PRIMITIVE &&
-                    JsonGetPrimitiveType(metric) == JSON_PRIMITIVE_TYPE_INTEGER &&
-                    (default_route == NULL ||
-                     JsonPrimitiveGetAsInteger(metric) < lowest_metric))
+                    JsonGetPrimitiveType(metric) == JSON_PRIMITIVE_TYPE_INTEGER)
                 {
-                    default_route = route;
+                    /* Not JsonPrimitiveGetAsInteger(): it reaches
+                     * StringToLongExitOnError(), which would terminate the
+                     * agent over a metric too large for a long. A metric we
+                     * cannot read simply does not win. */
+                    long metric_value;
+                    if (StringToLong(JsonPrimitiveGetAsString(metric),
+                                     &metric_value) == 0 &&
+                        (default_route == NULL || metric_value < lowest_metric))
+                    {
+                        default_route = route;
+                    }
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


olehermanse:

PR #6315, opened 2026-08-18
CFE-4719: Rejected CMDB data now names the offending entry and the file

djbclark:
A `host_specific.json` holding an unresolved variable reference is rejected with only the section name:

```
error: Invalid 'vars' CMDB data, cannot contain variable references
error: Failed to load CMDB data
```

Since one bad entry drops the whole section, an operator with a large CMDB file has to bisect it by hand to find which entry was at fault. This names the entry, and the file it came from.

Both `JsonWalk()` visitors already took a `void *data` parameter declared `ARG_UNUSED`, so the carrier was already in the signature and simply unused. They now record what stopped the walk, and the three call sites (`vars`, `variables`, `classes`) share one helper.

```
error: Invalid 'vars' CMDB data in '/var/cfengine/data/host_specific.json',
cannot contain variable references (offending value in entry
'key_with_unresolved_value': '$(sys.workdir)')
```

### Why the walk is per-entry

The helper walks each top-level entry separately rather than the section as a whole. A single walk can only report the offending primitive's *immediate* property name — which in the `variables` format (CFE-3633) is the metadata key `value`, not the variable's name, and for a value inside an array is nothing at all.

### Scope

The accepted and rejected sets are unchanged; only the log text differs. Which entry is named can differ when a document has several offenders, since the walk now finishes one entry before starting the next.

### Testing

Two acceptance tests, because one is not enough to pin the behaviour. `15-variable-references-metadata-format.cf` uses the `variables` format, where folding the walk back into a single section walk reports `value` instead of the variable's name — verified by simulating exactly that regression: test 15 fails under it, test 12 still passes.

All of `00_basics/06_host_specific_data` passes (13 tests, the full directory on this branch). Verified message shapes for: bad value in `vars`, bad key in `vars`, bad value in the `variables` metadata format, a value inside an array, a bad key nested inside an entry, and `classes`.

### Note

Developed and tested on macOS (arm64). Unit and acceptance suites were run locally; I have not been able to exercise this on the project's own CI.

Ticket: [CFE-4719](https://northerntech.atlassian.net/browse/CFE-4719)

[CFE-4719]: https://northerntech.atlassian.net/browse/CFE-4719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libpromises/cmdb.c
+++ b/libpromises/cmdb.c
@@ -67,17 +67,38 @@ JsonElement *ReadJsonFile(const char *filename, LogLevel log_level, size_t size_
     return doc;
 }
 
-static bool CheckPrimitiveForUnexpandedVars(JsonElement *primitive, ARG_UNUSED void *data)
+/* What made the unexpanded-variable walk below stop: exactly one of these is
+ * set, depending on which visitor stopped it. Both point into the JSON
+ * document being walked and must not be freed. */
+typedef struct
+{
+    const char *key;            /* set if a key held the reference */
+    const char *value;          /* set if a value held the reference */
+} CMDBUnexpandedVar;
+
+static bool CheckPrimitiveForUnexpandedVars(JsonElement *primitive, void *data)
 {
     assert(JsonGetElementType(primitive) == JSON_ELEMENT_TYPE_PRIMITIVE);
+    assert(data != NULL);
 
-    /* Stop the iteration if a variable expression is found. */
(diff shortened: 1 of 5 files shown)


olehermanse:

PR #6316, opened 2026-08-18
CFE-4738: Fixed a JSON null in host_specific.json crashing the agent

djbclark:
A `host_specific.json` containing a JSON `null` value segfaults `cf-promises` and `cf-agent`. The document is valid JSON and passes CFEngine's own unresolved-variable check.

```
$ cat $(sys.workdir)/data/host_specific.json
{"vars": {"k": null}}
$ cf-promises --show-classes -f promises.cf
Segmentation fault: 11
```

`cf-agent` dies inside `GenericAgentDiscoverContext()` — before any promise is evaluated — so an affected host stops converging and falls back to failsafe.

### Mechanism

```
AddCMDBVariable                          libpromises/cmdb.c:136
EvalContextVariablePutTagsSetWithComment libpromises/eval_context.c:2492
VariableTablePut                         libpromises/variable.c:255
RvalNewRewriter                          libntech/libutils/rlist.c:474
xstrdup                                  libntech/libutils/alloc.c:57
strdup -> strlen        <-- EXC_BAD_ACCESS at 0x0
```

`JsonPrimitiveToString()` returns NULL for `JSON_PRIMITIVE_TYPE_NULL`, and `AddCMDBVariable()` passes it straight in as the variable's value with no NULL check.

### Crashing inputs fixed

| input | before |
|---|---|
| `{"vars": {"k": null}}` | SIGSEGV |
| `{"variables": {"k": {"value": null}}}` | SIGSEGV |
| `{"variables": {"k": null}}` | SIGSEGV |
| `{"vars": {"ns:bundle.k": null}}` | SIGSEGV |
| `{"vars": {"k": [1, null]}}` | silently installed `{"1"}` |
| `{"vars": {"k": [null]}}` | silently installed an empty slist |

### Behaviour chosen

Skip the offending entry with an error and continue, rather than rejecting the section. This follows the file's own idiom — `cmdb.c` already `continue`s for every per-entry problem (bad var ref, missing `value` field, all five invalid-class cases) and reserves `return false` for section-level structural faults. A single bad entry in a generated CMDB file should not discard the rest of the host's data.

Nulls nested inside data containers are deliberately left alone — they are stored and rendered correctly today.

### Testing

`13-null-values.cf` asserts the full ordered diagnostic sequence *and* that the neighbouring good entries are still installed, so it pins the skip-and-continue semantics rather than merely the absence of a crash. It fails on unpatched code (which segfaults) and passes with the fix; the whole `00_basics/06_host_specific_data` directory passes.

### Related, not fixed here

The augments loader (`def.json`) has the same root cause at three sites in `libpromises/generic_agent.c` — `:454`, `:606` and `:723` all reach `JsonPrimitiveToString()` unchecked and segfault on the equivalent input. Different file and different feature, so it is left for a separate change rather than widening this one.

### Note

Ticket: [CFE-4738](https://northerntech.atlassian.net/browse/CFE-4738)

[CFE-4738]: https://northerntech.atlassian.net/browse/CFE-4738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libpromises/cmdb.c
+++ b/libpromises/cmdb.c
@@ -118,6 +118,22 @@ static VarRef *GetCMDBVariableRef(const char *key)
     return ref;
 }
 
+/** Does the given JSON array have a JSON null among its direct children? */
+static bool JsonArrayContainsNull(const JsonElement *array)
+{
+    assert(JsonGetType(array) == JSON_TYPE_ARRAY);
+
+    const size_t length = JsonLength(array);
+    for (size_t i = 0; i < length; i++)
+    {
+        if (JsonGetType(JsonArrayGet(array, i)) == JSON_TYPE_NULL)
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
 static bool AddCMDBVariable(EvalContext *ctx, const char *key, const VarRef *ref,
                             JsonElement *data, StringSet *tags, const char *comment)
 {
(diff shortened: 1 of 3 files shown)


olehermanse:

PR #6317, opened 2026-08-18
CFE-4722: Warned about CMDB keys whose dots are taken as a scope separator

djbclark:
`GetCMDBVariableRef()` hands a CMDB key to `VarRefParse()`, which splits it at the first dot. A flat-looking key such as `com.dotted.key` therefore installs the variable `dotted.key` in the scope `com`, reachable as `$(data:com.dotted.key)` and not as a variable of the name that was written. Nothing said so.

**Behaviour is unchanged.** Picking a scope this way is a supported feature that policies rely on; the same variable is installed at the same address as before. Only the logging is new.

```
warning: The '.' in the CMDB key 'com.dotted.key' was taken as a scope
separator: the variable is named 'dotted.key' in the 'com' scope,
addressed as 'data:com.dotted.key'
```

### Where the line is drawn

A warning on every agent run about a working configuration would be worse than no warning at all, so this reports at two levels:

- **`LOG_LEVEL_VERBOSE`** for a key of the documented `[namespace:]bundle.variable` form. That is indistinguishable from a deliberate scope qualification, so it is not warned about.
- **`LOG_LEVEL_WARNING`** only for keys that *cannot* be of that form: it has room for exactly one dot and neither half may be empty. So an empty scope (`.key`), an empty variable name (`key.`), or a variable name that still contains a dot (`com.dotted.key`).

The dotted-variable-name case is the strongest signal: a `vars:` promise splits its promiser the same way, so no policy can create a variable whose name contains a dot. Such an address can only come from CMDB data.

Keys with indices are unaffected — `idx[a.b]` has lval `idx`, and dots inside `[]` are never scope separators.

### Also fixed

The three `Installing CMDB variable` verbose messages printed `'%s:%s.%s'` of `ns`, `scope` and the raw key, which repeated the scope for any scope-qualified key (`data:com.com.dotted.key`) — misreporting the very address the operator needs. They now print `VarRefToString()` of the ref actually installed.

### Sections

`vars` and `variables` (CFE-3633) share `GetCMDBVariableRef()` and are both covered. `classes` is *not* affected: class names are not scope-parsed — `EvalContextClassPutTagsSet()` canonifies them, so `com.dotted.class` becomes `data:com_dotted_class` rather than moving scope. That canonification is also silent, but it is universal CFEngine behaviour for classes from every source, so it belongs in `eval_context.c` rather than the CMDB loader and is not touched here.

### Testing

`14-dotted-keys.cf` asserts both warnings fire, that all three variables still land at their original dotted addresses, and — as a false-positive guard — that `myscope.myvar` is **not** warned about. Fails on unpatched code, passes with the fix.

Behaviour non-regression was verified by diffing `--show-vars` and `--show-classes` output filtered to `source=cmdb`, before and after, across dotted, scoped, namespace-qualified, indexed and edge-case (`.leading`, `trailing.`, `a..b`) fixtures: byte-identical.

### Note

Ticket: [CFE-4722](https://northerntech.atlassian.net/browse/CFE-4722)

[CFE-4722]: https://northerntech.atlassian.net/browse/CFE-4722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libpromises/cmdb.c
+++ b/libpromises/cmdb.c
@@ -93,6 +93,45 @@ static bool CheckObjectForUnexpandedVars(JsonElement *object, ARG_UNUSED void *d
     return true;
 }
 
+/**
+ * Report that the dots in a CMDB key were taken as a scope separator.
+ *
+ * VarRefParse() splits a key at its first dot, so the flat-looking key
+ * "com.dotted.key" does not become a variable of that name, it becomes the
+ * variable "dotted.key" in the scope "com". Naming a scope like this is a
+ * supported feature, so a key of the documented "[namespace:]bundle.variable"
+ * form is only reported at the verbose level; warning on every agent run
+ * about a working policy would just be noise.
+ *
+ * The remaining keys cannot be of that form: it has room for exactly one dot,
+ * and neither the bundle name nor the variable name can be empty. A variable
+ * name that still contains a dot is a particularly good hint, since a vars
+ * promise cannot create one (the same splitting happens there), so nothing
+ * but CMDB data can address it. Those keys are much more likely to be flat
+ * keys meant literally, and get a warning.
+ */
+static void LogCMDBVariableScope(const char *key, const VarRef *ref)
+{
+    assert(key != NULL);
+    assert(ref != NULL);
+
+    char *address = VarRefToString(ref, true);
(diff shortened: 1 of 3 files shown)


olehermanse:

PR #6318, opened 2026-08-18
CFE-4721: Made policy function eval() return whole numbers without a fractional part

djbclark:
`eval()` always formats its result with `"%lf"`, so an integral result comes back with six decimal places:

```
R: eval(4-1)=[3.000000]  eval(2+2)=[4.000000]  eval(10/4)=[2.500000]
```

The consequence is that arithmetic cannot feed any function taking a count:

```
sublist(list, "tail", eval("$(n) - 1", "math", "infix"))
  -> Anomalous ending '.0' while parsing integer number: 4.000000
```

The working form today is `format("%d", eval(...))`. Notably, every other `eval()` call in the bundled acceptance suite already wraps the result in `int()` or `format()` — the six-decimal form was being worked around rather than relied on.

A new `FnReturnRealOrInt()` renders `"%ld"` when the result is integral and in range, and is otherwise unchanged.

### Range check

```c
isfinite(result) && result == floor(result) &&
result >= (double) LONG_MIN && result < (double) LONG_MAX + 1.0
```

The bounds are asymmetric deliberately. `LONG_MIN` is a power of two and converts to `double` exactly, so `>=` is correct. `(double) LONG_MAX` rounds *up* to `LONG_MAX + 1` on 64-bit `long`, so `result <= (double) LONG_MAX` would admit 2^63 and make the cast undefined.

| expression | before | after |
|---|---|---|
| `4-1` | `3.000000` | `3` |
| `10/4` | `2.500000` | `2.500000` |
| `-1*0` | `-0.000000` | `0` |
| `3/0` | `inf` | `inf` |
| `sqrt(-1)` | `nan` | `nan` |
| `2^62` | `4611686018427387904.000000` | `4611686018427387904` |
| `2^63` | `9223372036854775808.000000` | `9223372036854775808.000000` |
| `0-2^63` (`LONG_MIN`) | `-9223372036854775808.000000` | `-9223372036854775808` |

### Scope

`sum`, `product`, `mean` and `variance` keep `"%lf"` and are deliberately untouched. All four are declared `CF_DATA_TYPE_REAL` in the function table and policy assigns them to `real =>` variables, so a real number is their documented contract. `eval` is declared `CF_DATA_TYPE_STRING`, so changing its rendering stays within its contract. `"class"` mode is untouched.

### Compatibility

This is a visible change and worth calling out: any policy that string-compares or regex-matches `eval()` output will see the new form, and boolean-ish results shift too (`10 == 10` gives `1`, not `1.000000`). The bundled `eval.cf` was itself asserting `300.000000`, `100.000000` and similar, and its expectations are updated here. This probably belongs in a minor release rather than a patch release. I have not surveyed masterfiles or any Enterprise policy for `eval()` consumers.

### Testing

`test_eval_integral_result` in `tests/unit/evalfunction_test.c` drives the real `FnCallEval` rather than a reimplementation; it fails on unpatched code (`"3" != "3.000000"`) and passes with the fix, verified with the shared library forcibly relinked rather than only the test binary. Long-boundary cases are guarded for 32-bit `long`. Full unit suite passes.

### Note

Ticket: [CFE-4721](https://northerntech.atlassian.net/browse/CFE-4721)

[CFE-4721]: https://northerntech.atlassian.net/browse/CFE-4721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libpromises/evalfunction.c
+++ b/libpromises/evalfunction.c
@@ -194,6 +194,27 @@ static FnCallResult FnReturnF(const char *fmt, ...)
     return FnReturnNoCopy(buffer);
 }
 
+/*
+ * Return a double, rendered as an integer when it has no fractional part
+ * and fits in a long, and with "%lf" otherwise.
+ */
+static FnCallResult FnReturnRealOrInt(double result)
+{
+    /* The bounds are deliberately asymmetric. LONG_MIN is a power of two and
+     * converts to double exactly, but (double) LONG_MAX rounds up to
+     * LONG_MAX + 1 where long is 64 bits, so testing "result <= (double)
+     * LONG_MAX" would admit a value the cast cannot represent -- which is
+     * undefined behaviour, not a wrapped result. */
+    if (isfinite(result) && (result == floor(result)) &&
+        (result >= (double) LONG_MIN) &&
+        (result < (double) LONG_MAX + 1.0))
+    {
+        return FnReturnF("%ld", (long) result);
+    }
+
+    return FnReturnF("%lf", result);
+}
+
 static FnCallResult FnReturnContext(bool result)
 {
     return FnReturn(result ? "any" : "!any");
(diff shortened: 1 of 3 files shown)


olehermanse:

PR #6319, opened 2026-08-18
Fixed crash on JSON null values in def.json augments data

djbclark:
## Summary

The augments (`def.json`) loader has the same null-crash root cause as `host_specific.json` CMDB data ([CFE-4738](https://northerntech.atlassian.net/browse/CFE-4738), #6316): `JsonPrimitiveToString()` returns `NULL` for a JSON `null`, and that `NULL` was passed straight into `xstrdup()` (via `EvalContextVariablePut*`) or `CheckContextClassmatch()` without a check. Each of these crashed `cf-promises`/`cf-agent` with SIGSEGV:

```json
{"vars": {"k": null}}
{"variables": {"k": null}}
{"classes": {"k": null}}
```

- Reject a null value per entry and continue the loop -- the same skip-with-error idiom `LoadAugmentsData()` already uses for every other per-entry problem (invalid variable spec, missing `"value"` field). Whole-section rejection stays reserved for structural problems.
- An array of primitives containing a null is rejected the same way in `"vars"`, `"variables"`, and `"inputs"`: not a crash, but `RlistFromContainer()` silently drops nulls, so the installed list would otherwise be silently shorter than the source data -- the same defect class [CFE-4738](https://northerntech.atlassian.net/browse/CFE-4738) fixed in `cmdb.c`.
- The `"classes"` array path and the `class_expressions`/`regular_expressions`/further-`"augments"`-files array paths were already null-safe (`JsonIteratorNextValueByType(..., skip_null=true)`).
- `{"value": null}` under `"variables"` was already rejected safely before this change, via a `NULL_JSON()` check `cmdb.c` lacked.

Ticket: [CFE-4739](https://northerntech.atlassian.net/browse/CFE-4739)

## Test plan

- [x] Built unpatched, confirmed a real SIGSEGV (rc=139) loading the exact `def.json` from the new test
- [x] Rebuilt patched (forced full `libpromises` relink, not just recompile), confirmed the new acceptance test passes
- [x] Ran all 22 tests in `tests/acceptance/00_basics/def.json/`, zero regressions
- [x] Independent review (grok, on a disposable copy of the tree): SHIP -- no leaks on the new early-return paths, no missed null-crash site in the function

[CFE-4738]: https://northerntech.atlassian.net/browse/CFE-4738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CFE-4738]: https://northerntech.atlassian.net/browse/CFE-4738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libpromises/generic_agent.c
+++ b/libpromises/generic_agent.c
@@ -385,6 +385,22 @@ static inline const char *GetAugmentsComment(const char *item_type, const char *
     return JsonPrimitiveGetAsString(json_comment);
 }
 
+/** Does the given JSON array have a JSON null among its direct children? */
+static bool JsonArrayContainsNull(const JsonElement *array)
+{
+    assert(JsonGetType(array) == JSON_TYPE_ARRAY);
+
+    const size_t length = JsonLength(array);
+    for (size_t i = 0; i < length; i++)
+    {
+        if (JsonGetType(JsonArrayGet(array, i)) == JSON_TYPE_NULL)
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
 static bool LoadAugmentsData(EvalContext *ctx, const char *filename, const JsonElement* augment)
 {
     bool loaded = false;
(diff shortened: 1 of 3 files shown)


olehermanse:

PR #6320, opened 2026-08-18
CFE-4720: Fixed one bad CMDB entry discarding every entry in its section

djbclark:
Ticket: [CFE-4720](https://northerntech.atlassian.net/browse/CFE-4720)

Stacked on #6315 (CFE-4719) — this branch is #6315's commit plus one more. Diff will shrink to just the new commit once #6315 merges.

`CheckCMDBDataForUnexpandedVars()` rejected a whole `vars`/`variables`/`classes` section on the first entry containing an unresolved variable reference, discarding every other entry in that section along with it — a single bad `host_specific.json` key could silently drop all CMDB-sourced variables on a host. Checked per entry instead, inside each section's own install loop, matching this file's existing skip-and-continue idiom for other per-entry problems (an invalid variable specification, a missing `value` field).

New acceptance test: one good and one bad entry in the same `vars` section, asserting the bad entry is still rejected and named, and the good entry installs anyway. Discrimination proved: fails on the unpatched function, passes with the fix; full `00_basics/06_host_specific_data` suite still 14/14.

AI-assisted, reviewed and verified by me before submitting.

[CFE-4720]: https://northerntech.atlassian.net/browse/CFE-4720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libpromises/cmdb.c
+++ b/libpromises/cmdb.c
@@ -116,73 +116,66 @@ static bool CheckObjectForUnexpandedVars(JsonElement *object, void *data)
 }
 
 /**
- * Reject CMDB data containing unresolved variable references, naming the entry
- * that was rejected.
+ * Reject a single CMDB entry containing unresolved variable references,
+ * naming the entry that was rejected.
  *
- * Each top-level entry is walked separately so that the entry can be named,
- * which is what an operator needs to find the offending line. A single walk of
- * the whole section could only report the immediate property name, and in the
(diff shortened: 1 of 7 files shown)


olehermanse:

PR #6335, opened 2026-08-25
ENT-14434: Move reactor-plugin logic into cf-reactor daemon loop

victormlg:
**Context**

So originally, we wanted to have two daemons: one for cf-reactor and one for the agent driven cfengine code. Then it has been decided to merge these two binaries into a single one. In this PR, we move the cf-reactor daemon to core, split the reactor-plugin in different functions, and integrate them inside the daemon loop of cf-reactor

Merge together: https://github.com/cfengine/enterprise/pull/996 https://github.com/cfengine/nova/pull/2696

### DIFF
--- a/cf-reactor/cf-reactor.c
+++ b/cf-reactor/cf-reactor.c
@@ -33,12 +33,21 @@
 #include <man.h>
 #include <cleanup.h>
 #include <prototypes3.h>
+#include <signal.h>             /* signal, kill */
+#include <signals.h>            /* GetSignalPipe, MakeSignalPipe */
+#include <exec_tools.h>
 
 /*****************************************************************************/
 /* Globals                                                                   */
 /*****************************************************************************/
 
 int NO_FORK = false;
+
+#define DEFAULT_POLL_INTERVAL_SECS 30
+// this is just an arbitrary number that has to be higher or equal to the number of fds used by reactor-plugin
+#define N_ALL_FDS 8
+
+static volatile sig_atomic_t terminate = 0;
 
 /*******************************************************************/
 /* Command line options                                            */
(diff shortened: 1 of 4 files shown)


larsewi (commented):
Please add some more context. E.g., explain why chose to make cf-reactor plugin its own process. Also I expected cf-reactor in nova to already fork out and make it a daemon, but I have not seen this code removed.

cf-reactor/cf-reactor.c
```diff
@@ -185,10 +198,67 @@ int main(int argc, char *argv[])
...
+#endif /* !__MINGW32__ */
+
+    umask(077);
+    WritePID("cf-reactor.pid");
+
+    signal(SIGINT, HandleSignalsForDaemon);
+    signal(SIGTERM, HandleSignalsForDaemon);
+    signal(SIGBUS, HandleSignalsForDaemon);
+    signal(SIGHUP, HandleSignalsForDaemon);
+    signal(SIGPIPE, SIG_IGN);
```
larsewi, line 294: In nova there is a comment explaining why this needs to be set later

cf-reactor/cf-reactor.c
```diff
@@ -185,10 +198,67 @@ int main(int argc, char *argv[])
...
+        Log(LOG_LEVEL_VERBOSE, "Windows does not support starting processes in the background - starting in foreground");
+    }
+
+#else /* !__MINGW32__ */
+    pid_t existing_pid = ReadPID("cf-reactor.pid");
+    if ((existing_pid != -1) && (kill(existing_pid, 0) == 0))
+    {
+        Log(LOG_LEVEL_ERR, "Another instance of cf-reactor is already running, terminating");
+        return 1;
+    }
```
larsewi, line 245: Will the nova fork inherit `CleanPidFile` clean-up handler and remove core's PID file on exit?

cf-reactor/cf-reactor.c
```diff
@@ -185,10 +198,67 @@ int main(int argc, char *argv[])
...
+    int child = ReactorEnterpriseMain();
+    if (child == -1)
+    {
+        return 1;
+    }
+
+    for (int i = 0; !IsPendingTermination(); i++)
+    {
+        /* Do something */
+        Log(LOG_LEVEL_INFO, "cf-reactor: %d", i);
```
larsewi: This will log one line per second. That becomes quite a few log messages to syslog in a day.

cf-reactor/cf-reactor.c
```diff
@@ -185,10 +198,67 @@ int main(int argc, char *argv[])
...
+        sleep(1);
+    }
+
+    /* child == 0 means in this case that no child process was created.
+       It is the default value returned by enterprise stubs */
+    if (child > 0)
+    {
+        TerminateReactorEnterprise(child);
```
larsewi: Core should reap the child instead of leaving a zombie

larsewi (changes requested):
There one major issue here. If the nova fork dies, then core will continue running and no one will restart the nova reactor again. With this architecture, core must be responsible for restarting the nova fork if it dies.

cf-reactor/cf-reactor.c
```diff
@@ -185,10 +263,118 @@ int main(int argc, char *argv[])
...
+        if ((child > 0) && ReactorEnterpriseHasExited((pid_t) child))
+        {
+            if (IsPendingTermination())
+            {
+                /* Already shutting down, no point in restarting it. */
+                child = 0;
+                break;
+            }
+
+            if (time(NULL) - child_started_at > REACTOR_MIN_UPTIME_SECS)
```
larsewi:
```suggestion
            if ((time(NULL) - child_started_at) > REACTOR_MIN_UPTIME_SECS)
```

cf-reactor/cf-reactor.c
```diff
@@ -185,10 +263,118 @@ int main(int argc, char *argv[])
...
+#endif /* !__MINGW32__ */
+
+    while (!IsPendingTermination())
+    {
+        /* Do something */
+        sleep(1);
+
+#ifndef __MINGW32__
+        /* Here cf-reactor tries to restart the reactor plugin if it exited. It retries
+         * 10 times before giving up. The counter is set back to 0 after some time without failure */
```
larsewi: Maybe the core process should die itself. Otherwise, systemd will think everything is fine. Not sure. We are basically implementing `systemd` ourself by now. Could be worth discussing with PM if this is the correct way to go.

larsewi (commented):
Please write a test where you kill the nova cf-reactor and see that it comes back. Maybe also test reaching that limit that you set.

victormlg:
I updated the context of this PR to match what it currently contains

cf-reactor/cf-reactor.c
```diff
@@ -179,16 +186,162 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
...
+    }
+
+#endif /* !__MINGW32__ */
+
+    umask(077);
+    WritePID("cf-reactor.pid");
+
+    int all_fds[N_ALL_FDS];
+    // the first num_nova_fds fds are populated with nova fds
+    size_t num_nova_fds = ReactorNovaInitialize(all_fds, N_ALL_FDS);
```
larsewi: We probably need to differentiate between `ReactorNovaInitialize` failed and `ReactorNovaInitialize` has no FDs to contribute.

cf-reactor/cf-reactor.c
```diff
@@ -179,16 +186,162 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
...
+    signal(SIGTERM, HandleReactorSignals);
+    signal(SIGBUS, HandleReactorSignals);
+    signal(SIGHUP, HandleReactorSignals);
+    signal(SIGUSR1, HandleReactorSignals);
+    signal(SIGUSR2, HandleReactorSignals);
+    /* Writing to a pipe whose spawned process already exited (e.g. cfbs
+     * rejecting its arguments before reading its stdin) must fail with EPIPE
+     * rather than terminate the whole daemon. Set after ReactorNovaInitialize(),
+     * so that the spawner and the processes it execs keep the default handling. */
+    signal(SIGPIPE, SIG_IGN);
```
larsewi: These should probably happen before `ConnectAndListen` because that's the one place where the can sit for an unbound time. And right now it sits there without a way to shut it down cleanly.

cf-reactor/cf-reactor.c
```diff
@@ -179,16 +186,162 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
...
 /*****************************************************************************/
 
+static void HandleReactorSignals(int signum)
+{
+    HandleSignalsForDaemon(signum);
+    if (IsPendingTermination())
+    {
+        ReactorNovaTerminate();
+    }
+}
```
larsewi:
```suggestion
static void HandleReactorSignals(int signum)
{
    HandleSignalsForDaemon(signum);
    if (IsPendingTermination())
    {
        TERMINATE = 1;
    }
    signal(signum, HandleReactorSignals);   /* re-arm the wrapper */
}
```
You will need to re-arm this wrapper, otherwise the internal handler will get call directly the next time the signal is triggered

libpromises/prototypes3.h
```diff
@@ -85,7 +85,11 @@ ENTERPRISE_VOID_FUNC_0ARG_DECLARE(void, ReloadHAConfig);
 ENTERPRISE_VOID_FUNC_2ARG_DECLARE(void, Nova_ClassHistoryAddContextName, const StringSet *, list, const char *, context_name);
 ENTERPRISE_VOID_FUNC_2ARG_DECLARE(void, Nova_ClassHistoryEnable, StringSet **, list, bool, enable);
 
-ENTERPRISE_FUNC_1ARG_DECLARE(int, ReactorEnterpriseMain, bool, no_fork);
+ENTERPRISE_VOID_FUNC_0ARG_DECLARE(void, ReactorNovaTerminate);
+ENTERPRISE_FUNC_2ARG_DECLARE(size_t, ReactorNovaInitialize, ARG_UNUSED int*, fds, ARG_UNUSED size_t, max_size);
```
larsewi: Put `ARG_UNUSED` in the declarations. It belongs on stub definitions only.

cf-reactor/cf-reactor.c
```diff
@@ -179,16 +186,162 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
...
+    if (!NO_FORK)
+    {
+        Log(LOG_LEVEL_VERBOSE, "Windows does not support starting processes in the background - starting in foreground");
+    }
+
+#else /* !__MINGW32__ */
+    pid_t existing_pid = ReadPID("cf-reactor.pid");
+    if ((existing_pid != -1) && (kill(existing_pid, 0) == 0))
+    {
+        Log(LOG_LEVEL_ERR, "Another instance of cf-reactor is already running, terminating");
```
craigcomstock: Should this message be rather "was already running, terminated" since you are calling kill() In the if expression? Maybe it would be good to mention what PID it was that you killed for debugging issues with unexpected running cf-reactor processes.

craigcomstock (approved):
Looks good.

cf-reactor/cf-reactor.c
```diff
@@ -179,16 +188,179 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
...
+    if (!NO_FORK)
+    {
+        ActAsDaemon();
+    }
+
+#endif /* !__MINGW32__ */
+
+    umask(077);
+    WritePID("cf-reactor.pid");
+    MakeSignalPipe();
```
larsewi, line 262: Nothing writes the signal pipe anylonger. The signal handlers need to do this in order to wake the daemon when it's blocked on select. The previous signal handler wrapper you had looked good. It was only missing the re-registration at the end.

```C
static void HandleReactorSignals(int signum)
{
    HandleSignalsForDaemon(signum);

    if (IsPendingTermination())
    {
        TERMINATE = 1;
    }

    /* HandleSignalsForDaemon() re-arms itself, which would take this wrapper
     * out of the chain. Re-arm ours afterwards, so that a second delivery
     * still reaches the code above. */
    signal(signum, HandleReactorSignals);
}
```
Point all six registrations at it

cf-reactor/cf-reactor.c
```diff
@@ -179,16 +188,179 @@ static GenericAgentConfig *CheckOpts(int argc, char **argv)
...
+        int ret = select(max_fd, &readfds, NULL, NULL, &timeout);
+
+        next_tick = last_tick + DEFAULT_POLL_INTERVAL_SECS;
+
+        if (ret < 0)
+        {
+            /*** error ***/
+            Log(LOG_LEVEL_ERR, "Failed to poll events");
+            continue;
+        }
```
larsewi, line 333: > EINTR  A signal was caught; see signal(7).

Is  not an error. Also you should investigate whether a transient error can cause this to become an infinite loop.

cf-reactor/cf-reactor.c
```diff
@@ -33,13 +33,22 @@
...
 /* Globals                                                                   */
 /*****************************************************************************/
 
 int NO_FORK = false;
 
+#define DEFAULT_POLL_INTERVAL_SECS 30
+// this is just an arbitrary number that has to be higher or equal to the number of fds used by reactor-plugin
+#define N_ALL_FDS 8
+
+volatile sig_atomic_t terminate = 0;
```
larsewi:
```suggestion
static volatile sig_atomic_t terminate = 0;
```
Also common to use uppercase

libpromises/prototypes3.h
```diff
@@ -85,7 +85,11 @@ ENTERPRISE_VOID_FUNC_0ARG_DECLARE(void, ReloadHAConfig);
 ENTERPRISE_VOID_FUNC_2ARG_DECLARE(void, Nova_ClassHistoryAddContextName, const StringSet *, list, const char *, context_name);
 ENTERPRISE_VOID_FUNC_2ARG_DECLARE(void, Nova_ClassHistoryEnable, StringSet **, list, bool, enable);
 
-ENTERPRISE_FUNC_1ARG_DECLARE(int, ReactorEnterpriseMain, bool, no_fork);
+ENTERPRISE_VOID_FUNC_0ARG_DECLARE(void, ReactorNovaTerminate);
```
larsewi: This one is never called any longer. Maybe it should be removed, now that we have the terminate parameter?

larsewi (approved):

larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14626/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14626/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14626/

PR #6345, opened 2026-09-02
Work with variable references as long as possible (commands promises)

nickanderson:
A commands promise now works with the *reference* to a secret-tagged variable
instead of its value: the promiser is expanded with `ExpandScalarKeepSecrets()`,
and `RepairExec()` resolves the references only in the string it hands to
`cf_popen()`. The lock name, `$(this.promiser)`, the promise banner,
`Executing ...`, `Would execute script ...` and the module output all keep
`$(password)`, so the value no longer reaches the log at inform level, nor the
lock database, nor a `--dry-run` report.

Ordinary variables are untouched — `ExpandScalar()` behaves exactly as before,
and a promiser with no references left skips the extra pass. `args` and
`arglist` are ordinary constraints and still expand eagerly.

`31_tickets/CFE-3298` covers all three shell types and pairs each absent value
with the reference that replaced it, so a build that expanded nothing fails it
too; disabling `PromiseTypeResolvesSecretsAtUse()` turns it red naming both
leaks. No enterprise or nova changes needed.

### DIFF
--- a/cf-agent/verify_exec.c
+++ b/cf-agent/verify_exec.c
@@ -43,6 +43,7 @@
 #include <scope.h>
 #include <ornaments.h>
 #include <eval_context.h>
+#include <expand.h>
 #include <retcode.h>
 #include <timeout.h>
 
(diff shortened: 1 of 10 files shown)


cf-bottom:
Thank you for submitting a pull request! Maybe @larsewi can review this?

PR #6346, opened 2026-09-03
ENT-14195: Event handling file deletion

victormlg:
The function `WatcherRegister` is what allows to add watchers to the reactor loop. Its definition looks like:

```C
void WatcherRegister(const char *key, EventType type, void *payload,
  Bundle *bundle, time_t interval);
```

Where `key` is the name of the promise, `type` is the attribute in the `when` body, `payload` is some data used specifically for the event type, for example in our case the name of the file watched. `bundle` is the bundle to be run, for now only a single one, and finally `interval` is how often do we want to check for the file.

Exemple:

```C
WatcherRegister("a.txt deleted", FILE_DELETED, 
  FileWatcherPayloadNew("/tmp/a.txt"), create_file_bundle, 1);
```

This maps to events promises:

```cf3
body when file_deleted(filename)
{
  file_deleted => "$(filename)";
}

bundle reactor A
{
  vars:
    "filename" string => "/tmp/a.txt";

  events:
    "a.txt deleted"
      when => file_deleted("$(filename)"),
      then => create_file("$(filename)");
}
```

**Implementation**

We have a thread iterating through the Watcher list. When an event is registered, it pushes the key on a queue and signals the daemon that there is something on the queue. Then, inside `ReactorContextHandleEvents`, the bundle that should be run is looked up from the key in a hashmap.

### DIFF
--- a/cf-reactor/cf-reactor.h
+++ b/cf-reactor/cf-reactor.h
@@ -1,0 +1,60 @@
+/*
+  Copyright 2026 Northern.tech AS
+
+  This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+  This program is free software; you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by the
+  Free Software Foundation; version 3.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#ifndef CFENGINE_REACTOR_H
+#define CFENGINE_REACTOR_H
+
+#include <platform.h>
+
+/**
+ * @brief Shared state for the cf-reactor daemon's single select(2) loop.
+ *
+ * `all_fds` is a single flat array shared by every event source the daemon
+ * watches. Nova's fds always occupy the first `num_nova_fds` slots (Nova
+ * owns that sub-range and is the only thing allowed to populate it); any
+ * other event source (currently just the watcher subsystem, see watcher.h)
+ * appends its own fd(s) after that, and `num_fds` tracks the total number
+ * of slots in use. Adding a new event source means:
+ *
+ *   1. Have it report how many fds it needs, and add that to the capacity
+ *      computed in ReactorContextInitialize() (reactor_context.c).
+ *   2. Give it an Initialize(fds, max_size, num_fds)/HandleEvents(readfds)/
+ *      Finalize(void) triplet shaped like ReactorNova*() or EventWatcher*(),
+ *      and wire the three calls into reactor_context.c next to the existing
+ *      ones.
+ *
+ * No other file needs to know how many event sources exist or in what
+ * order their fds appear.
+ */
+typedef struct ReactorContext
+{
+    int *all_fds;
+    size_t all_fds_capacity;
+    size_t num_nova_fds;
+    size_t num_fds;
+
+    fd_set readfds;
+} ReactorContext;
+
+#endif
--- a/cf-reactor/file_watcher.c
+++ b/cf-reactor/file_watcher.c
@@ -1,0 +1,88 @@
+/*
+  Copyright 2026 Northern.tech AS
+
+  This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+  This program is free software; you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by the
+  Free Software Foundation; version 3.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#include <file_watcher.h>
+#include <watcher.h>
+#include <logging.h>
+#include <alloc.h>
+#include <sys/stat.h>
+#include <errno.h>
+
+// =========== code for EVENT_FILE_DELETED event type ===========
+
+typedef struct
+{
+    char *path;
+    bool existed_last_check;
+} FileWatcherPayload;
+
+/* Only ENOENT/ENOTDIR mean the path genuinely doesn't exist. Any other
+ * stat() failure (e.g. EACCES, ESTALE) is a transient/permission error, not
+ * a deletion, so treat the file as still present rather than misreporting
+ * it as deleted. */
+static bool FileExists(const char *path)
+{
+    struct stat sb;
+    if (stat(path, &sb) == 0)
+    {
+        return true;
+    }
+
+    if (errno == ENOENT || errno == ENOTDIR)
+    {
+        return false;
+    }
+
+    Log(LOG_LEVEL_ERR, "Unable to stat '%s' while checking for file deletion: %s", path, GetErrorStr());
+    return true;
+}
+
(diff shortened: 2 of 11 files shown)


craigcomstock (approved):

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14636/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14636/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14636/

olehermanse (changes requested):
I think this PR changes too many things in one PR and one commit; I would split it up like this:

## 1. README

Add `cf-reactor/README.md` with an explanation of how this should work. This could explain the different sections I've listed below.

## 2. Moving stuff around

Looks like you are moving some stuff around for having the "Nova" functions in the right place. This could be its own commit.

## 3. Empty main loop

You could set up an empty infinite main loop, with just a sleep statement and `TODO` comments for what we're going to implement below. At this point, cf-reactor only does the nova reactor-plugin stuff, and runs forever (sleeping most of the time). This is the same as we expect in the future as well (as long as user didn't write any new event promises).

## 4. Tracking spec

Policy needs to be converted to a data structure of what we want to track.
Events promises may have conditions (class guards), so those should be included.
Policy may also be changed, so how to reload should be considered, or there should be a TODO comment to implement that later.

## 5. Polling

Implement the part where you actually check if the file exists, and record that in a current state, and do this again and again. At this point there should be TODO comments for actually creating events. Also some TODO comments about how this would be implemented for the non-polling way might be appropriate.

## 6. Events

Implement the part where you generate events from changes in the current state between polling. I would even go so far as to just creating the events and sending them to a queue / handler that does nothing, just deletes them with a TODO comment to actually handle the events.

## 7. Actually running the bundles

Implement the last step of actually running the bundle.

## Why / how

I would start by opening a PR with only the README (1), to get that reviewed as soon as possible, and to be able to incorporate any feedback you get into future commits / PRs.

After that / while you wait for review, I would start the second PR with commit 2 (maybe 3 and maybe 4 as well). Finally, you can keep working on a third PR with the remaining commits. When PR 1 and 2 are merged, you can consider whether PR 3 is ready to be reviewed and merged, or if you should split it up further.

I think this would make it easier for us to review, and easier for you to incorporate our feedback on the early steps into the later steps.

cf-reactor/cf-reactor.h
```diff
@@ -0,0 +1,60 @@
...
+#ifndef CFENGINE_REACTOR_H
+#define CFENGINE_REACTOR_H
+
+#include <platform.h>
+
+/**
+ * @brief Shared state for the cf-reactor daemon's single select(2) loop.
+ *
+ * `all_fds` is a single flat array shared by every event source the daemon
+ * watches. Nova's fds always occupy the first `num_nova_fds` slots (Nova
```
olehermanse, line 34: Instead of referring to Nova, I think the thing you should refer to is the reactor-plugin. At least to me that is more clear.

cf-reactor/file_watcher.c
```diff
@@ -0,0 +1,88 @@
...
+#include <sys/stat.h>
+#include <errno.h>
+
+// =========== code for EVENT_FILE_DELETED event type ===========
+
+typedef struct
+{
+    char *path;
+    bool existed_last_check;
+} FileWatcherPayload;
```
olehermanse, line 38: To me, `Watcher` and `Payload` are a bit vague / ambiguous.

I would suggest simply:

```suggestion
typedef struct
{
    char *path;
    bool exists;
} FileState;
```

cf-reactor/cf-reactor.h
```diff
@@ -0,0 +1,60 @@
...
+ * order their fds appear.
+ */
+typedef struct ReactorContext
+{
+    int *all_fds;
+    size_t all_fds_capacity;
+    size_t num_nova_fds;
+    size_t num_fds;
+
+    fd_set readfds;
```
olehermanse, line 57: Clearly indicate which fields are related to reactor-plugin and which are not.

cf-reactor/watcher.h
```diff
@@ -0,0 +1,56 @@
...
+typedef enum
+{
+  EVENT_FILE_DELETED,
+} EventType;
+
+typedef bool (*WatcherCheckFn)(void *payload);
+typedef void (*WatcherPayloadDestroyFn)(void *payload);
+
+void WatcherRegistryInitialize(void);
+void WatcherRegistryFinalize(void);
```
olehermanse, line 40: IMO it is not completely obvious what `Watcher` means.

Is this polling only, or does it include future implementations which are not based on polling?

Please add a comment to the top of this file explaining this.

cf-reactor/watcher.h
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * @brief Register a specific watcher instance.
+ * 
+ * @param key the events promise identifier
+ * @param type the type of watcher, defined in when bodies
+ * @param payload the data used for by the watcher, depending on the type
+ * @param bundle the bundle to run on event
+ * @param interval interval between runs
+ */
+void WatcherRegister(const char *key, EventType type, void *payload, Bundle *bundle, time_t interval);
```
olehermanse, line 51: Looks like interval does not need to be a parameter (yet)?

If watcher is supposed to cover non-polling implementations, it looks out of place.

If it is always a hard coded value, it makes more sense to just use that macro name.

If interval is configurable, or varies, maybe we should mention how / where. I.e. you can set it in policy, you can set it in env var, it varies based on event type, etc.

cf-reactor/reactor_context.c
```diff
@@ -0,0 +1,130 @@
...
+#include <watcher.h>
+#include <file_watcher.h>
+#include <alloc.h>
+
+bool ReactorContextInitialize(ReactorContext *ctx)
+{
+    assert(ctx != NULL);
+
+    WatcherRegistryInitialize();
+    // TODO: Register watchers with `WatcherRegister()`
```
olehermanse, line 37: What does this mean? Looking at the PR and commit title and description it's not very clear to me what is implemented and what is missing and why this is still a TODO.

cf-reactor/watcher.h
victormlg: interval is useful for debugging and it could be added in the future, so I think it makes sense to have it.

cf-reactor/reactor_context.c
victormlg: This just indicates where one should manually register events promise "watchers", and will be removed when we will create them from parsing the policy

cf-reactor/cf-reactor.h
```diff
@@ -0,0 +1,60 @@
...
+ *      computed in ReactorContextInitialize() (reactor_context.c).
+ *   2. Give it an Initialize(fds, max_size, num_fds)/HandleEvents(readfds)/
+ *      Finalize(void) triplet shaped like ReactorNova*() or EventWatcher*(),
+ *      and wire the three calls into reactor_context.c next to the existing
+ *      ones.
+ *
+ * No other file needs to know how many event sources exist or in what
+ * order their fds appear.
+ */
+typedef struct ReactorContext
```
larsewi: Would it make sense to have two `Seq *`, one for reactor-plugin FDs, and one for the other FDs? Or one `Seq *` where FDs are wrapped in a struct with flags determining who owns them?

cf-reactor/file_watcher.c
```diff
@@ -0,0 +1,88 @@
...
+
+/* Only ENOENT/ENOTDIR mean the path genuinely doesn't exist. Any other
+ * stat() failure (e.g. EACCES, ESTALE) is a transient/permission error, not
+ * a deletion, so treat the file as still present rather than misreporting
+ * it as deleted. */
+static bool FileExists(const char *path)
```
larsewi: Should there be a flag `follow_symlink`? Also, don't we already have this function implemented somewhere?

cf-reactor/file_watcher.h
```diff
@@ -0,0 +1,35 @@
...
+#define CFENGINE_FILE_WATCHER_H
+
+#include <platform.h>
+
+
+void *FileWatcherPayloadNew(const char *path);
+bool CheckFileExists(void *payload);
+void DestroyFileWatcherPayload(void *payload);
+
+#endif
```
larsewi:
```suggestion
#endif /* CFENGINE_FILE_WATCHER_H */
```

cf-reactor/reactor_context.c
```diff
@@ -0,0 +1,130 @@
...
+
+    FD_ZERO(&ctx->readfds);
+    int signal_pipe = GetSignalPipe();
+    FD_SET(signal_pipe, &ctx->readfds);
+
+    int max_fd = signal_pipe;
+    for (size_t i = 0; i < ctx->num_fds; i++)
+    {
+        FD_SET(ctx->all_fds[i], &ctx->readfds);
+        max_fd = MAX(ctx->all_fds[i], max_fd);
```
larsewi: Max, makes it sound like a limit. But it appears to be the number of FDs.
```suggestion
        num_fd = MAX(ctx->all_fds[i], max_fd);
```

victormlg:
Divided the PR into:

- https://github.com/cfengine/core/pull/6353
- https://github.com/cfengine/core/pull/6354
- https://github.com/cfengine/core/pull/6355

PR #6352, opened 2026-09-07
ENT-14195 0. README for cf-reactor event handling

victormlg:


cf-reactor/README.md
```diff
@@ -0,0 +1,55 @@
...
+- `ReactorContextInitialize()`: initializes the reactor-plugin and event-driven code. Wraps `ReactorNovaInitialize()`
+- `ReactorContextSetupFileDescriptors()`: populates readfds with the file descriptors to monitor, prior to the select() call.
+- `ReactorContextHandleEvents()`: iterates over the file descriptors and dispatches the appropriate action based on which ones were signaled as ready. Wraps `ReactorNovaHandleTimeout` and `ReactorNovaHandleEvents()`.
+- `ReactorContextFinalize()`: releases the daemon's associated resources. Wraps `ReactorNovaFinalize()`.
+
+# Tracking spec & Events
+
+In order to track all the events promises, we use two datastructures: a global list of `"Watcher"`, which is a struct associated with an event type and the promise name (also called `key`) and a global hashmap mapping this `key` to a `bundle` which is parsed from the policy.
+
+On an agent run, cf-agent makes cf-reactor read the policy, and rebuilds the list of watchers and the hashmap using the single function `WatcherRegister(key, event_type, payload, bundle, interval)`. Each events promise is associated with an event type, which is defined in `when` bodies:
```
olehermanse: Why should this happen on an agent run?

cf-reactor/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```cf3
+body when file_deleted(filename)
+{
+    file_deleted => "$(filename)";
+}
+```
+
+Every event type is must have defined:
```
olehermanse:
```suggestion
Every event type must have defined:
```

cf-reactor/README.md
```diff
@@ -0,0 +1,55 @@
...
+- A check function (called `check_fn` in `Watcher`): This is a function defined specifically for the event that checks if the conditions holds. For example, in case of file deletion, we check if the file doesn't exist anymore compare to the last time we checked. If yes, then it returns `true`.
+- A state (called `"payload"`): This is a struct whose interpretation depends on the event type (thus being declared as `void *`). We typically need some state that we compare between each event-check. In the case of file deletion, we need to know the name of the file we are watching, and whether the file existed last time we checked.
```
olehermanse: The name state is already quite overloaded - this feature is about current state and desired state and change in state. Maybe we should just use payload as the name here?

cf-reactor/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```cf3
+body when file_deleted(filename)
+{
+    file_deleted => "$(filename)";
+}
+```
+
+Every event type is must have defined:
+- A check function (called `check_fn` in `Watcher`): This is a function defined specifically for the event that checks if the conditions holds. For example, in case of file deletion, we check if the file doesn't exist anymore compare to the last time we checked. If yes, then it returns `true`.
```
olehermanse, line 65: IMO this part is a bit unclear;

Does check_fn only return true once when the state changes from exists to not exists? Or will it keep returning true as long as the file does not exist? Also might be worth mentioning how this is intended to work for the non-polling case (i.e. inotify where we just get an event, no check function / polling happening.

olehermanse (approved):
This README is a good starting point. I would prefer if it focused a bit more on the high level architecture and terminology, instead of going straight into structs and specific function names.

cf-reactor/README.md
```diff
@@ -0,0 +1,55 @@
+
+# Moving stuff around
+
+Rather than exposing the raw file-descriptor bookkeeping required for `select(2)`, we introduce a unified interface that serves both the reactor-plugin and event-driven code paths. This is achieved by encapsulating all relevant state in a context struct, `ReactorContext`:
+
+```C
+typedef struct ReactorContext
```
larsewi: Would it make sense to have two `Seq *`, one for reactor-plugin FDs, and one for the other FDs? Or perhaps one `Seq *` where FDs are wrapped in a struct with flags determining who owns them?

cf-reactor/README.md
```diff
@@ -0,0 +1,82 @@
...
+On dispatch, the bundle corresponding to an event is run, however it doesn't do a full agent run.
+
+## Implementation details
+
+### Moving stuff around
+
+Rather than exposing the raw file-descriptor bookkeeping required for `select(2)`, we introduce a unified interface that serves both the reactor-plugin and event-driven code paths. This is achieved by encapsulating all relevant state in a context struct, `ReactorContext`:
+
+```C
+typedef struct ReactorContext
```
larsewi, line 34: Would it make sense to have two `Seq *`, one for reactor-plugin FDs, and one for the other FDs? Or perhaps one `Seq *` where FDs are wrapped in a struct with flags determining who owns them?

larsewi (approved):

PR #6353, opened 2026-09-07
ENT-14195 1. Refactored cf-reactor event handling with one unified interface

victormlg:
Rather than exposing the raw file-descriptor bookkeeping required for `select(2)`, we introduce a unified interface that serves both the reactor-plugin and event-driven code paths. This is achieved by encapsulating all relevant state in a context struct, `ReactorContext`:

```C
typedef struct ReactorContext
{
    Seq *fds // array of ReactorFd, which holds the fd and some metadata
    fd_set readfds;
    size_t max_nova_fds;
} ReactorContext;
```

- `ReactorContextInitialize()`: initializes the reactor-plugin and event-driven code. Wraps `ReactorNovaInitialize()`
- `ReactorContextSetupFileDescriptors()`: populates readfds with the file descriptors to monitor, prior to the select() call.
- `ReactorContextHandleEvents()`: iterates over the file descriptors and dispatches the appropriate action based on which ones were signaled as ready. Wraps `ReactorNovaHandleTimeout` and `ReactorNovaHandleEvents()`.
- `ReactorContextFinalize()`: releases the daemon's associated resources. Wraps `ReactorNovaFinalize()`.


cf-reactor/cf-reactor.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <platform.h>
+
+/**
+ * @brief Shared state for the cf-reactor daemon's single select(2) loop.
+ *
+ * `all_fds` is a single flat array shared by every event source the daemon
+ * watches. Nova's fds always occupy the first `num_nova_fds` slots (Nova
+ * owns that sub-range and is the only thing allowed to populate it); any
+ * other event source (currently just the watcher subsystem, see watcher.h)
```
larsewi, line 36: Did we not recently remove the watcher code you are referencing here?

cf-reactor/cf-reactor.h
```diff
@@ -0,0 +1,60 @@
...
+ *      computed in ReactorContextInitialize() (reactor_context.c).
+ *   2. Give it an Initialize(fds, max_size, num_fds)/HandleEvents(readfds)/
+ *      Finalize(void) triplet shaped like ReactorNova*() or EventWatcher*(),
+ *      and wire the three calls into reactor_context.c next to the existing
+ *      ones.
+ *
+ * No other file needs to know how many event sources exist or in what
+ * order their fds appear.
+ */
+typedef struct ReactorContext
```
larsewi, line 50: Seems like this belongs in `reactor_context.h`. What is the reason for having it here?

larsewi (commented):
What if we stored FDs in a dynamic list instead (e.g., `Seq`)? Furthermore, maybe the FD itself could be wrapped in a struct with flags to tell what type of FD it is, and whether or not it is owned by any external plugin.

cf-reactor/cf-reactor.h
victormlg: The comment here is quite confusing, but the "watcher" and "cf-watchd" are too different things in my implementation, and the watcher is how I called the logic that registers instances that the daemon will react on, on event

cf-reactor/reactor_context.c
```diff
@@ -0,0 +1,171 @@
...
+*/
+
+#include <reactor_context.h>
+#include <prototypes3.h>        /* ReactorNova*() */
+#include <signals.h>            /* GetSignalPipe() */
+#include <alloc.h>
+
+#define INIT_FD_COUNT 8
+
+static size_t max_nova_fds = 0;
```
larsewi: Should this be `num_nova_fds`? Should probably live in the struct rather than file static.

cf-reactor/reactor_context.c
```diff
@@ -0,0 +1,171 @@
...
+    assert(ctx != NULL);
+
+    if (NovaHasTimedOut(ctx))
+    {
+        ReactorNovaHandleTimeout(next_tick);
+    }
+    else
+    {
+        int *nova_fds = GetNovaFds(ctx->fds);
+        ReactorNovaHandleEvents(&ctx->readfds, nova_fds, next_tick);
```
larsewi, line 160: If i remember correctly, the second argument is an in/out parameter. I.e., `ReactorNovaHandleEvents` will try to reconnect bad connections. Right now we just throw away `nova_fds` on the next line. What is the plan here?

larsewi (commented):
Please make sure unit tests pass before we can merge

olehermanse (approved):

olehermanse:
macos unit test failure seems unrelated.

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14655/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14655/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14655/

larsewi (approved):
🚀

cf-reactor/reactor_context.c
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        for (size_t i = 0; i < num_nova_fds; i++)
+        {
+            SeqAppend(ctx->fds, ReactorFdNew(nova_fds[i], REACTOR_FD_NOVA));
+        }
+        free(nova_fds);
+    }
+
+    // TODO: initialize other event sources here.
```
craigcomstock, line 67: Maybe change this from a TODO to a comment describing that this is the location that "other event sources" would be initialized. Maybe include some description of WHAT an "event source" is exactly so the comment is fully understandable.

cf-reactor/reactor_context.c
```diff
@@ -0,0 +1,189 @@
...
+    
+    return true;
+}
+
+int ReactorContextSetupFileDescriptors(ReactorContext *ctx)
```
craigcomstock, line 72: I would suggest not using `ctx` since that is used often in core for EvalContext. Maybe reactor_context.

cf-reactor/reactor_context.c
```diff
@@ -0,0 +1,189 @@
...
+        if (rfd->type != REACTOR_FD_NOVA)
+        {
+            continue;
+        }
+        // Since we came so far, this should be always true
+        assert(num_nova_fds < ctx->max_nova_fds);
+        nova_fds[num_nova_fds++] = rfd->fd;
+    }
+
+    return nova_fds;
```
craigcomstock, line 129: nova_fds is allocated to ctx->max_nova_fds but only initializes num_nova_fds. ReactorNovaHandleEvents() is not given any clue as to how many fds might have been initialized so uninitialized memory could be read in that function? Maybe the check for FD_ISSET() handles uninitialized fds? -craig[claude]

PR #6354, opened 2026-09-07
ENT-14195 2. cf-reactor data structure and polling

victormlg:
# Spec

In order to track all the events promises, we use two datastructures: a global list of `"Watcher"`, which is a struct associated with an event type and the promise name (also called `key`) and a global hashmap mapping this `key` to a `bundle` which is parsed from the policy.

cf-reactor reads the policy periodically, and when it does, rebuilds the list of watchers and the hashmap using the single function `WatcherRegister(key, event_type, state, bundle, interval)`. Each events promise is associated with an event type, which is defined in `when` bodies:

```cf3
body when file_deleted(filename)
{
    file_deleted => "$(filename)";
}
```

Every event type must have defined:
- A check function (called `check_callback` in `Watcher`): This is a function defined specifically for the event that checks if the conditions holds. For example, in case of file deletion, we check if the file doesn't exist anymore compare to the last time we checked. If yes, then it returns `true`.
- A `state`: This is a struct whose interpretation depends on the event type (thus being declared as `void *`). We typically need some state that we compare between each event-check. In the case of file deletion, we need to know the name of the file we are watching, and whether the file existed last time we checked.
- A state destroying function (called `destroy_state`): This is simply a function to free the state associated with the event type.

Also, we need a function that will create the state. That's what `FileWatcherPayloadNew()` does.

So each event type we add in the future just need to have these four things defined, and we need to create the `Watcher` object with the right functions inside `WatcherRegister()` and also call the right `"...PayloadNew()"` function.

# Polling

`ReactorContextInitialize()` sets up all the necessary data structures for polling, and then starts `WatcherThreadMain`, which polls for events as follows:

- It iterates through each watcher in the global list of watchers.
- If the elapsed time exceeds the watcher's `interval`, it runs `check_callback` to determine whether an event has been triggered.
- If an event was triggered, it pushes the watcher's `key` (the promise name) onto a thread-safe queue, then signals the file descriptor via `WakeupChannelNotify`, which `select(2)` will pick up on its next iteration. It then goes back to sleep.

In parallel, `EventWatcherHandleEvents`, called from within `ReactorContextHandleEvents`, reads from the file descriptor with `WakeupChannelReadFd()` once notified that an event has occurred, and pops the thread-safe queue until it's empty. Each key popped from the queue is looked up in the global hashmap to retrieve the corresponding bundle, which `cf-reactor` then runs (in another thread or subprocess)

### DIFF
--- a/cf-reactor/reactor_context.c
+++ b/cf-reactor/reactor_context.c
@@ -25,13 +25,16 @@
 #include <reactor_context.h>
 #include <prototypes3.h>        /* ReactorNova*() */
 #include <signals.h>            /* GetSignalPipe() */
+#include <watcher.h>
+#include <file_watcher.h>
 #include <alloc.h>
 
 bool ReactorContextInitialize(ReactorContext *ctx)
 {
     assert(ctx != NULL);
 
-    // TODO: initialize watcher registry
+    WatcherRegistryInitialize();
+    // TODO: Register watchers with `WatcherRegister()`
 
     size_t max_nova_fds = ReactorNovaMaxFds();
     ctx->all_fds_capacity = max_nova_fds + 1;
@@ -48,7 +51,14 @@ bool ReactorContextInitialize(ReactorContext *ctx)
     ctx->num_nova_fds = num_nova_fds;
     ctx->num_fds = num_nova_fds;
 
-    // TODO: intialize event watcher
+    if (!EventWatcherInitialize(ctx->all_fds, ctx->all_fds_capacity, &ctx->num_fds))
+    {
+        ReactorNovaFinalize();
+        WatcherRegistryFinalize();
+        free(ctx->all_fds);
+        ctx->all_fds = NULL;
+        return false;
+    }
 
     return true;
 }
@@ -106,14 +116,14 @@ void ReactorContextHandleEvents(ReactorContext *ctx, time_t *next_tick)
         while (recv(GetSignalPipe(), &buf, 1, 0) > 0) { /* drain */ }
     }
 
-    // TODO: handle watcher events
+    EventWatcherHandleEvents(&ctx->readfds);
 }
 
 void ReactorContextFinalize(ReactorContext *ctx)
 {
     assert(ctx != NULL);
 
-    // TODO: finalize watcher events
+    EventWatcherFinalize();
     ReactorNovaFinalize();
     free(ctx->all_fds);
     ctx->all_fds = NULL;
--- a/cf-reactor/reactor_context.c
+++ b/cf-reactor/reactor_context.c
@@ -1,0 +1,120 @@
+/*
+  Copyright 2026 Northern.tech AS
+
+  This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+  This program is free software; you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by the
+  Free Software Foundation; version 3.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#include <reactor_context.h>
+#include <prototypes3.h>        /* ReactorNova*() */
+#include <signals.h>            /* GetSignalPipe() */
+#include <alloc.h>
+
+bool ReactorContextInitialize(ReactorContext *ctx)
+{
+    assert(ctx != NULL);
+
+    // TODO: initialize watcher registry
+
+    size_t max_nova_fds = ReactorNovaMaxFds();
+    ctx->all_fds_capacity = max_nova_fds + 1;
+    ctx->all_fds = xmalloc(ctx->all_fds_capacity * sizeof(int));
+
+    size_t num_nova_fds = 0;
+    if (!ReactorNovaInitialize(ctx->all_fds, max_nova_fds, &num_nova_fds))
+    {
+        WatcherRegistryFinalize();
+        free(ctx->all_fds);
+        ctx->all_fds = NULL;
+        return false;
+    }
+    ctx->num_nova_fds = num_nova_fds;
+    ctx->num_fds = num_nova_fds;
+
+    // TODO: intialize event watcher
+
+    return true;
+}
+
+int ReactorContextSetupFileDescriptors(ReactorContext *ctx)
+{
+    assert(ctx != NULL);
+
+    FD_ZERO(&ctx->readfds);
+    int signal_pipe = GetSignalPipe();
+    FD_SET(signal_pipe, &ctx->readfds);
+
+    int max_fd = signal_pipe;
+    for (size_t i = 0; i < ctx->num_fds; i++)
+    {
+        FD_SET(ctx->all_fds[i], &ctx->readfds);
+        max_fd = MAX(ctx->all_fds[i], max_fd);
+    }
+    return max_fd + 1;
+}
+
+static bool NovaHasTimedOut(ReactorContext *ctx)
+{
+    assert(ctx != NULL);
+    for (size_t i = 0; i < ctx->num_nova_fds; i++)
+    {
+        if (FD_ISSET(ctx->all_fds[i], &ctx->readfds))
+        {
+            return false;
+        }
+    }
+    return true;
+}
+
+void ReactorContextHandleEvents(ReactorContext *ctx, time_t *next_tick)
+{
+    assert(ctx != NULL);
+
+    if (NovaHasTimedOut(ctx))
+    {
+        ReactorNovaHandleTimeout(next_tick);
+    }
+    else
+    {
+        ReactorNovaHandleEvents(&ctx->readfds, ctx->all_fds, next_tick);
+    }
+
+    /* The signal pipe is always in the watched set so we wake up
+    * promptly on a pending signal, but (per its own contract in
+    * signals.c) it must be drained or it stays "ready" forever, which
+    * would stop select() from ever blocking again. */
+    if (FD_ISSET(GetSignalPipe(), &ctx->readfds))
+    {
+        unsigned char buf;
+        while (recv(GetSignalPipe(), &buf, 1, 0) > 0) { /* drain */ }
+    }
+
+    // TODO: handle watcher events
+}
+
+void ReactorContextFinalize(ReactorContext *ctx)
+{
+    assert(ctx != NULL);
+
+    // TODO: finalize watcher events
+    ReactorNovaFinalize();
+    free(ctx->all_fds);
+    ctx->all_fds = NULL;
+}
--- a/cf-reactor/wakeup_channel.c
+++ b/cf-reactor/wakeup_channel.c
@@ -1,0 +1,106 @@
+/*
+  Copyright 2026 Northern.tech AS
+
+  This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+  This program is free software; you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by the
+  Free Software Foundation; version 3.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#include <wakeup_channel.h>
+#include <logging.h>
+#include <prototypes3.h>       /* cf_closesocket() */
+
+bool WakeupChannelOpen(WakeupChannel *channel)
+{
+    assert(channel != NULL);
+
+    channel->fds[0] = -1;
+    channel->fds[1] = -1;
+
+    /* Windows' select() only works on sockets, and a plain pipe(2) isn't
+     * one, so this always goes through socketpair() -- which works just as
+     * well on POSIX -- rather than keeping two code paths in sync. This
+     * mirrors MakeSignalPipe() in libpromises/signals.c. */
+    if (socketpair(AF_UNIX, SOCK_STREAM, 0, channel->fds) != 0)
+    {
+        Log(LOG_LEVEL_ERR, "Could not create wakeup channel (socketpair: '%s')", GetErrorStr());
+        return false;
+    }
+
+    for (int i = 0; i < 2; i++)
+    {
+#ifdef __MINGW32__
+        u_long enable = 1;
+        int ret = ioctlsocket(channel->fds[i], FIONBIO, &enable);
+#define CNTLNAME "ioctlsocket"
+#else
+        int ret = fcntl(channel->fds[i], F_SETFL, O_NONBLOCK);
+#define CNTLNAME "fcntl"
+#endif
+        if (ret != 0)
+        {
+            Log(LOG_LEVEL_ERR, "Could not set wakeup channel to non-blocking (" CNTLNAME ": '%s')",
+                GetErrorStr());
+            WakeupChannelClose(channel);
+            return false;
+        }
+#undef CNTLNAME
+    }
+
+    return true;
+}
+
+int WakeupChannelReadFd(const WakeupChannel *channel)
+{
+    assert(channel != NULL);
+    return channel->fds[0];
+}
+
+void WakeupChannelNotify(const WakeupChannel *channel)
+{
+    assert(channel != NULL);
+
+    /* One byte is enough to wake the reader up; if the channel happens to
+     * already be full, the reader is already guaranteed to wake up because
+     * of what's queued, so a transient EAGAIN/EWOULDBLOCK here is fine. */
+    unsigned char byte = 1;
+    send(channel->fds[1], (const char *) &byte, sizeof(byte), 0);
+}
+
+void WakeupChannelDrain(const WakeupChannel *channel)
+{
+    assert(channel != NULL);
+
+    unsigned char buf;
+    while (recv(channel->fds[0], (char *) &buf, sizeof(buf), 0) > 0) { /* drain */ }
+}
+
+void WakeupChannelClose(WakeupChannel *channel)
+{
+    assert(channel != NULL);
+
+    for (int i = 0; i < 2; i++)
+    {
+        if (channel->fds[i] != -1)
+        {
+            cf_closesocket(channel->fds[i]);
+            channel->fds[i] = -1;
+        }
+    }
+}
--- a/cf-reactor/wakeup_channel.h
+++ b/cf-reactor/wakeup_channel.h
@@ -1,0 +1,49 @@
+/*
+  Copyright 2026 Northern.tech AS
+
+  This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+  This program is free software; you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by the
+  Free Software Foundation; version 3.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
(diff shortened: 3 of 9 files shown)


cf-bottom:
Thank you for submitting a pull request! Maybe @craigcomstock can review this?

olehermanse:
@victormlg please rebase this one :)

cf-reactor/reactor_context.c
```diff
@@ -37,106 +38,129 @@ static ReactorFd *ReactorFdNew(int fd, ReactorFdType type)
     return rfd;
 }
 
-bool ReactorContextInitialize(ReactorContext *ctx)
+bool ReactorContextInitialize(ReactorContext *reactor_context)
```
larsewi, line 41: This refactor should be on it's own commit

cf-reactor/wakeup_channel.h
```diff
@@ -0,0 +1,49 @@
...
+    int fds[2]; /* [0] = read end, add to the select() fd_set; [1] = write end */
+} WakeupChannel;
+
+bool WakeupChannelOpen(WakeupChannel *channel);
+int WakeupChannelReadFd(const WakeupChannel *channel);
+void WakeupChannelNotify(const WakeupChannel *channel);
+void WakeupChannelDrain(const WakeupChannel *channel);
+void WakeupChannelClose(WakeupChannel *channel);
+
+#endif
```
larsewi:
```suggestion
#endif /* CFENGINE_WAKEUP_CHANNEL_H */
```

cf-reactor/watcher.c
```diff
@@ -0,0 +1,251 @@
...
+/* Not currently caller-configurable -- WatcherRegister() doesn't take an
+ * interval parameter -- so every watcher polls at this rate for now. */
+#define DEFAULT_WATCHER_POLL_INTERVAL_SECS 10
+
+typedef struct
+{
+    char *key;
+    WatcherCheckFn check_fn;       /* resolved from `type` at WatcherRegister() time */
+    WatcherPayloadDestroyFn destroy_payload;
+    void *payload;
```
larsewi: A more common generic name for this is `userdata`. `state` would also be nice, since that is what it holds. The term `payload` is confusing in this context.

cf-reactor/watcher.c
```diff
@@ -0,0 +1,251 @@
...
+#define MAX_WATCHER_THREAD_SLEEP_SECS 1
+
+/* Not currently caller-configurable -- WatcherRegister() doesn't take an
+ * interval parameter -- so every watcher polls at this rate for now. */
+#define DEFAULT_WATCHER_POLL_INTERVAL_SECS 10
+
+typedef struct
+{
+    char *key;
+    WatcherCheckFn check_fn;       /* resolved from `type` at WatcherRegister() time */
```
larsewi: The term `callback` is often used for this
```suggestion
    WatcherCheckFn check_callback_fn;       /* resolved from `type` at WatcherRegister() time */
```

larsewi (commented):
Did not review the whole thing yet. Here is what I have so far

olehermanse:
> On an agent run, cf-agent makes cf-reactor read the policy, and rebuilds the list of watchers and the hashmap using the single function

I commented on this in the README PR as well - why should cf-agent be involved? cf-reactor can do this on its own.

victormlg:
> > On an agent run, cf-agent makes cf-reactor read the policy, and rebuilds the list of watchers and the hashmap using the single function
> 
> I commented on this in the README PR as well - why should cf-agent be involved? cf-reactor can do this on its own.

I forgot to update the description in the PR. The README doesn't mention this anymore.

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14671/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14671/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14671/

cf-reactor/reactor_context.c
```diff
@@ -37,106 +38,129 @@ static ReactorFd *ReactorFdNew(int fd, ReactorFdType type)
...
         {
-            SeqAppend(ctx->fds, ReactorFdNew(nova_fds[i], REACTOR_FD_NOVA));
+            SeqAppend(reactor_context->fds, ReactorFdNew(nova_fds[i], REACTOR_FD_NOVA));
         }
         free(nova_fds);
     }
 
-    // TODO: initialize other event sources here.
-    
+    // Initialize watcher fd
```
craigcomstock, line 75:
```suggestion
    // Initialize event watcher fd
```

cf-reactor/reactor_context.c
```diff
@@ -37,106 +38,129 @@ static ReactorFd *ReactorFdNew(int fd, ReactorFdType type)
...
+            SeqDestroy(reactor_context->fds);
+            reactor_context->fds = NULL;
             return false;
         }
 
+        // num_nova_fds can never end up less than max_nova_fds here: Nova
+        // asserts internally that it always reports back the same fd count
+        // it advertises as its max (see the assert on poll_fd_idx in
+        // SetupEventProcessing(), nova/reactor-plugin/cf-reactor.c),
+        // so this loop always appends exactly max_nova_fds entries.
```
craigcomstock, line 67: Ok. I agree with this. SetupEventProcessing() uses all macros to determine the number of "things" involved and so num_nova_fds won't ever be "wrong". Ack.

cf-reactor/reactor_context.c
```diff
@@ -145,21 +169,39 @@ static void SetNovaFds(ReactorContext *ctx, const int *nova_fds)
...
+
+        if (rfd->type == REACTOR_FD_WATCHER)
+        {
+            return rfd->fd;
+        }
+    }
+    /**
+     * We should never reach this point: ReactorContextInitialize would fail if no watcher fd was set up.
+     */
+    ProgrammingError("Couldn't find the watcher fd");
```
craigcomstock, line 187: maybe change this to say "ReactorContextInitialize() was not called" ? So that the future programming will no why the programming error happened most likely? Then you could remove the comment above this as well possibly if you included that in the message instead.

cf-reactor/README.md
```diff
@@ -6,7 +6,7 @@
...
 
-Events promises are parsed into watcher instances, each holding: the event `key` from the events promise, an event `type` (e.g. file deletion), a `payload` with whatever state the check needs (e.g. the name of the file being watched), and a `check function` that detects the event by comparing the current state to the state recorded at the last check (e.g. whether the file still exists). It returns `true` only on the specific transition being watched for, so it fires once per change and does not stay `true` afterwards.
+Events promises are parsed into watcher instances, each holding: the event `key` from the events promise, an event `type` (e.g. file deletion), a `state` with whatever state the check needs (e.g. the name of the file being watched), and a `check function` that detects the event by comparing the current state to the state recorded at the last check (e.g. whether the file still exists). It returns `true` only on the specific transition being watched for, so it fires once per change and does not stay `true` afterwards.
```
craigcomstock, line 9: How is "state" the name of a file? Seems like it should just be named "data". And sadly I think someone else asked you to change it already from payload. So this is optional "bike shedding" :'(

cf-reactor/wakeup_channel.c
```diff
@@ -0,0 +1,106 @@
...
+    assert(channel != NULL);
+
+    channel->fds[0] = -1;
+    channel->fds[1] = -1;
+
+    /* Windows' select() only works on sockets, and a plain pipe(2) isn't
+     * one, so this always goes through socketpair() -- which works just as
+     * well on POSIX -- rather than keeping two code paths in sync. This
+     * mirrors MakeSignalPipe() in libpromises/signals.c. */
+    if (socketpair(AF_UNIX, SOCK_STREAM, 0, channel->fds) != 0)
```
craigcomstock, line 40: Curious why you don't refactor MakeSignalPipe() so that it works as-is and for this case as well? It isn't a LOT of code but bugs in the copied code might not get fixed in both places in the future. FWIW.

cf-reactor/wakeup_channel.c
```diff
@@ -0,0 +1,106 @@
...
+}
+
+void WakeupChannelNotify(const WakeupChannel *channel)
+{
+    assert(channel != NULL);
+
+    /* One byte is enough to wake the reader up; if the channel happens to
+     * already be full, the reader is already guaranteed to wake up because
+     * of what's queued, so a transient EAGAIN/EWOULDBLOCK here is fine. */
+    unsigned char byte = 1;
```
craigcomstock, line 82: do you need some of the "safety" that is present in SignalNotify() in libpromises/signals.c?

cf-reactor/watcher.c
```diff
@@ -0,0 +1,251 @@
...
+    assert(watchers != NULL && event_to_bundle != NULL);
+
+    WatcherCheckFn check_callback = NULL;
+    WatcherStateDestroyFn destroy_state = NULL;
+    switch (type)
+    {
+
+    // TODO: add more cases
+
+    default:
```
craigcomstock, line 101: So... why not include EVENT_FILE_DELETED here?

cf-reactor/watcher.c
```diff
@@ -0,0 +1,251 @@
...
+    WatcherStateDestroyFn destroy_state;
+    void *state;
+    time_t poll_interval_secs;
+    time_t next_due;
+} Watcher;
+
+static void WatcherDestroy(void *item); /* defined below, next to WatcherRegister() */
+
+static Seq *watchers = NULL;
+static Map *event_to_bundle = NULL;
```
craigcomstock, line 59: claude seems to think that these two globals are at risk for multi-thread mutations and lack locking.

When trying to see if this was true I could not find a call to WatcherRegister() which is what claude was concerned about, maybe due to this comment in the README

```
cf-reactor/README.md:33:cf-reactor reads the policy periodically, and when it does, rebuilds the list of watchers and the hashmap using the single function `WatcherRegister(key, event_type, state, bundle, interval)`. Each events promise is associated with an event type, which is defined in `when` bodies:
```

So... did I miss the call to WatcherRegister() or ???

cf-reactor/watcher.c
```diff
@@ -0,0 +1,251 @@
...
+        {
+            Watcher *w = SeqAt(watchers, i);
+
+            if (now >= w->next_due)
+            {
+                bool fired = w->check_callback(w->state);
+                w->next_due = now + w->poll_interval_secs;
+                if (fired)
+                {
+                    ThreadedQueuePush(event_queue, w->key);
```
craigcomstock, line 167: Claude seems to think that this raw pointer, `w->key`, is added to the queue with a null destructor so if the watcher is destroyed during a policy re-read before the queue entry is popped, it is a use-after-free issue. Thoughts? Interesting to due to my other finding that WatcherRegister() that would be called during policy re-read seems to not actually be called anywhere?

cf-reactor/wakeup_channel.c
```diff
@@ -0,0 +1,106 @@
...
+    send(channel->fds[1], (const char *) &byte, sizeof(byte), 0);
```
craigcomstock, line 83: yeah, claude noticed that you throw away the return value of send() which might cause silent errors and lost wakeup notifications.

craigcomstock (commented):
I read through about half of the changes and researched claude's "low" effort code review results.

cf-reactor/README.md
victormlg: I think state is more fitting, since it not only stores the data about the events we listen (like a filename) but it also stores some state describing how it looked like in the previous call. For files, it's quite minimal,but I imagine that in the future with more complex events, this would look more like a state.

cf-reactor/watcher.c
victormlg: Yes, there _will be_ a race condition: cf-reactor rebuilds the `watchers` list and the `event_to_bundle` hashmap from the policy while the event loop is reading from it. In the future I will have to add some locks here. However, that it is out of scope of this PR, because cf-reactor doesn't read from the policy yet.

cf-reactor/watcher.c
victormlg: This is a real issue, but I think it's also out of scope because the `event_to_bundle` hashmap and the `watchers` list do not get destroyed yet before reading the policy.  This will be fixed when it does using a locking system.

cf-reactor/README.md
craigcomstock: sounds good. thanks!

cf-reactor/watcher.c
craigcomstock: Ah, ok! So maybe a TODO somewhere?

cf-reactor/watcher.c
craigcomstock: So a TODO would be good, right?

cf-reactor/watcher.c
```diff
@@ -0,0 +1,253 @@
...
+/*****************************************************************************/
+
+bool EventWatcherInitialize(int *fd)
+{
+    assert(fd != NULL);
+    assert(watchers != NULL && event_to_bundle != NULL); /* WatcherRegistryInitialize() first */
+
+    if (!WakeupChannelOpen(&wakeup_channel))
+    {
+        return false;
```
craigcomstock, line 195: I think an additional log statement that the WakeupChannelOpen() failed in EventWatcherInitialize() might be helpful in debugging, especially if WakeupChannelOpen() gets used in more than one place (in the future).

cf-reactor/watcher.c
```diff
@@ -0,0 +1,253 @@
...
+    return true;
+}
+
+void EventWatcherHandleEvents(int fd, fd_set *readfds)
+{
+    assert(readfds != NULL);
+
+    if (!FD_ISSET(fd, readfds))
+    {
+        return;
```
craigcomstock, line 222: Might be nice to add a verbose or debug log message here stating that no events need to be handled right? I suppose if we might rate-limit the message to avoid excessive noise, like log once every five minutes if no events "No events triggered in the past five minutes".

craigcomstock: Granted, you have the message below which does mention when a watcher would be fired so maybe that's enough.

craigcomstock (changes requested):
thanks for the signals refactor.

All that I'd like to see is a few TODO comments.

cf-reactor/watcher.c
```diff
@@ -0,0 +1,259 @@
...
+    void *state;
+    time_t poll_interval_secs;
+    time_t next_due;
+} Watcher;
+
+static void WatcherDestroy(void *item); /* defined below, next to WatcherRegister() */
+
+/** TODO: potential race condition. If the policy is reparsed and watchers are re-registered 
+/* while the watcher thread is iterating over this Seq, 
+/* it may read a Watcher that is being freed or reallocated concurrently */
```
olehermanse: Compiler doesn't like this, you´re starting a new multiline comment with `/*` while you are already inside one. In general, I think you could use more single line comments `//`.

cf-reactor/watcher.c
```diff
@@ -0,0 +1,259 @@
...
+#define MAX_WATCHER_THREAD_SLEEP_SECS 1
+
+/* Not currently caller-configurable -- WatcherRegister() doesn't take an
+ * interval parameter -- so every watcher polls at this rate for now. */
+#define DEFAULT_WATCHER_POLL_INTERVAL_SECS 10
+
+typedef struct
+{
+    char *key;
+    WatcherCheckFn check_callback;       /* resolved from `type` at WatcherRegister() time */
```
olehermanse:
```suggestion
    WatcherCheckFn check_callback; // resolved from `type` at WatcherRegister() time
```

cf-reactor/watcher.c
```diff
@@ -0,0 +1,259 @@
...
+  included file COSL.txt.
+*/
+
+#include <watcher.h>
+#include <wakeup_channel.h>
+#include <signal_lib.h>         /* MaskTerminationSignalsInThread() */
+#include <signals.h>            /* IsPendingTermination() */
+#include <logging.h>
+#include <alloc.h>
+#include <string_lib.h>         /* StringHash_untyped(), StringEqual_untyped() */
```
olehermanse:
```suggestion
#include <signal_lib.h>         // MaskTerminationSignalsInThread()
#include <signals.h>            // IsPendingTermination()
#include <logging.h>
#include <alloc.h>
#include <string_lib.h>         // StringHash_untyped(), StringEqual_untyped()
```

cf-reactor/watcher.c
```diff
@@ -0,0 +1,259 @@
...
+{
+    SeqDestroy(watchers);
+    watchers = NULL;
+    MapDestroy(event_to_bundle);
+    event_to_bundle = NULL;
+}
+
+/*****************************************************************************/
+/* WatcherRegister() / WatcherDestroy() -- create and destroy one Watcher. */
+/*****************************************************************************/
```
olehermanse: I think these banners with lines of asterisks are unnecessary. (Also the ending `*/` on the middle line is misaligned.)

craigcomstock (approved):

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14678/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14678/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14678/

cf-reactor/watcher.c
```diff
@@ -0,0 +1,247 @@
...
+        time_t sleep_for = MAX_WATCHER_THREAD_SLEEP_SECS;
+        bool any_event = false;
+
+        for (size_t i = 0; i < SeqLength(watchers); i++)
+        {
+            Watcher *w = SeqAt(watchers, i);
+
+            if (now >= w->next_due)
+            {
+                bool fired = w->check_callback(w->state);
```
larsewi, line 163: `WatcherRegister()` set's `check_callback` to `NULL`. So this will segfault.

cf-reactor/watcher.c
```diff
@@ -0,0 +1,247 @@
...
+                if (fired)
+                {
+                    /** TODO: potential use after free. If a policy reparse destroys this Watcher (and frees w->key)
+                     * before the queued key is consumed, the consumer will read freed memory */
+                    ThreadedQueuePush(event_queue, w->key);
+                    any_event = true;
+                }
+            }
+
+            time_t until_due = (w->next_due > now) ? (w->next_due - now) : 0;
```
larsewi, line 174: If the interval is intended to be set in policy, we should probably make sure it's never `<= 0` which can cause busy spin at 100% CPU

cf-reactor/watcher.c
```diff
@@ -0,0 +1,247 @@
...
+#include <pthread.h>
+
+/* Upper bound on how long the watcher thread ever sleeps in one go, so that
+ * IsPendingTermination() is re-checked at least this often during shutdown,
+ * regardless of what poll intervals watchers asked for. */
+#define MAX_WATCHER_THREAD_SLEEP_SECS 1
+
+/* Not currently caller-configurable -- WatcherRegister() doesn't take an
+ * interval parameter -- so every watcher polls at this rate for now. */
+#define DEFAULT_WATCHER_POLL_INTERVAL_SECS 10
```
larsewi: This one appears to be unused

cf-reactor/reactor_context.c
```diff
@@ -37,106 +38,129 @@ static ReactorFd *ReactorFdNew(int fd, ReactorFdType type)
...
         size_t num_nova_fds = 0;
 
-        if (!ReactorNovaInitialize(nova_fds, ctx->max_nova_fds, &num_nova_fds))
+        if (!ReactorNovaInitialize(nova_fds, reactor_context->max_nova_fds, &num_nova_fds))
         {
             free(nova_fds);
-            SeqDestroy(ctx->fds);
-            ctx->fds = NULL;
+            SeqDestroy(reactor_context->fds);
+            reactor_context->fds = NULL;
```
larsewi, line 59: Potential memory leak
```suggestion
            reactor_context->fds = NULL;
            WatcherRegistryFinalize();
```

cf-reactor/watcher.c
```diff
@@ -0,0 +1,247 @@
...
+
+static void WatcherDestroy(void *item); /* defined below, next to WatcherRegister() */
+
+// TODO: potential race condition. If the policy is reparsed and watchers are re-registered 
+// while the watcher thread is iterating over this Seq, 
+// it may read a Watcher that is being freed or reallocated concurrently
+static Seq *watchers = NULL;
+static Map *event_to_bundle = NULL;
+
+static WakeupChannel wakeup_channel;
```
larsewi: Maybe we should initialize it to `{-1, -1}` so that we don't accidentally close stdin twice

cf-reactor/watcher.c
```diff
@@ -0,0 +1,247 @@
...
+        const char *key = item;
+        ARG_UNUSED const Bundle *bundle = MapGet(event_to_bundle, key);
+        Log(LOG_LEVEL_NOTICE, "Reactor watcher '%s' fired", key);
+        // TODO: run bundle
+    }
+}
+
+void EventWatcherFinalize(void)
+{
+    pthread_join(watcher_thread, NULL);
```
larsewi: If the main thread wants to exit -- e.g. `cf-reactor.c:286`
```C
            /*** error ***/
            Log(LOG_LEVEL_ERR, "Failed to poll events: %s", GetErrorStr());
            break;
```
-- then what would stop the thread? Because `IsPendingTermination()` would still return false.

Maybe the main thread could set it. But, also, we should not trust that the thread will exit. Hence, we should probably have a timeout followed by pthread_kill or something similar

larsewi (commented):
Some smaller comments

olehermanse:
@cf-bottom please test this in jenkins.

cf-reactor/watcher.c
```diff
@@ -0,0 +1,254 @@
...
+
+void WatcherRegistryFinalize(void)
+{
+    SeqDestroy(watchers);
+    watchers = NULL;
+    MapDestroy(event_to_bundle);
+    event_to_bundle = NULL;
+}
+
+// Expects interval to be strictly superior to 0, otherwise the watcher thread will busy spin
```
olehermanse: superior => greater than

olehermanse (approved):

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14692/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14692/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14692/

larsewi (approved):

PR #6355, opened 2026-09-07
ENT-14195 3. Event handling for file deletion

victormlg:
# File deletion

In this PR we create a new event type for file deletion with all the necessary code to be working in the logic of watcher.c

### DIFF
--- a/cf-reactor/file_watcher.c
+++ b/cf-reactor/file_watcher.c
@@ -1,0 +1,88 @@
+/*
+  Copyright 2026 Northern.tech AS
+
+  This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+  This program is free software; you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by the
+  Free Software Foundation; version 3.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#include <file_watcher.h>
+#include <watcher.h>
+#include <logging.h>
+#include <alloc.h>
+#include <sys/stat.h>
+#include <errno.h>
+
+// =========== code for EVENT_FILE_DELETED event type ===========
+
+typedef struct
+{
+    char *path;
+    bool existed_last_check;
+} FileWatcherPayload;
+
+/* Only ENOENT/ENOTDIR mean the path genuinely doesn't exist. Any other
+ * stat() failure (e.g. EACCES, ESTALE) is a transient/permission error, not
+ * a deletion, so treat the file as still present rather than misreporting
+ * it as deleted. */
+static bool FileExists(const char *path)
+{
+    struct stat sb;
+    if (stat(path, &sb) == 0)
+    {
+        return true;
+    }
+
+    if (errno == ENOENT || errno == ENOTDIR)
+    {
+        return false;
+    }
+
+    Log(LOG_LEVEL_ERR, "Unable to stat '%s' while checking for file deletion: %s", path, GetErrorStr());
+    return true;
+}
+
+void *FileWatcherPayloadNew(const char *path)
+{
+    FileWatcherPayload *pl = (FileWatcherPayload *) xmalloc(sizeof(FileWatcherPayload));
+
+    pl->path = xstrdup(path);
+    pl->existed_last_check = FileExists(path);
+
+    return (void *) pl;
+}
+
+bool CheckFileExists(void *payload)
+{
+    FileWatcherPayload *fwp = payload;
+
+    bool exists_now = FileExists(fwp->path);
+    bool deleted = fwp->existed_last_check && !exists_now;
+
+    fwp->existed_last_check = exists_now;
+    return deleted;
+}
+
+void DestroyFileWatcherPayload(void *payload)
+{
+    FileWatcherPayload *fwp = payload;
+    free(fwp->path);
+    free(fwp);
+}
+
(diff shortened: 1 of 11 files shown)


cf-bottom:
Thank you for submitting a pull request! Maybe @larsewi can review this?

cf-reactor/file_watcher.c
```diff
@@ -0,0 +1,88 @@
...
+{
+    char *path;
+    bool existed_last_check;
+} FileWatcherPayload;
+
+/* Only ENOENT/ENOTDIR mean the path genuinely doesn't exist. Any other
+ * stat() failure (e.g. EACCES, ESTALE) is a transient/permission error, not
+ * a deletion, so treat the file as still present rather than misreporting
+ * it as deleted. */
+static bool FileExists(const char *path)
```
craigcomstock: Maybe change the name of this and reverse the logic to be FileDeleted() since that a special thing you are wanting to know from this function and you have built in some special logic here.

If you wanted FileExists() you should probably use PathExists() from libntech.

craigcomstock: If you want to keep the logic the same direction you could change this to FileNotDeleted() or PathNotDeleted().

cf-reactor/watcher.c
```diff
@@ -94,11 +95,13 @@ void WatcherRegister(const char *key, EventType type, void *state, Bundle *bundl
     WatcherStateDestroyFn destroy_state = NULL;
     switch (type)
     {
-
     case EVENT_FILE_DELETED:
-        // TODO: initialize check_callback and destroy_state
+        check_callback = CheckFileExists;
```
craigcomstock: Seems you should change this to CheckFileDeleted() since I would imagine we could have a watcher for FileCreated, FileExists, FileModified, FilePermissionsChanged, FileOwnershipChanged, etc...

cf-reactor/file_watcher.c
```diff
@@ -0,0 +1,88 @@
...
+{
+    FileWatcherPayload *pl = (FileWatcherPayload *) xmalloc(sizeof(FileWatcherPayload));
+
+    pl->path = xstrdup(path);
+    pl->existed_last_check = FileExists(path);
+
+    return (void *) pl;
+}
+
+bool CheckFileExists(void *payload)
```
craigcomstock: As I mentioned in watcher.c I think this file should handle many types of situations and to make it easier to find the code this should be called CheckFileDeleted() unless you were thinking that one check would handle many different situations like create, delete, chmod, chown changes, but that would seem to break the pattern you use in watcher.c with `case EVENT_FILE_DELETED:`.

cf-reactor/file_watcher.c
```diff
@@ -0,0 +1,88 @@
...
+{
+    FileWatcherPayload *fwp = payload;
```
craigcomstock: assert(payload != null)?

cf-reactor/file_watcher.c
```diff
@@ -0,0 +1,88 @@
...
+
+    bool exists_now = FileExists(fwp->path);
+    bool deleted = fwp->existed_last_check && !exists_now;
+
+    fwp->existed_last_check = exists_now;
```
craigcomstock: I wonder if here it would be nice to save the timestamp instead of a boolean so that you could add debug or verbose logs about when this code specifically observes the change.

craigcomstock (commented):
Looks good to me. Maybe some semantic word changes to be clear that all of this is about files being deleted instead of files existing.

cf-reactor/file_watcher.c
```diff
@@ -0,0 +1,107 @@
...
+
+/* Only ENOENT/ENOTDIR mean the path genuinely doesn't exist. Any other
+ * stat() failure (e.g. EACCES, ESTALE) is a transient/permission error, not
+ * a deletion, so don't misreport the file as deleted.
+ *
+ * Follows symlinks */
+static bool FileDeleted(const char *path)
+{
+    struct stat sb;
+    if (stat(path, &sb) == 0)
```
larsewi: Should we use `lstat(2)` instead?

cf-reactor/file_watcher.c
```diff
@@ -0,0 +1,107 @@
...
+            fwp->path,
+            (now_ts != NULL) ? now_ts : "unknown time",
+            (last_seen_ts != NULL) ? last_seen_ts : "unknown time");
+    }
+
+    fwp->last_seen = deleted_now ? 0 : now;
+    return deleted;
+}
+
+void DestroyFileWatcherPayload(void *payload)
```
larsewi: Normally these types of functions support passing `NULL` as argument

cf-reactor/file_watcher.c
```diff
@@ -0,0 +1,107 @@
...
+#include <sys/stat.h>
+#include <errno.h>
+
+// =========== code for EVENT_FILE_DELETED event type ===========
+
+typedef struct
+{
+    char *path;
+    time_t last_seen; /* Time the file was last observed present, 0 if absent */
+} FileWatcherPayload;
```
larsewi: How about `FileWatcherState`?

cf-reactor/file_watcher.c
```diff
@@ -0,0 +1,110 @@
...
+
+/* Only ENOENT/ENOTDIR mean the path genuinely doesn't exist. Any other
+ * stat() failure (e.g. EACCES, ESTALE) is a transient/permission error, not
+ * a deletion, so don't misreport the file as deleted.
+ *
+ * Follows symlinks */
+static bool FileDeleted(const char *path)
```
larsewi: Don't we already have a `FileExists` function in libpromises or libntech?

cf-reactor/file_watcher.c
victormlg: I think this already does support `NULL` as an argument? I don't see what you mean

cf-reactor/file_watcher.c
victormlg: We have PathExists and Nova_FileExists, but one doesn't distinguish between reasons of the error, and the other is Nova only

cf-reactor/file_watcher.c
victormlg: Ah, I see now, I had the fix in the wrong commit. Fixed it.

larsewi (approved):

larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14701/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14701/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14701/

PR #6359, opened 2026-09-21
cf-check: Fix segfault on missing state directory

btriller:


cf-bottom:

cf-check/utilities.c
```diff
@@ -15,6 +15,7 @@ Seq *default_lmdb_files()
     if (files == NULL)
     {
         Log(LOG_LEVEL_ERR, "Could not open %s", state);
+        files = SeqNew(1, NULL);
```
larsewi:
```suggestion
        return NULL;
```

larsewi (changes requested):
Thanks @btriller 🚀  Although it should probably return `NULL` so that the failure can be detected by the callers.

cf-check/utilities.c
btriller: Ah, didn't check for other uses of `argv_to_lmdb_files()`. Will squash if updated change is ok. I guess, leaking sequence is no real issue on `argv_to_lmdb_files()` usage?
```
% valgrind --leak-check=full ./cf-check/cf-check dump
==1477289== Memcheck, a memory error detector
==1477289== Copyright (C) 2002-2024, and GNU GPL'd, by Julian Seward et al.
==1477289== Using Valgrind-3.24.0 and LibVEX; rerun with -h for copyright info
==1477289== Command: ./cf-check/cf-check dump
==1477289==
    info: No filenames specified, defaulting to .lmdb files in /home/btriller/.cfagent/state
No database files to dump.
==1477289==
==1477289== HEAP SUMMARY:
==1477289==     in use at exit: 136 bytes in 3 blocks
==1477289==   total heap usage: 34 allocs, 31 frees, 53,996 bytes allocated
==1477289==
==1477289== 112 (32 direct, 80 indirect) bytes in 1 blocks are definitely lost in loss record 3 of 3
==1477289==    at 0x4844818: malloc (vg_replace_malloc.c:446)
==1477289==    by 0x11209C: xmalloc (alloc.c:42)
==1477289==    by 0x11BBA7: SeqNew (sequence.c:32)
==1477289==    by 0x112AE3: ListDir (file_lib.c:634)
==1477289==    by 0x10F750: default_lmdb_files (utilities.c:14)
==1477289==    by 0x10F4C7: dump_main (dump.c:556)
==1477289==    by 0x10CD49: main (cf-check.c:209)
==1477289==
==1477289== LEAK SUMMARY:
==1477289==    definitely lost: 32 bytes in 1 blocks
==1477289==    indirectly lost: 80 bytes in 1 blocks
==1477289==      possibly lost: 0 bytes in 0 blocks
==1477289==    still reachable: 24 bytes in 1 blocks
==1477289==         suppressed: 0 bytes in 0 blocks
==1477289== Reachable blocks (those to which a pointer was found) are not shown.
==1477289== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==1477289==
==1477289== For lists of detected and suppressed errors, rerun with: -s
==1477289== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```

cf-check/dump.c
```diff
@@ -556,6 +556,12 @@ int dump_main(int argc, const char *const *const argv)
     }
 
     Seq *files = argv_to_lmdb_files(argc, argv, offset);
+    if (files == NULL || SeqLength(files) == 0)
+    {
+        printf("No database files to dump.\n");
+        return 1;
+    }
```
larsewi, line 563: The `argv_to_lmdb_files` already logs an error. Hence, there is no need to log another one here.

```suggestion
    if (files == NULL || SeqLength(files) == 0)
    {
        return 1;
    }
```

larsewi (commented):
This is good. However, I still think `argv_to_lmdb_files` should return `NULL` on failure, so that the call sites can handle it.

The memory leak is preexisting. But, feel free to fix it too, preferably in a separate commit.

Thanks again!

